PR #1, opened 2018-08-09
Feature/ap 95

a5rar:
This pull request includes is for the create application service.  Best way to review this is to go through the commits in order so you can hopefully understand what i was thinking.


bleach:
First up, github is telling me there are conflicts, can you resolve them please?

spec/requests/legalaid-application_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'rails_helper'
+
+RSpec.describe 'Legal aid applications' do
+  let(:response_json) {JSON.parse(response.body)}
+
+  describe  'POST /api/v1/applications' do
+    it 'returns an application reference' do
+
+      post '/api/v1/applications'
+      expect(response_json['data']['attributes']['application_ref']).to be
```
bleach: I'm not sure that it is sufficient to test that the string you get is truthy. Won't this pass if `application_ref` is an empty string?

Maybe an approach would be to create two and ensure they have unique  `application_refs`?

app/models/legal_aid_application.rb
```diff
@@ -0,0 +1,13 @@
...
+
+
+  before_create :create_app_ref
+
+
+
+  private
+
+   def create_app_ref
+     self.application_ref = SecureRandom.uuid
```
bleach: A note that we might want to try [this approach](23891181-62ef-4e27-ad6d-c13d2159ee81) when we move to postgres.

db/migrate/20180807211758_create_legal_aid_applications.rb
```diff
@@ -0,0 +1,9 @@
+class CreateLegalAidApplications < ActiveRecord::Migration[5.2]
+  def change
+    create_table :legal_aid_applications do |t|
+      t.string :application_ref
+
+      t.timestamps
```
bleach: Just a note for anyone else puzzled by this. It's an ActiveRecord macro that results in the `created_at` and `updated_at` columns.

README.md
```diff
@@ -1,4 +1,43 @@
...
+* Ruby version 
+    * Ruby version 2.5.1
+    * Rails 5
+
+
+* System dependencies
+    * none
+
+* Configuration
+    * ```bundle install```
```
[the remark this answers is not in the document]
a5rar: The purpose of that bundle install i think is to update the gemfile lock and make sure the developer has the required gems locally.

bin/spring
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env ruby
```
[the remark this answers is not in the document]
a5rar: does your gem file have spring in it? when you create new project.  I thought the purpose of the spring is to keep the application alive in the background whilst changes are being made.  I don't remember explicitly adding this.  It may have come in when i did bundle install or rspec install

spec/requests/legalaid-application_spec.rb
a5rar: I thought that was not empty equivalent.  Although i think ```not_to be_empty``` is probably the better one to use here

app/models/legal_aid_application.rb
a5rar: if this approach is a url link it doesn't go anywhere other than a 404

db/migrate/20180807211758_create_legal_aid_applications.rb
a5rar: yep rails gives you these audit columns as part of its conventions

Gemfile.lock
```diff
@@ -0,0 +1,172 @@
...
+    minitest (5.11.3)
+    msgpack (1.2.4)
+    nio4r (2.3.1)
+    nokogiri (1.8.4)
+      mini_portile2 (~> 2.3.0)
+    puma (3.12.0)
+    rack (2.0.5)
+    rack-test (1.1.0)
+      rack (>= 1.0, < 3)
+    rails (5.2.0)
```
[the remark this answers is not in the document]
a5rar: i think updating the rails version should be a cross team thing so there are no surprises on other developers machine.  We will do this soon

README.md
```diff
@@ -1,4 +1,43 @@
-# laa-apply-for-legalaid-api
-This api will povide various services to create a legal aid application
 
-Test
+#LAA apply for legal aid
+
+
+This is the service api for persisting case related information to the back end database and
```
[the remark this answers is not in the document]
a5rar: cool i have changed this to application

adrianclay:
As a general note, I'd recommend keeping PRs small.  The smaller a PR, the easier it is to review.  1.3K lines requires a large block of time to review, which makes it harder for somebody to schedule into their day.

On this PR I'd have raised one PR which created the Rails app and a second PR which focuses on adding the functionality.

README.md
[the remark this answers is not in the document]
bleach: So, after conversations offline, this happens because `rbenv` sets a non-default `BUNDLE_PATH`. I'm using `chruby` which doesn't. Everyone else is using `rbenv` so we can ignore this for now.

bleach:
Paired with @a5rar to remove `.idea` directory from all commits using `git filter-branch --tree-filter 'rm -rf .idea' --prune-empty -f BRANCHNAME` (NB - this command should be used with caution and never on master, as we'll lose history).

PR #3, opened 2018-08-14
feature/AP-96  Added dockerfile to build containers.

a5rar:
Added instructions to readme

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,6 @@ RUN apk --update add --virtual build-dependencies \
                                     libxslt-dev \
 #                                    postgresql-dev \
                                     sqlite-dev \
-
                                     tzdata \
                                     && rm -rf /var/cache/apk/*
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,6 @@ RUN apk --update add --virtual build-dependencies \
                                     build-base \
                                     libxml2-dev \
                                     libxslt-dev \
-#                                    postgresql-dev \
                                     sqlite-dev \
                                     tzdata \
                                     && rm -rf /var/cache/apk/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,12 @@ FROM ruby:2.5.1-alpine
 MAINTAINER apply for legal aid team
 
 RUN apk --update add --virtual build-dependencies \
-                                    build-base \
-                                    libxml2-dev \
-                                    libxslt-dev \
-                                    sqlite-dev \
-                                    tzdata \
-                                    && rm -rf /var/cache/apk/*
+                               build-base \
+                               libxml2-dev \
+                               libxslt-dev \
+                               sqlite-dev \
+                               tzdata \
+                               && rm -rf /var/cache/apk/*
 
 RUN bundle config build.nokogiri --use-system-libraries
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM ruby:2.5.1-alpine
+MAINTAINER apply for legal aid team
+
+RUN apk --update add --virtual build-dependencies \
+                                    build-base \
+                                    libxml2-dev \
+                                    libxslt-dev \
+#                                    postgresql-dev \
+                                    sqlite-dev \
+                                    nodejs \
+                                    tzdata \
+                                    && rm -rf /var/cache/apk/*
+
+RUN bundle config build.nokogiri --use-system-libraries
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM ruby:2.5.1-alpine
+MAINTAINER apply for legal aid team
+
+RUN apk --update add --virtual build-dependencies \
+                                    build-base \
+                                    libxml2-dev \
+                                    libxslt-dev \
+#                                    postgresql-dev \
+                                    sqlite-dev \
+                                    nodejs \
```
adrianclay: What necessitates NodeJS being installed?

a5rar: nothing it has now been removed.  I copied it over form another config i had for a web project.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM ruby:2.5.1-alpine
+MAINTAINER apply for legal aid team
+
+RUN apk --update add --virtual build-dependencies \
+                                    build-base \
+                                    libxml2-dev \
+                                    libxslt-dev \
+#                                    postgresql-dev \
```
bleach: It is my opinion that we shouldn't commit commented-out lines, however, on the assumption that we're going to use this in the near future, it can stay.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM ruby:2.5.1-alpine
+MAINTAINER apply for legal aid team
+
+RUN apk --update add --virtual build-dependencies \
+                                    build-base \
```
bleach: Could we make the indentation of this and the following lines match the line above? I think it would be easier to read.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+                                    libxml2-dev \
+                                    libxslt-dev \
+#                                    postgresql-dev \
+                                    sqlite-dev \
+
```
bleach: I think we should remove this blank line. I get this warning:

```[WARNING]: Empty continuation line found in:
    RUN apk --update add --virtual build-dependencies                                     build-base                                     libxml2-dev                                     libxslt-dev                                     sqlite-dev                                     tzdata                                     && rm -rf /var/cache/apk/*
[WARNING]: Empty continuation lines will become errors in a future release.```

If we get used to seeing red text printed to the screen then we'll end up missing something important. Also, the message says it'll definitely break later.

File: Dockerfile [hunk 5]
```diff
@@ -0,0 +1,28 @@
...
+RUN bundle config build.nokogiri --use-system-libraries
+
+RUN mkdir /usr/src/app
+
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
+
+RUN gem update bundler
```
bleach: This line doesn't currently have any practical effect, as the base ruby image already contains bundler 1.16.3. Whether the line is there or not, after running a `docker build -t laa-apply-for-legalaid-api` this always shows the same output:

```$ docker run   laa-apply-for-legalaid-api gem list bundler
bundler (1.16.3)
```

However, if we leave it in then I think that we'll get the latest bundler version whenever it's released. I think what we should do is rely on the base image for bundler version bumps for now ie. remove this line.

Dockerfile
adrianclay: Blank line

Dockerfile
adrianclay: I'd remove commented code

README.md
```diff
@@ -45,3 +45,20 @@ may well be used to fire requests to other services.
         * Only GET is supported at the moment not sure anything else is needed here
 
 
+#Docker
```
adrianclay: Missing space between the hash and the "Docker" makes it not look like a title:

image

Dockerfile
bleach: Removed.

Dockerfile
bleach: Removed.

bleach: I revised my opinion after I saw that @adrianclay thought the same. Removed.

Dockerfile
bleach: I've pushed a commit to do this.

README.md
bleach: I did this and also made it a sub-heading.

bleach:
I've resolved both mine and @adrianclay's comments as @a5rar isn't available.

I also pushed one other unrelated README change in 3232e37 after I noticed that we seemed to have two headings and introductions (possibly the result of a merge that I missed in #2 )

PR #4, opened 2018-08-16
feature/AP-96  Added postgres as a dependency and provided instructio…

a5rar:
…ns for getting up and running

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,14 @@ FROM ruby:2.5.1-alpine
 MAINTAINER apply for legal aid team
 
 RUN apk --update add --virtual build-dependencies \
-                               build-base \
-                               libxml2-dev \
-                               libxslt-dev \
-                               sqlite-dev \
-                               tzdata \
-                               && rm -rf /var/cache/apk/*
+                                    build-base \
+                                    libxml2-dev \
+                                    libxslt-dev \
+                                    postgresql-dev \
+#                                    sqlite-dev \
+
+                                    tzdata \
+                                    && rm -rf /var/cache/apk/*
 
 RUN bundle config build.nokogiri --use-system-libraries
 
@@ -16,6 +18,8 @@ RUN mkdir /usr/src/app
 WORKDIR /usr/src/app
 
 COPY . /usr/src/app
+
+RUN gem update bundler
 
 RUN  bundle install
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,12 @@ FROM ruby:2.5.1-alpine
 MAINTAINER apply for legal aid team
 
 RUN apk --update add --virtual build-dependencies \
-                                    build-base \
-                                    libxml2-dev \
-                                    libxslt-dev \
-                                    postgresql-dev \
-                                    tzdata \
-                                    && rm -rf /var/cache/apk/*
+                               build-base \
+                               libxml2-dev \
+                               libxslt-dev \
+                               postgresql-dev \
+                               tzdata \
+                               && rm -rf /var/cache/apk/*
 
 RUN bundle config build.nokogiri --use-system-libraries
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,8 +6,6 @@ RUN apk --update add --virtual build-dependencies \
                                     libxml2-dev \
                                     libxslt-dev \
                                     postgresql-dev \
-#                                    sqlite-dev \
-
                                     tzdata \
                                     && rm -rf /var/cache/apk/*
 
@@ -18,8 +16,6 @@ RUN mkdir /usr/src/app
 WORKDIR /usr/src/app
 
 COPY . /usr/src/app
-
-RUN gem update bundler
 
 RUN  bundle install
 
--- a/README.md
+++ b/README.md
@@ -12,17 +12,10 @@ may well be used to fire requests to other services.
     * postgres 10.5  -> see setup below
 
 * Configuration
-    you may have to install postgres gem if you dont already have it 
-    
-    ```brew install libpq ```                                         # needed for the pg gem's native extensions
    
-    ```PATH=$PATH:/usr/local/opt/libpq/bin```        # so that the pg gem can find the pq library
+    ```brew install postgres```
     
-    ```gem install pg```
-    
-    ```rbenv rehash```
-    
-    * ```bundle install```
+    ```bundle install```
 
 
 * How to run the test suite
--- a/README.md
+++ b/README.md
@@ -9,16 +9,17 @@ may well be used to fire requests to other services.
 
 
 * System dependencies
-    * none
+    * postgres 10.5  -> see setup below
 
 * Configuration
+    you may have to install pstgres gem if you dont already have it 
+    
+    ```gem install pg```
+    
+    ```rbenv rehash```
+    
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -2,12 +2,14 @@ FROM ruby:2.5.1-alpine
...
-                               libxml2-dev \
-                               libxslt-dev \
-                               sqlite-dev \
-                               tzdata \
-                               && rm -rf /var/cache/apk/*
+                                    build-base \
+                                    libxml2-dev \
+                                    libxslt-dev \
+                                    postgresql-dev \
+#                                    sqlite-dev \
```
bleach: Re-adds lines that was removed in #3 - see ba56a87ef2ed87a6c1e5794971eff451942927a1 and 8d043ea6d5fa07341eae65f323a846525f2c75d3

Dockerfile
```diff
@@ -17,6 +19,8 @@ WORKDIR /usr/src/app
 
 COPY . /usr/src/app
 
+RUN gem update bundler
```
bleach: This was removed in #3 - see d87a7f28ce2314c9bf17a6485d61cf0639354e13

README.md
```diff
@@ -9,16 +9,17 @@ may well be used to fire requests to other services.
 
 
 * System dependencies
-    * none
+    * postgres 10.5  -> see setup below
 
 * Configuration
+    you may have to install pstgres gem if you dont already have it
```
bleach: typo: pstgres

bleach: Can you explain why the installation of `pg` from the bundle would not suffice?

README.md
```diff
@@ -56,3 +57,34 @@ And then run the image using
 
 ```docker container run  -d -p 3000:3000 laa-apply-for-legalaid-api```
 
+
+## Postgres Sql
```
bleach: I think these instructions currently cover two things:

1. bootstrapping project from scratch in dev
2. migrating from current sqlite to postgres

I think in the README, we should have just 1), maybe we have a "getting started" section for that. Put 2) in the PR description and point people at it in slack. It's just a one-off thing.

Also, I think it'll be simpler if we make a strong opinion about how the team *should* run postgres, because it makes the README easier to follow. If people really want to run a local non-Docker postgres, then we can tuck the instructions to do that away somewhere.

README.md
a5rar: I will aim to fix the readme changes in one go.  As i was expecting there to be differing views on this.
so as for the second comment around pg, i kept getting issues with bundler complaining the gem doesnt exist maybe there was something else i was doing wrong, we can remove ths iif others who haven't installed postgres gem in any way yet have the same issue

README.md
a5rar: Lets wait for a few other opinions on this first infact i want everyone to help write this bit up.  My aim was to provide a base sert of instructions which can get everyone going.

README.md
```diff
@@ -56,3 +57,34 @@ And then run the image using
...
+
+If you already have an installation of postgres you might want to use it.
+For development mode i am going for the simplest setup without setting any specific roles or users
+
+If you dont have postgres installed then you can run the following command to get up and running quickly
+
+```docker run  -d -v pg-data:/var/lib/postgresql/data -p 5432:5432  --name postgres  postgres:10.5```
+ 
+ the 10.5 version appears to be the latest and most likely the one we will be using in RDS.
```
adrianclay: FYI You can see a full list of supported versions here https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.DBVersions

a5rar: Interesting they don't have 10.5 there.  I wonder why that is

README.md
```diff
@@ -9,16 +9,17 @@ may well be used to fire requests to other services.
 
 
 * System dependencies
-    * none
+    * postgres 10.5  -> see setup below
 
 * Configuration
+    you may have to install pstgres gem if you dont already have it 
+    
+    ```gem install pg```
```
bleach: So both this and `bundle install` fail for me with this error:

```
$ gem install pg -- --with-pg-config /usr/local/opt/libpq/bin/pg_config
Building native extensions with: '--with-pg-config /usr/local/opt/libpq/bin/pg_config'
This could take a while...
ERROR:  Error installing pg:
	ERROR: Failed to build gem native extension.

    current directory: /Users/grahambleach/.gem/ruby/2.5.1/gems/pg-1.0.0/ext
/Users/grahambleach/.rubies/ruby-2.5.1/bin/ruby -r ./siteconf20180816-32064-1la9lpe.rb extconf.rb --with-pg-config /usr/local/opt/libpq/bin/pg_config
Using config values from true
checking for libpq-fe.h... no
Can't find the 'libpq-fe.h header
*** extconf.rb failed ***
Could not create Makefile due to some reason, probably lack of necessary
...
```

bleach: So the commands that are currently missing are:

```
brew install libpq                                          # needed for the pg gem's native extensions
PATH=$PATH:/usr/local/opt/libpq/bin        # so that the pg gem can find the pq library
```

After this, `bundle install` works fine.

README.md
bleach: Outcome from our real-life discussion:

- prefer getting started & how to run sections with a list of all the commands needed to get things up and running
- you're going to do a future PR to condense them down to a `make command` anyway, so we don't need to fix anything until then

bleach:
This is good to go when:

- it doesn't reverse the changes from 09cd806
- it includes instructions to allow us to install the pg gem on MacOS (install libpq and add it to the PATH)

README.md
mpw5: I didn't have any problems with `gem install pg`, which is what's in the readme. I get the same error as you when I run `gem install pg -- --with-pg-config /usr/local/opt/libpq/bin/pg_config` though. Why do we neeed the `--with-pg-config` flag?

README.md
```diff
@@ -9,16 +9,21 @@ may well be used to fire requests to other services.
...
+    * postgres 10.5  -> see setup below
 
 * Configuration
+    you may have to install postgres gem if you dont already have it 
+    
+    ```brew install libpq ```                                         # needed for the pg gem's native extensions
+   
+    ```PATH=$PATH:/usr/local/opt/libpq/bin```        # so that the pg gem can find the pq library
+    
+    ```gem install pg```
```
mpw5: `gem install pg` runs fine for me without the two new preceding libpq lines. It looks like we only need those if `gem install pg -- --with-pg-config /usr/local/opt/libpq/bin/pg_config` is run instead? If that's the approach we're taking then shouldn't we also update line 21 to reflect that?

README.md
bleach: We don't. The `--with-pg-config` option was an unsuccessful attempt to fix the failure to build native extension.

mpw5: Ah, I see. Ignore my subsequent comments in that case.

README.md
bleach: If we don't need libpq installed, then the way to test it is a fresh bundle install without libpq installed:

```bundle config path
mv $BUNDLE_PATH $BUNDLE_PATH.old   # get rid of existing bundle, to force a gem install
brew uninstall libpq
bundle install```

I tried this and it failed for me. I'd be interested to see what happens on your machines.

PR #5, opened 2018-08-16
Added a model for client details along with validation

DamanFergus:
Changed the port for the API to 3001
Generates a model and added validation
Wrote passing validation tests against the model

Mark Whitaker 20/08/2018 -  I've made a number subsequent changes to revise the model and tests.

POST request is still working as expected following these changes:

screen shot 2018-08-20 at 14 29 39


app/models/client_detail.rb
```diff
@@ -0,0 +1,15 @@
...
+  validates :name,  presence: true
+
+  validates :dob_day, numericality: {
+                    greater_than_or_equal_to: 1, less_than_or_equal_to: 31,
+                  }
+  validates :dob_month, numericality: {
+                    greater_than_or_equal_to: 1, less_than_or_equal_to: 12,
+                   }
+   validates :dob_year, numericality: {
+                    greater_than_or_equal_to: 1900, less_than_or_equal_to: 2099,
```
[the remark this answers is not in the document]
DamanFergus: Implementation on the front end is three separate fields

spec/models/client_detail_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  it "is not valid without a name" do
+    subject.name = ""
+    expect(subject).to_not be_valid
+  end
+
+  it "is not valid without a dob_day" do
+    subject.dob_day = ""
+    expect(subject).to_not be_valid
+    print subject.errors.to_a
```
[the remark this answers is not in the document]
DamanFergus: That is a mistake, I was testing something and forgot to remove it.

app/models/client_detail.rb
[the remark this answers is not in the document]
willc-work: this was done to be inline with the gov.uk design standards, making it one field would probably be easier, alternatively we could combine them in the front end and just pass one value to the db

config/puma.rb
```diff
@@ -9,7 +9,7 @@
 
 # Specifies the `port` that Puma will listen on to receive requests; default is 3000.
 #
-port        ENV.fetch("PORT") { 3000 }
+port        ENV.fetch("PORT") { 3001 }
```
mpw5: 3001 is being used by the Citizen app, I'd suggest 3002 instead.

spec/models/client_detail_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    expect(subject).to_not be_valid
+  end
+
+  it "is not valid if the date of birth is in the future" do
+    subject.date_of_birth = Date.today + 1
+    expect(subject).to_not be_valid
+  end
+
+  it "is not valid if the date of birth is before 01/01/1900" do
+    subject.date_of_birth = "31/12/1899".to_date
```
adrianclay: Minor point, but I'd be consistent about assigning the `date_of_birth` attribute either as a string (line 9) or Date object (lines 27 and 32).  I'd prefer using one which will be used by the production code, which I imagine will be a string via the JSON PATCH request.

adrianclay: Also, I believe a convention of using ["YYYY-MM-DD"](https://en.wikipedia.org/wiki/ISO_8601) would be the least surprising format for a JSON API.

adrianclay: Also, great work on the tests 🎉

config/puma.rb
DamanFergus: I had to change this locally while development due to the provider frontend and API still being stuck on the same port. Afterwards, a local port hierarchy was created

app/models/client_detail.rb
```diff
@@ -0,0 +1,15 @@
+class ClientDetail < ApplicationRecord
```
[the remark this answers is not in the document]
DamanFergus: It was an agnostic choice to reflect what Will had created in the frontend. Happy to change it to whatever you prefer

app/models/legal_aid_application.rb
```diff
@@ -1,4 +1,7 @@
 class LegalAidApplication < ApplicationRecord
+
+  has_one :applicant
```
a5rar: minor cosmetic change, i think the before** methods should be at the top of a class

app/models/applicant.rb
```diff
@@ -1,6 +1,8 @@
 require 'date'
 
-class ClientDetail < ApplicationRecord
+class Applicant < ApplicationRecord
+
+  belongs_to :legal_aid_application
```
bleach: Should it be this way around? Would it make more sense for an application to belong to an applicant? I talked to @mpw5 and I think we should keep it this way around as long as it's not going to be painful to change later

DamanFergus: Pretty sure a person can have multiple applications, but an application can only be for one person at a time. Something to ask the BAs

bleach:
As discussed in person, the tests fail after migrating from current master. The changes to the original application migration should be reverted and we need a new migration to add the applicant reference to `legal_aid_applications`

app/models/applicant.rb
```diff
@@ -0,0 +1,18 @@
+require 'date'
```
lostie: As spoken offline, this require should not necessary as `date` belongs to the ruby core libraries, so should be required already by default.

spec/models/applicant_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'rails_helper'
+require 'date'
```
lostie: [Ditto](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/5/files#r211595776)

app/models/applicant.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  validates :name, :date_of_birth, presence: true
+
+  validate :validate_date_of_birth
+
+  private
+
+  def validate_date_of_birth
+    if date_of_birth.present? && (date_of_birth < Date.new(1900, 01, 01) || date_of_birth > Date.today)
+       errors.add(:date_of_birth, "Date of birth is not valid")
```
lostie: **[not a blocker]** Going forward consider using error codes instead of strings, making use of [Rails i18n](https://guides.rubyonrails.org/i18n.html#error-message-scopes) so you can define the error messages in the appropriate locales (normally set in `config/locales/en.yml` for en-GB)

bleach:
@mpw5 only your review is now preventing merge 😉

PR #7, opened 2018-08-17
Added docker-compose and makefile script to make getting up and runni…

a5rar:
…ng faster

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,7 @@
+.git*
+db/*.sqlite3
+db/*.sqlite3-journal
+log/*
+tmp/*
+Dockerfile
+README.md
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,31 @@
+docker-build:
+	docker-compose build
+
+docker-stop:
+	docker-compose stop
+
+initial-setup:
+	docker-compose run api rake db:setup
+	docker-compose run api rake db:migrate
+
+update-db:
+	docker-compose run api rake db:migrate
+
+setup-local-db:
+	docker run  -d -v pg-data:/var/lib/postgresql/data -p 5432:5432  --name postgres  postgres:10.5
+	docker-compose run api rake db:setup db:migrate
+
+destroy-local-db:
+	@docker container stop postgres || true
+	@docker container rm postgres || true
+
+start-local-server: destroy-local-db setup-local-db
+	rails s
+
+
+serve: docker-stop docker-build update-db
+	docker-compose run --service-ports api
+
+
+test: docker-stop docker-build update-db
+	docker-compose run -e "RAILS_ENV=test" api rake  spec
--- a/README.md
+++ b/README.md
@@ -18,70 +18,53 @@ may well be used to fire requests to other services.
     ```bundle install```
 
 
-* How to run the test suite
-    * ```bundle exec rspec spec```
 
-* Services (job queues, cache servers, search engines, etc.)
+## Initial setup
+
+Run ```make initial-setup```
+
+This will create the db schema  and run a  migration. you shouldn't have to run this again.
+
+## Running the tests
+
+In order to run the tests run
+
+ * ```make test```
+ 
+ This is currently setup to run all rspec tests
+
+## Running the application
+
+ * ```make serve```
+ 
+ This will use docker compose to start postgres and the api project.
+ requests will be served on port localhost:3000
+ 
+
+## Other options
+
+There are other configurations in the makefile which you can use if you want 
+
+i.e. ```start-local-server```
+
+This will start a server without docker and setup a postgres db exposed at localhost:5432.  
+Benefit of this is you dont have to build container everytime you make a change to a file.  handy for the html/css changes.
+
+
+
+## Developer local Endpoints
+
+* Post an application 
+
+```http://localhost:3000/api/v1/applications```    
+
+* Get status of the service 
+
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -18,70 +18,53 @@ may well be used to fire requests to other services.
...
- 
- ```rake db:migrate```
- 
- 
-For every other time you do a pull it might be a good idea to 
-```rake db:migrate```
+* Get status of the service 
 
-It's also good idea at this pint to rerun the tests.
+```http://localhost:3000/api/v1/status```
```
mpw5: I think we need to be careful about running everything on 3000 - it will conflict with our other apps. Can we change this to 3002?

That will meam
Provider frontend = 3000
Citizen frontend = 3001
API = 3002

.dockerignore
```diff
@@ -0,0 +1,7 @@
+.git*
+db/*.sqlite3
```
lostie: If we're switching to `postgres` completely this line and the following one should no longer be required (I think).

README.md
mpw5: `/api/v1/status` should be `/v1/status`

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+services:
+  postgres:
+    image: postgres:10.5
+    volumes:
+    - pgdata:/var/lib/postgresql/data
+  api:
+    environment:
+    - DB_HOST=postgres
+    build: .
+    command: rails server -p 3000 -b '0.0.0.0'
```
mpw5: Port 3002?

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    ports:
+    - "3000:3000"
```
mpw5: Port 3002?

README.md
```diff
@@ -18,70 +18,53 @@ may well be used to fire requests to other services.
...
 
-## Postgres Sql
+## Developer local Endpoints
 
-If you already have an installation of postgres you might want to use it.
-For development mode i am going for the simplest setup without setting any specific roles or users
+* Post an application 
 
-If you dont have postgres installed then you can run the following command to get up and running quickly
+```http://localhost:3000/api/v1/applications```
```
willc-work: I thought that 'api' had been removed so it should now be http://localhost:3000/v1/applications

README.md
willc-work: I thought that 'api' had been removed so it should now be http://localhost:3000/v1/status

Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+update-db:
+	docker-compose run api rake db:migrate
+
+setup-local-db:
+	docker run  -d -v pg-data:/var/lib/postgresql/data -p 5432:5432  --name postgres  postgres:10.5
+	docker-compose run api rake db:setup db:migrate
+
+destroy-local-db:
+	@docker container stop postgres || true
```
lostie: `@` here (and in the following line), is it just a typo?

Makefile
```diff
@@ -0,0 +1,31 @@
...
+	@docker container rm postgres || true
+
+start-local-server: destroy-local-db setup-local-db
+	rails s
+
+
+serve: docker-stop docker-build update-db
```
lostie: **[typo]** `server`

README.md
```diff
@@ -18,70 +18,53 @@ may well be used to fire requests to other services.
...
+ * ```make test```
+ 
+ This is currently setup to run all rspec tests
 
-The docker file is created against a ruby alpine image using version 2.5.1 which is consistent
-with the latest version at the time of writing.
+## Running the application
 
-Alpine images are usually very lightweight and such are preferred choice for deploying containers.
+ * ```make serve```
```
lostie: [Ditto](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/7/files#r210875636)

Makefile
```diff
@@ -0,0 +1,31 @@
...
+setup-local-db:
+	docker run  -d -v pg-data:/var/lib/postgresql/data -p 5432:5432  --name postgres  postgres:10.5
+	docker-compose run api rake db:setup db:migrate
+
+destroy-local-db:
+	@docker container stop postgres || true
+	@docker container rm postgres || true
+
+start-local-server: destroy-local-db setup-local-db
+	rails s
```
bleach: Missing `bundle exec` here?

a5rar: i didn't need bundle exec here

Makefile
a5rar: it should be serve.  i.e. serving the content

Makefile
bleach:
```
$ make start-local-server
[...]
rails s
Rails is not currently installed on this system. To get the latest version, simply type:

    $ sudo gem install rails

You can then rerun your "rails" command.
```

I'll try switching to `rbenv` and see if it goes away.

bleach: Nope, switching to rbenv does not fix it.

But adding `./bin` to my `$PATH` does. Perhaps you already set that, or you have installed something that does it automatically?

bleach: `gem install rails` to install the gem into my rbenv-managed ruby (ie. outside of the bundle) made this command work (I think because `~/.rbenv/shims/rails` appeared), so maybe that's why it was working on your machine?

I favour changing this to be './bin/rails s' as it doesn't seem ideal to require us to install gems directly into the ruby - we want applications to exclusively use bundler managed gems.

PR #8, opened 2018-08-17
Add Rubocop (static code analyser) to the project

lostie:
#### Why

Mainly because it enforces the whole team to follow the community Ruby Style Guide and reduces the unnecessary amount of PR comments related with code style/formatting.

#### TODO (Future Feature)

The idea is to also run `Rubocop` as part of the CI process alongside with the test suite runs.

#### NOTE 👨‍👧‍👦 

Rubocop does not solve all the issues with style/formatting. It's advised for devs that work on the project to have their environments, IDE, etc configured with the appropriate setup (in this case for Ruby) to avoid creating some of the offences fixed here.


bleach:
:+1: for using rubocop

Does this get automatically run by CircleCI somehow? And should there be instructions in the README explaining that we enforce it and how to run it?

PR #9, opened 2018-08-20
Feature/ap99 circleci

a5rar:


a5rar:
I have created an issue in github to address the separation of build and test jobs in circleci. this can be addressed at a later stage.
https://github.com/ministryofjustice/laa-apply-for-legalaid-api/issues/10

.circleci/config.yml
```diff
@@ -0,0 +1,75 @@
+version: 2.0
+
+references:
+
+  build_containers_config: &build_container_config
+    docker:
+    - image: circleci/ruby:2.5.1
+      environment:
+        GITHUB_TEAM_NAME_SLUG: apply-for-legalaid
```
lostie: Just noticed this env variable, is it actually required for this image?

a5rar: i think it was used at image level because multiple steps can perhaps share it rather than keep defining it at step level for each step

README.md
```diff
@@ -61,7 +61,7 @@ Benefit of this is you dont have to build container everytime you make a change
 
 * Get status of the service 
 
-```http://localhost:3002/v1/status```
+```http://localhost:3002/v1/status```
```
lostie: **[not a blocker]** Is this an actual change? If not, consider discarding it.

.circleci/config.yml
lostie: Not marking it as a blocker, but if that's the intent, that should still be defined in a shared scope that contains the environment variable(s) that are shared across the different jobs/steps. The ruby image itself does not require it so I don't think it should be there.

PR #11, opened 2018-08-21
Remove Unit test suite

lostie:
Because **RSpec** was choosen as the test suite and the unit test suite was automatically generated when the Rails app skeleton was created, so it's not actually being used.


adrianclay:
I like the PR title + description.  Very minor point, it doesn't match the commit message.

Keeping the commit message + PR similar is useful in situations where devs can use `git blame` and see the same info that GitHub has.

bleach:
Looks like there's an incomplete sentence in the commit message now? 
> The unit test suite that was automatically generated

should probably be followed by 

> is not used and is being removed

PR #12, opened 2018-08-21
Fix Rubocop errors

mpw5:
The current CircleCI build is failing due to Rubocop errors (my fault - Sergio had introduced Rubocop in the interim, I didn't rebase from master before pushing so all tests passed in my branch but fail after the merge). These changes will fix those errors.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,3 +4,7 @@ AllCops:
   Exclude:
   - 'db/schema.rb'
   - 'vendor/**/*'
+
+Style/Documentation:
+  Exclude:
+  - '**/*'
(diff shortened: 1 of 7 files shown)


.rubocop.yml
```diff
@@ -4,3 +4,7 @@ AllCops:
   Exclude:
   - 'db/schema.rb'
   - 'vendor/**/*'
+
+Style/Documentation:
```
adrianclay: I'd want to highlight this in the "Why" of the commit, in case somebody in the future wanted to understand the reasoning behind excluding this.

adrianclay:
I'd also recommend highlighting how we got into this situation where the CI build is failing in master, share that with the rest of the team so we can understand the limitations of our build pipeline.

.rubocop.yml
mpw5: Thanks Adrian. Comments added in response to both points.

bleach:
Merging, as master is a bit broken until we do.

PR #13, opened 2018-08-22
Update readme run tests

bleach:
## What

- run `rake` not `rake spec`
- make running tests locally the preferred way

## Why

- `rake` runs rubocop, `rake spec` does not. We want to run rubocop for the reasons given in #8 
- it's a lot faster to run them locally

## How to review

Ensure the readme still makes sense


README.md
```diff
@@ -68,4 +67,4 @@ Benefit of this is you dont have to build container everytime you make a change
 
     
  
-
```
lostie: ✂️

PR #14, opened 2018-08-23
Provision a database with helm alongside the API server

lostie:
[JIRA Ticket - AP-100](https://dsdmoj.atlassian.net/browse/AP-100)

#### What

Adds an [Helm](https://helm.sh/) configuration to the application to provision the application server and a postgresql database into `staging` environment (can do to `production` as well, but there's more configurations that need to be addressed maybe as part of [AP-98](https://dsdmoj.atlassian.net/browse/AP-98))

#### How

* Helm provision configurations includes a postgresql server chart as a dependency, which means, when the application is deployed, a postgresql server pod instance is also provisioned (as it can be seen by the screenshot). This, as noted in the JIRA ticket **is a temporary solution** and we will be moving to a RDS solution in the foreseeable future. 
  * Changes to this can be found under the `helm_deploy/` directory
* `Dockerfile` and the `docker-compose.yml` now use a binary script `bin/run` as the start point of the server which includes some tasks such as:
  * 1. checking that the database server is up and accepting connections
  * 2. setup and/or migrate the database. this ensures that when the app is being started the migrations are always up to date
  * 3. start the application server as per usual
  *  Changes to this can be found in the Dockerfile itself, the docker-compose.yml and the scripts under `docker/` directory

To deploy the application, we executed the following command (this will be moved to an automated deploy as part of [AP-98](https://dsdmoj.atlassian.net/browse/AP-98)):
```bash
          helm upgrade ${APPLICATON_DEPLOY_NAME} ./helm_deploy/apply-for-legalaid-api/ \
                        --install \
                        --tiller-namespace=${KUBE_ENV_STAGING_NAMESPACE} \
                        --namespace=${KUBE_ENV_STAGING_NAMESPACE} \
                        --set image.repository="${ECR_ENDPOINT}/${GITHUB_TEAM_NAME_SLUG}/${REPO_NAME}" \
                        --set image.tag="${CIRCLE_SHA1}" \
                        --set ingress.hosts="{${STAGING_HOST}}" \
                        --set replicaCount="2" \
                        --debug

```

screen shot 2018-08-23 at 17 12 36


spec/services/healthcheck_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+RSpec.describe Healthcheck, type: :service do
+  describe '.perform' do
+    it 'returns an hash with the healthy status' do
+      result = described_class.perform
+      expect(result).to eq(database: true)
+    end
+
+    context 'when the database is not healthy' do
+      let(:mocked_connection) { double(:active_record_connection) }
```
adrianclay: There is a mantra of "[Don't mock what you don't own](https://8thlight.com/blog/eric-smith/2011/10/27/thats-not-yours.html)" which applies in this case.

In this example, if the rails connection class broke the `active?` method during an upgrade, your test suite wouldn't notice.

I'd recommend investigating if using the `disconnect!` method would allow you to test the behaviour without mocking out the `active?` method.

docker/wait_for_pg
```diff
@@ -0,0 +1,40 @@
...
+
+export PGPASSWORD=${POSTGRES_PASSWORD-postgres}
+
+PROBE_TIMEOUT=${PROBE_TIMEOUT-15}
+POSTGRES_HOST=${POSTGRES_HOST-postgres}
+POSTGRES_PORT=${POSTGRES_PORT-5432}
+POSTGRES_USER=${POSTGRES_USER-postgres}
+
+log() {
+  echo "[wait-for] [`date +'%Y%m%d%H%M%S'`] $@"
```
bleach: Can we use `$( .. )` instead of backticks?

(doesn't block merging)

docker/wait_for_pg
```diff
@@ -0,0 +1,40 @@
...
+}
+
+wait_for() {
```
bleach: This function seems meaty enough that we'd want a test.

helm_deploy/apply-for-legalaid-api/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+  #  memory: 128Mi
+
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+deploy:
+  host: ""
+  secretKey: "EdocGMybHpWR2hsVTJWamNtVjBTMlY1VG05MGFHbHVaMVJ2VTJWbFNHVnlaUzQ9"
```
bleach: Is this being used?

helm_deploy/apply-for-legalaid-api/values.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+postgresql:
+  imageTag: "10.5"
+  postgresUser: "postgres"
+  postgresPassword: "hdjhfsjkdfgsghfjgshdfgsdfkhkjdf"
```
bleach: I'm assuming that this is going to go away in the upcoming `git-crypt` work.

docker/run
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -e
+
+./bin/wait_for_pg
+./bin/setup_db
```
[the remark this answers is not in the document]
a5rar: the number of app containers is 2 the way deployments work is it starts up containers 1 at a time so there is no concurrent startup, so if the migration has already run the second container will detect its already run.  Migration run record is recorded in the database so i don't think they will run again.

app/services/healthcheck.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def call
+    { database: database_alive? }
+  end
+
+  private
+
+  def database_alive?
+    ActiveRecord::Base.connection.active?
```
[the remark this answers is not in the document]
a5rar: Is this really a blocker, we can make these amendments down the line if we find it's an issue

helm_deploy/apply-for-legalaid-api/values.yaml
lostie: Going forward, this file won't be tracked anymore, and we will track one for each environment (which will be encrypted). 

This is already addressed in the work that follows this where we setup the helm deployment through the CircleCI pipeline.

a5rar: for this PR yes but it is changing  in the next

helm_deploy/apply-for-legalaid-api/values.yaml
a5rar: yes

app/services/healthcheck.rb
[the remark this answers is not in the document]
lostie: For simplicity I'm going to keep the status endpoint as it was before removing the check for the database.

lostie:
@bleach here's a summary of how I've addressed your bullet points:

> * move db health out of liveness check (if we want it to be in readiness check that's 👌 )

I've removed the database healthcheck from the status endpoint preserving the existent functionality

> * double check that it's safe to keep the db migration in the startup script for now (assumption: only one app container will run)

It's not safe, and it's something we might need to address going forward, I don't think it's necessarily a problem just now. I'm happy to create an issue to address this later on.

> * test on the shell that waits for postgres

I've tested getting the following output when the postgresql is not available:

```
[wait-for] [20180829130759] wait 'apply-for-legalaid-api-postgresql':'5432' up to '15'
[wait-for] [20180829130759] wait attempt 'apply-for-legalaid-api-postgresql:5432' [1]
[wait-for] [20180829130800] wait attempt 'apply-for-legalaid-api-postgresql:5432' [2]
[wait-for] [20180829130801] wait attempt 'apply-for-legalaid-api-postgresql:5432' [3]
[wait-for] [20180829130802] wait attempt 'apply-for-legalaid-api-postgresql:5432' [4]
[wait-for] [20180829130803] wait attempt 'apply-for-legalaid-api-postgresql:5432' [5]
[wait-for] [20180829130804] wait attempt 'apply-for-legalaid-api-postgresql:5432' [6]
[wait-for] [20180829130805] wait attempt 'apply-for-legalaid-api-postgresql:5432' [7]
[wait-for] [20180829130806] wait attempt 'apply-for-legalaid-api-postgresql:5432' [8]
[wait-for] [20180829130807] wait attempt 'apply-for-legalaid-api-postgresql:5432' [9]
[wait-for] [20180829130808] wait attempt 'apply-for-legalaid-api-postgresql:5432' [10]
[wait-for] [20180829130809] wait attempt 'apply-for-legalaid-api-postgresql:5432' [11]
[wait-for] [20180829130810] wait attempt 'apply-for-legalaid-api-postgresql:5432' [12]
[wait-for] [20180829130811] wait attempt 'apply-for-legalaid-api-postgresql:5432' [13]
[wait-for] [20180829130812] wait attempt 'apply-for-legalaid-api-postgresql:5432' [14]
[wait-for] [20180829130813] wait attempt 'apply-for-legalaid-api-postgresql:5432' [15]
[wait-for] [20180829130814] [ERROR] wait timeout of '15' on 'apply-for-legalaid-api-postgresql:5432'
```

bleach:
Thanks!

Agreed offline that we will merge as-is and separately I will:
- produce a branch demonstrating how to test shell
- create a story for moving the db setup out of app startup

docker/wait_for_pg
sldblog: 🙋‍♂️ https://pkgs.alpinelinux.org/package/edge/testing/x86_64/dockerize is something that has this functionality out of the box if required.

PR #15, opened 2018-08-28
Add endpoint to return an application

mpw5:
The Citizen application needs to be able to retrieve application and
applicant data from the database based on application reference.

This adds an endpoint to do this: /v1/applications/:id

* Add show action to applications resource in routes.rb.
* Define show method in ApplicationsController.rb.
* Add test to requests/legal_aid_application_spec.rb.
* Install webmock and faker gems to support testing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,8 @@ gem 'bootsnap', '>= 1.1.0', require: false
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
+  gem 'faker', '>=1.9.1'
+  gem 'json_expressions'
   gem 'rspec-rails', '~> 3.7'
   gem 'rubocop', require: false
 end
--- a/app/controllers/v1/applications_controller.rb
+++ b/app/controllers/v1/applications_controller.rb
@@ -1,15 +1,17 @@
 class V1::ApplicationsController < ApplicationController
   def create
-    legalaid_application = LegalAidApplication.new
+    application = LegalAidApplication.new
 
-    if legalaid_application.save
-      # TODO: figure out why this render function doesnt automatically use  the serializer.
-      render json: LegalAidApplicationSerializer.new(legalaid_application).serialized_json, status: :created, serializer: LegalAidApplicationSerializer
+    if application.save
+      render json: LegalAidApplicationSerializer.new(application).serialized_json, status: :created
     else
-      # TODO: probably a better way to do this error return something like a global errrors handler if save fails
-      render json: { status: 'ERROR', message: 'Failed to create application', data: legalaid_application.errors }, status: :bad_request
+      render json: { status: 'ERROR', message: 'Failed to create application', data: application.errors }, status: :bad_request
     end
+  end
 
-    # respond_with @legalaid_application
+  def show
+    application = LegalAidApplication.find_by!(application_ref: params[:id])
+    options = { include: [:applicant] }
+    render json: LegalAidApplicationSerializer.new(application, options).serialized_json
   end
 end
--- a/spec/requests/legalaid_application_spec.rb
+++ b/spec/requests/legalaid_application_spec.rb
@@ -1,4 +1,5 @@
 require 'rails_helper'
+require 'json_expressions/rspec'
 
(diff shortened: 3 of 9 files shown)


spec/requests/legalaid_application_spec.rb
```diff
@@ -10,4 +13,27 @@
...
+          "date_of_birth": applicant_date_of_birth,
+          "created_at": '2018-08-26T22:47:06.093Z',
+          "updated_at": '2018-08-26T22:47:06.093Z',
+          "application_ref": application_ref,
+          "applicant_id": 2
+        }].to_json)
+    end
+
+    it 'returns a complete application' do
+      get '/v1/applications/' + application_ref
```
adrianclay: I’m guessing you’re wanting to test the logic in `application_controller.rb`. You shouldn’t need the `stub_request`. 

`stub_request` is used to mock out HTTP requests. Useful if your production code is making calls to an API. E.g. if we needed to make an API request to a DWP service.

You would need to have an Applcant and LegalAidApplication record in place to drive your current production code.

app/controllers/v1/applications_controller.rb
```diff
@@ -12,4 +12,10 @@ def create
 
     # respond_with @legalaid_application
   end
+
+  def show
+    render json: Applicant.joins(:legal_aid_application)
```
adrianclay: If you’ve got an application GET I’d expect it to select the Application and left join an Applicant.  This is assuming the applicant is assigned after the Application is created.

I might also expect the applicant to be a object which is returned on an application object. E.g.

```
{
  id: 2,
  applicant: {
    name: “Adrian”
  }
}
```

Gemfile
```diff
@@ -33,8 +33,10 @@ gem 'bootsnap', '>= 1.1.0', require: false
 group :development, :test do
   # Call 'byebug' anywhere in the code to stop execution and get a debugger console
   gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
+  gem 'faker', '>=1.9.1'
   gem 'rspec-rails', '~> 3.7'
   gem 'rubocop', require: false
+  gem 'webmock'
```
adrianclay: As my comment below, I doubt you’ll need `webmock` as part of this PR.

spec/requests/legalaid_application_spec.rb
```diff
@@ -8,6 +9,53 @@
...
+
+  describe 'GET /v1/applications/:id' do
+    let(:applicant_name) { Faker::Name.name }
+    let(:applicant_date_of_birth) { Faker::Date.birthday(18, 100) }
+    let(:applicant) { Applicant.create(name: applicant_name, date_of_birth: applicant_date_of_birth) }
+    let(:legal_aid_application) { LegalAidApplication.create(applicant: applicant) }
+    let(:application_ref) { legal_aid_application.application_ref }
+
+    context 'with a non-existant application reference' do
+      let(:application_ref) { SecureRandom.uuid }
```
adrianclay: Before you read below, I'd like to highlight I'm not recommending any change here.  I am using this opportunity to highlight the dangers of flaky tests 👻.

A test which gives false positives, i.e. it randomly fails when there isn't a defect with the production code, causes the devs to lose confidence in the test.  A test which exhibits this behaviour is described as flaky.  Having flaky tests is bad. 

Symptoms of this can include your CI build fails when it was passing locally, re-running the same build on CI causes the build to pass again.

Whenever I see a dynamic expression being generated and used in a test I make sure to confirm that using a dynamic value is:

A) More valuable than using a constant value.
B) Won't create a situation where the test wouldn't randomly fail.

In this situation, the test _could_ fail if the `application_ref` was valid.  An extremely unlikely situation given the nature of UUIDs.

Use of `Faker` above also introduces randomness.  Imagine an absurd situation where a new business rule was introduced which said applicants with the name "Adrian" gave a validation error, this code would become flaky as a result.

Developers should prevent flaky tests from being introduced.  If at a later date flakiness is observed with a test, the fixing of this should be prioritised with high importance.  The longer the flakiness lives in a test suite, the more time will be wasted on re-running tests.

Reading material on flaky tests:
https://testing.googleblog.com/2016/05/flaky-tests-at-google-and-how-we.html
https://blogs.dropbox.com/tech/2018/05/how-were-winning-the-battle-against-flaky-tests/

PR #16, opened 2018-08-30
Feature/app-98 Added Staging and production deployment to pipeline

a5rar:
As part of the circle ci build we can now deploy to staging and production.

This change also takes advantage of git crypt to hide secrets so when we deploy to staging and production we are not exposing passwords etc. 

Worth noting that in order to allow others to test the functionality i have left a commit inside which allows deployment from the branch, this will need to be removed before we merge and going forward only master branch can be deployed to staging and production


adrianclay:
If your automated tools are failing, it means the branch requires some more work before it can be reviewed. Before creating a PR & asking for a review you should make sure that the automated checks are passing.

Asking for a review should be done once you are confident there aren't any more improvements you should be making for the branch.

spec/requests/status_spec.rb
```diff
@@ -7,17 +7,7 @@
     it 'returns a status message' do
       get('/status')
       expect(response).to be_successful
-<<<<<<< HEAD
```
adrianclay: Ideally the marker changes should be squished into commit d9516f1429ad0ec84b5e64a903d485b8b6038d97.

The README change feels separate to the marker fixes, might make sense being it's own commit with a relevant message.

"Manually fixed rebase" doesn't feel like a good commit message for somebody trying to understand the documentation change.

config/routes.rb
```diff
@@ -2,7 +2,6 @@
   resources :status, only: [:index]
 
   namespace 'v1' do
-    resources :status, only: [:index]
```
lostie: 👍 (I forgot to remove this in a previous PR)

PR #18, opened 2018-09-04
Ap 122

ujoshi9999:
This task is about an association between application and proceeding type codes when an application is created. commits from this branch has been cherry-picked to - https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/19


README.md
```diff
@@ -49,6 +49,12 @@ Benefit of this is you dont have to build container everytime you make a change
 ## Developer local Endpoints
 
 * Post an application 
+ ##### you can post with or without proceeding types, example
```
lostie: **[not required for this PR]** As spoken offline, we should probably introduce https://swagger.io/docs/ to the app so that we don't have to manually update the README every single time a new endpoint is created/changed.

I'll create a github issue/story in JIRA to try and address that.

app/controllers/v1/applications_controller.rb
```diff
@@ -1,8 +1,14 @@
 class V1::ApplicationsController < ApplicationController
   def create
     application = LegalAidApplication.new
+    proceeding_types_found = true
 
-    if application.save
+    if params[:proceeding_type_codes]
+      proceeding_types_found = process_proceeding_types(application)
+      application.proceeding_types << ProceedingType.where(code: params[:proceeding_type_codes])
```
lostie: Between this line and https://github.com/ministryofjustice/laa-apply-for-legalaid-api/blob/5fdc0d560faa4d43337526dc0e8bdc7b28fb61d1/app/controllers/v1/applications_controller.rb#L25 they're executing exactly the same code, feels like this shouldn't be necessary.

Can't this just be simplified to something like the following? (just a suggestion):

```ruby
  proceeding_types = ProceedingType.where(code: params[:proceeding_type_codes])
  all_match = proceeding_types.size == params[:proceeding_type_codes].size
  application.proceeding_types = proceeding_types if all_match

  if application.save
  ...
  end
```

app/controllers/v1/applications_controller.rb
```diff
@@ -14,4 +20,9 @@ def show
     options = { include: [:applicant] }
     render json: LegalAidApplicationSerializer.new(application, options).serialized_json
   end
+
+  def process_proceeding_types(_application)
```
lostie: If you address the [previous comment](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/18/files#r214990391), this might not be necessary at all, but if you're not using `application` in the method it shouldn't be passed as an argument at all.

spec/models/application_proceeding_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+require 'spec_helper'
```
lostie: ✏️ The `rails_helper` already calls the `spec_helper` so this require seems unnecessary.

spec/models/application_proceeding_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+RSpec.describe ApplicationProceedingType, type: :model do
+  subject { described_class.new }
+
+  let!(:applicationProceedingType) { ApplicationProceedingType.create }
```
lostie: please lets not Javify our Ruby code 😅 , the `let` variables should be defined like any normal ruby variable, not camelcase, but instead with underscores 👍 .

spec/models/application_proceeding_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  it 'should belong to an proceeding_type and legal_aid_application' do
+    expect(subject.legal_aid_application_id).to eq(applicationProceedingType.id)
```
lostie: 😕 Are we sure this test is doing something useful?

```ApplicationProceedingType.create``` 

seems to just be creating an empty record so this assertion:

```ruby
expect(subject.legal_aid_application_id).to eq(applicationProceedingType.id)
```

is probably just passing because both sides of the expectation are `nil`

what you probably would like to assert IMO is:

* create a legal aid application record
* create a proceeding type record
* create an application proceeding type record that takes as arguments the two models created in the previous steps
* now you can make the assertion where you check that 

```ruby
expect(application_proceeding_type.legal_aid_application).to eq(legal_aid_application)
expect(application_proceeding_type.proceeding_type).to eq(proceeding_type)
```

spec/models/legalaid_application_spec.rb
```diff
@@ -16,4 +16,11 @@
   it 'should belong to an applicant' do
     expect(subject.applicant_id).to eq(applicant.id)
   end
+
+  describe 'Associations' do
+    it 'has many proceeding_types' do
```
lostie: You should probably consider to use [this](https://github.com/thoughtbot/shoulda-matchers/blob/master/lib/shoulda/matchers/active_record/association_matcher.rb) instead to do these sort of relation assertions.

spec/requests/legalaid_application_spec.rb
```diff
@@ -59,3 +50,44 @@
...
+      expect(response_json['data']['attributes']['application_ref']).not_to be_empty
+      expect(response.status).to eql(201)
+      expect(response.content_type).to eql('application/json')
+    end
+
+    context 'when proceedings are saved along with application reference' do
+      let(:proceeding_type_first) { ProceedingType.create(code: 'PR0001', ccms_code: 'PBM23', meaning: 'test meaning1', description: 'test desc1') }
+      let(:proceeding_type_second) { ProceedingType.create(code: 'PR0002', ccms_code: 'PBM24', meaning: 'test meaning2', description: 'test desc2') }
+
+      it 'creates a new legal aid application without proceeding_type_codes ' do
```
lostie: **[not a blocker]** ideally these descriptions should be fully human readable, so we should avoid using variable names in it and just use its human readable form.

spec/requests/legalaid_application_spec.rb
```diff
@@ -59,3 +50,44 @@
...
+        expect(LegalAidApplication.find_by(application_ref: application_ref).proceeding_types.size).to eq(1)
+      end
+
+      it 'creates a new legal aid application with more than one proceeding_type_codes' do
+        post '/v1/applications', params: { proceeding_type_codes: [proceeding_type_first.code, proceeding_type_second.code] }
+        application_ref = response_json['data']['attributes']['application_ref']
+        expect(LegalAidApplication.find_by(application_ref: application_ref).proceeding_types.size).to eq(2)
+      end
+
+      it 'creates a new legal aid application when proceeding_type_codes param is not valid, response should be bad data ' do
```
lostie: This description doesn't seem to make sense, the response status if `400` which normally something wrong happened, but the description states a new legal application was still created? 😕

spec/requests/legalaid_application_spec.rb
```diff
@@ -59,3 +50,44 @@
...
+        post '/v1/applications', params: { proceeding_type_codes: ['invalid_proceeding_type_code'] }
+        expect(response.status).to eql(400)
```
lostie: Also, for error cases we should probably be asserting on the format of the error JSON payload returned to the client.

app/controllers/v1/applications_controller.rb
ujoshi9999: done

app/controllers/v1/applications_controller.rb
ujoshi9999: done

ujoshi9999: done

spec/requests/legalaid_application_spec.rb
ujoshi9999: I am going to change this to- "response should be bad data when proceeding_type_codes param is not valid "

spec/requests/legalaid_application_spec.rb
ujoshi9999: done

spec/requests/legalaid_application_spec.rb
```diff
@@ -59,3 +50,44 @@
...
+      expect(response.status).to eql(201)
+      expect(response.content_type).to eql('application/json')
+    end
+
+    context 'when proceedings are saved along with application reference' do
+      let(:proceeding_type_first) { ProceedingType.create(code: 'PR0001', ccms_code: 'PBM23', meaning: 'test meaning1', description: 'test desc1') }
+      let(:proceeding_type_second) { ProceedingType.create(code: 'PR0002', ccms_code: 'PBM24', meaning: 'test meaning2', description: 'test desc2') }
+
+      it 'creates a new legal aid application without proceeding_type_codes ' do
+        expect do
```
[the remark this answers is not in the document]
ujoshi9999: RuboCop is a pain, let's have a chat about it.

spec/requests/legalaid_application_spec.rb
ujoshi9999: done

spec/requests/legalaid_application_spec.rb
ujoshi9999: let's leave it for now.

app/controllers/v1/applications_controller.rb
ujoshi9999: I realised that RuboCop has issues -
app/controllers/v1/applications_controller.rb:2:3: C: Metrics/AbcSize: Assignment Branch Condition size for create is too high. [18.17/18]

Gemfile.lock
```diff
@@ -148,6 +148,12 @@ GEM
       unicode-display_width (~> 1.0, >= 1.0.1)
     ruby-progressbar (1.10.0)
     ruby_dep (1.5.0)
+    shoulda (3.6.0)
```
bleach: This and some of the following lines, get removed when I run `bundle install` on this branch. It looks like you had these gems in your `Gemfile` and committed the resulting `Gemfile.lock`

Please re-run `bundle install` and commit the result.

README.md
```diff
@@ -3,17 +3,17 @@
 This is the service api for persisting application related information to the back end database and
 may well be used to fire requests to other services.
 
-* Ruby version
+* Ruby version
```
bleach: As discussed face to face, there are still whitespace changes in this branch that should not be there.

spec/models/application_proceeding_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe ApplicationProceeding, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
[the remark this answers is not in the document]
ujoshi9999: I would recommend doing a diff against the master, that would give better picture this test does not exist it has been renamed to "ApplicationProceedingType"

spec/requests/legalaid_application_spec.rb
```diff
@@ -59,3 +50,51 @@
...
+          post '/v1/applications'
+        end.to change { LegalAidApplication.count }.by(1)
+      end
+
+      it 'creates a new legal aid application with proceeding_type_codes' do
+        post '/v1/applications', params: { proceeding_type_codes: [proceeding_type_first.code] }
+        application_ref = response_json['data']['attributes']['application_ref']
+        expect(LegalAidApplication.find_by(application_ref: application_ref).proceeding_types.size).to eq(1)
```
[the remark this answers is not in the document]
ujoshi9999: I will try this when we pair for a couple of minutes later maybe on Monday.

lostie:
Closing in favour of https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/19 which replaces this.

PR #19, opened 2018-09-06
Ap 122 Consume proceeding type data in the Applications API endpoint

ujoshi9999:
This request is about enhancement in the application resource creation. With this implementation, if correct proceeding_type_codes are supplied in the request then the application will be created with proceeding type codes(an association between the application and proceeding types.)
for more details - https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-122


app/controllers/v1/applications_controller.rb
```diff
@@ -2,6 +2,11 @@ class V1::ApplicationsController < ApplicationController
   def create
     application = LegalAidApplication.new
 
+    if params[:proceeding_type_codes]
```
bjpirt: I suspect the changes in this file might be better placed in the model where a validation can run if any of the passed in IDs do not already exist - does anyone else have an opinion on this?

ujoshi9999:
Merging to master as pull request has been approved.

PR #22, opened 2018-09-12
Feature/ap 199- GET end point to pull proceeding types from API

ujoshi9999:
This is the first part of the story where I have provided a GET endpoint on API.
This will be utilised by the provider frontend rather than reading from json file.

https://dsdmoj.atlassian.net/browse/AP-119

### DIFF
--- a/app/controllers/proceeding_types_controller.rb
+++ b/app/controllers/proceeding_types_controller.rb
@@ -1,0 +1,6 @@
+class ProceedingTypesController < ApplicationController
+  def index
+    proceeding_types = ProceedingType.all
+    render json: ProceedingTypesSerializer.new(proceeding_types).serialized_json, status: :ok
+  end
+end
--- a/app/controllers/proceeding_types_controller.rb
+++ b/app/controllers/proceeding_types_controller.rb
@@ -1,6 +1,6 @@
 class ProceedingTypesController < ApplicationController
   def index
     proceeding_types = ProceedingType.all
-    render json: ProceedingTypesSerializer.new(proceeding_types).serialized_json, status: :ok
+    render json: ProceedingTypesSerializer.new(proceeding_types).serialized_json
   end
 end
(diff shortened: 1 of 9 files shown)


app/controllers/proceeding_types_controller.rb
```diff
@@ -0,0 +1,6 @@
+class ProceedingTypesController < ApplicationController
+  def index
+    proceeding_types = ProceedingType.all
+    render json: ProceedingTypesSerializer.new(proceeding_types).serialized_json, status: :ok
```
lostie: **[note/not a blocker]** More for spreading the knowledge than being required as a change, by default render json already returns a successful response status code unless we tell it otherwise, so `status: :ok` I believe can be optional in the cases we're returning a successful response.

config/routes.rb
```diff
@@ -1,5 +1,6 @@
 Rails.application.routes.draw do
   resources :status, only: [:index]
+  resources :proceeding_types
```
lostie: This generates all actions for proceeding types, if we're only expected to list them this should be restricted as such:

```ruby
resources :proceeding_types, only: [:index]
```

spec/requests/proceeding_types_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+                            ccms_category_law: 'Family',
+                            ccms_category_law_code: 'MAT',
+                            ccms_matter: 'Public Law - Family',
+                            ccms_matter_code: 'KPBLB',
+                            meaning: 'Application for a care order',
+                            description: 'to be represented on an application for a care order.')
+    end
+
+    it 'returns a proceeding types' do
+      get '/proceeding_types'
```
lostie: **[suggestion/not a blocker/does not enforce code changes]** 💭 In cases where we returning a list of values, I find it more _accurate_ to at least expected two element to be returned.

PR #23, opened 2018-09-13
Split client name into first_name and last_name

willc-work:
Client name has been split out into two separate fields for first and last name

[Link to story](https://dsdmoj.atlassian.net/browse/AP-128)

As it stands the benefit checker requires letters from the surname in order to complete a successful request, and it was proving difficult to extract this information from a single client name field.

## Checklist

- [x] There should be no uneccessary whitespace changes. These make diffs harder to read and conflicts more likely. 

### DIFF
--- a/spec/requests/create_applicant_spec.rb
+++ b/spec/requests/create_applicant_spec.rb
@@ -17,7 +17,8 @@ RSpec.describe 'Legal aid applications' do
       'data': {
         'type': 'legal_aid_applicant',
         'attributes': {
-          'name': 'John Doe',
+          'first_name': 'John',
+          'last_name': 'Doe',
           'date_of_birth': '1991-12-01',
           'application_ref': application.application_ref
         }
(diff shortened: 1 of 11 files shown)


spec/requests/create_applicant_spec.rb
```diff
@@ -30,8 +31,9 @@
       post '/v1/applicants', params: body, headers: headers
       expect(response.status).to eql(201)
       expect(response.content_type).to eql('application/json')
-      expect(response_json['data']['attributes']['name']).to eq('John Doe')
-      expect(response_json['data']['attributes']['date_of_birth']).to eq('1991-12-01')
+      expect(response_json.dig('data', 'attributes', 'first_name')).to eq('John')
```
lostie: **[not a blocker]** 💭 Consider [this approach](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/blob/master/spec/requests/legalaid_application_spec.rb#L47) as an alternative way of asserting the same content && schema without having to assert each field separately.

willc-work: is this possible on a POST request? as it is generating a new id that we dont know how to match on

lostie: Yes, if you look at the gem in [more detail](https://github.com/chancancode/json_expressions#usage) you will find some example where we can define a more dynamic way of expecting a given value we don't know its exact content, but we know it's a valid UUID.

PR #24, opened 2018-09-17
added /v1 in request path for proceeding types

ujoshi9999:
https://dsdmoj.atlassian.net/browse/AP-119

before proceeding types were retrieved with request path - /proceeding_types
with this change, it will be retrieved with request path - /v1/proceeding_types
and
removed the status endpoint which was hanging inside v1 namespace, it already exists outside.

this will fix the provider frontend rest call to API (for proceeding types)

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,9 +1,8 @@
 Rails.application.routes.draw do
   resources :status, only: [:index]
-  resources :proceeding_types, only: [:index]
 
   namespace 'v1' do
-    resources :status, only: [:index]
+    resources :proceeding_types, only: [:index]
     resources :applications, only: [:create, :show]
     resources :applicants
   end
--- a/app/controllers/proceeding_types_controller.rb
+++ b/app/controllers/proceeding_types_controller.rb
@@ -1,6 +1,0 @@
-class ProceedingTypesController < ApplicationController
-  def index
-    proceeding_types = ProceedingType.all
-    render json: ProceedingTypesSerializer.new(proceeding_types).serialized_json
-  end
-end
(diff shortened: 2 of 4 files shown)


config/routes.rb
```diff
@@ -1,9 +1,8 @@
 Rails.application.routes.draw do
   resources :status, only: [:index]
-  resources :proceeding_types, only: [:index]
 
   namespace 'v1' do
-    resources :status, only: [:index]
```
bleach: Is it intentional that you are removing `/v1/status` as well as adding `/v1/proceeding_types`?

DamanFergus: Not sure why you deleted this. Wont status only be able to be accessed from the base url now?

ujoshi9999: status is not related to versioning which is / v1, it should be outside on root level.
why do you think status should be like / v1/status?

adrianclay:
This highlights the perils of introducing 2 logical changes in 1 PR.  One change (the removal of /v1/status/) can hold up the deployment of the other.

config/routes.rb
ujoshi9999: I had a chat with Sergio, load balance will call /status not /v1/status, so I removed it.
proceeding types are part of API which is at the moment represented as /v1 , so makes sense to change it like /v1/proceeding_types

DamanFergus: Ok that sounds reasonable. Thanks for clarifying

ujoshi9999:
> This highlights the perils of introducing 2 logical changes in 1 PR. One change (the removal of /v1/status/) can hold up the deployment of the other.

as discussed with Sergio, status is a health check which has nothing to do with versioning of API, so I added that change as part of this PR ( trying to avoid another PR). If this is not correct process then I will not include this change as part of PR?

config/routes.rb
ujoshi9999: I think I would have done two commits as you suggested but I did one commit, tried to explain 2 changes in one commit message: "added /v1 in request path for proceeding types and removed status from v1 namespace because (it exist outside v1 namespace which is the correct location)"

PR #25, opened 2018-09-18
Added endpoint to update applicant

a5rar:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-8)

Added email address to the model and the applicant params so we can collect it as part of the 
applicant data.

## Checklist


app/controllers/v1/applications_controller.rb
```diff
@@ -15,7 +15,7 @@ def create
   end
 
   def show
-    application = LegalAidApplication.find_by!(application_ref: params[:id])
+    application = LegalAidApplication.find_by!(application_ref: params[:ref])
```
lostie: Is this a breaking change? Do both frontend apps need to be updated to supply the `:ref` instead of `:id`?

a5rar: were not passing application id at any point,  we were always using application ref, so i thought it's a good idea to make this consistent

lostie: ⚠️ Seems like [we're passing `:id` as the param in the citizen frontend](https://github.com/ministryofjustice/laa-apply-for-legal-aid-citizen-frontend/blob/master/app/controllers/citizen_controller.rb#L3) (regardless of the fact that ID is the application ref). I would say that citizen frontend doesn't currently work anymore due to these changes.

PR #27, opened 2018-10-02
Add National Insurance number to applicant

willc-work:
Add National Insurance number to applicant, validate and persist

[Link to story](https://dsdmoj.atlassian.net/browse/AP-9)

[Linked to provider PR](https://github.com/ministryofjustice/laa-apply-for-legal-aid-provider-frontend/pull/55)
Add NINO to applicant, add validation and tests
Add database migration to update applicants table

- [ ] There should be no unneccessary whitespace changes. These make diffs harder to read and conflicts more likely. 

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -5,7 +5,7 @@ class Applicant < ApplicationRecord
 
   validates :first_name, :last_name, :date_of_birth, :national_insurance_number, presence: true
 
-  validate :validate_date_of_birth
+  validate :validate_date_of_birth, :validate_national_insurance_number
 
   validates :email_address, format: { with: URI::MailTo::EMAIL_REGEXP }, allow_nil: true
 
@@ -13,5 +13,18 @@ class Applicant < ApplicationRecord
 
   def validate_date_of_birth
     errors.add(:date_of_birth, 'is not valid') if date_of_birth.present? && (date_of_birth < Date.new(1900, 0o1, 0o1) || date_of_birth > Date.today)
+  end
+
+  def normalise_national_insurance_number
+    return unless national_insurance_number
+    national_insurance_number.delete!(' ')
+    national_insurance_number.upcase!
+  end
+
+  def validate_national_insurance_number
+    return if national_insurance_number.blank?
+    national_insurance_number_regex = /^[A-CEGHJ-PR-TW-Z]{1}[A-CEGHJ-NPR-TW-Z]{1}[0-9]{6}[A-DFM]{1}$/
+    normalise_national_insurance_number
+    errors.add(:national_insurance_number, :invalid) if (national_insurance_number_regex =~ national_insurance_number).nil?
   end
 end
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -10,18 +10,31 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_09_17_084624) do
+ActiveRecord::Schema.define(version: 2018_09_29_123828) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "addresses", force: :cascade do |t|
```
LAACharlesBoardman: Is this related to AP-133?
Adding client address.
Should be included with this PR.

spec/models/applicant_spec.rb
```diff
@@ -52,8 +53,19 @@
...
 
-  it 'is valid  when email address is not provided' do
+  it 'is valid when email address is not provided' do
     subject.email_address = nil
     expect(subject).to be_valid
   end
+
+  it 'is not valid when a national insurance number is not provided' do
+    subject.national_insurance_number = nil
+    expect(subject).to_not be_valid
```
LAACharlesBoardman: Add include clause as to match the next test.
Without it this test may pass if the model is invalid for another reason.

spec/requests/applicant_spec.rb
```diff
@@ -97,7 +103,9 @@
             "first_name": existing_applicant.first_name,
             "last_name": existing_applicant.last_name,
             "email_address": 'test@test.com',
+            "national_insurance_number": existing_applicant.national_insurance_number,
             "date_of_birth": existing_applicant.date_of_birth.to_s
+
```
LAACharlesBoardman: Blank line

app/models/applicant.rb
```diff
@@ -14,4 +14,17 @@ class Applicant < ApplicationRecord
   def validate_date_of_birth
     errors.add(:date_of_birth, 'is not valid') if date_of_birth.present? && (date_of_birth < Date.new(1900, 0o1, 0o1) || date_of_birth > Date.today)
   end
+
+  def normalise_national_insurance_number
```
lostie: **[not a blocker]** I would normally normalise the NINo on initialisation of the object if the value was present. It took me a little bit to realise we're effectively changing the value during the validation itself.

db/schema.rb
lostie: Seems like this needs to be ✂️ as it is not part of this PR like Charlie mentioned.

db/schema.rb
```diff
@@ -55,4 +68,5 @@
     t.index ["code"], name: "index_proceeding_types_on_code"
   end
 
+  add_foreign_key "addresses", "applicants"
```
lostie: ✂️  Ditto

spec/models/applicant_spec.rb
lostie: ☝️ What Charlie said.

PR #28, opened 2018-10-03
[AP-141] API responses consistency

lostie:
## What

**DO NOT MERGE**: Needs rebasing and needs changes in the client gem and the frontend apps to match this API changes.

[AP-141 - Uniformization of the API request/response payload contracts](https://dsdmoj.atlassian.net/browse/AP-141)

This PR refactors all the existent API endpoints to consume/return flat JSON payloads instead of the existent mix of JSONAPI/non-JSONAPI ones. The idea behind this is to make the payloads simpler and uniform across the app.

* Changes the endpoint to create an **applicant** to be under 

```
POST /v1/applications/:application_id/applicant
```

instead of 

```
POST /v1/applicant`
```

as the original endpoint already depended on the `application_ref/application_id`.

* Changes the identifier required to be passed/return to/by the applications endpoint to be `id` instead of `application_ref` as to the outside world that's what the field is. Internally, for now, I've decided to keep the field with the same name

* Renders 404 responses whenever the endpoint path is provided with some data that does not match an existent resource, e.g:

```
PATCH /v1/applications/:application_id/applicant
```

will return a 404 response when either there's no application for the provided `:application_id` or there's no applicant associated with the application to be _patched_.

* For validation error responses (400s), an error serializer was introduced to format the errors in a way that the client can determine which fields have validation errors and what sort of errors are. Before this change we were just returning error messages leaving the client with no way to determine what field or exactly what kind of validation error happened.

Also:

* Introduces [FactoryBot](https://github.com/thoughtbot/factory_bot) to make it easier to manage _factories_ for the resources we need to setup in the test suite.
* Refactored some of the request specs structure so each of the request actions is tested in isolation and the files themselves become more readable.

## Checklist


app/serializers/applicant_serializer.rb
```diff
@@ -1,5 +1,3 @@
-class ApplicantSerializer
-  include FastJsonapi::ObjectSerializer
-  attribute :first_name, :last_name, :email_address
-  attribute :date_of_birth
+class ApplicantSerializer < ActiveModel::Serializer
+  attributes :first_name, :last_name, :email_address, :date_of_birth
```
mpw5: Should National Insurance number be included here now too? (Just a reminder - I'm aware that you haven't rebased so might not have those changes yet.)

PR #30, opened 2018-10-16
Add address table and endpoint to handle addresses

willc-work:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-153)

AP153 requires that an address can be entered which is associated with an applicant, an applicant can have more than one address. A new table has been created which references the applicants table and validation applied in line with the design requirements.

## Checklist

### DIFF
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -5,20 +5,21 @@ class Address < ApplicationRecord
   # of the validation process we are changing postcode to uppercase and removing spaces this allows
   # for some simplification of the regular expression
 
-  POSTCODE_REGEXP = /^([A-Z][A-HJ-Y]?[0-9][A-Z0-9]? ?[0-9][A-Z]{2}|GIR ?0A{2})$/
+  POSTCODE_REGEXP = /([A-Z][A-HJ-Y]?[0-9][A-Z0-9]? ?[0-9][A-Z]{2}|GIR ?0A{2})/
 
   belongs_to :applicant
 
   validates :address_line_one, :address_line_two, :city, :postcode, presence: true
 
-  validate :validate_postcode
+  before_validation :normalize_postcode
+
+  validates :postcode, format: { with: POSTCODE_REGEXP }
 
   private
 
-  def validate_postcode
+  def normalize_postcode
     return if postcode.blank?
     postcode.delete!(' ')
     postcode.upcase!
-    errors.add(:postcode, :invalid) unless POSTCODE_REGEXP =~ postcode
   end
 end
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -1,5 +1,4 @@
 class Address < ApplicationRecord
-
   # The REGEXP for the postcode validation is taken from this very detailed stack overflow answer:
   # https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive/51885364#51885364
   # The second REGEXP from the answer section of that comment was used because as part
@@ -10,10 +9,9 @@ class Address < ApplicationRecord
 
   belongs_to :applicant
 
-  validates :address_line_one, :city, :postcode, presence: true
+  validates :address_line_one, :address_line_two, :city, :postcode, presence: true
 
   validate :validate_postcode
-
 
   private
 
@@ -23,5 +21,4 @@ class Address < ApplicationRecord
     postcode.upcase!
     errors.add(:postcode, :invalid) unless POSTCODE_REGEXP =~ postcode
   end
-
 end
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -9,13 +9,19 @@ class Address < ApplicationRecord
 
   belongs_to :applicant
 
-  validates :address_line_one, :address_line_two, :city, :postcode, presence: true
+  validates :city, :postcode, presence: true
 
   before_validation :normalize_postcode
 
   validates :postcode, format: { with: POSTCODE_REGEXP }
+  validate :validate_address_lines
 
   private
+
+  def validate_address_lines
+    return if address_line_one.present? || address_line_two.present?
+    errors.add(:address_line_one, :blank)
+  end
 
   def normalize_postcode
     return if postcode.blank?
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -1,0 +1,27 @@
+class Address < ApplicationRecord
+
+  # The REGEXP for the postcode validation is taken from this very detailed stack overflow answer:
+  # https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive/51885364#51885364
+  # The second REGEXP from the answer section of that comment was used because as part
+  # of the validation process we are changing postcode to uppercase and removing spaces this allows
+  # for some simplification of the regular expression
+
+  POSTCODE_REGEXP = /^([A-Z][A-HJ-Y]?[0-9][A-Z0-9]? ?[0-9][A-Z]{2}|GIR ?0A{2})$/
+
+  belongs_to :applicant
+
+  validates :address_line_one, :city, :postcode, presence: true
+
+  validate :validate_postcode
+
+
+  private
+
+  def validate_postcode
+    return if postcode.blank?
+    postcode.delete!(' ')
+    postcode.upcase!
+    errors.add(:postcode, :invalid) unless POSTCODE_REGEXP =~ postcode
+  end
+
+end
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -5,7 +5,7 @@ class Address < ApplicationRecord
   # of the validation process we are changing postcode to uppercase and removing spaces this allows
   # for some simplification of the regular expression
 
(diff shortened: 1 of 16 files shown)


config/locales/en.yml
```diff
@@ -1,36 +1,7 @@
...
-# en:
-#   'true': 'foo'
-#
-# To learn more, please read the Rails Internationalization guide
-# available at http://guides.rubyonrails.org/i18n.html.
-
 en:
   activerecord:
+    postcode:
+      invalid: 'is not in the right format'
```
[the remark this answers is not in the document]
willc-work: I will remove this one as it is not required.

app/models/address.rb
```diff
@@ -0,0 +1,24 @@
...
+  validate :validate_postcode
+
+  private
+
+  def validate_postcode
+    return if postcode.blank?
+    postcode.delete!(' ')
+    postcode.upcase!
+    errors.add(:postcode, :invalid) unless POSTCODE_REGEXP =~ postcode
+  end
```
reggieb: The modification of postcode (removal of spaces and upcase) is a processing step and not something that's fundamental to the validation process. If it is done via a callback the postcode validation can then be done with a standard `validates` call.

I'd suggest doing this:

```ruby
before_validation :normalize_postcode

validates :postcode, format: { with: POSTCODE_REGEXP }

private

def normalize_postcode
  postcode.delete!(' ')
  postcode.upcase!
end

File: db/migrate/20181011143410_create_addresses.rb [hunk 3]
```diff
@@ -0,0 +1,13 @@
+class CreateAddresses < ActiveRecord::Migration[5.2]
+  def change
+    create_table :addresses do |t|
+      t.string :address_line_one, null: false
```
reggieb: Why is `address_line_one` set to `null: false`, but the other parameters that are validated in the model for presence (`address_line_two`, `city`, `postcode`) not set the same way. I'd suggest removing the `, null: false` from `address_line_one` and letting the model control this.

spec/requests/applicants/create_address_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+  end
+
+  it 'returns a successful address payload' do
+    post_request.call
+
+    expect(response).to have_http_status(201)
+    expect(response.content_type).to eql('application/json')
+    json = JSON.parse response.body
+    expect(json['address_line_one']).to eq params[:address_line_one]
```
lostie: **[not a blocker]** 💭 Consider using the `match_json_expression` instead so that you assert the content of the whole response payload, rather than partial matches (an example can be found [here](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/blob/81446e67b02b7e42692eb466b51b58bba97a463b/spec/requests/applicants/patch_applicant_spec.rb#L55-L65)).

spec/models/address_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  subject { build(:address) }
+
+  it 'is valid with all valid attributes' do
+    expect(subject).to be_valid
+  end
+
+  it 'is not valid without a building name' do
+    subject.address_line_one = ''
+    expect(subject).to_not be_valid
+    expect(subject.errors[:address_line_one]).to include("can't be blank")
```
reggieb: What are you testing here? That the rails method `validates` works?

spec/models/address_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    subject.postcode = nil
+    expect(subject).to_not be_valid
+    expect(subject.errors[:postcode]).to include("can't be blank")
+  end
+
+  it 'is not valid if the postcode entered is not in the correct format' do
+    subject.postcode = '1GIR00A'
+    expect(subject).to_not be_valid
+    expect(subject.errors[:postcode]).to include('is not in the right format')
+  end
```
reggieb: This is the main spec that is required here - as it's test non-standard (app specific) behaviour.

spec/requests/applicants/create_address_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      address_line_one: '123',
+      address_line_two: 'High Street',
+      city: 'London',
+      county: 'Greater London',
+      postcode: 'SW1H9AJ'
+    }
+  end
+  let(:post_request) do
+    -> { post "/v1/applicants/#{applicant_id}/addresses", params: params.to_json, headers: json_headers }
+  end
```
reggieb: I think this is over complicating things. If you change it to:
```ruby
let(:post_request) do
  post "/v1/applicants/#{applicant_id}/addresses", params: params.to_json, headers: json_headers
end
```
It can be called with `post_request`, rather than `post_request.call`

lostie: This is more to express explicitly that we're actually performing the request, rather than defining it, I don't mind either way.

db/migrate/20181011143410_create_addresses.rb
willc-work: postcode did have this as well but I have remove it from both now

spec/models/address_spec.rb
lostie: If you don't have any tests in place, you don't really know if or which validations are being performed. Where as it they can be done some other way, that's another question, but I believe these are still valid tests.

spec/requests/applicants/create_address_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  context 'when there is no applicant with the provided reference' do
+    let(:applicant_id) { SecureRandom.uuid }
```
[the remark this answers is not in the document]
willc-work: added valid invalid ID

spec/requests/applicants/create_address_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect(response.body).to match_json_expression(expected_json)
+    end
+  end
+
+  it 'returns a successful address payload' do
+    post_request.call
+
+    expect(response).to have_http_status(201)
+    expect(response.content_type).to eql('application/json')
+    json = JSON.parse response.body
```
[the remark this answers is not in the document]
willc-work: removed unnecessary JSON.parse

spec/models/address_spec.rb
reggieb: My main point, is that if you start adding specs that just test rails functionality, and not the app specific logic, you end up with a lot of unnecessary specs and they just act as noise to make it more difficult to identify the important specs.

If you want to know which validations are applied to a model - go and look in the model code.

In general - if a validation is standard out-of-the-box rails, testing it is testing rails and not your code. If you create a custom validator (for example with `validate :my_custom_validator`) that needs testing.

The same with joins. `has_many :foos` doesn't need testing. Testing it just tests how `has_many` works and is therefore unnecessary. However, `has_many :bars, -> { where('bars.names LIKE "%thimble"') }` that's got app specific logic and needs testing.

db/schema.rb
```diff
@@ -10,11 +10,23 @@
...
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_09_29_123828) do
+ActiveRecord::Schema.define(version: 2018_10_11_143410) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "addresses", force: :cascade do |t|
+    t.string "address_line_one", null: false
```
lostie: ⚠️ Seems like the schema is not up to date with the migration. Ensure the newly generated schema.rb is also pushed to the remote.

willc-work: pushed changed schema

spec/models/address_spec.rb
willc-work: 👍

app/models/address.rb
```diff
@@ -0,0 +1,31 @@
+class Address < ApplicationRecord
+  # The REGEXP for the postcode validation is taken from this very detailed stack overflow answer:
+  # https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive/51885364#51885364
+  # The second REGEXP from the answer section of that comment was used because as part
+  # of the validation process we are changing postcode to uppercase and removing spaces this allows
+  # for some simplification of the regular expression
+
+  POSTCODE_REGEXP = /([A-Z][A-HJ-Y]?[0-9][A-Z0-9]? ?[0-9][A-Z]{2}|GIR ?0A{2})/
```
lostie: The regexp should enforce that the postcode **starts** (`\A`) and **ends** (`\z`) with the defined pattern. As it stands, something like `bla bla SW1H 9AJ bla bla`:

screen shot 2018-10-17 at 09 24 35

whereas the restricted version does match it correctly:

screen shot 2018-10-17 at 09 26 59

screen shot 2018-10-17 at 09 26 39

lostie: You can find more info on this [here](https://ruby-doc.org/core-2.5.1/Regexp.html#class-Regexp-label-Anchors).

PR #31, opened 2018-10-17
Demo: Moving Provider Front end functionality into core API application

reggieb:
This is a proof of concept. It shows how namespacing can be used to bring front end functionality into the API application in a structured way.

- Provider controllers are in `app/controllers/provider`
- Provider views are in `app/views/provider`

This imported work flow works up to the creation of the Applicant.

A few things to note:

- I've been lazy with specs - so only a couple of very basic request specs are included.
- Note that I have used `laa` to replaces `legal_aid_application` in the urls.
- The way `ApplicationController` was overridden so that it defaults to API only caused a couple of issues. I reworked it so that there is now an `ApiController` that provides this functionality only where it is needed.


willc-work:
Just a note that ''laa'' is also the acronym for Legal Aid Agency, i dont think it matters, just wanted to highlight it

reggieb:
The latest update to this PR adds a second copy of the applicant new form in [app/views/provider/applicants/new.html.erb](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/31/files#diff-c3815c7c4b895f34d7cd5aec341be52d). This allow comparison of the existing technique used to render forms, with one based on [simple_form](https://github.com/plataformatec/simple_form). The modification to simple_form required to get the styling to closely match the GOV.UK designs can be found in [config/initializers/simple_form.rb](https://github.com/ministryofjustice/laa-apply-for-legalaid-api/pull/31/files#diff-754d908a36f8f0af1316fe4b5ecd2c9b)

PR #34, opened 2018-10-19
Migrate the database id's from sequential bigint to UUID

willc-work:
The tables currently use sequential bigint to set the id. This change modifies the behaviour to use uuid instead.

The pgcrypto gem sets uuid as the default for all future tables that are created.
The webdack gem converts existing id fields to uuid and can be removed once the migrations have been run
Add foreign keys to existing references
Converts foreign keys on existing references to uuid
Convert all other id's (that are not foreign keys) to uuid

## Checklist

### DIFF
--- a/db/migrate/20181018165405_id_to_uuid_fk.rb
+++ b/db/migrate/20181018165405_id_to_uuid_fk.rb
@@ -1,0 +1,18 @@
+require 'webdack/uuid_migration/helpers'
+
+class IdToUuidFk < ActiveRecord::Migration[5.2]
+  def change
+    reversible do |dir|
+      dir.up do
+
+        primary_key_and_all_references_to_uuid :applicants
+        primary_key_and_all_references_to_uuid :legal_aid_applications
+        primary_key_and_all_references_to_uuid :proceeding_types
+      end
+
+      dir.down do
+        raise ActiveRecord::IrreversibleMigration
+      end
+    end
+  end
+end
--- a/db/migrate/20181018165405_id_to_uuid_fk.rb
+++ b/db/migrate/20181018165405_id_to_uuid_fk.rb
@@ -4,7 +4,6 @@ class IdToUuidFk < ActiveRecord::Migration[5.2]
   def change
     reversible do |dir|
       dir.up do
-
         primary_key_and_all_references_to_uuid :applicants
         primary_key_and_all_references_to_uuid :legal_aid_applications
         primary_key_and_all_references_to_uuid :proceeding_types
--- a/db/migrate/20181018165537_uuid_migration.rb
+++ b/db/migrate/20181018165537_uuid_migration.rb
@@ -1,0 +1,17 @@
+require 'webdack/uuid_migration/helpers'
+
+class UuidMigration < ActiveRecord::Migration[5.2]
+  def change
+    reversible do |dir|
+      dir.up do
+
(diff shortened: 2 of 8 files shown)


bleach:
It would be good to understand a bit more about why you're changing this. Neither the PR description nor the commit message on ca3e4cdb7592fd28119f9635886ed52de3eae85b make this very clear.

db/migrate/20181018165405_id_to_uuid_fk.rb
```diff
@@ -0,0 +1,17 @@
...
+        primary_key_and_all_references_to_uuid :applicants
+        primary_key_and_all_references_to_uuid :legal_aid_applications
+        primary_key_and_all_references_to_uuid :proceeding_types
+      end
+
+      dir.down do
+        raise ActiveRecord::IrreversibleMigration
+      end
+    end
+  end
```
reggieb: Is there any reason why you didn't just use[ `up` and `down` methods](https://edgeguides.rubyonrails.org/active_record_migrations.html#using-the-up-down-methods) instead of `change` to get differing up and down behaviours:
```ruby
class IdToUuidFk < ActiveRecord::Migration[5.2]
  def up
    primary_key_and_all_references_to_uuid :applicants
    primary_key_and_all_references_to_uuid :legal_aid_applications
    primary_key_and_all_references_to_uuid :proceeding_types
  end

  def down
    raise ActiveRecord::IrreversibleMigration
  end
end
```
The result is a lot clearer I would suggest.

db/migrate/20181018165537_uuid_migration.rb
```diff
@@ -0,0 +1,16 @@
+require 'webdack/uuid_migration/helpers'
+
+class UuidMigration < ActiveRecord::Migration[5.2]
+  def change
```
reggieb: You can use `up` and `down` here too.

db/migrate/20181018165405_id_to_uuid_fk.rb
willc-work: I thought that because down was not actually possible (throws an error) it didn't make sense to use it in this case. However if it is best practice to use up/down even when you cannot run down to change the db I am happy to change it.

willc-work: 👍

reggieb: `reversible` is useful if you only need that functionality for part of the migration. Using it for the whole migration isn't wrong - it just noisy.

lostie:
💁  The drive among other things, was to stop creating extra columns in the table that define a `UUID` identifier for the models, where with this setup by default, both the primary keys and the foreign keys would provide that out of the box. 

More so, by converting the sequential IDs to UUIDs, when they're being used in the web pages, they're unique and someone can't just replace the ID with its precedent number.

Not to claim this article is by any means the truth, but some more info about the subject can be found [here|https://medium.com/@Mareks_082/auto-increment-keys-vs-uuid-a74d81f7476a].

@bleach I'm happy to discuss this offline, if you have any concerns about it.

db/migrate/20181018165405_id_to_uuid_fk.rb
willc-work: yes I can see how much clearer it looks your way

PR #46, opened 2018-10-31
Granting access to bank accounts through true layer

reggieb:
## What

[JIRA AP-27](https://dsdmoj.atlassian.net/browse/AP-27)

- [x] Add devise to manage user authentication (required renaming `applicant.email_address` to `applicant.email`)
- [x] Add omniauth strategy to manage OAUTH2 authentication and integrate with Devise
- [x] Gather user data from True Layer using token received via authentication.
- [x] Identify applicant via user data
- [x] Redirect to success page on successful login in and applicant identification
- [x] Redirect to root with error message on failure.
- [x] Persist True Layer token and expiry time on Applicant model
- [x] Set authentication settings via ENV


reggieb:
Login failure due to issue at True Layer login causes redirect to root with this error:
true_layer_login_failed

If True Layer login is successful, but the return credential do not match an Applicant, the error message looks like this:
credential_do_not_match_error

Kalagan:
I'm a bit concerned about using Devise here. I wonder how it is going to behave when an applicant is going to need several true_layer tokens. And when we are going to implement provider authentication.
You are setting the token in the applicants table but I wonder if we should create a separate table.
In the meeting this morning you mentionned that Devise has a lot of out-of-the-box useful features.
But which ones do we need in that specific case?

app/services/true_layer/identify_applicant.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    def response
+      @response ||= conn.get '/data/v1/info'.freeze
+    end
+
+    def conn
+      @conn ||= Faraday.new(url: TRUE_LAYER_URL) do |conn|
+        conn.authorization :Bearer, token
+        conn.response :logger if Rails.env.development?
```
lostie: **[not a blocker/nit pick]** 💭 I would consider avoiding conditional code based on environment checks, and instead check some configuration that is setup to be enable for the environments you need, e.g.:

```ruby
...
conn.response :logger if debug?
...

def conf
  # either read from something that got set in the Rails app configuration or from an ENV variable (which is set to false by default unless for development)
  @conf ||= Rails.configuration.x.true_layer
end

def debug?
  !!conf.debug
end
```

That way we're not tied to a particular environment in the code and gives us flexibility to define when this options are enabled/disabled.

app/services/true_layer/identify_applicant.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    attr_reader :token, :error
+
+    def initialize(token)
+      @token = token
+    end
+
+    def applicant
+      return login_error unless response.success?
+      Applicant.find_by(email: emails) || applicant_lookup_error
```
lostie: 🤔 I have a feeling that this might cause us more trouble right now than actually help us, as I'm not sure there's any  guarantee that the email address registered in our system is going to be the same as the one the applicant is currently using to login to TrueLayer/Online Bank Account.

**not necessarily a blocker**, but I would consider to review this identification process with Neil, to confirms what are the service needs.

app/controllers/citizens/accounts_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Citizens
+  class AccountsController < ApplicationController
+    before_action :authenticate_applicant!
```
lostie: 😕 I'm a bit confused as how we're authenticating the applicant here, since our app has (**or more accurately** will have) authentication for the applicant which is _supposed_ to be completely independent from the applicant providing us access to his/her bank account through TrueLayer.

config/application.rb
```diff
@@ -20,5 +20,7 @@ class Application < Rails::Application
     config.govuk_notify_templates = config_for(
       :govuk_notify_templates, env: ENV.fetch('GOVUK_NOTIFY_ENV', 'development')
     ).symbolize_keys
+
+    config.eager_load_paths << Rails.root.join('lib/omniauth')
```
lostie: **not a blocker** FYI, I came across this previously in the project, it seems Rails has removed autoload from `lib` directories, and they recommend having this sort of functionality in `app/lib` 🤷‍♂️ : https://stackoverflow.com/questions/38198668/rails-5-load-lib-files-in-production

config/initializers/devise.rb
```diff
@@ -0,0 +1,291 @@
...
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
lostie: Do we need to change this to some sensible configuration?

config/initializers/devise.rb
```diff
@@ -0,0 +1,291 @@
...
+  # config.rememberable_options = {}
+
+  # ==> Configuration for :validatable
+  # Range for password length.
+  config.password_length = 6..128
+
+  # Email regex used to validate email formats. It simply asserts that
+  # one (and only one) @ exists in the given string. This is mainly
+  # to give user feedback and not to assert the e-mail validity.
+  config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
```
lostie: **not a blocker** Not entirely sure of what's the impact of this, but seems like it makes sense to use the same regexp as the one [already in place](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/29dce121c8c27f238e004951cc331ade8f4533d4/app/forms/applicants/email_form.rb#L9) in the forms?

spec/requests/applicants/omniauth_callbacks_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    example.run
+    OmniAuth.config.mock_auth[:true_layer] = nil
+    OmniAuth.config.test_mode = false
+  end
+
+  let(:token) { SecureRandom.uuid }
+  let(:expires_at) { 1.hour.from_now.round }
+  let(:true_layer_expires_at) { expires_at.to_i }
+  let(:applicant) { create :applicant }
+  let(:identification_tool) { double applicant: applicant }
```
lostie: 💭 If possible avoid the use of doubles with no connection to the real object we're mocking. If we can use `instance_double(TrueLayer::IdentifyApplicant)` instead, as it enforces the methods defined in the double to exist in the original object, which is not the case for a simple `double`.

spec/requests/applicants/omniauth_callbacks_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+    context 'with nil time' do
+      let(:true_layer_expires_at) { nil }
+      it 'should persist expires at in applicant' do
+        subject
+        expect(applicant.reload.true_layer_token_expires_at).to be_nil
+      end
+    end
+
+    context 'on lookup failure' do
+      let(:identification_tool) { double applicant: nil, error: 'foo' }
```
lostie: [Ditto](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/46/files#r231114895)

app/controllers/v1/applicants_controller.rb
```diff
@@ -6,20 +6,25 @@ def create
...
       else
-        render_400 applicant.errors
+        render_400 applicant_errors
       end
     end
 
     private
 
+    def applicant_errors
+      applicant.errors.details[:email_address] = applicant.errors.details.delete(:email) if applicant.errors.details[:email]
```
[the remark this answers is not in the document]
reggieb: :+1:

spec/services/true_layer/identify_applicant_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+        it 'returns an errors type' do
+          subject.applicant
+          expect(subject.error).to eq('Credentials do not match')
+        end
+      end
+    end
+
+    describe '#error' do
+      it 'nil until something goes wrong' do
```
lostie: 😈 This description seems odd specially if there's nothing changing what `until` means :P

reggieb:
@Kalagan To be honest, using Devise made sense when I started building this, but the more I built, the less sure I was in that assumption. I've left it there until we've had the authentication discussion.

app/services/true_layer/identify_applicant.rb
reggieb: As discussed - I'll strip this out.

config/initializers/devise.rb
reggieb: I left it place because it's a less stringent check than the existing, but I'll change it to match.

app/controllers/citizens/accounts_controller.rb
reggieb: I've updated the controller at the start of the Citizen journey so that it signs the applicant into Devise and stores the current applicant details in session. It is then these credentials that are used to get the correct `applicant` when the user returns from TrueLayer.

spec/lib/omniauth/strategies/true_layer_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+          expect(client_options.token_url).to match('connect/token')
+        end
+      end
+
+      describe '#authorize_params' do
+        before do
+          allow(subject).to receive(:session).and_return({})
+        end
+        it 'disables mock as default' do
+          with_modified_env TRUE_LAYER_ENABLE_MOCK: nil do
```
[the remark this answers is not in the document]
reggieb: As discussed: I've left this unchanged as I think it's clearer that what is being mocked are changes to the environment variables, and that the specs are describing the change of behaviour that results from these changes.

spec/rails_helper.rb
```diff
@@ -6,6 +6,7 @@
 abort("The Rails environment is running in production mode!") if Rails.env.production?
 require 'rspec/rails'
 require 'json_expressions/rspec'
+require 'webmock/rspec'
```
lostie: **[nit pick/not a blocker]** I believe this is already required in the [spec_helper](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/29dce121c8c27f238e004951cc331ade8f4533d4/spec/spec_helper.rb#L17), so shouldn't really be necessary to require it here.

reggieb:
- **cancel/go back buttons** Failures now redirect back to `/citizens/consent` (the start of the True Layer journey)
- **handling error scenarios** These "failures" still result in a token being returned. It is only on the subsequent call out to get data from the API that the failure can be detected. I've passed this on to Julien who is working on the API call out steps.

PR #49, opened 2018-11-01
More small improvements to the setup process and README file

Kalagan:
## What

1. Add `postgresql` to `Brewfile` and add `brew bundle` instruction to README
2. change `system` to `system!` in `bin/setup` for consistency with other commands and so it shows an error if the command fails
3. Remove incomplete endpoint documentation from README as we should generate documentation from the code with something like Swagger

## Checklist

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -42,7 +42,7 @@ chdir APP_ROOT do
   system! 'bin/rails log:clear tmp:clear'
 
   puts "\n== Installing JavaScript dependencies =="
-  system('npm install')
+  system! 'npm install'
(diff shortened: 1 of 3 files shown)


bin/setup
```diff
@@ -42,7 +42,7 @@ chdir APP_ROOT do
   system! 'bin/rails log:clear tmp:clear'
 
   puts "\n== Installing JavaScript dependencies =="
-  system('npm install')
+  system! 'npm install'
```
lostie: 👌

PR #51, opened 2018-11-01
Bugfix/postcode not found

willc-work:
The design requires the form to show the lookup postcode and inset text to the user after they submit an address manually, if the form contains errors. Currently the validation of the form will remove the postcode and inset text from the users view.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-170)

Add lookup_postcode and lookup_error to parameters so they can be rendered with the new page
after a validation error.
Exclude the new lookup_* parameters from being passed to the model

## Checklist

### DIFF
--- a/app/controllers/providers/addresses_controller.rb
+++ b/app/controllers/providers/addresses_controller.rb
@@ -16,6 +16,7 @@ module Providers
         redirect_to action_for_next_step(options: { application: applicant.legal_aid_application })
       else
         @lookup_postcode = params[:address][:lookup_postcode]
+        @lookup_error = params[:address][:lookup_error]
         render :new
       end
     end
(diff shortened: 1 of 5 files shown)


app/forms/base_form.rb
```diff
@@ -42,7 +42,7 @@ def save
 
     # List of form attributes not to be passed to model
     def exclude_from_model
-      []
+      %i[lookup_postcode lookup_error]
```
lostie: This class (`BaseForm`) holds the default configuration for that method. If you want a specific configuration for your class, you should be overriding this in the class that requires it (`AddressForm`).

willc-work: changed👍

app/controllers/providers/addresses_controller.rb
```diff
@@ -15,14 +15,16 @@ def create
       if @form.save
         redirect_to action_for_next_step(options: { application: applicant.legal_aid_application })
       else
+        @lookup_postcode = params[:address][:lookup_postcode]
```
lostie: If `:lookup_postcode` and `:lookup_error` are already being submitted through the form, 

`@form = Applicants::AddressForm.new(form_params)`

is already getting old of those values as it's initialised with them (`form_params`), so when rendering the view, there's no need to have these instance variables declared (they're not being used, and the form object already has the values for them anyway.

PR #58, opened 2018-11-05
Feature/ap 106 107 benefit checker result

Kalagan:
[AP-106 client is passported](https://dsdmoj.atlassian.net/browse/AP-106)
[AP-107 client is not passported](https://dsdmoj.atlassian.net/browse/AP-107)

After the Benefits Check step, show the result and indicate what is the next step.

The API doesn't return what kind of benefit the citizen receives so I could not implement exactly what's on the prototype.
Prototype: https://provider-apply-for-legal-aid.herokuapp.com/passported

To get a positive result (citizen receiving benefits), use these values:
Last name: Walker
DoB: 10/01/1980
NIN: JA293483A
You can find more example in this JIRA spike: https://dsdmoj.atlassian.net/browse/AP-84

**Positive Result:**
image

**Negative Result:**
image

## Checklist

### DIFF
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -15,6 +15,8 @@ linters:
         Enabled: false
       Layout/TrailingWhitespace:
         Enabled: false
+      Layout/LeadingBlankLines:
+        Enabled: false
       Lint/UselessAssignment:
         Enabled: false
       Metrics/LineLength:
--- a/features/providers/step_definitions/civil_journey_steps.rb
+++ b/features/providers/step_definitions/civil_journey_steps.rb
@@ -94,7 +94,7 @@ Then('I am on the postcode entry page') do
 end
 
 Then('I am on the benefit check results page') do
-  page.should have_content("Your client's tax and benefits status")
+  expect(page).to have_content("Your client's tax and benefits status")
 end
 
 Then(/^I enter a valid postcode '(.*)'$/) do |postcode|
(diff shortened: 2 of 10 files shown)


.erb-lint.yml
```diff
@@ -15,6 +15,8 @@ linters:
         Enabled: false
       Layout/TrailingWhitespace:
         Enabled: false
+      Layout/LeadingBlankLines:
+        Enabled: false
```
Kalagan: I didn't manage to make this rule happy with my code on app/views/providers/check_benefits/index.html.erb line 5 to 20 (see below)

Kalagan:
Good point. I squashed all these small commits into one

features/providers/step_definitions/civil_journey_steps.rb
```diff
@@ -94,7 +94,7 @@
 end
 
 Then('I am on the benefit check results page') do
-  page.should have_content('Benefit check results')
+  page.should have_content("Your client's tax and benefits status")
```
willc-work: The should syntax is deprecated, these can be changed to use the expect syntax

```Then('I am on the benefit check results page') do
  expect(page).to have_content("Your client's tax and benefits status")```

File: features/providers/step_definitions/civil_journey_steps.rb [hunk 3]
```diff
@@ -108,3 +108,11 @@
 Then(/^I select an address '(.*)'$/) do |address|
   select(address, from: 'address_selection[address]')
 end
+
+Then(/^I see a notice saying that the citizen receives benefits$/) do
+  page.should have_content('Your client receives benefits that qualify for legal aid.')
```
willc-work: as above re: should/expect

PR #64, opened 2018-11-07
AP-71 Allow applicant to select whether they have additional accounts

reggieb:
## What

[JIRA AP-71](https://dsdmoj.atlassian.net/browse/AP-71)

The start of these steps is at `/citizens/additional_accounts`

Adds two screens to allow a user to:

- Confirm they have no more accounts and continue
- Start the True Layer add another account journey
- Confirm they have other accounts, that they will need to supply other evidence, and continue

At the moment the next step in the process and the "add another account" journey are not defined, so the endpoints instead display simple landing pages.

### DIFF
--- a/app/controllers/base_controller.rb
+++ b/app/controllers/base_controller.rb
@@ -1,3 +1,8 @@
 class BaseController < ApplicationController
   layout 'application'
+
+  def current_legal_aid_application
+    return unless session[:current_application_ref].present?
+    @current_legal_aid_application ||= LegalAidApplication.find(session[:current_application_ref])
+  end
 end
--- a/app/controllers/citizens/additional_accounts_controller.rb
+++ b/app/controllers/citizens/additional_accounts_controller.rb
@@ -1,0 +1,35 @@
+module Citizens
+  class AdditionalAccountsController < BaseController
+    def index; end
+
+    def create
+      case params[:additional_account]
+      when 'yes'
+        redirect_to new_citizens_additional_account_path
+      when 'no'
+        # TODO: - set redirect path when known
+        render plain: 'Landing page: Next step in Citizen journey'
+      else
+        flash[:error] = 'You must select either Yes or No'
+        render :index
+      end
+    end
+
+    def new; end
+
+    def update
+      case params[:additional_account]
+      when 'yes'
+        # TODO: - set redirect path when known
+        render plain: 'Landing page: Return to True Layer steps'
+      when 'no'
+        current_legal_aid_application&.update(has_offline_accounts: true)
+        # TODO: - set redirect path when known
+        render plain: 'Landing page: Next step in Citizen journey'
+      else
+        flash[:error] = 'You must select either Yes or No'
+        render :index
+      end
+    end
+  end
+end
--- a/app/views/citizens/additional_accounts/index.html.erb
+++ b/app/views/citizens/additional_accounts/index.html.erb
@@ -1,0 +1,43 @@
+<% content_for :navigation do %>
+  <%#= link_to 'Back', path-to-be-determined, class: 'govuk-back-link', id: 'back' %>
+<% end %>
+
+  <%= form_tag(citizens_additional_accounts_path) do %>
+
+    <div class="govuk-form-group">
+      <fieldset class="govuk-fieldset">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+          <h1 class="govuk-fieldset__heading">
+            Do you have accounts with other banks?
+          </h1>
+        </legend>
+        <br />
+        <span class="govuk-hint">
+          Share details of any accounts you have with another bank or building society
+        </span>
+        <% if flash[:alert] %>
+          <%= flash[:alert] %>
+        <% end %>
+        <div class="govuk-radios">
+          <div class="govuk-radios__item">
+            <input class="govuk-radios__input" id="additional-account-1" name="additional_account" type="radio" value="yes" />
+            <label class="govuk-label govuk-radios__label" for="additional-account-1">
+              Yes
+            </label>
+          </div>
+          <div class="govuk-radios__item">
+            <input class="govuk-radios__input" id="additional-account-2" name="additional_account" type="radio" value="no" />
+            <label class="govuk-label govuk-radios__label" for="additional-account-2">
+              No
+            </label>
+          </div>
+        </div>
+      </fieldset>
+    </div>
+
+    <%= submit_tag 'Continue', class: 'govuk-button govuk-button' %>
+
+  <% end %>
+
+  </div>
+</div>
--- a/app/views/citizens/additional_accounts/new.html.erb
+++ b/app/views/citizens/additional_accounts/new.html.erb
@@ -1,0 +1,43 @@
+<% content_for :navigation do %>
+  <%#= link_to 'Back', path-to-be-determined, class: 'govuk-back-link', id: 'back' %>
+<% end %>
+
+  <%= form_tag(citizens_additional_account_path(id: :update), method: :patch) do %>
+
+    <div class="govuk-form-group">
+      <fieldset class="govuk-fieldset">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+          <h1 class="govuk-fieldset__heading">
+            Do you have online access to your accounts with other banks?
+          </h1>
+        </legend>
+        <br />
+        <span class="govuk-hint">
+          If you do not, you will need to provide paper evidence for these accounts.
+        </span>
+        <% if flash[:alert] %>
+          <%= flash[:alert] %>
+        <% end %>
+        <div class="govuk-radios">
+          <div class="govuk-radios__item">
+            <input class="govuk-radios__input" id="additional-account-1" name="additional_account" type="radio" value="yes" />
+            <label class="govuk-label govuk-radios__label" for="additional-account-1">
+              Yes
+            </label>
+          </div>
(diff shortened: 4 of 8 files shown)


app/controllers/base_controller.rb
```diff
@@ -1,3 +1,8 @@
 class BaseController < ActionController::Base
   layout 'application'
+
+  def current_legal_aid_application
+    return unless session[:current_application_ref].present?
```
lostie: If either the `session[:current_application_ref]` or the application itself don't exist, isn't the expected behaviour the same? For the user to be shown an error page? 

I would say, if the application existence is a pre-condition to access these page, this should be defined as a `before_action` to enforce its existence or redirect the user to an error page.

app/views/citizens/additional_accounts/index.html.erb
```diff
@@ -0,0 +1,43 @@
+<% content_for :navigation do %>
+  <%#= link_to 'Back', path-to-be-determined, class: 'govuk-back-link', id: 'back' %>
+<% end %>
+
+  <%= form_tag(citizens_additional_accounts_path) do %>
```
lostie: **[not a blocker]** One of the things we discussed offline as a team was that there was quite a lot of repetition across the multiple form pages we have in the app.

💭 It would be 👍 to slowly try to reduce that repetition, things like:

```html
    <div class="govuk-form-group">
      <fieldset class="govuk-fieldset">
        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
          <h1 class="govuk-fieldset__heading">
            Do you have accounts with other banks?
          </h1>
        </legend>
        <br />
       ...
```

are the same across all form pages and only the heading copy changes (which could even be set in the locales).

Also, this file and https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/64/files#diff-e7165391fd39abd65998b4a03e647975 seem very similar which only re-enforces this point. DRY.

app/views/citizens/additional_accounts/index.html.erb
```diff
@@ -0,0 +1,43 @@
...
+          </h1>
+        </legend>
+        <br />
+        <span class="govuk-hint">
+          Share details of any accounts you have with another bank or building society
+        </span>
+        <% if flash[:alert] %>
+          <%= flash[:alert] %>
+        <% end %>
+        <div class="govuk-radios">
```
lostie: 💭 Given this is the first time we're defining a radio button with the new designs, it would be 👍 to try to abstract that similarly to [what was done to text fields](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/ap-71-add-accounts-for-more-than-one-bank/app/views/shared/forms/_text_fields.html.erb), so that going forward anyone could just define a new radio group by supplying the the field and the radio values allowed for it (similary to what Rails does by default). The idea behind this, again, would be to avoid having the exactly same HTML across multiple templates where we're effectively just changing the values for the radios, not the radios HTML itself. DRY.

app/views/citizens/additional_accounts/index.html.erb
```diff
@@ -0,0 +1,43 @@
...
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+          <h1 class="govuk-fieldset__heading">
+            Do you have accounts with other banks?
+          </h1>
+        </legend>
+        <br />
+        <span class="govuk-hint">
+          Share details of any accounts you have with another bank or building society
+        </span>
+        <% if flash[:alert] %>
```
lostie: I'm not entirely sure this is needed. The controllers are defining `flash[:error]` and the main layout [already has](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/ap-71-add-accounts-for-more-than-one-bank/app/views/layouts/application.html.erb#L60-L64) a general way of displaying those.

app/controllers/citizens/additional_accounts_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    def new; end
+
+    def update
+      case params[:additional_account]
+      when 'yes'
+        # TODO: - set redirect path when known
+        render plain: 'Landing page: Return to True Layer steps'
+      when 'no'
```
lostie: **[not a blocker]** I personally find it a bit confusing to use the same form for two different types of resource (they're both have a Yes/No radios, that's true), but then performing this something like:

```ruby
if params[:additional_account] == 'no'
  current_legal_aid_application&.update(has_offline_accounts: true)
end
```

Wouldn't be simpler and clearer to submit this data as `params[:has_offline_accounts]` ? which expresses exactly what we're asking?

app/controllers/citizens/additional_accounts_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    def create
+      case params[:additional_account]
+      when 'yes'
+        redirect_to new_citizens_additional_account_path
+      when 'no'
+        # TODO: - set redirect path when known
+        render plain: 'Landing page: Next step in Citizen journey'
+      else
+        flash[:error] = 'You must select either Yes or No'
```
lostie: **[suggestion]** If we were to handle this submission as normal forms that need to be validated, we could leverage out of the box from the form validation errors, the already existent way to display the form errors if it's not valid and all this copy would be set in the locales, reducing this action code to something like the following:

```ruby
def create
  @form = Citizens::AdditionalAccountsForm.new(form_params)
  if @form.valid?
     redirect_to new_citizens_additional_account_path
  else
    render :new
  end  
end
```

app/controllers/base_controller.rb
reggieb: This method just gets a `legal_aid_application`. I was expecting the specific controller to decide on how to deal with it's absence. Also as the behaviour through the Citizen Journey didn't seem to be finalised I didn't want to put too many restrictions on its use.

app/views/citizens/additional_accounts/index.html.erb
reggieb: OK

app/controllers/citizens/additional_accounts_controller.rb
reggieb: I'd agree with you if this wasn't such a simple case - in that there is only one error condition (submit is clicked without choosing either Yes or No).

app/views/citizens/additional_accounts/new.html.erb
```diff
@@ -0,0 +1,29 @@
+<% content_for :navigation do %>
+  <%#= link_to 'Back', path-to-be-determined, class: 'govuk-back-link', id: 'back' %>
+<% end %>
+
+  <%= form_tag(citizens_additional_account_path(id: :update), method: :patch) do %>
```
lostie: **[suggestion/not a blocker]** If we were conforming with the conventions associated with the form template name and the associated RESTFUL action, this should probably be a **post**.

app/helpers/gov_uk_form_helper.rb
```diff
@@ -0,0 +1,54 @@
+module GovUkFormHelper
```
lostie: 👍 This is a great starting point to define abstractions for our forms.

app/views/citizens/additional_accounts/index.html.erb
```diff
@@ -0,0 +1,29 @@
...
+          ) %>
+      <%= govuk_error_message @error %>
+      <%= govuk_radio_inputs :additional_account, yes: 'Yes', no: 'No' %>
+    <% end %>
+
+    <%= govuk_submit_button 'Continue' %>
+
+  <% end %>
+
+  </div>
```
lostie: ✂️ Are these `divs` required anymore?

app/views/citizens/additional_accounts/new.html.erb
```diff
@@ -0,0 +1,29 @@
...
+          ) %>
+      <%= govuk_error_message(@error) %>
+      <%= govuk_radio_inputs :has_offline_accounts, yes: 'Yes', no: 'No' %>
+    <% end %>
+
+    <%= govuk_submit_button 'Continue' %>
+
+  <% end %>
+
+  </div>
```
lostie: [Ditto](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/64/files#r232273315)

app/controllers/citizens/additional_accounts_controller.rb
lostie: Let's agree to disagree on this :bowtie: , a lot of the custom code, including the case statements in the controller, the custom error display, the bespoke copy set directly in the views, could just be managed automatically by driving the form submission through the conventional Rails model validations/locales.

app/views/citizens/additional_accounts/new.html.erb
reggieb: It could be a post, but then I'd have to either customise the routes or put the second action into another controller. I could argue that the resource is the citizen's decision on whether to add additional accounts, and this action is a modification of that decision.

app/helpers/gov_uk_form_helper.rb
reggieb: I'm glad you like it. I think helpers allow the code to be tidied up and remove logic from the view partials.

app/views/citizens/additional_accounts/index.html.erb
reggieb: Good spot. Thank you.

app/views/citizens/additional_accounts/new.html.erb
reggieb: Ditto

PR #69, opened 2018-11-09
AP-174 Retrieve bank data from TrueLayer

Kalagan:
## What

[AP-174](https://dsdmoj.atlassian.net/browse/AP-174)

Once the applicant is redirected from TrueLayer back to us, we use the TrueLayer token to fetch: 
- bank provider (HSBC, Barclays, etc..)
- bank accounts (Current Account, Savings, etc...)
- bank account balances
- bank account holders (can be more than 1) and their info (name, address, DoB, email)
- transactions

For now I kept it simple and everything is done synchronously but we'll probably need to improve that at some point because it obviously take several seconds to retrieve all this data.

For each model, the raw TrueLayer API response is store as json in a `true_layer_response` attribute and I only added specific attributes for the ones that will be used in the following screen. We can easily extract more specific attributes later by simply adding the DB columns and modifying the mapping methods.

I also kept the error handling super simple for now.
For now, if the API raises an error, it stops the import, logs the error in Sentry, and return it.
This needs some consideration because different errors need to be treated differently and we need to decide what to present to the user.
I also noticed that some API endpoints return "Feature not supported by the provider" for certain banks.

You can test the integration by using the Mock bank and using one of these users: https://docs.truelayer.com/#mock-users

## Checklist


app/services/true_layer/bank_data_import_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    def initialize(applicant:, token:)
+      @applicant = applicant
+      @token = token
+    end
+
+    def call
+      return unless bank_provider
+
+      import_bank_accounts
+      import_account_holders
```
lostie: ⚠️ There's a chance that if any of the these steps fail, while a previous step succeeded, given they're not wrapped in a transaction, that we might end up with inconsistent data (e.g. imported bank account and holders but the transactions pull fails).

💭 Consider having some process in place to trigger a rollback so that we either import everything or rollback to the initial state, to guarantee consistency of data.

app/services/true_layer/importers/import_account_holders_service.rb
```diff
@@ -0,0 +1,36 @@
...
+      private
+
+      attr_reader :api_client, :bank_provider
+
+      def mapped_resources
+        true_layer_resources.map do |resource|
+          {
+            true_layer_response: resource,
+            full_name: resource[:full_name],
+            full_address: resource[:addresses]&.map do |address|
```
lostie: 💭 If we are storing **multiple** addresses in this field, would it be better just to call it `addresses` to be clear as what the fields has? (`full_address` seems a bit deceiving).  

Also, I'm not entirely familiar with the format of the returned addresses, but is there any chance that any of the `address.values` already has commas on it? (if so, I would consider storing these addresses in a more structured way, serialised).

db/migrate/20181107110620_create_true_layer_models.rb
```diff
@@ -0,0 +1,42 @@
...
+      t.string :currency
+      t.string :account_number
+      t.string :sort_code
+      t.decimal :balance
+      t.timestamps
+    end
+
+    create_table :bank_transactions, id: :uuid do |t|
+      t.belongs_to :bank_account, foreign_key: true, null: false, type: :uuid
+      t.json :true_layer_response
```
lostie: 💭 Wouldn't a transaction have some basic attributes we're expecting to have by default like:

* operation - `credit/debit`
* value
* name/reference

?

app/controllers/applicants/omniauth_callbacks_controller.rb
```diff
@@ -6,10 +6,7 @@ class OmniauthCallbacksController < Devise::OmniauthCallbacksController
 
     def true_layer
       if applicant
-        applicant.update!(
-          true_layer_token: token,
-          true_layer_token_expires_at: token_expires_at
-        )
+        import_bank_data
```
lostie: ❓ If this process fails (an exception is raised for example), is the expected outcome to show a 500 page or display an error page to the user stating there was a problem fetching their bank details?

app/services/true_layer/api_client.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        Rails.logger.warn("TrueLayer Error : #{response.body}")
+        []
+      end
+    end
+
+    def connection
+      @connection ||= Faraday.new(url: TRUE_LAYER_URL) do |conn|
+        conn.authorization :Bearer, token
+        conn.response :logger if Rails.env.development?
```
lostie: **[not a blocker]** I think I mentioned this in a previous PR, would be preferable to not tie the code to specific env, but instead configure the environments themselves to have a (just an example): `ENV['TRUE_LAYER_LOGGER']` which is disabled by default in all environment (it doesn't even have to be defined) and it's activated in the development configuration. Makes it more flexible for us to control it as well.

app/controllers/applicants/omniauth_callbacks_controller.rb
Kalagan: Looks like there's a story about showing a message to the user: https://dsdmoj.atlassian.net/browse/AP-173

I'm going to modify this PR to show the raw errors to the user and AP-173 can refine it.

app/services/true_layer/bank_data_import_service.rb
Kalagan: I'm thinking of refactoring the code to do all API requests first and then do the DB changes. That way we don't even need transactions and most importantly, we avoid long deadlocks caused by API requests in-between DB calls.
What do you think?

spec/services/true_layer/importers/import_transactions_service_spec.rb
```diff
@@ -1,44 +1,33 @@
...
+    let(:mock_transaction_2) { mock_account[:transactions][1] }
+    let(:transaction_1) { bank_account.bank_transactions.find_by(true_layer_id: mock_transaction_1[:transaction_id]) }
+    let(:transaction_2) { bank_account.bank_transactions.find_by(true_layer_id: mock_transaction_2[:transaction_id]) }
     let!(:existing_transaction) { create :bank_transaction, bank_account: bank_account }
 
     before do
       allow(Time).to receive(:now).and_return(now)
-      stub_request(:get, endpoint).to_return(body: response_body)
+      stub_true_layer_transactions
+      # stub_request(:get, endpoint).to_return(body: response_body)
```
lostie: ✂️ (I'm assuming this is not needed anymore).

PR #71, opened 2018-11-09
Address lookup and selection page styling amendments (AP-179)

LAACharlesBoardman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-179)

- Added a new local variable to the Text Field partial that determines if a hint is displayed
This means on one page we can display the postcode field with a hint and on another without the hint
- Correcting the postcode text field partial call seemed to have fixed the error anchor link
- Added some logic that determines which title the manual address page should display depending on whether the user chose to input the address manual or was redirected after a lookup fail

## Checklist

### DIFF
--- a/app/views/shared/forms/_text_fields.html.erb
+++ b/app/views/shared/forms/_text_fields.html.erb
@@ -6,8 +6,10 @@
 <div class="govuk-form-group <%= group_error_class %>">
   <%= form.label field_name, class: ['govuk-label'] %>
   <% hint = t(".#{field_name}_hint", default: '') %>
-    <% if hint.present? %>
-      <%= content_tag(:span, hint, id: "#{field_name}-hint".dasherize, class: ['govuk-hint']) %>
+    <% if local_assigns[:hint?] == true %>
+      <% if hint.present? %>
+        <%= content_tag(:span, hint, id: "#{field_name}-hint".dasherize, class: ['govuk-hint']) %>
+      <% end %>
     <% end %>
   <% if form.object.errors[field_with_error].any? %>
     <%= content_tag(:span, form.object.errors[field_name].first, class: ['govuk-error-message']) %>
--- a/app/views/shared/forms/_text_fields.html.erb
+++ b/app/views/shared/forms/_text_fields.html.erb
@@ -6,7 +6,7 @@
 <div class="govuk-form-group <%= group_error_class %>">
   <%= form.label field_name, class: ['govuk-label'] %>
   <% hint = t(".#{field_name}_hint", default: '') %>
-  <% if local_assigns[:hint?] && hint.present? %>
+  <% if !local_assigns[:hide_hint?] && hint.present? %>
     <%= content_tag(:span, hint, id: "#{field_name}-hint".dasherize, class: ['govuk-hint']) %>
   <% end %>
   <% if form.object.errors[field_with_error].any? %>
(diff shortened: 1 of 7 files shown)


spec/requests/providers/addresses_request_spec.rb
```diff
@@ -34,7 +34,7 @@
       get_request
 
       expect(response).to be_successful
-      expect(response.body).to include("Enter your client's address manually")
+      expect(response.body).to include(CGI.escapeHTML("Enter your client's address manually"))
```
lostie: 💭 [There's already an helper method](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/spec/support/request_helpers.rb#L22-L24) which can be used for this purpose, like in [this example](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/2e1471215a2e3abc3900dfe33a00484aff35fe79/spec/requests/providers/address_selections_spec.rb#L59).

spec/requests/providers/addresses_request_spec.rb
```diff
@@ -71,7 +71,7 @@
 
       it 'renders the form again if validation fails' do
         post_request
-        expect(response.body).to include("Enter your client's address manually")
+        expect(response.body).to include(CGI.escapeHTML("Enter your client's address manually"))
```
lostie: [Ditto](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/71/files#r232283856)

app/views/shared/forms/_text_fields.html.erb
```diff
@@ -6,8 +6,10 @@
 <div class="govuk-form-group <%= group_error_class %>">
   <%= form.label field_name, class: ['govuk-label'] %>
   <% hint = t(".#{field_name}_hint", default: '') %>
-    <% if hint.present? %>
-      <%= content_tag(:span, hint, id: "#{field_name}-hint".dasherize, class: ['govuk-hint']) %>
+    <% if local_assigns[:hint?] == true %>
```
lostie: 💭 This block can probably be simplified to something like the following:

```ruby
<% if local_assigns[:hint?] && hint.present? %>
  <%= content_tag(:span, hint, id: "#{field_name}-hint".dasherize, class: ['govuk-hint']) %>
<% end %>
```

lostie: ❓ Is the expected behaviour that by default the hint is not shown for **all text fields** unless we specifically set `hint?: true` in the form for those fields? Even when the hint text for those text fields are defined in the locale file?

spec/requests/providers/addresses_request_spec.rb
LAACharlesBoardman: I've updated the code here.
I hope I've understood this correctly.

app/views/shared/forms/_text_fields.html.erb
mpw5: I think it might be better if we flipped this logic, so that hints are shown by default and only hidden if explicitly set by the developer - I think this situation is a minority case, in most cases if a hint is being defined then it probably should be shown.

LAACharlesBoardman: Reversed the logic here.
Local is now called hide_hint and will hide the text field hint when set to true.

PR #73, opened 2018-11-12
AP 148 add secure link for client access

reggieb:
[JIRA AP-148](https://dsdmoj.atlassian.net/browse/AP-148)

- [x] we are able to send a user a unique link to access their application
- [x] if I go directly to an application page then I get an authentication error
- [x] application links are unique
- [x] we generate a new link for every new application

This system uses a new model object - `SecureData` that stores a mapping between a UUID and a JWT token. The data within the token is then used to identify the matching `LegalAidApplication`. This ensures that without decoding the JWT, there is no connection between the UUID and the `legal_aid_application`. It also means that the UUID/token can be managed (expired for example) separately from the application.

With this in place:

1. Within the **provider** journey, once a `legal_aid_application` has been created, a `secure_id` (UUID) can be generated with `legal_aid_application.generate_secure_id`. 
2. The start of the citizen journey is then: `<host>/citizens/legal_aid_applications/<secure_id>`, and this url can be emailed to the citizen (that action has not yet been created).

For now, in the **provider** journey, once the Citizen's email is entered, a _notice_ is used to provide the link in the next page.

Note - that if a controller action has `before_action :authenticate_applicant!` a user trying to access the rendered page will get an authentication error. However, there is still work to do to modify the default Devise authentication error pages.


app/services/citizen_email_service.rb
```diff
@@ -32,4 +32,8 @@ def applicant
     # application.
     @applicant ||= application&.applicant
   end
+
+  def secure_id
+    application.generate_secure_id
```
lostie: ❓ Isn't this code and [the email form controller](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/14cc44122efdd89d144bc3db5a8e52e72dbf0946/app/controllers/providers/applicants_controller.rb#L65) generating two distinct secure IDs? Is this intended or once the secure ID is generated once it should stand as the only one (unless we forcefully re-generated it)?

app/models/secure_data.rb
```diff
@@ -0,0 +1,22 @@
...
+  def self.for(id)
+    find(id).retrieve
+  end
+
+  def store(hash)
+    self.data = JWT.encode hash, nil, 'none'
+  end
+
+  def retrieve
+    JWT.decode(data, nil, false).first.symbolize_keys
```
lostie: 📝 Memo for signing the JWT (as spoken offline), as without it we can leave ourselves vulnerable by dynamically picking the algorithm, [as stated in the gem itself](https://github.com/jwt/ruby-jwt#algorithms-and-usage).

app/services/citizen_email_service.rb
reggieb: That's correct. In the final code it should only be needed once. I've added a comment to the applicants controller code stating that this should be removed when we no longer need an easy way to display a working link.

app/models/legal_aid_application.rb
```diff
@@ -10,6 +10,22 @@ class LegalAidApplication < ApplicationRecord
...
+  def self.find_by_secure_id(secure_id)
+    secure_data = SecureData.find(secure_id)
+    find_by! secure_data.retrieve[:find_by]
+  end
+
+  def generate_secure_id
+    SecureData.create_and_store(
+      find_by: {
+        id: id,
+        applicant_id: applicant_id
```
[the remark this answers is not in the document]
reggieb: I think it may be simpler to just store the id for now. We can increase the amount of data stored if there is a requirement to do so.

app/models/secure_data.rb
```diff
@@ -0,0 +1,28 @@
+# Note: model uses UUID for id which is generated by Postgres. Note that any changes to
+# the way the UUID is generated need to ensure that it generated by a secure random system.
+class SecureData < ApplicationRecord
+  def self.create_and_store(hash)
+    secure_data = new
+    secure_data.store(hash)
+    secure_data.save
```
Kalagan: If you are not using the boolean returned by `.save`, I suggest to use `.save!` instead

app/models/secure_data.rb
```diff
@@ -0,0 +1,28 @@
...
+    secure_data.id
+  end
+
+  def self.for(id)
```
Kalagan: I don't see this method used anywhere

spec/models/legal_aid_application_spec.rb
```diff
@@ -89,4 +89,34 @@
...
+    end
+
+    it 'returns the generated id' do
+      expect(subject).to eq(secure_data.id)
+    end
+
+    it 'generates data that can be used to find application' do
+      id = subject
+      secure_data = SecureData.find(id)
+      expect(described_class.find_by(secure_data.retrieve[:find_by])).to eq(legal_aid_application)
```
Kalagan: This should be failing, shouldn't it?

spec/services/citizen_email_service_spec.rb
```diff
@@ -14,6 +15,7 @@
         .with(application.id, 'test@example.com', citizen_url, 'John Doe')
         .and_return(message_delivery)
       allow(message_delivery).to receive(:deliver_later)
+      allow(application).to receive(:generate_secure_id).and_return(secure_id)
```
Kalagan: shouldn't it be `expect` here instead of `allow` (and the above line as well)?
I tend to use `allow` when stubbing behaviour in the `before` blocks.
And `expect` in the specs.

app/models/secure_data.rb
reggieb: Thanks for that. I remember where I meant to use it - I've updated the `LegalAidApplication` model to make it a little cleaner.

app/models/secure_data.rb
reggieb: Reasonable point - but to match, I've added a band to the method too `SecureData.create_and_store!`

spec/models/legal_aid_application_spec.rb
reggieb: Weirdly `LegalAidApplication.find_by nil` is returning an instance, and as there is only one instance, it's the matching instance.

app/models/legal_aid_application.rb
```diff
@@ -10,6 +10,19 @@ class LegalAidApplication < ApplicationRecord
...
+  def self.find_by_secure_id!(secure_id)
+    secure_data = SecureData.for(secure_id)
+    find_by! secure_data[:legal_aid_application]
+  end
+
+  def generate_secure_id
+    SecureData.create_and_store!(
+      legal_aid_application: { id: id },
+      # Timestamp so token unique for each secure id
+      timestamp: Time.now.to_s
```
Kalagan: If the only goal of this is to generate a unique hash, we could use `SecureRandom.hex`. Because `Time.now.to_s` is only unique per second.

PR #78, opened 2018-11-14
AP-70 Confirm details of accounts that have been shared

tmckernan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-70)

Display ALL the accounts that the client has shared with the LAA.

A bank account can have multi names and addresses,  I have only displayed the last name and address link to this bank account as most bank accounts should only have one name and address. 

Created a small helper to display bank balance with the correct symbol if the currency is GBP, USD or EUR. If isn't one of these currency is just display the currency code before the balance.  If you want to display the all currencies  correct symbol, we could use the [money gem](https://github.com/RubyMoney/money). 

## Checklist


app/views/citizens/accounts/index.html.erb
```diff
@@ -11,8 +11,48 @@
...
+
+  <table class="govuk-table">
+    <tbody class="govuk-table__body">
+      <tr class="govuk-table__row">
+        <th class="govuk-table__header" scope="row"><%= t '.acount_holder_name_heading' %></th>
+        <td class="govuk-table__cell"><%= applicant_bank.bank_account_holders.last.full_name %></td>
+      </tr>
+      <tr class="govuk-table__row">
+        <th class="govuk-table__header" scope="row"><%= t '.account_holder_address_heading' %></th>
+        <td class="govuk-table__cell"><%= applicant_bank.bank_account_holders.last.addresses.first.values.join(', ') %></td>
```
lostie: **[not a blocker/suggestion]** Both here and line 24 we seems to be using very specific logic to display information on the page. 💭 Consider having a presenter (or helper methods, although I personally recommend the former) for the `applicant_bank` where `applicant_bank.bank_account_holders.last` could be the implementation for a **applicant_bank.main_bank_account_holder** for example, removing the chained logic from the view.

app/views/citizens/accounts/index.html.erb
```diff
@@ -11,8 +11,48 @@
...
+          <td class="govuk-table__cell"><%= display_value_with_symbol(bank_account.balance, bank_account.currency) %></td>
+        </tr>
+      <% end %>
+    </tbody>
+
+  </table>
+<% end %>
+
+<div class="govuk-!-padding-top-3">
+<%= link_to 'Continue', new_citizens_additional_account_path, class: 'govuk-button', id: 'continue' %>
```
lostie: **[not a blocker/nit pick]** Probably the lint doesn't detect this, but if we're wrapping something within the div, I would say it should properly indented.

spec/helpers/money_helper_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+RSpec.describe MoneyHelper, type: :helper do
+  let(:bank_account_gbp) { create(:bank_account, currency: 'GBP') }
+  let(:bank_account_eur) { create(:bank_account, currency: 'EUR') }
+  let(:bank_account_usd) { create(:bank_account, currency: 'USD') }
+  let(:bank_account_rsa) { create(:bank_account, currency: 'RSA') }
+
+  describe '#display_value_with_symbol' do
+    it 'return value with pound symbol' do
+      expect(display_value_with_symbol(bank_account_gbp.balance, bank_account_gbp.currency)).to eq "£#{bank_account_gbp.balance}"
```
lostie: **[improvement]** Given we're just testing an helper, I don't believe there's any need here to create records in the database as the helper itself just needs an integer and a string as arguments.

**[not a blocker/nit pick]** In terms of readability it would be clearer to have something like the following instead:

```ruby
describe '#display_value_with_symbol' do
  let(:balance) { 325.42 }

  subject(:humanized_balance) { display_value_with_symbol(balance, currency) }

  context 'when the currency is Euro' do
    let(:currency) { 'EUR' }

    it { is_expected.to eq("€#{balance}") } 
  end
end
```

spec/helpers/money_helper_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    it 'return value with euro symbol' do
+      expect(display_value_with_symbol(bank_account_eur.balance, bank_account_eur.currency)).to eq "€#{bank_account_eur.balance}"
+    end
+
+    it 'return value with dollor symbol' do
```
Kalagan: Typo:  `dollar` instead of `dollor `

spec/requests/citizens/accounts_spec.rb
```diff
@@ -12,5 +25,30 @@
     it 'returns http success' do
       expect(response).to have_http_status(:ok)
     end
+
+    it 'display account holder name' do
+      expect(response.body).to include(applicant_bank_account_holder.full_name)
```
lostie: Just me wary that (even this is not the case here), if the account holder name contains special characters and they're therefore encoded in the page, the current expectation is likely to fail, so I would suggest to use the [existent helper](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/2e1471215a2e3abc3900dfe33a00484aff35fe79/spec/support/request_helpers.rb#L22-L25) (this applies to the other expectations as well) to perform the assertion against. e.g:

```ruby
 expect(unescaped_response_body).to include(applicant_bank_account_holder.full_name)
```

📝 It's unfortunate that Rails/RSpec used to have a `have_content` helper method on these sort of tests that did exactly that, and they seem to have removed support for it inline with the very confusing (at least to me) Controller Tests vs Requests Tests vs Feature Tests :S

spec/helpers/money_helper_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      let(:currency) { 'EUR' }
+      it { is_expected.to eq("€#{balance}") }
+    end
+
+    context 'when the currency is USD' do
+      let(:currency) { 'USD' }
+      it { is_expected.to eq("$#{balance}") }
+    end
+
+    context 'when the currency is USD' do
```
reggieb: I think that label is wrong. Should it be something like:
```
context 'when currency is not recognised' do
```

app/helpers/money_helper.rb
```diff
@@ -0,0 +1,5 @@
+module MoneyHelper
+  def display_value_with_symbol(value, currency)
```
reggieb: I'd suggest `value_with_currency_unit` as being a little clearer

app/presenters/applicant_account_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+  def initialize(applicant_bank = nil)
+    @applicant_bank = applicant_bank
+  end
+
+  def main_account_holder_name
+    @applicant_bank.bank_account_holders.last.full_name
+  end
+
+  def main_account_holder_address
+    @applicant_bank.bank_account_holders.last.addresses.first.values.join(', ')
```
reggieb: That's a very long chain of methods without any checks that they each exist.

Also `@applicant_bank.bank_account_holders.last` is used in a couple of methods. If the definition of `main bank account` changes (to `first` for example), both would need to be updated. I'd suggest

```ruby
def main_account_holder
  @applicant_bank.bank_account_holders.&last
end

def main_account_holder_name
  main_account_holder.&full_name
end

def main_account_holder_address
  main_account_holder.&addresses.&first.&values.&join(', ')
end
```
For the last method - even better may be to have an `Address` presenter:
```ruby
def main_account_holder_address
  address = AddressPresenter.new(main_account_holder.&addresses.&first)
  address.values
end
```

app/presenters/applicant_account_presenter.rb
```diff
@@ -0,0 +1,21 @@
+class ApplicantAccountPresenter
+  def initialize(applicant_bank = nil)
```
reggieb: If no `applicant_bank` is passed in, all the other methods fail. So using `= nil` here moves the errors to the individual methods where the underlying problem (that no `applicant_bank` was passed on initialization) is less obvious.

I'd suggest:
```ruby
  def initialize(applicant_bank)
```

app/presenters/applicant_account_presenter.rb
```diff
@@ -0,0 +1,21 @@
...
+    @applicant_bank.bank_account_holders.last.addresses.first.values.join(', ')
+  end
+
+  def bank_accounts
+    @applicant_bank.bank_accounts
+  end
+
+  def name
+    @applicant_bank.name
+  end
```
reggieb: When you are just calling a method on a sub-object, a useful pattern is to use `attr_reader` to create a getter for the object, and then `delegate` to access the individual methods. Especially in a presenter as that also gives you the [options `:prefix` and `allow_nil`](https://api.rubyonrails.org/classes/Module.html#method-i-delegate). So:
```ruby
  attr_reader :applicant_bank
  delegate :name, :bank_accounts, to: :applicant_bank
```

app/presenters/applicant_account_presenter.rb
reggieb: Also there is no `applicant_bank` model. Should that be `bank_provider` to maintain consistency.

app/presenters/applicant_account_presenter.rb
reggieb: Actually, thinking about this a little more: I don't think the definition of "Main account holder" belongs in a presenter. That belongs in the Bank Provider. I'd suggest updating the model:

```ruby
class BankProvider < ApplicationRecord
  ...
  has_one :main_account_holder, class_name: :BankAccountHolder, -> { order: (created_at: :desc) }
```

reggieb:
Can you hold fire on merging this. I'm mid-review and need to track something through.

spec/presenters/applicant_account_presenter_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  let(:addresses) do
+    [{ address: Faker::Address.building_number,
+       city: Faker::Address.city,
+       zip: Faker::Address.zip }]
+  end
+  let!(:applicant_bank_provider) { create(:bank_provider, applicant_id: applicant.id) }
+
+  let!(:applicant_bank_account_holder) do
+    create(:bank_account_holder, bank_provider_id: applicant_bank_provider.id,
+                                 addresses: addresses)
```
reggieb: `BankAccountHolder` doesn't have an `addresses` instance method.

app/presenters/applicant_account_presenter.rb
reggieb: Also, the `BankAccountHolder` model doesn't have an `addresses` field.

reggieb:
Now I see how the addresses are being gathered, my only blockers is that I think the `ApplicantAccountPresenter` initializer should be defined as:

```ruby
def initialize(applicant_bank)
```

And that the definition of the "Main" account holder belongs in the model and not the presenter.

The rest of my comments should be considered as suggestions.

PR #82, opened 2018-11-19
AP-177 improve proceeding search behaviour

reggieb:
[Jira AP-177](https://dsdmoj.atlassian.net/browse/AP-177)

Previously, each time the search string changed, a JSON call was being made to gather all the Proceeding Type data. In this PR, the JSON data is gathered once (on completion of page load). As the load only happens once, a debounce has not been added to reduce the frequency that it is being called.

A search tool ([js-search](https://github.com/bvaughn/js-search)) is used rather than basic JavaScript `indexOf` search. This provides a couple of advantages:

- It allows searching across multiple fields (so `code`, `meaning` and `description` rather than just `description`).
- The results are ordered by how good a match was made.  

The HTML for each Proceeding Type was generated using a JavaScript template. They are now generated via an ERB template. This was done to:

- Make rendering inline with the rest of the app (this was the only place JavaScript templates were used).
- Provide better degradation behaviour. If JavaScript is disabled, all the Proceeding Types are displayed.


app/views/providers/legal_aid_applications/_proceeding_type.html.erb
```diff
@@ -0,0 +1,24 @@
+<div id="<%= proceeding_type.code %>" class="govuk-grid-row proceeding-item" style="display: block;">
```
lostie: **[not a blocker/minor]** This was probably in the original JS template, but if we can avoid the use of embedded styling (`style="display: block;"`) in the page it would be 🏅 .

reggieb: I kept this bit of markup as it was, but I've just tested and in fact the style is defaulting to `block` anyway. So I've removed it.

lostie:
**[not a blocker]** I've raised https://github.com/ministryofjustice/laa-apply-for-legal-aid/issues/83 as something to raise awareness that anything that includes substantial JS functionality should probably have their own Unit Tests. Something for us to consider going forward.

app/assets/javascripts/proceeding_types.js
```diff
@@ -1,56 +1,45 @@
...
+
+    if (inputText.length > 2) {
+      let codes = search.search(inputText).map(function(obj) {return obj["code"];});
+
+      if (codes.length > 0) {
+        // Iterate through each code, find the matching element, move it to the
+        // top and display it
+        $.each(codes.reverse(), function(_i, code){
+          let element = $('#' + code);
+          let parent = element.parent("div");
```
lostie: **[not a blocker]** For more reliability, consider defining the ID of the object to prepend the _element_ to as a **data** attribute (this avoids trying to navigate through the DOM and just _link_ the two objects semantically), e.g:

```ruby
# proceeding type partial view
<div id="<%= proceeding_type.code %>" class="govuk-grid-row proceeding-item" data-list-container="proceeding-list">
...
</div>

```

```js
# this JS file
let element = $('#' + code);
let container = element.data("list-container");
element.detach().prependTo(container);
```

app/views/providers/legal_aid_applications/_proceeding_type.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  </div>
+  <% end %>
+</div>
+
+<script>
+  <%#
+    Hidding div as soon as possible to minimize appearance on page transition.
+    `raw ... to_json` to keep erblint happy
+  %>
+  $(<%= raw "##{proceeding_type.code}".to_json %>).hide();
```
lostie: ❓ Is the intention here to achieve something functionally equivalent to a:

```js
$( document ).ready(function() {
    $('.proceeding-item').hide();
});

reggieb: Yes and no. The two do the same thing - hide the proceeding item elements. My first version had exactly that code (`$('.proceeding-item').hide();`) at the start of `app/assets/javascripts/proceeding_types.js`. 

However, there is an issue in that it can take a fraction of a second for the items to be hidden after they are rendered, and on Chrome this is visible as a flash of green on page load.

This code is an attempt to minimise this effect. It has two effects: first, it hides each element as soon as it is rendered (rather than after the whole page has rendered). Second, it ensures there is only ever one element visible on the page, so the size of the "green" block is minimised.

If I'm honest, it's not a particularly elegant bit of code and I'd welcome a better suggestion. It's made more ugly by erblint insisting that any ruby code rendered into a `<script>` block must have `to_json` called on it.

lostie:
🚢

PR #84, opened 2018-11-20
[AP-175] [SPIKE] LAA Portal integration

lostie:
## What

**This is not ready to be merged, mainly for the reasons described in the JIRA ticket. It should serve though as the base to the actual integration when the other environments are setup.**

[AP-175](https://dsdmoj.atlassian.net/browse/AP-175)

Also: [AP-10](https://dsdmoj.atlassian.net/browse/AP-10)

* Sets up integration with the _dev_ environment of [LAA Portal](https://github.com/ministryofjustice/laa-portal) which will enable **providers** to authenticate.

**Outcomes/TODOS:**

* Need to chase Get Access/LAA Portal team to setup staging/prod environments and collect all the credentials necessary to setup our app to connect to those LAA Portal environments.

**Additional things:**

* Renames CircleCI cache and forces the bundle path so that CircleCI can actually make use of previously cached bundles.
* Enforces Rails generators to use RSpec as the default test framework (was TestUnit so far)

## Checklist


app/controllers/saml_sessions_controller.rb
```diff
@@ -0,0 +1,7 @@
+class SamlSessionsController < Devise::SamlSessionsController
+
+  # def create
```
mpw5: Did you mean to leave this commented code here?

config/application.rb
```diff
@@ -13,11 +13,18 @@ class Application < Rails::Application
...
+    config.x.application.host             = ENV['HOST']
     config.x.benefit_check.service_name   = ENV['BC_LSC_SERVICE_NAME']
     config.x.benefit_check.client_org_id  = ENV['BC_CLIENT_ORG_ID']
     config.x.benefit_check.client_user_id = ENV['BC_CLIENT_USER_ID']
     config.x.benefit_check.wsdl_url       = ENV['BC_WSDL_URL']
 
+    config.x.laa_portal.idp_slo_target_url             = ENV['LAA_PORTAL_IDP_SLO_TARGET_URL']
+    config.x.laa_portal.idp_sso_target_url             = ENV['LAA_PORTAL_IDP_SSO_TARGET_URL']
+    config.x.laa_portal.idp_cert = ENV['LAA_PORTAL_IDP_CERT']
+    # config.x.laa_portal.idp_cert_fingerprint           = ENV['LAA_PORTAL_IDP_CERT_FINGERPRINT']
```
mpw5: Did you mean to leave this commented code here?

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
   root to: 'home#index'
 
+  devise_for :providers, controllers: { saml_sessions: 'saml_sessions' }
+  devise_for :users
```
mpw5: I think we talked about removing this because it's not needed at present. Did we decide to keep it in the end?

config/environments/development.rb
```diff
@@ -56,5 +56,6 @@
   # routes, locales, etc. This feature depends on the listen gem.
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
-  Rails.application.routes.default_url_options[:host] = ENV['HOST'] || 'localhost:3002'
+  config.x.application.host = ENV['HOST'] || 'localhost:3002'
```
reggieb: I may be wrong here:
There is a `PORT` environment variable too. Should the development host be `'localhost'` with no port. And then `Rails.application.routes.default_url_options[:port]` defined separately?

features/support/env.rb
```diff
@@ -101,4 +102,12 @@
 
 Before do |_scenario|
   load Rails.root.join('db/seeds.rb')
+  Warden.test_mode!
 end
+
+After do |_scenario|
+  Warden.test_reset!
+end
```
reggieb: Not a blocker: Would [an around block](https://relishapp.com/cucumber/cucumber/docs/writing-support-code/around-hooks) be good here, as it would have both the set and reset actions within the same block:
```ruby
Around do |_scenario, block|
  Warden.test_mode!
  block.call
  Warden.test_reset!
end
```

laa-portal.cert
```diff
@@ -0,0 +1 @@
+MIIDODCCAiCgAwIBAgIJAJSRXNZd/kGVMA0GCSqGSIb3DQEBBQUAMBwxGjAYBgNVBAoTEUxBQSwgQ049UG9ydGFsLVNQMB4XDTE4MDEyNTE2MjE1NloXDTI4MDEyNTE2MjE1NlowHDEaMBgGA1UEChMRTEFBLCBDTj1Qb3J0YWwtU1AwggEiMA0GCSqGSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDRa4Yfdy7Isbb58LZ4GsXQiDDJIaVXrYHsda3PcKjN8h6svGDSNjWmk4+Q4cCQHPB92EKkHsI7uqpIDooIXeS4Ko71hGsfy+/H2UC0nrqptoTZLNK0tTRBpjbmdhtz5giM919N6RfKt2rrw577XzqOKzshiomFu08mM9pP/OPXmGQm5D4H4g4Y7u8hkBqg8mv48hYJxLLRnl9Y1JugkVDA5Mm4V/GRMS6ok6kGa2uWitA8keRlfRx6xsFgNluay1uosDo2tjQBwBXyaVO9dfbJDxceKKz3wlJby6Ma5i45i41WXXFagSnEg89mCKL/AP7QyJQGgTQo0eeSycK0XNv/AgMBAAGjfTB7MB0GA1UdDgQWBBRdWLZVn0h0Lla/y3Cn1Rm2gswARTBMBgNVHSMERTBDgBRdWLZVn0h0Lla/y3Cn1Rm2gswARaEgpB4wHDEaMBgGA1UEChMRTEFBLCBDTj1Qb3J0YWwtU1CCCQCUkVzWXf5BlTAMBgNVHRMEBTADAQH/MA0GCSqGSIb3DQEBBQUAA4IBAQBA75RB+4NXKbOV5pxpz6ZGv6rxZbCVSqq5yowyQFFw5jv6W/Ga3Rfumgl4On53cKrzhRuDLbneNrjvcB0rQRSnBms0X/Otc8asXABLOf4W3ONKPn5O0jd5En8D+SauolCfpfomcRTj6XxxQe6NaIgoWf9SkKiDxWy7AYQnn6Y+c4E+r0lLcYKmC+GXKyzCjdEj418f0Zflh3+nklsK/oNtpC8+K/h568TQHEWCIbMjYQ0T5ai8M9LmqGSFFH3ScQlWx++93rWOfAT+jF4EOrkL0aFULyatalAnUuJNhA1W1Fk8qUZiOTQKlUuYu/4IyY2t6YHBKgId0r0pfXekCPBD
```
reggieb: Should this be in Git?

spec/support/shared_examples/provider_auth_shared_examples.rb
```diff
@@ -0,0 +1,6 @@
+RSpec.shared_examples 'a provider not authenticated' do
+  it 'should redirect the user to the login page' do
+    perform_request
+    expect(response).to redirect_to(new_provider_session_path)
+  end
```
reggieb: Two things:

1. If you pass the request into the shared example, you won't have to force the request to be wrapped as `perform_request`
2. If you log the user off before calling the request, you won't need to separate the logged on behaviour into a separate context. You'll be able to just have a general `before { login_as create(:provider) }` at the start of each spec.

So:
```ruby
RSpec.shared_examples 'a provider not authenticated' do |current_request|
   it 'should redirect the user to the login page' do
     sign_out :provider
     current_request
     expect(response).to redirect_to(new_provider_session_path)
  end
end
```
Then in the specs you call:
```ruby
it_behaves_like 'a provider not authenticated', perform_request
```

laa-portal.cert
lostie: Speaking with @bleach about this, the idea I got is that this is the public certificate and therefore there's no security implications. I think the fact that https://samlmock.dev.legalservices.gov.uk/metadata exposes that same information also re-enforced that idea.

reggieb: Fair enough. Thought it worth checking.

reggieb:
This has been superseded and therefore is no longer needed.

PR #86, opened 2018-11-20
Add basic state machine to the applications model

lostie:
## What

🚧 

[AP-197](https://dsdmoj.atlassian.net/browse/AP-197)

Right now it sets a default state when an application is created and changes the state when the provider submits the application.

**TODOS:**

* _state_ names subject to change
* Do we required to also mark the actual timestamp at which the application changed to that state?

## Checklist


app/models/legal_aid_application.rb
```diff
@@ -10,6 +10,12 @@ class LegalAidApplication < ApplicationRecord
 
   validate :proceeding_type_codes_existence
 
+  state_machine initial: :initiated do
```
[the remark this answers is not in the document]
reggieb: I'd also suggest moving the state machine logic into a separate object and mixing it into the model as my experience is that it can bloat the model very quickly.

Gemfile
```diff
@@ -14,6 +14,7 @@ gem 'sass-rails', '~> 5.0'
 gem 'savon', '~> 2.12.0'
 gem 'sentry-raven'
 gem 'simple_command', '~> 0.0.9'
+gem 'state_machines-activerecord', '~> 0.5.2'
```
tmckernan: why have you selected to use the gem `state_machines-activerecord` over [AASM](https://github.com/aasm/aasm)?

lostie: The gem itself is [state_machines](https://github.com/state-machines/state_machines), `state_machines-activerecord` is the specific `activerecord` integration. As for the choice itself, there's a bunch of state machines gems out there such as:

- [state_machine](https://github.com/pluginaweek/state_machine)
- [state_machines](https://github.com/state-machines/state_machines)
- [aasm](https://github.com/aasm/aasm)
- [statesman](https://github.com/gocardless/statesman)
- [transitions](https://github.com/troessner/transitions)
- ...

I've used 3 or 4 out of the ones in the list before, and there's not many differences between most of them, the basic features I was expected to support:

* state machine definition
* helper methods such as transition methods and <state>?
* callbacks

are supported by it. 

So, the question is, is `AASM` (or any other gem) giving us anything else that **we need** and this one does support? If that's the case, I'm more than happy to change to it.

app/models/legal_aid_application.rb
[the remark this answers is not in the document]
lostie: I don't see any reason, just now, to pre-optimise and move the state machine to a new file. When the bloat happens, I agree, we can move it across to a separate file.

lostie:
♻️

app/views/providers/check_your_answers/index.html.erb
```diff
@@ -3,7 +3,7 @@
     <h1 class="govuk-heading-xl">Check your answers</h1>
     <h2 class="govuk-heading-l">Coming soon</h2>
     <p class="govuk_body">This doesn't exist yet, but click Continue to send an email with a link to the citizen application</p>
+
+    <%= button_to 'Submit', confirm_providers_legal_aid_application_check_your_answers_path(@legal_aid_application), class: 'govuk-button' %>
```
reggieb: Does that button require `method: :post` as the confirm endpoint is a POST?

PR #91, opened 2018-11-22
Accessibility runner

lostie:
## What

🚧 

Proof of Concept to incorporate some accessibility checks on RSpec feature specs making use of the Google Accessibility Developer Tools.

screen shot 2018-11-22 at 17 33 40


mpw5:
This is very out of date, and we now have Webhint which is doing something similar. I'm going to close this PR and delete the branch.

PR #92, opened 2018-11-23
AP 183 provider back navigation

reggieb:
[JIRA AP-183](https://dsdmoj.atlassian.net/browse/AP-183)

Provided a single place where the sequence of backward steps could be defined.

## What I did:

- Moved applicant email update to its own controller - so that there was a 1:1 relationship between form and controller through the Provider workflow
- Moved mounting of address controllers to within `application` rather than `applicant` so that behaviour consistent through the Provider workflow
- Changed the step definition so that it could use `controller_name`.
- Changed the step logic so that no longer order dependant.
- Modified the step logic to allow both forward and backward paths to be defined.
- Updated the back links in the views to use the new logic.


app/controllers/concerns/providers/steppable.rb
```diff
@@ -2,29 +2,71 @@ module Providers
...
+        @current_back_method ||= STEPS.dig controller_name.to_sym, :back
+      end
+
+      def with_legal_aid_application_if_needed(current_method)
+        return if paths_not_requiring_legal_aid_application_instance.include?(current_method)
+        legal_aid_application
+      end
+
+      def paths_not_requiring_legal_aid_application_instance
+        %i[providers_legal_aid_applications_path providers_root_path]
```
Kalagan: Should this be defined as a constant below the `STEPS` constant?

app/controllers/providers/emails_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      email_params.merge(model: applicant)
+    end
+
+    def email_params
+      params.require(:applicant).permit(:email)
+    end
+
+    def applicant
+      @applicant ||= legal_aid_application.applicant
+    end
```
Kalagan: You don't need this method thanks to the delegate you added in `ApplicationDependable`

app/helpers/providers_helper.rb
```diff
@@ -0,0 +1,5 @@
+module ProvidersHelper
+  def provider_back_link(text = 'Back')
```
lostie: **[nit pick/not a blocker]** Consider setting the default to something that's configured in a locale? This is tying the code to English only.

lostie:
👏 🏆 Good work!

app/helpers/providers_helper.rb
reggieb: Sorry - I missed that. I'll slip it into the work I'm doing next.

PR #94, opened 2018-11-23
Feature/ap178 storing users consent choice

willc-work:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-178)

Add an open_banking_consent and timestamp to legal_aid_applications table to store the users choice
Currently there is no path for the users doesn't provide consent so it redirects to a plain page.
Refactor the form page to make use of the helpers and partials

## Checklist


app/models/legal_aid_application.rb
```diff
@@ -7,9 +7,9 @@ class LegalAidApplication < ApplicationRecord
   has_one :benefit_check_result
 
   before_create :create_app_ref
+  before_save :set_consent_choice_date
 
-  attr_reader :proceeding_type_codes
-
+  attr_reader :proceeding_type_codes, :open_banking_consent, :consent_choice_timestamp
```
reggieb: You've added `open_banking_consent` and `consent_choice_timestamp` as attributes (in the migration). So they should not have a getter defined here.

app/models/legal_aid_application.rb
```diff
@@ -61,4 +61,8 @@ def proceeding_type_codes_existence
   def create_app_ref
     self.application_ref = SecureRandom.uuid
   end
+
+  def set_consent_choice_date
```
reggieb: The attribute is `consent_choice_timestamp`, so the method should match - `set_consent_choice_timestamp`

db/migrate/20181123111150_add_consent_to_legal_aid_applications.rb
```diff
@@ -0,0 +1,6 @@
+class AddConsentToLegalAidApplications < ActiveRecord::Migration[5.2]
+  def change
+    add_column :legal_aid_applications, :open_banking_consent, :string
+    add_column :legal_aid_applications, :consent_choice_timestamp, :datetime
```
reggieb: The rails convention is to use the `_at` suffix with datetime fields. Also the consent type being timestamped should be identified. So perhaps: `open_banking_consent_given_at`

db/migrate/20181123111150_add_consent_to_legal_aid_applications.rb
```diff
@@ -0,0 +1,6 @@
+class AddConsentToLegalAidApplications < ActiveRecord::Migration[5.2]
+  def change
+    add_column :legal_aid_applications, :open_banking_consent, :string
```
reggieb: This should be a boolean and not a string. If you keep it as a string, the code is going to be littered with string comparisons each time code need to determine whether consent was given.

app/views/shared/forms/_check_boxes.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-form-group">
+  <fieldset class="govuk-fieldset" aria-describedby="open-banking-consent">
+    <div class="govuk-checkboxes">
+      <div class="govuk-checkboxes__item">
+        <%= form.check_box :open_banking_consent, { class: 'govuk-checkboxes__input' }, 'YES', 'NO' %>
+        <%= form.label :open_banking_consent, t('.label.html'), class: 'govuk-label govuk-checkboxes__label' %>
+      </div>
+    </div>
+ </fieldset>
+</div>
```
reggieb: If this is to be a generic bit of code used where checkboxes are required, the field name and the check box options need to be passed into the partial. At the moment, the "shared" partial can only be used for the field `open_banking_consent`.  So something like:
```erb
<div class="govuk-checkboxes">
  <% field_names_and_choices.each do | name , choices|
    <div class="govuk-checkboxes__item">
      <%= form.check_box name, { class: 'govuk-checkboxes__input' }, *choices %>
      <%= form.label :name, t('.label.html'), class: 'govuk-label govuk-checkboxes__label' %>
    </div>
  <% end %>
</div>
```
Then you can call it with:
```
<%= render partial 'shared/forms/check_boxes', field_names_and_choices: {  open_banking_consent: ['YES', 'NO'] } %>
```
Which can be wrapped in a helper as others are.
Also, the form group can be set with the existing helper method.

app/forms/applicants/open_banking_consent_form.rb
```diff
@@ -0,0 +1,9 @@
+module Applicants
+  class OpenBankingConsentForm
+    include BaseForm
+
+    form_for LegalAidApplication
+
+    attr_accessor :open_banking_consent, :consent_choice_timestamp
```
reggieb: `consent_choice_timestamp` isn't used by this form as it is set via a callback in the model. So just:
```ruby
attr_accessor :open_banking_consent
```

app/controllers/citizens/consents_controller.rb
```diff
@@ -1,5 +1,31 @@
...
   class ConsentsController < BaseController
-    def show; end
+    def show
+      @applicant = current_applicant
+    end
+
+    def create
+      @form = Applicants::OpenBankingConsentForm.new(edit_params)
+
+      case params[:legal_aid_application][:open_banking_consent]
```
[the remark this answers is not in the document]
willc-work: 👍

db/migrate/20181123111150_add_consent_to_legal_aid_applications.rb
willc-work: I initially used open_banking_consented_at but that implied consent and as consent can be denied I revised the name. I will rethink it.

db/migrate/20181123111150_add_consent_to_legal_aid_applications.rb
willc-work: 👍

app/views/shared/forms/_body.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
```
reggieb: The indentation here is out of alignment

spec/requests/citizens/consents_spec.rb
```diff
@@ -1,11 +1,42 @@
...
+      post citizens_consent_path, params: params
+    end
+
+    context 'when consent is granted' do
+      let(:params) { { legal_aid_application: { open_banking_consent: 'YES' } } }
+
+      it 'redirects to new action' do
+        expect(response).to redirect_to(applicant_true_layer_omniauth_authorize_path)
+        expect(unescaped_response_body).to include('true')
+      end
```
reggieb: I think there should also be a spec to check that the decisions have been persisted. Something like:
```ruby
it 'records decision on the legal aid application' do
  legal_aid_application.reload
  expect(legal_aid_application.open_banking_consent).to eq(true)
  expect(legal_aid_application.open_banking_consent_choice_at.to be_between(2.seconds.ago, 1.second.from_now)
end
```

spec/requests/citizens/consents_spec.rb
```diff
@@ -1,11 +1,42 @@
...
+    end
+
+    context 'when consent is not granted' do
+      let(:params) { { legal_aid_application: { open_banking_consent: 'NO' } } }
+
```
reggieb: Again - add a spec for how the choice is persisted.

lostie:
Just bare in mind that once this consent is in place, it should be used as a pre-condition to pull redirect the user to the open banking page and to pull the information from open banking for the user.

app/views/citizens/consents/show.html.erb
```diff
@@ -2,19 +2,18 @@
...
-        <%= render partial: 'shared/forms/check_boxes', locals: { form: form } %>
-      <% end %>
+  <%= govuk_form_group(
+        input: :open_banking_consent
+      ) do %>
+    <%= govuk_fieldset_header t('.field_set_header'), padding_below: 6 %>
+    <%= govuk_body t('.body') %>
+    <%= render partial: 'shared/forms/detailed_list', locals: { translation_path: 'citizens.consents.show' } %>
+    <%= render partial: 'shared/forms/check_boxes', locals: { field_name: :open_banking_consent, form: form, checked: 'YES', unchecked: 'NO' } %>
+    <%#= render partial: 'shared/forms/check_boxes', locals: { field_names_and_choices: {  open_banking_consent: ['YES', 'NO'] } } %>
```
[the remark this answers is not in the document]
mpw5: 👍

app/views/shared/forms/_check_boxes.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="govuk-checkboxes">
+  <div class="govuk-checkboxes__item">
+    <%= form.check_box field_name, { class: 'govuk-checkboxes__input' }, checked, unchecked %>
+    <%= form.label field_name, t('.label.html'), class: 'govuk-label govuk-checkboxes__label' %>
```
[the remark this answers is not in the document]
mpw5: 👍

app/views/shared/forms/_body.html.erb
mpw5: 👍

spec/requests/citizens/consents_spec.rb
mpw5: 👍

spec/requests/citizens/consents_spec.rb
mpw5: 👍

app/controllers/citizens/consents_controller.rb
```diff
@@ -1,5 +1,31 @@
...
+
+      @form.save
+
+      case @form.open_banking_consent
+      when 'true'
+        redirect_to applicant_true_layer_omniauth_authorize_path
+      when 'false'
+        render plain: 'Landing page: No Consent provided'
+        # redirect_to citizens_information_path
+      end
```
reggieb: `case` statements make sense if there are more than two options, and/or you want to do multiple `===` comparisons. `open_banking_consent` is a boolean! You should be able to do:
```
if @form.open_banking_consent?
  ...
else
  ...
end
```
Obviously you'll need to ensure `open_banking_consent?` return `false` for `"false"` Have a look at `ActiveModel::Type::Boolean.new.cast`

app/controllers/citizens/consents_controller.rb
```diff
@@ -1,5 +1,31 @@
 module Citizens
   class ConsentsController < BaseController
-    def show; end
+    def show
+      @applicant = current_applicant
```
reggieb: Looking at the view, it is no longer using `@applicant`. So this can be:
```ruby
def show; end
```

config/locales/en.yml
```diff
@@ -187,6 +187,17 @@ en:
...
+      show:
+        open_banking_consent:
+          label:
+            html: <b>I agree for you to check 3 months of bank transactions</b> <br /> using temporary access to my online banking
+        field_set_header: 'Do you agree to share your bank transactions with us?'
+        body: 'We only use your bank transactions to support your legal aid application. We do not use the information for anything else.'
+        list_heading: 'Find out how we use and store your personal data'
+        list_1: 'We do not save your online banking username and password'
+        list_2: 'We do not have any ongoing access to your bank accounts'
+        list_3: 'We only use your details to assess your eligibility for legal aid'
```
reggieb: Worth noting that you don't have to put delimiters around text values in YAML files unless some interpolation is required or the text contains unusual characters. So this will work just as well:
```yaml
list_1: We do not save your online banking username and password
list_2: We do not have any ongoing access to your bank accounts
list_3: We only use your details to assess your eligibility for legal aid
```

spec/requests/citizens/consents_spec.rb
```diff
@@ -1,11 +1,53 @@
 require 'rails_helper'
 
-RSpec.describe 'citizen consents request test', type: :request do
-  describe 'GET /citizens/consent' do
-    before { get citizens_consent_path }
+RSpec.describe Applicants::OpenBankingConsentForm, type: :request do
```
reggieb: This spec is describing the output from a controller action. It is not a direct test of the form. So:
```
RSpec.describe Citizens::ConsentsController, type: :request do
```

app/controllers/citizens/consents_controller.rb
willc-work: removed the case statement and replaced with an if/else, it doesn't use the boolean operator though

PR #95, opened 2018-11-23
Ap 191 check your answers

mpw5:
## What

[AP-197](https://dsdmoj.atlassian.net/browse/AP-197)

Implements the Check Your Answers page for providers:

- the page already exists as a placeholder, this adds content to that page;
- tests already exist, this refines them to test the new content;
- the check your answers design pattern does not currently exist in GOV.UK Frontend, so import the styling from the GOV.UK Prototype Kit as an interim solution;
- renames all references to check_your_answers to check_provider_answers as required in the design.

## Checklist


app/controllers/providers/check_provider_answers_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    def confirm
+      legal_aid_application.provider_submit!
+      CitizenEmailService.new(legal_aid_application).send_email
+    end
+
+    private
+
+    def html_address(address)
+      [address.organisation, address.address_line_one, address.address_line_two, address.city, address.postcode].compact.reject(&:blank?).join('</br>').html_safe
+    end
```
reggieb: Two things ...

First, this method is very similar to `address#full_address`. I'd suggest modifying that method and using it here. Something like:
```ruby
class Address < ApplicationRecord
  ...
  def full_address(joined_by: ', ')
    [organisation, address_line_one, address_line_two, city, postcode].compact.reject(&:blank?).join(joined_by)
  end
```
Then here you can do:
```
  address.full_address(joined_by: '</br>')
```

Second you must not use `.html_safe` here as address is user input. So it is not safe to blanket sanitize it like this. I expect the thing you are trying to sanitize is `</br>` so just sanitize that:

```ruby
sanitize address.full_address(joined_by: '</br>'), tags: ['br']
```
As you are doing this in the controller, you'll probably need to `include ActionView::Helpers::SanitizeHelper`. Alternatively and probably better:

1. Assign the address like this: `@address = @applicant.addresses.first`
2. Then use a helper method to display the address as needed (`sanitize` will be available within a helper).

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -0,0 +1,127 @@
...
+        </dt>
+        <dd class="app-check-your-answers__answer">
+          <%= @applicant.first_name %>
+        </dd>
+        <dd class="app-check-your-answers__change">
+          <a href="#">
+            Change<span class="govuk-visually-hidden"><%= t '.section_2.first_name' %></span>
+          </a>
+        </dd>
+      </div>
```
reggieb: Each of the contents div have the same structure - I'd suggest using a partial.

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -0,0 +1,127 @@
...
+    </h2>
+
+    <p>
+      <%= t '.section_3.text' %>
+    </p>
+
+  </div>
+</div>
+
+<%= button_to 'Submit', confirm_providers_legal_aid_application_check_provider_answers_path(@legal_aid_application), class: 'govuk-button' %>
```
reggieb: As you're using translations for everything else - why not here and for the back button too! That is, both the word "Submit" and "Back" should be defined via a translation. I'd suggest adding them to the `generic` section: `t 'generic.back'` for example.

app/assets/stylesheets/providers/_check-your-answers.scss
```diff
@@ -0,0 +1,88 @@
+@import "govuk-frontend/settings/all";
```
lostie: Just a couple of questions:

* Is there any particular reason for having the filename with a prefixed underscore? Also, I would suggest we keep the filename to the more conventional `check_your_answers.scss`. 
* Should all the imports set here already been imported by the global [application.scss](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/assets/stylesheets/application.scss#L24) one?

app/controllers/providers/check_provider_answers_controller.rb
```diff
@@ -0,0 +1,16 @@
+module Providers
+  class CheckProviderAnswersController < BaseController
+    include Providers::ApplicationDependable
+
+    def index
+      @proceeding = @legal_aid_application.proceeding_types.first
```
lostie: **[nit pick/not a blocker]** **For consistency**, given we're using the helper method `legal_aid_application` in the confirm action, we should probably use the same one here instead of using the instance variable directly. **Definitely not a blocker though.**

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -0,0 +1,43 @@
+<% content_for :navigation do %>
+  <%= link_to (t 'generic.back'), edit_providers_legal_aid_application_applicant_path, class: 'govuk-back-link' %>
```
lostie: **[nit pick/not a blocker]** Was this supposed to be `link_to t('generic.back'),` ? The brackets surrounding the translation seem unnecessary.

lostie:
I spoke with Simon offline about the display of the page, and I think it is expected for all the answers to be aligned the same way, regardless if they're in a different _section_, in this case the proceeding should be aligned the same way as the applicant details answers.

screen shot 2018-11-26 at 10 42 24

app/views/providers/check_provider_answers/_form.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= form_with(model: check_provider_answers, local: true) do |form| %>
```
lostie: ❓ I can't see this template being used anywhere. Is it actually required? If not, let's ✂️ it!

app/views/providers/check_provider_answers/index.html.erb
lostie: I've seen this is actually across the whole template.

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    </dl>
+
+    <h2 class="govuk-heading-m">
+      <%= t '.section_2.heading' %>
+    </h2>
+
+    <dl class="app-check-your-answers app-check-your-answers--long">
+      <%= render partial: 'check_your_answers_item', locals: { question: (t '.section_2.first_name'), answer: @applicant.first_name } %>
+      <%= render partial: 'check_your_answers_item', locals: { question: (t '.section_2.last_name'), answer: @applicant.last_name } %>
+      <%= render partial: 'check_your_answers_item', locals: { question: (t '.section_2.dob'), answer: @applicant.date_of_birth.strftime('%d/%m/%Y') } %>
```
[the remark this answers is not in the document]
reggieb: If you are playing with date and time formats, these two commands are very useful:
```ruby
# To display the current date and time in all the formats in your current environment:
Time::DATE_FORMATS.keys.each{|k| puts [k,Date.today.to_s(k)].join(':- ')}
Time::DATE_FORMATS.keys.each{|k| puts [k,Time.now.to_s(k)].join(':- ')}
```

app/assets/stylesheets/providers/_check-your-answers.scss
mpw5: I just copied and pasted the file from the GDS Prototype Kit, where it had an underscore. I've removed it for consistency.

The styling of the page fails if those imports aren't included. I discussed it with someone at GDS on x-gov slack and they confirmed that they were explicitly needed.

app/controllers/providers/check_provider_answers_controller.rb
mpw5: 👍

app/views/providers/check_provider_answers/index.html.erb
mpw5: 👍

app/views/providers/check_provider_answers/_form.html.erb
mpw5: 👍

app/views/providers/check_provider_answers/index.html.erb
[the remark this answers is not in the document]
mpw5: 👍

PR #96, opened 2018-11-26
AP-194 provider home navigation

tmckernan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-194)

- add a link to header to link to  provider home page 

## Checklist

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -17,10 +17,7 @@
       </a>
     </div>
     <div class="govuk-header__content">
-
-      <a href="#" class="govuk-header__link govuk-header__link--service-name">
-        <%= t('layouts.application.header.title') %>
-      </a>
+      <%= home_link %>
     </div>
   </div>
 </header>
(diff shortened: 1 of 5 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -18,7 +18,7 @@
     </div>
     <div class="govuk-header__content">
 
-      <a href="#" class="govuk-header__link govuk-header__link--service-name">
+      <a href="<%= providers_legal_aid_applications_url %>" class="govuk-header__link govuk-header__link--service-name">
```
lostie: **[not a blocker/nit pick]** 💭  If we're regulary embedding ERB code into the raw HTML, specially in these link cases, might be worth using the Rails helpers instead (e.g. `link_to`).

⚠️ Also, I'm not sure if this is obvious for everyone, but this URL should **be only for providers**, so if you're on the **citizen/applicant side** of the application the URL **should be different**, as far as I'm aware.

lostie: 💭 Might be worth considering adjusting the layout to fork between **providers** and **applicant**.

PR #97, opened 2018-11-26
AP-196: Redirect back to check your answers when checking answers

reggieb:
[JIRA AP-196](https://dsdmoj.atlassian.net/browse/AP-196)

The main change this PR adds is to add the Check Answers functionality. Now, once a provider gets to the Check Answers page, they can click on a Change link and they will be taken to the form to change that associated attribute. On form submission they will be returned to the Check Answers page.

## In detail:

- Added a new state `checking_answers` and use it to modify steppable behaviour
- modified forward actions so redirect to the Check Answers page while checking answers.
- modified back actions so redirect to the Check Answers page while checking answers.
- Over rode above forward behaviour for `address_lookups` so next steps (of actually modify the address) would complete before returning to Check Answers page
- Refactored address pages so they comply to the new show/update pattern
- Added a Proceedings Types controller, so that they can be modified via checking answers, the same way as other attributes.
- Created a set of cucumber specs to test the new Check Answer journeys.
- Updated cucumber steps to make it a little easier to get to a particular point in the flow, and some tidying up of steps.


app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -12,18 +12,58 @@
...
+      <%= check_provider_answer_link(
+            name: 'address',
+            url: providers_legal_aid_application_address_lookup_path,
+            question: t('.section_2.address'),
+            answer: address_with_line_breaks(@address)
+          ) %>
+
+      <%= check_provider_answer_link(
+            name: 'email',
+            url: providers_legal_aid_application_email_path,
```
Kalagan: This needs to go to the client details page now (`providers_legal_aid_application_applicant_path`).
Would you mind changing it? :smiley:

reggieb: Yes - I noticed that. If your PR is merged in, I'll do it as part of the rebase work.

features/providers/civil_application_journey.feature
```diff
@@ -16,108 +16,192 @@ Feature: Civil application journeys
...
+  Scenario: I want to return to the check your answers page without changing proceeding type
+    Given I complete the journey as far as check your answers
+    And I click Check Your Answers Change link for 'Proceeding Type'
+    And I search for proceeding 'Application for a care order'
+    Then proceeding suggestions has results
+    Then I click link "Back"
+    Then I should be on the Check Your Answers page
+
+  @javascript @vcr
+  Scenario: I want to change email from the check your answers page
```
Kalagan: With `email` now being on **Client's details** page, maybe we don't need this scenario and the one above. They are similar to the **first name** one

reggieb:
@Kalagan Can you have another look as I've had to make a couple of changes to get through design review.

The two main changes are:

- Change links on the Review Answers page now use anchors to make focus on the returned to form, match the element being selected to change.
- The back link on Review Answers behaves as "normal". To achieve that I've added a reset action on the controller, that sets the legal aid application back to `initiated` and the follows the `Steppable` back action.

I've also rebased, so your change to the local of the email input has been incorporated.

Kalagan:
@reggieb I had another look. It still looks good. Good job!

PR #100, opened 2018-11-27
[AP-195] Back navigation from passported/not passported

Kalagan:
## What

[AP-195](https://dsdmoj.atlassian.net/browse/AP-195)

- Modify the address pages to update the address if there is one rather than always creating a new one
- Implement correct backlink URL on check_benefits page

## Checklist


app/forms/applicants/address_selection_form.rb
```diff
@@ -10,14 +10,19 @@ class AddressSelectionForm
 
     validates :address, presence: true
 
+    def initialize(*args)
+      attributes[:lookup_used] = true
+      super
+    end
```
reggieb: I think this will work instead of overriding `initialize`:
```
def lookup_used
  true
end
```

spec/requests/providers/address_selections_spec.rb
```diff
@@ -62,13 +108,16 @@
...
     end
 
     it 'redirects to next submission step' do
-      post_request
+      patch_request
 
       expect(response).to redirect_to(providers_legal_aid_application_check_benefits_path(legal_aid_application))
     end
+
+    context 'when an address already exists' do
```
lostie: ❓ Are there tests missing for this context or this was just left here by mistake?

app/models/applicant.rb
```diff
@@ -25,6 +25,10 @@ def full_name
     "#{first_name} #{last_name}".strip
   end
 
+  def address
+    addresses.first
+  end
```
reggieb: I think using a `has_one` association would be better - as it wouldn't gather all the addresses and then discard all but the first:
```
has_one :address, -> { order(created_at: :desc) }
```

spec/requests/providers/address_selections_spec.rb
Kalagan: Test missing!

spec/requests/providers/address_lookups_spec.rb
```diff
@@ -64,36 +64,19 @@
       end
     end
 
-    context 'with a valid postcode', :vcr do
+    context 'with a valid postcode' do
```
reggieb: If you've turned of VCR here, does that mean there is a cassette than is no longer used? If so, it needs deleting.

app/forms/applicants/address_selection_form.rb
Kalagan: OK I'll try

app/models/applicant.rb
Kalagan: I don't think `addresses.first` gather all the address, SQL only fetch the first address with `LIMIT 1`.
But anyway, your `has_one` suggesting is more elegant so I'll do that

spec/requests/providers/address_lookups_spec.rb
Kalagan: I deleted it. But this logic moved to another controller and it created a new VCR cassette.
Git is registering this as moved files : 
image

app/models/applicant.rb
reggieb: You're right (as usual), but another problem with `addresses.first` is that it finds the first via `id`, and as addresses use UUIDs as ids, that means it plucks a random address out.

app/forms/applicants/address_selection_form.rb
reggieb: Booo :sob:

Kalagan:
@reggieb, @lostie, I addressed your feedbacks. Could you have another look when you have a moment?

app/forms/applicants/address_selection_form.rb
```diff
@@ -17,20 +22,24 @@ def applicant
...
+      attributes[:lookup_id] = selected_address.lookup_id
     end
 
     def exclude_from_model
-      %i[address]
+      %i[addresses]
+    end
+
+    def selected_address
+      @selected_address ||= addresses.find { |address| address.lookup_id == lookup_id }
```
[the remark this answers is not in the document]
Kalagan: `addresses` is a standard ruby array, not an ActiveRecord one. So it doesn't have `find_by`

reggieb:
:+1:

PR #101, opened 2018-11-27
AP-199 show previously entered information

tmckernan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-199)

- on navigating to any page in the application any previously entered information is visible
- update specs

## Checklist

### DIFF
--- a/app/controllers/providers/emails_controller.rb
+++ b/app/controllers/providers/emails_controller.rb
@@ -4,11 +4,11 @@ module Providers
     include Steppable
 
     def show
-      @form = Applicants::EmailForm.new(model: applicant)
+      @form = Applicants::EmailForm.new(current_params)
     end
 
     def update
-      @form = Applicants::EmailForm.new(edit_params)
+      @form = Applicants::EmailForm.new(update_params)
 
       if @form.save
         redirect_to(
@@ -23,12 +23,17 @@ module Providers
 
     private
 
-    def edit_params
+    def update_params
       email_params.merge(model: applicant)
     end
 
     def email_params
       params.require(:applicant).permit(:email)
+    end
+
+    def current_params
+      return nil unless applicant
+      applicant.attributes.symbolize_keys.slice(:email)
     end
   end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,7 +27,7 @@ Rails.application.routes.draw do
     root to: 'start#index'
 
     resources :legal_aid_applications, path: 'applications', only: %i[index new create] do
-      resource :applicant, only: %i[new create]
+      resource :applicant, only: %i[show update]
 
       resources :check_provider_answers, only: [:index] do
         post :confirm, on: :collection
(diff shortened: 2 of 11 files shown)


app/controllers/providers/emails_controller.rb
```diff
@@ -17,18 +17,23 @@ def update
...
+    def update_params
       email_params.merge(model: applicant)
     end
 
     def email_params
       params.require(:applicant).permit(:email)
     end
+
+    def current_params
+      return nil unless applicant
```
lostie: **[nit pick/not a blocker]** (something I forgot to mention in https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/100). `nil` is irrelevant here, `return unless applicant` already returns nil 💅  .

Kalagan: Good point. I'll fix it in my PR

spec/requests/providers/applicants_spec.rb
```diff
@@ -21,9 +21,20 @@
         expect(response).to redirect_to(providers_legal_aid_applications_path)
       end
     end
+
+    context 'has already got applicant infro' do
```
lostie: **[nit pick/typo]** _info_ (instead of _infro_). 

The phrasing itself looks a bit odd IMO, would something like _when the application already has applicant info_ be a bit more clearer?

spec/requests/providers/applicants_spec.rb
```diff
@@ -21,9 +21,20 @@
...
+      let(:applicant) { create(:applicant) }
+      let(:laa) { create(:legal_aid_application, applicant: applicant) }
+      let(:get_request) { get "/providers/applications/#{laa.id}/applicant/edit" }
+
+      it 'display first_name' do
+        get_request
+        expect(response.body).to include(applicant.first_name)
```
lostie: **[nit pick]** Consider using the `unescaped_response_body` helper here to ensure this test doesn't randomly fail if the applicant's first name has special character that are encoded differently.

app/controllers/providers/emails_controller.rb
```diff
@@ -3,12 +3,12 @@ class EmailsController < ApplicationController
     include ApplicationDependable
     include Steppable
 
-    def show
-      @form = Applicants::EmailForm.new(model: applicant)
+    def edit
```
reggieb: I don't think you should change this one. In routes this is defined via `resource :email ...` - not the singular. That gives a path ending `/email` for the show view. If you change it to `edit` the path becomes `/email/edit`.

Following this afternoon's meeting, I wonder if we should be using more routes defined by `resource` (singular) and then use `show` rather than `edit` as the default landing page for the controllers.

config/routes.rb
```diff
@@ -27,15 +27,15 @@
     root to: 'start#index'
 
     resources :legal_aid_applications, path: 'applications', only: %i[index new create] do
-      resource :applicant, only: %i[new create]
+      resource :applicant, only: %i[edit update]
```
reggieb: Make application use show rather than edit too!

PR #102, opened 2018-11-28
[AP-208] Move email address entry to client_details page

Kalagan:
## What

[AP-208](https://dsdmoj.atlassian.net/browse/AP-208)

- Remove the **applicant email** page
- Add email input to **client details** page

## Checklist


spec/forms/applicants/basic_details_form_spec.rb
```diff
@@ -26,7 +26,7 @@
     let(:applicant) { Applicant.last }
 
     it 'creates a new applicant' do
-      expect { subject.save }.to change { Applicant.count }
+      expect { subject.save }.to change { Applicant.count }.by(1)
```
lostie: 👏

spec/forms/applicants/basic_details_form_spec.rb
```diff
@@ -65,6 +65,18 @@
       end
     end
 
+    context 'with an invalid email' do
+      before { attributes[:email] = 'not-a-valid-email' }
```
lostie: **[not a blocker/nit pick]** Consider re-defining the `attributes` variable instead of using before blocks, e.g.:

```ruby
let(:attributes) { attributes_for(:applicant).merge(email: 'not-a-valid-email') }
```

In my experience, I find it nicer to read and more manageable.

spec/forms/applicants/basic_details_form_spec.rb
```diff
@@ -65,6 +65,18 @@
...
+      it 'does not persist model' do
+        expect { subject.save }.not_to change { Applicant.count }
+      end
+
+      it 'generate an error' do
+        subject.save
+        expect(subject.errors[:email]).to be_present
```
lostie: 💭 Consider asserting on the error you're actually expecting to avoid false positives (e.g. the model has an error on the email but it's because it's not present where your expectation is that the format is invalid)

app/forms/applicants/basic_details_form.rb
```diff
@@ -23,9 +23,11 @@ class BasicDetailsForm
     )
 
     validates :national_insurance_number, presence: true
-
     validates :national_insurance_number, format: { with: NINO_REGEXP, message: :not_valid }, allow_blank: true
 
+    validates :email, format: { with: URI::MailTo::EMAIL_REGEXP }, allow_blank: true
+    validates :email, presence: true
+
     # rubocop:disable Lint/DuplicateMethods
```
lostie: **[This is not part of the PR so don't feel like you need to address it]** If we're just re-defining the reader method for `date_of_birth`, changing the definition of it above from:

```ruby
attr_accessor :date_of_birth
```

to 

```ruby
attr_writer :date_of_birth
```

should get rid of `# rubocop:disable Lint/DuplicateMethods`. 

Unless there's a VERY strong reason to have those, I would avoid using them all together.

PR #103, opened 2018-11-29
Feature/ap 210 about financial assessment

mpw5:
**Add 'about-the-financial-assessment' screen**

[AP-210](https://dsdmoj.atlassian.net/browse/AP-210)

Add a new screen to give providers information about the financial assessment. 

This page should exist between the 'does-client-use-online-banking' screen and the 'confirmation' page., however the 'does-client-use-online-banking' screen does not currently exist - it is being added in [AP-209](https://dsdmoj.atlassian.net/browse/AP-209). As an interim step this change places the 'about-the-financial-assessment' between the 'check-provider-answers' page and the 'confirmation' page. This will need to be changed as part of AP-209 or as a separate change.

This change:

- adds a controller and view for the new page
- moves functionality for sending a citizen email from the controller for 'check-provider-answers' to the new controller
- amends navigation appropriately
- adds and amends rpsec and feature tests appropriately

## Checklist


app/controllers/concerns/providers/steppable.rb
```diff
@@ -33,8 +33,12 @@ module Steppable
         # Back determined by controller action logic
       },
       check_provider_answers: {
-        forward: :providers_legal_aid_applications_path,
+        forward: :providers_legal_aid_application_about_the_financial_assessment_path,
         back: :providers_legal_aid_application_check_benefits_path
+      },
+      about_the_financial_assessments: {
+        forward: :submit_providers_legal_aid_application_about_the_financial_assessment_path,
```
Kalagan: `forward` of this new page is not use, is it? If so, maybe it would make more sense to not define it.

PR #104, opened 2018-11-30
[AP-209] Page "does-client-use-online-banking"

Kalagan:
## Confirm client uses online banking

[AP-209](https://dsdmoj.atlassian.net/browse/AP-209)

- Add a new page asking the provider if the client uses online banking. It has a yes/no radio button
- For now, it is after the **chek benefits** page but I will modify it as soon as AP-210 is merged
- If **Yes** is selected, it goes to **Check answers** page.
- If **No** is selected, it shows a placeholder message indicating to use CCMS
- Fix some sporadically failing tests (`spec/requests/citizens/accounts_spec.rb`)

## Checklist

### DIFF
--- a/app/views/providers/online_bankings/show.html.erb
+++ b/app/views/providers/online_bankings/show.html.erb
@@ -1,0 +1,46 @@
+<% content_for(:navigation) { provider_back_link } %>
+
+<%= render partial: 'shared/forms/error_summary', locals: { model: @form } %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+      <h1 class="govuk-fieldset__heading">
+         <%= t '.heading' %>
+      </h1>
+    </legend>
+
+    <%= form_with(
+          model: @form,
+          url: providers_legal_aid_application_online_banking_path(@legal_aid_application),
+          method: :patch,
+          local: true
+        ) do |form| %>
+
+      <div class="govuk-!-padding-bottom-8"></div>
+
+      <% error = form.object.errors[:uses_online_banking].first %>
+      <% group_error_class = error ? 'govuk-form-group--error' : '' %>
+      <% input_error_class = error ? 'govuk-input--error' : '' %>
+
+      <div class="govuk-form-group <%= group_error_class %>">
+        <div class="govuk-radios">
+          <%= content_tag(:span, error, class: 'govuk-error-message') if error %>
+          <div class="govuk-radios__item">
+            <%= form.radio_button 'uses_online_banking', 'true', class: "govuk-radios__input #{input_error_class}" %>
+            <%= form.label 'uses_online_banking_true', t('generic.yes'), class: 'govuk-label govuk-radios__label' %>
+          </div>
+          <div class='govuk-radios__item'>
+            <%= form.radio_button 'uses_online_banking', 'false', class: "govuk-radios__input #{input_error_class}" %>
+            <%= form.label 'uses_online_banking_false', t('generic.no'), class: 'govuk-label govuk-radios__label' %>
+          </div>
+        </div>
+      </div>
+
+      <div class="govuk-!-padding-bottom-6"></div>
+
+      <%= form.submit t('generic.continue'), id: 'continue', class: 'govuk-button' %>
+    <% end %>
+  </div>
+</div>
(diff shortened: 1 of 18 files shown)


config/application.rb
```diff
@@ -22,5 +22,9 @@ class Application < Rails::Application
     config.govuk_notify_templates = config_for(
       :govuk_notify_templates, env: ENV.fetch('GOVUK_NOTIFY_ENV', 'development')
     ).symbolize_keys
+
+    # Prevents surrounding of erroneous inputs with <div class="field_with_errors">
+    # https://guides.rubyonrails.org/configuring.html#configuring-action-view
+    config.action_view.field_error_proc = proc { |html_tag| html_tag.html_safe }
```
Kalagan: I had to add this otherwise the div added by Rails when there is an error messes up with the gov-uk styling

reggieb: Again, have a look at the `GovUkFormHelper` as I did manage to get GovUK styled error to work with radio buttons.

config/routes.rb
```diff
@@ -32,8 +32,8 @@
       resource :address, only: %i[show update]
       resource :address_lookup, only: %i[show update]
       resource :address_selection, only: %i[show update]
-      resource :email, only: %i[show update]
       resources :check_benefits, only: [:index]
+      resource :online_banking, only: %i[show update], path: 'does-client-use-online-banking'
```
reggieb: :+1:

spec/requests/providers/online_bankings_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'rails_helper'
+
+RSpec.describe 'does client use online banking requests', type: :request do
+  let(:applicant) { create :applicant }
+  let(:application) { create :legal_aid_application, applicant: applicant }
+  let(:application_id) { application.id }
+
+  describe 'GET /providers/applications/:legal_aid_application_id/applicant' do
+    let(:get_request) { get "/providers/applications/#{application_id}/does-client-use-online-banking" }
```
reggieb: I think `subject` makes more sense here. I know `<HTTP VERB>_request` is the pattern used across a lot of this code base, but it has problems - for example the headache of having to change all the instances of `post_request` to `patch_request` when the HTTP method changed. The fact that the name of the `describe` block is exactly the request specified in `get_request` implies that the request is the subject of the spec.

spec/requests/providers/online_bankings_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      it 'redirects the user to the applications page with an error message' do
+        get_request
+        expect(response).to redirect_to(providers_legal_aid_applications_path)
+      end
+    end
+
+    it 'displays the correct page' do
+      get_request
+      expect(unescaped_response_body).to include('Does your client use online banking?')
+    end
```
reggieb: A pedantic point perhaps, but I think `it` within `context`s should be clearly separated from those without context. Personally I like the pattern:
```ruby
describe 'the main container describe the scope of the spec' do
  it 'specs that describe default or no parameters entered behaviour` do
  end

  context 'behaviours that deviate from default/no-parameter' do
  end
end
```
In other words put all the `it` blocks that are not modified by contexts at the start of the description block, and start the context blocks after all of the not-context `it` blocks.

app/views/providers/online_bankings/show.html.erb
```diff
@@ -0,0 +1,46 @@
...
+          <div class='govuk-radios__item'>
+            <%= form.radio_button 'uses_online_banking', 'false', class: "govuk-radios__input #{input_error_class}" %>
+            <%= form.label 'uses_online_banking_false', t('generic.no'), class: 'govuk-label govuk-radios__label' %>
+          </div>
+        </div>
+      </div>
+
+      <div class="govuk-!-padding-bottom-6"></div>
+
+      <%= form.submit t('generic.continue'), id: 'continue', class: 'govuk-button' %>
```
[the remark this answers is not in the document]
Kalagan: I tried using the existing radio_button helper but is a different use case. The existing one generates a `radio_button_tag`. Here I need a `form.radio_button` that can be prepopulated with an existing value.

config/application.rb
Kalagan: But the existing component doesn't prepopulate the radio button with an existing value

app/views/providers/online_bankings/show.html.erb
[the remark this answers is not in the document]
Kalagan: Ultimately I would like to create some versatile form components similar to https://github.com/sgruhier/foundation_rails_helper or https://github.com/bootstrap-ruby/bootstrap_form

PR #107, opened 2018-11-30
AP-218 Run benefit checker again only if details change

reggieb:
[JIRA AP-218](https://dsdmoj.atlassian.net/browse/AP-218)

Adds a check to see if the benefit check needs to be re-run based of two criteria:

1. Whether the check has already been run
2. If the applicant's details have been updated since the check result was last updated

It then uses that check in `CheckBenefitsController`.


spec/cassettes/check_benefits_requests/GET_/providers/applications/_application_id/check_benefits/when_the_check_benefit_result_already_exists/and_the_applicant_has_since_been_modified/generates_a_new_check_benefit_result.yml
```diff
@@ -0,0 +1,199 @@
+---
```
Kalagan: I think this file might not be needed. Maybe you rename the test and it generated another cassette but left this one?

reggieb: There was an extra cassette that I've now removed.

PR #110, opened 2018-12-04
Update govuk-frontend to the most recent version (2.4.0)

willc-work:
Update govuk to the most recent version

[Link to story](https://dsdmoj.atlassian.net/browse/AP-219)

### DIFF
--- a/app/views/providers/address_selections/show.html.erb
+++ b/app/views/providers/address_selections/show.html.erb
@@ -44,7 +44,7 @@
             <%= content_tag(:span, @form.errors[:address].first, class: ['govuk-error-message']) %>
           <% end %>
           <% input_error_class = @form.errors[:address].any? ? 'govuk-select--error' : '' %>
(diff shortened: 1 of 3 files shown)


reggieb:
The main issues should be as described in the [2.0.0 release notes](https://github.com/alphagov/govuk-frontend/releases/tag/v2.0.0). Looking through them the only obvious issue I could find is:

- **Remove the width declaration from the `<select>` component** - I think `api/app/views/providers/address_selections/show.html.erb:47` applies a width style to a `select` tag.

I think the main vulnerability will be when behaviour is modified via JavaScript, and that that will be most apparent in cross browser tests - particularly in how pages behave with changing window sizes.

The current page that seems to rely most heavily on behaviour changes due to window sizes is `check_provider_answers`. If you look at `app/assets/stylesheets/providers/check-your-answers.scss` it contains more calls to dynamic SASS method like `govuk-media-query` than the rest of the app put together.

The elements that are most likely to be problematic are error and hint elements in forms. As the most complicated form is probably the `/providers/applicant` form, I'd suggest testing the behaviour of that when an error is shown, and the window size is changed - across a number of browsers.

In summary - I don't think the update will cause any significant problems. However, I recommend:

- [ ] remove the width class declaration in the select element highlighted above.
- [ ] test the app in a number of browsers, paying particular attention to `providers/check_provider_answers` and `providers/applicants` (the latter with errors being displayed (submit the form with no input)

PR #114, opened 2018-12-05
Introducing a govuk_text_field that generates a text_field with the right GovUK style

Kalagan:
Modifying our current text_fields from using shared views to  using a custome form builder like that: `form.govuk_text_field(:first_name)`

We can add other inputs later (radio button, check boxes)

Heavily inspired by https://github.com/ministryofjustice/govuk_elements_form_builder

## Checklist

- [ ] There should be no unneccessary whitespace changes. These make diffs harder to read and conflicts more likely. 


app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    # Given an attributes hash that could include any number of arbitrary keys, this method
+    # ensure we merge one or more 'default' attributes into the hash, creating the keys if
+    # don't exist, or merging the defaults if the keys already exists.
+    # It supports strings or arrays as values.
+    #
+    def merge_attributes(attributes, default:)
+      hash = attributes || {}
+      hash.merge(default) { |_key, oldval, newval| Array(newval) + Array(oldval) }
```
reggieb: I'm not sure you need to use the `Array(x)` syntax. The more usual way to create an array us `[x]`. Also `merge!` is more memory efficient. So:
```
hash.merge!(default) { |_key, oldval, newval| [newval] + [oldval] }
```

app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -0,0 +1,89 @@
...
+        merge_attributes(options, default: { class: default_classes })
+      )
+    end
+
+    def label_classes!(options)
+      options ||= {}
+      options[:label_options] ||= {}
+      options[:label_options].merge!(
+        merge_attributes(options[:label_options], default: { class: 'govuk-label' })
+      )
```
reggieb: I think if you use `merge!` within `merge_attributes` this can be replaced with just:
```ruby
merge_attributes(options[:label_options], default: { class: 'govuk-label' })
```
No outer `options[:label_options].merge!`

PR #115, opened 2018-12-06
AP-211 Add Do  you own your own home? page

stephenrichards:
## What

[Link to story]https://dsdmoj.atlassian.net/browse/AP-211

Added the page "Do you own the home you live in?"  This is the first of the capital questions that 
an applicant should answer to determine his eligibility for legal aid.

screenshot 2018-12-06 12 25 02

## Checklist


app/controllers/citizens/own_homes_controller.rb
```diff
@@ -0,0 +1,39 @@
+module Citizens
+  class OwnHomesController < BaseController
+    def show
+      @page_heading = "#{I18n.t('citizens.own_homes.show.h1-heading')} - #{I18n.t('shared.page-title.suffix')}"
```
Kalagan: I think this should be in the update action as well.
Maybe use a `before_action`
image 3

config/locales/en.yml
```diff
@@ -217,6 +224,14 @@ en:
         list_1: We do not save your online banking username and password
         list_2: We do not have any ongoing access to your bank accounts
         list_3: We only use your details to assess your eligibility for legal aid
+    own_homes:
+      show:
+        h1-heading: Do you own the home that you live in?
+        own_home:
+          mortgage: Yes, with a mortage or loan
+          owned_outright: Yes, owned outright
+          not: No
```
willc-work: in the view this shows up as 'Own home no'.
it might be possible to use the generic no in the en.yml file for this radio button

PR #117, opened 2018-12-06
Add property value form and store users entry

willc-work:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-212)

Citizen needs to add the value of their home if they are a home owner. Value is stored in legal_aid_applications table.

Form limits users to entering a number and throws errors if it is empty or is not a valid number. property_value is stored as a decimal with precision 10 and scale 2 to allow for units after the decimal point. Happy to revise this if e.g. we only want value to nearest whole pound.

Changes will be required to link this into the capital questions flow once the preceding and subsequent pages are available

screen shot 2018-12-06 at 15 07 59

## Checklist

### DIFF
--- a/app/controllers/citizens/additional_accounts_controller.rb
+++ b/app/controllers/citizens/additional_accounts_controller.rb
@@ -24,7 +24,13 @@ module Citizens
         render plain: 'Landing page: Return to True Layer steps'
       when 'no'
         legal_aid_application.update(has_offline_accounts: true)
+<<<<<<< HEAD
         redirect_to citizens_own_home_path
+=======
+        # TODO: - set redirect path when known
+        # render plain: 'Landing page: Next step in Citizen journey'
+        redirect_to citizens_property_value_path
+>>>>>>> Add property value to applicant and legal aid applictions table
       else
         @error = I18n.t('generic.errors.yes_or_no')
         render :new
--- a/app/controllers/citizens/additional_accounts_controller.rb
+++ b/app/controllers/citizens/additional_accounts_controller.rb
@@ -24,13 +24,7 @@ module Citizens
         render plain: 'Landing page: Return to True Layer steps'
       when 'no'
         legal_aid_application.update(has_offline_accounts: true)
-<<<<<<< HEAD
         redirect_to citizens_own_home_path
-=======
-        # TODO: - set redirect path when known
-        # render plain: 'Landing page: Next step in Citizen journey'
-        redirect_to citizens_property_value_path
->>>>>>> Add property value to applicant and legal aid applictions table
       else
         @error = I18n.t('generic.errors.yes_or_no')
         render :new
--- a/app/controllers/citizens/property_values_controller.rb
+++ b/app/controllers/citizens/property_values_controller.rb
@@ -1,10 +1,10 @@
 module Citizens
   class PropertyValuesController < BaseController
     def show
-      @form = Applicants::PropertyValueForm.new # this is so errors are visible
+      @form = Applicants::PropertyValueForm.new
     end
 
-    def create
+    def update
       @form = Applicants::PropertyValueForm.new(edit_params)
 
       if @form.save
@@ -17,7 +17,7 @@ module Citizens
     private
 
     def property_value_params
-      params.fetch(:legal_aid_application, {}).permit(:property_value)
+      params.require(:legal_aid_application).permit(:property_value)
     end
 
     def edit_params
--- a/app/controllers/citizens/property_values_controller.rb
+++ b/app/controllers/citizens/property_values_controller.rb
@@ -1,32 +1,20 @@
 module Citizens
   class PropertyValuesController < BaseController
     def show
-      # @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
       @form = Applicants::PropertyValueForm.new # this is so errors are visible
     end
 
     def create
-      # @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
       @form = Applicants::PropertyValueForm.new(edit_params)
 
       if @form.save
         render plain: 'Navigate to question 1c; What is the outstanding mortgage?'
       else
-        # @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
         render :show
       end
     end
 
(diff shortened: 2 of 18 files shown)


app/controllers/citizens/additional_accounts_controller.rb
```diff
@@ -25,7 +25,8 @@ def update
       when 'no'
         legal_aid_application.update(has_offline_accounts: true)
         # TODO: - set redirect path when known
-        render plain: 'Landing page: Next step in Citizen journey'
+        # render plain: 'Landing page: Next step in Citizen journey'
```
reggieb: Delete the comments if they are no longer needed.

app/controllers/citizens/property_values_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Citizens
+  class PropertyValuesController < BaseController
+    def show
+      @form = Applicants::PropertyValueForm.new # this is so errors are visible
```
reggieb: Remove the comment. It is wrong. The show action will never show an error. The errors are rendered when the create action renders the show template (and therefore the `show` method is not called)

app/controllers/citizens/property_values_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+        render plain: 'Navigate to question 1c; What is the outstanding mortgage?'
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def property_value_params
+      params.fetch(:legal_aid_application, {}).permit(:property_value)
```
reggieb: The correct syntax for handling params is:
```ruby
params.require(:legal_aid_application).permit(:property_value)
```
See: https://edgeguides.rubyonrails.org/action_controller_overview.html#strong-parameters

config/routes.rb
```diff
@@ -18,6 +18,7 @@
   namespace :citizens do
     resources :legal_aid_applications, only: [:show]
     resource :consent, only: %i[show create]
+    resource :property_value, only: %i[show create]
```
reggieb: The pattern being used in the providers journey is show and update. So I think the form should be submitted to `update` rather than `create`. Not that the form method will need to change to PATCH to match.

spec/requests/citizens/additional_accounts_spec.rb
```diff
@@ -101,8 +101,9 @@
 
       it 'displays holding page' do
         # TODO: Delete when redirect set
-        expect(response).to have_http_status(:ok)
-        expect(response.body).to match('Landing page')
+        # expect(response).to have_http_status(:ok)
+        # expect(response.body).to match('Landing page')
```
reggieb: Remove redundant comments

spec/requests/citizens/property_value_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        expect(unescaped_response_body).to include('Navigate to')
+        # TO DO
+        # expect(response).to redirect_to(to_be_determined_path)
+      end
+
+      it 'records the value in the legal aid application table' do
+        legal_aid_application.reload
+        expect(legal_aid_application.property_value).to be_within(0.01).of(123_456.78)
+        expect(legal_aid_application.updated_at.utc.to_i).to be_within(1).of(Time.now.to_i)
+        # expect(legal_aid_application.updated_at.to_s(be_between(2.hours.ago, 1.hour.ago)))
```
reggieb: Remove commented out text

spec/requests/citizens/property_value_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      post citizens_property_value_path, params: params
+    end
+
+    context 'when a property value is entered' do
+      let(:params) { { legal_aid_application: { property_value: 123_456.78 } } }
+
+      it 'redirects to new action' do
+        expect(unescaped_response_body).to include('Navigate to')
+        # TO DO
+        # expect(response).to redirect_to(to_be_determined_path)
```
reggieb: I think using an `xit` is a better pattern (than just adding a comment), as it is a look more obvious that there is a behaviour that needs to be used in the future.

spec/requests/citizens/property_value_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'rails_helper'
+
+RSpec.describe Citizens::PropertyValuesController, type: :request do
+  describe 'citizen property value test' do
+    describe 'GET /citizens/property_value' do
+      before { get citizens_property_value_path }
+
+      it 'returns http success' do
+        expect(response).to have_http_status(:ok)
+        expect(unescaped_response_body).to include('How much is your home worth')
```
reggieb: A pedantic point, but you don't need to use `unescaped_response_body` if you know the text you are matching does not contain any problematic characters. As you know the matching text is `'How much is your home worth'` So `response.body` would be fine here.

app/forms/applicants/property_value_form.rb
```diff
@@ -0,0 +1,29 @@
+module Applicants
```
reggieb: Other forms are in the module `Applicants` and use `form_for :applicant`. As the object this form applies to is a `LegalAidApplication`, I think this form should be in the module `LegalAidApplications` (also sub-folder `legal_aid_applications`).

spec/requests/citizens/own_home_spec.rb
```diff
@@ -21,35 +21,24 @@
       let(:params) { { 'legal_aid_application' => { 'own_home' => 'owned_outright' } } }
 
       it 'returns http_success' do
-        expect(response).to have_http_status(:ok)
+        expect(response).to have_http_status(:found)
```
reggieb: `:found` matches 302, which is a redirect. It's usually a good idea to specify where the redirect points at in the spec as there are a number of reasons why a redirect may occur (authentication failure for example). So just detecting a redirect isn't particularly useful.

Also - if the response is a redirect, the `it` text `'returns http_success'` is no longer valid.

So this should probably be replaced with something like:
```ruby
it 'redirects to ...<the destination>' do	          
    expect(response).to redirect_to(<destination_path>)
end
```
Replaces the `<..>` content with the appropriate entries.

PR #118, opened 2018-12-06
AP-213 Add Outstanding Mortgage page to Citizen journey

reggieb:
[AP-213](https://dsdmoj.atlassian.net/browse/AP-213)

Added a new citizens controller with show and update views.

Some extra complexity was added to get the £ sign within the input field.

GET renders:

show_outstanding_mortgate

PATCH with invalid input renders: 

show_outstanding_mortgage_error

### DIFF
--- a/app/views/citizens/outstanding_mortgages/show.html.erb
+++ b/app/views/citizens/outstanding_mortgages/show.html.erb
@@ -1,4 +1,4 @@
-<%= render 'shared/forms/error_summary', model: @form, attribute_prefix: :legal_aid_application_ %>
+<%= render 'shared/forms/error_summary', model: @form %>
 
 <fieldset class="govuk-fieldset">
   <div class="govuk-!-padding-bottom-2"></div>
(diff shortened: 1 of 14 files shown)


app/views/shared/forms/_error_summary.html.erb
```diff
@@ -7,7 +8,7 @@
       <ul class="govuk-list govuk-error-summary__list">
         <% model.errors.each do |attribute, message| %>
           <li>
-            <%= link_to message, "##{attribute}" %>
+            <%= link_to message, "##{attribute_prefix}#{attribute}" %>
```
willc-work: this will break the links from the error message in the summary to the relevant field on existing forms, definitely affects, on the provider side, applicant details and manual address.

app/views/citizens/outstanding_mortgages/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+          ) do |form| %>
+        <div class="govuk-!-padding-bottom-2"></div>
+
+        <%= govuk_form_group(
+              show_error_if: form.object.errors[:outstanding_mortgage_amount].present?,
+              input: :legal_aid_application_outstanding_mortgage_amount,
+              hint_id: :outstanding_mortgage_amount_hint
+            ) do %>
+
+          <%= govuk_fieldset_header t('.outstanding_mortgage_amount_heading'), padding_below: 6 %>
```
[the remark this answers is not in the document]
reggieb: I'm confused then. If you look at https://design-system.service.gov.uk/components/error-message/
it clearly shows the error bar extending past the header text.

app/views/citizens/outstanding_mortgages/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+        <% end %>
+
+        <div class="govuk-!-padding-bottom-2"></div>
+
+        <div class="govuk-grid-row">
+          <div class="govuk-grid-column-two-thirds">
+            <%= form.submit t('generic.continue'), id: 'continue', class: 'govuk-button govuk-button' %>
+          <% end %>
+         </div>
+       </div>
```
reggieb: Note to self: I can't see an `end` closure to the `form_with`

reggieb: `end` on line 38 should have been on line 40

app/views/citizens/outstanding_mortgages/show.html.erb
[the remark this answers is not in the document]
reggieb: Will has confirmed that is a single entry - error bar should extend to header.

PR #123, opened 2018-12-11
AP 217 Add restrictions that mean a Citizen cannot use capital assets

reggieb:
[Jira AP-217](https://dsdmoj.atlassian.net/browse/AP-217)

This PR adds a `Restriction` model, and a `has_many :restrictions, through: :legal_aid_application_restrictions` on `LegalAidApplication`. The new page then allows a Citizen to choose to associate all, any or none of those restrictions with their current application.

`Restriction` has a number of predefined instances (named) to describe each restriction condition, and there is a class method to generate these. This has been added to seeding, so `rake db:seed` adds the standard `restrictions`. The generation task can be run multiple times without generating duplicates.

I cannot think of a normal invalid condition (the Citizen selects from pre-existings objects via checkboxes generated by the app). So there is no validation - and therefore no form object is needed. Also no standard error handling - instead errors such as `ActiveRecord::RecordNotFound` (when an unrecognised id is submitted) bubble to the surface rather than being trapped.

citizens_restrictions

### DIFF
--- a/app/models/restriction.rb
+++ b/app/models/restriction.rb
@@ -1,0 +1,25 @@
+class Restriction < ApplicationRecord
+  NAMES = %i[
+    reposession_or_deferred_interest
+    subject_matter_of_dispute
+    foreign_exchange_controls
+    restraint_or_freezing_order
+    bankruptcy
+    held_overseas
+  ].freeze
+
+  has_many :legal_aid_application_restrictions
+  has_many :legal_aid_applications, through: :legal_aid_application_restrictions
+
+  validates :name, presence: true, uniqueness: true
+
+  def self.populate
+    # using pluck rather than `find_or_create` in the iteration to reduce SQL calls
+    existing = pluck(:name).map(&:to_sym)
+    (NAMES - existing).each { |name| create!(name: name) }
+  end
+
+  def label_name
+    I18n.t("restrictions.names.#{name}")
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -27,6 +27,7 @@ Rails.application.routes.draw do
     resource :outstanding_mortgage, only: %i[show update]
     resource :savings_and_investment, only: %i[show update]
     resource :shared_ownership, only: %i[show update]
+    resources :restrictions, only: %i[index create] # as multiple restrictions
   end
 
   namespace :providers do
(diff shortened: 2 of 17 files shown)


app/models/restriction.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  has_many :legal_aid_application_restrictions
+  has_many :legal_aid_applications, through: :legal_aid_application_restrictions
+
+  validates :name, presence: true, uniqueness: true
+
+  def self.populate
+    # using pluck rather than `find_or_create` in the iteration to reduce SQL calls
+    existing = pluck(:name).map(&:to_sym)
+    (NAMES - existing).each { |name| create(name: name) }
```
Kalagan: I suggest always using `create!` rather than `create` if you're not doing anything with what is returned by this method.

config/routes.rb
```diff
@@ -22,6 +22,7 @@
     resources :accounts, only: [:index]
     resources :additional_accounts, only: %i[index create new update]
     resource :own_home, only: %i[show update]
+    resources :restrictions, only: %i[index create] # as multiple restrictions
```
Kalagan: why `update` instead of `create` wouldn't work?

db/migrate/20181207162311_create_legal_aid_application_restrictions.rb
```diff
@@ -0,0 +1,14 @@
+class CreateLegalAidApplicationRestrictions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :legal_aid_application_restrictions, id: :uuid do |t|
+      t.references(
+        :legal_aid_application,
+        foreign_key: true,
+        type: :uuid,
+        index: { name: 'laa_id_laa_restriction_id' }
```
Kalagan: the default index name is too long?

reggieb: Exactly. The migration wouldn't migrate until I set a shorter name.

app/models/restriction.rb
reggieb: That's not the first time you've pull me up on that ... and correctly again 😄

config/routes.rb
reggieb: As I am using `resources` here, `update` would require an `:id` within the url. So I've used `create` as that does not require the `:id`.

The alternative would be to use `resource` but then the url ends `\restriction` and I think it should be `\restrictions` (plural). I could use override the path construction to return a plural with `resource` here, but using the `:index, :create` pair here rather than `:show, :update` seemed to make sense to me - especially as multiple restrictions are being managed by these controller actions.

PR #124, opened 2018-12-11
AP-237 review use of translation framework

tmckernan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-237)

- move views text to translation file

## Checklist


app/views/citizens/information/show.html.erb
```diff
@@ -1,32 +1,32 @@
 <% content_for :navigation do %>
-  <%= link_to 'Back', citizens_legal_aid_application_path(id: session[:current_application_ref]), class: 'govuk-back-link', id: 'back' %>
+  <%= link_to t('generic.back'), citizens_legal_aid_application_path(id: session[:current_application_ref]), class: 'govuk-back-link', id: 'back' %>
```
reggieb: Really minor - if there's only one thing passed into to path you don't have to specify what it is. So I think this will work: `citizens_legal_aid_application_path(session[:current_application_ref])` (no `id:`
)

PR #125, opened 2018-12-11
AP-216 Implementation of Other Assets page

stephenrichards:
## Implementation of Other Assets Page
[Link to story](https://dsdmoj.atlassian.net/browse/AP-216)

This page comprises 9 checkboxes which expand to expose 1 to 3 edit boxes
when checked.  The data is stored in an OtherAssetsDeclaration model which
is linked to the legal aid application on a one-to-one bases (legal aid application
has one other_assets_declaration) - this is to help prevent the legal aid declaration
model ending up with 50 columns.

Although working, the markup for the revealing checkboxes may need to be tweaked 
in order to make appearance and behaviour the same as on the other page with 
revealing checkboxes, probably by implementing this element in the FormBuilder 
and using the FormBuilder on both pages.

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -34,6 +34,7 @@ gem 'sprockets', '>= 3.0.0'
 gem 'sprockets-es6'
 
 group :development, :test do
+  gem 'awesome_print', '~> 1.8.0'
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
   gem 'dotenv-rails'
   gem 'erb_lint', require: false
@@ -50,7 +51,6 @@ group :development, :test do
 end
 
 group :development do
-  gem 'awesome_print', '~> 1.8.0'
   gem 'listen', '>= 3.0.5', '< 3.2'
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
--- a/app/controllers/citizens/other_assets_controller.rb
+++ b/app/controllers/citizens/other_assets_controller.rb
@@ -1,0 +1,43 @@
+module Citizens
+  class OtherAssetsController < BaseController
+    def show
+      declaration
+      @form = Citizens::OtherAssetsForm.new(current_params)
+    end
+
+    def update
+      @form = Citizens::OtherAssetsForm.new(form_params.merge(model: declaration))
+      if @form.save
+        render plain: 'Navigate to next question after any other assets'
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def legal_aid_application
+      @legal_aid_application ||= LegalAidApplication.find(session[:current_application_ref])
+    end
+
+    def declaration
+      @declaration ||= legal_aid_application.other_assets_declaration
+    end
+
+    def attributes
+      @attributes ||= Citizens::OtherAssetsForm::ALL_ATTRIBUTES
+    end
+
+    def current_params
+      @declaration.attributes.symbolize_keys.slice(*attributes)
+    end
+
+    def other_asset_params
+      params[:other_assets_declaration].permit(*(Citizens::OtherAssetsForm::ALL_ATTRIBUTES + Citizens::OtherAssetsForm::CHECK_BOXES_ATTRIBUTES))
+    end
+
+    def form_params
+      other_asset_params.merge(model: @declaration)
+    end
+  end
+end
--- a/app/controllers/citizens/other_assets_controller.rb
+++ b/app/controllers/citizens/other_assets_controller.rb
@@ -1,12 +1,11 @@
 module Citizens
   class OtherAssetsController < BaseController
     def show
-      declaration
-      @form = Citizens::OtherAssetsForm.new(current_params)
+      @form = Citizens::OtherAssetsForm.new(model: declaration)
     end
 
     def update
-      @form = Citizens::OtherAssetsForm.new(form_params.merge(model: declaration))
+      @form = Citizens::OtherAssetsForm.new(form_params)
       if @form.save
         render plain: 'Navigate to next question after any other assets'
       else
@@ -24,20 +23,12 @@ module Citizens
       @declaration ||= legal_aid_application.other_assets_declaration
     end
 
-    def attributes
-      @attributes ||= Citizens::OtherAssetsForm::ALL_ATTRIBUTES
-    end
-
-    def current_params
-      @declaration.attributes.symbolize_keys.slice(*attributes)
-    end
-
     def other_asset_params
       params[:other_assets_declaration].permit(*(Citizens::OtherAssetsForm::ALL_ATTRIBUTES + Citizens::OtherAssetsForm::CHECK_BOXES_ATTRIBUTES))
     end
 
     def form_params
-      other_asset_params.merge(model: @declaration)
+      other_asset_params.merge(model: declaration)
     end
   end
 end
--- a/app/forms/citizens/other_assets_form.rb
+++ b/app/forms/citizens/other_assets_form.rb
@@ -1,0 +1,104 @@
+module Citizens
+  class OtherAssetsForm
+    include BaseForm
+
+    form_for OtherAssetsDeclaration
+
+    SINGLE_VALUE_ATTRIBUTES = %i[
+      timeshare_value
+      land_value
+      jewellery_value
+      vehicle_value
+      classic_car_value
+      money_assets_value
+      money_owed_value
+      trust_value
+    ].freeze
+
+    SECOND_HOME_ATTRIBUTES = %i[
+      second_home_value
+      second_home_mortgage
+      second_home_percentage
+    ].freeze
+
+    ALL_ATTRIBUTES = (SECOND_HOME_ATTRIBUTES + SINGLE_VALUE_ATTRIBUTES).freeze
+
+    CHECK_BOXES_ATTRIBUTES = (SINGLE_VALUE_ATTRIBUTES.map { |attribute| "check_box_#{attribute}".to_sym } + [:check_box_second_home]).freeze
+
+    SINGLE_VALUE_ATTRIBUTES.each do |attribute|
+      check_box_attribute = "check_box_#{attribute}".to_sym
+      validates attribute, numericality: { greater_than_or_equal_to: 0 }, allow_blank: true
+      validates attribute, presence: true, if: proc { |form| form.send(check_box_attribute).present? }
+    end
+
+    validates :second_home_value, :second_home_mortgage, :second_home_percentage, numericality: { greater_than_or_equal_to: 0 }, allow_blank: true
+
+    attr_accessor(*ALL_ATTRIBUTES)
+    attr_accessor(*CHECK_BOXES_ATTRIBUTES)
+    attr_accessor :check_box_second_home
+
+    before_validation :normalize_amounts, :empty_unchecked_values
+
+    validate :all_second_home_values_present_if_any_are_present
+
+    def second_home_checkbox_status
+      any_second_home_value_present? ? 'checked' : nil
+    end
+
+    def any_second_home_value_present?
+      present_and_not_zero?(@second_home_value) ||
+        present_and_not_zero?(@second_home_mortgage) ||
+        present_and_not_zero?(@second_home_percentage)
+    end
+
+    def exclude_from_model
+      CHECK_BOXES_ATTRIBUTES
+    end
+
+    private
+
+    def empty_unchecked_values
+      SINGLE_VALUE_ATTRIBUTES.each do |attribute|
+        check_box_attribute = "check_box_#{attribute}".to_sym
+        if send(check_box_attribute).blank?
+          attributes[attribute] = nil
+          send("#{attribute}=", nil)
+        end
+      end
+
+      nullify_second_home_attrs if @check_box_second_home.blank?
+    end
+
+    def nullify_second_home_attrs
+      attributes[:second_home_value] = nil
+      attributes[:second_home_mortgage] = nil
+      attributes[:second_home_percentage] = nil
+      @second_home_value = nil
+      @second_home_mortgage = nil
+      @second_home_percentage = nil
+    end
+
+    def normalize_amounts
+      ALL_ATTRIBUTES
+        .map { |attribute| send(attribute) }
+        .compact
+        .each { |value| value.delete!(',') }
+    end
+
+    def all_second_home_values_present_if_any_are_present
+      return unless any_second_home_value_present?
+
+      %i[second_home_value second_home_mortgage second_home_percentage].each do |attr|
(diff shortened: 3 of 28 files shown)


app/forms/citizens/other_assets_form.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    CHECKBOX_STATUS_REGEX = Regexp.new(/_checkbox_status\z/)
+
+    def respond_to_missing?(meth, *args)
+      if CHECKBOX_STATUS_REGEX.match?(meth.to_s)
+        true
+      else
+        super
+      end
+    end
```
reggieb: A guard clause would be good here:
```ruby
def respond_to_missing?(meth, *args)
  return true if CHECKBOX_STATUS_REGEX.match?(meth.to_s)
  super
end
```

app/helpers/money_helper.rb
```diff
@@ -1,5 +1,6 @@
 module MoneyHelper
-  def value_with_currency_unit(value, currency)
+  def value_with_currency_unit(value, currency = nil)
+    currency = 'gbp' if currency.nil?
```
reggieb: It might be worth look at [this article](https://www.grzegorowski.com/rails-number_to_currency-change-locale-defaults/) that describes a more standard way of defining `number_to_currency` behaviour.

app/models/legal_aid_application.rb
```diff
@@ -92,4 +94,8 @@ def create_app_ref
   def set_open_banking_consent_choice_at
     self.open_banking_consent_choice_at = Time.current if will_save_change_to_open_banking_consent?
   end
+
+  def create_other_assets_declaration
+    OtherAssetsDeclaration.create(legal_aid_application_id: id)
+  end
```
reggieb: `create_other_assets_declaration` should already be created as one of the dynamic methods generated by the `has_one` declaration. See [docs](https://guides.rubyonrails.org/association_basics.html#methods-added-by-has-one)

I'd also suggest that `build_other_assets_declaration` and/or `create_other_assets_declaration` are used in either the controller or the associated form. That way - if this object is removed in the future the dev doing it only has to look in the controller responsible for adding the declarations - rather than looking for `after_create` callbacks on other models.

app/views/citizens/other_assets/_second_home_conditional_checkbox.html.erb
```diff
@@ -0,0 +1,47 @@
+<div class="govuk-checkboxes__item">
+  <input class="govuk-checkboxes__input"
+         id="second_home"
+         name="second_home_checkbox"
+         type="checkbox"
+         value="yes"
+         <%= @form.second_home_checkbox_status %>
```
reggieb: As mentioned above - I think `@form.checkbox_status(:second_home)` would be a better pattern.

app/views/citizens/other_assets/_second_home_conditional_checkbox.html.erb
```diff
@@ -0,0 +1,47 @@
...
+
+  <div class="govuk-form-group">
+    <label class="govuk-label" for="second_home_mortgage">
+      <%= t('citizens.other_assets.show.mortgage') %>
+    </label>
+    <input class="govuk-input govuk-!-width-one-third"
+           id="second_home_mortgage"
+           name="second_home_mortgage"
+           value="<%= value_with_currency_unit(@form.second_home_mortgage) %>"
+           pattern="\A[0-9][0-9,]+(.[0-9]{0,2})?\z" />
```
reggieb: I think the `+` needs to be `*` in the middle, as currently single figure values fail the pattern. For example `1.0`

Also as it is likely this will be used a lot, I'd suggest moving it to a help something like:
```
def currency_pattern
  '\A[0-9][0-9,]*(.[0-9]{0,2})?\z'
end
```
Then here the code becomes:
`pattern="<%= currency_pattern %>" />`

app/views/citizens/other_assets/_second_home_conditional_checkbox.html.erb
```diff
@@ -0,0 +1,47 @@
...
+
+  <div class="govuk-form-group">
+    <label class="govuk-label" for="second_home_percentage">
+      <%= t('citizens.other_assets.show.percentage') %>
+    </label>
+    <input class="govuk-input govuk-!-width-one-third"
+           id="second_home_percentage"
+           name="second_home_percentage"
+           value="<%= number_to_percentage(@form.second_home_percentage) %>"
+           pattern="\A[0-9]{1,3}(.[0-0]{1,3})?\z" />
```
reggieb: And perhaps a `percentage_pattern` helper?

db/migrate/20181206142708_create_other_asset_declaration.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.decimal :second_home_mortgage, precision: 14, scale: 2
+      t.decimal :second_home_percentage, precision: 14, scale: 2
+      t.decimal :timeshare_value, precision: 14, scale: 2
+      t.decimal :land_value, precision: 14, scale: 2
+      t.decimal :jewellery_value, precision: 14, scale: 2
+      t.decimal :vehicle_value, precision: 14, scale: 2
+      t.decimal :classic_car_value, precision: 14, scale: 2
+      t.decimal :money_assets_value, precision: 14, scale: 2
+      t.decimal :money_owed_value, precision: 14, scale: 2
+      t.decimal :trust_value, precision: 14, scale: 2
```
reggieb: Is there any reason for the precision and scale values being specifically set as 14 and 2 respectively? Current `decimal` setting are using default values. If `precision: 14` is needed for these values, should the existing attributes also be updated?

spec/factories/other_assets_declarations.rb
```diff
@@ -0,0 +1,25 @@
+FactoryBot.define do
+  factory :other_assets_declaration do
+    initialize_with { create(:legal_aid_application).other_assets_declaration }
```
reggieb: Is this necessary?

spec/factories/other_assets_declarations.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    trait :with_second_home do
+      second_home_value { 850_000.00 }
```
reggieb: I'd recommend using a random number generator in factories (native ruby or via Faker for example), as that avoids the possibility that the specific number specified in the factory is some sort of magic number that means in allows specs to pass when other numbers would cause an error.

app/forms/citizens/other_assets_form.rb
```diff
@@ -0,0 +1,108 @@
...
+
+    def method_missing(meth, *args)
+      if CHECKBOX_STATUS_REGEX.match?(meth.to_s)
+        attr = "#{meth.to_s.sub(CHECKBOX_STATUS_REGEX, '')}_value"
+        present_and_not_zero?(__send__(attr)) ? 'checked' : nil
+      else
+        super
+      end
+    end
```
reggieb: I think this meta-programming could be replaced with:
```ruby
def checkbox_status?(attribute)
  attr = "#{attribute}_value"
  present_and_not_zero?(__send__(attr)) ? 'checked' : nil
end
```
Then later in the code:
```ruby
<%= @form.__send__("#{attr}_checkbox_status") %>
```
Would become:
```ruby
<%= @form.checkbox_status?(attr) %>
```

Gemfile
```diff
@@ -34,13 +34,15 @@ gem 'sprockets', '>= 3.0.0'
 gem 'sprockets-es6'
 
 group :development, :test do
+  gem 'awesome_print', '~> 1.8.0'
```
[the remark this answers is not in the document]
stephenrichards: agreed.

app/controllers/citizens/other_assets_controller.rb
```diff
@@ -0,0 +1,53 @@
+module Citizens
+  class OtherAssetsController < BaseController
+    before_action :set_page_heading,
+                  :set_application_and_declaration
+
+    def show
+      @form = Citizens::OtherAssetsForm.new(current_params)
```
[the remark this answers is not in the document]
reggieb: I made a similar point on one of Julien's PRs. To get it work, override the form `initialize` method. That's easy with a single input, but more of a problem with multiple inputs. So ignore this comment.

Gemfile
```diff
@@ -34,13 +34,15 @@ gem 'sprockets', '>= 3.0.0'
...
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
   gem 'dotenv-rails'
   gem 'erb_lint', require: false
   gem 'faker', '>=1.9.1'
   gem 'i18n-tasks'
   gem 'json_expressions'
   gem 'pry-byebug'
+  gem 'rails-controller-testing'
```
tmckernan: Why do you need this gem if your not using `assigns` or `assert_template` in your request spec?  Also we are using request specs to test the behaviour. Both `assert_template` and `assigns` assert the implementation, not the behaviour.

app/forms/citizens/other_assets_form.rb
stephenrichards: yep.
Out of interest, how do we feel about `CHECKBOX_STATUS_REGEX.match?(meth.to_s) ? true : super`

app/models/legal_aid_application.rb
stephenrichards: Agreed

app/views/citizens/other_assets/show.html.erb
```diff
@@ -0,0 +1,42 @@
+
+<%= render partial: 'shared/forms/error_summary', locals: { model: @form } %>
```
Kalagan: This page is missing the back link and the title, and the indentation is not 100% correct.
Also, FYI, from talking with Simon, The form and its title should be included in a "two-third" div:
```
<div class="govuk-grid-row">
```
However, the error summary should be full width (above these divs)

app/views/citizens/other_assets/show.html.erb
```diff
@@ -0,0 +1,42 @@
...
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+          <h1 class="govuk-fieldset__heading">
+            <%= t('.h1-heading') %>
+          </h1>
+        </legend>
+
+        <div class="govuk-checkboxes" data-module="checkboxes">
+          <%= render partial: 'second_home_conditional_checkbox' %>
+          <%= render partial: 'shared/forms/conditional_checkbox_with_numeric_input',
+                     locals: { attr: :timeshare, label: t('.timeshare'), hint: t('.value_after_loan') } %>
```
Kalagan: I just merged my PR so if you want you can reuse the partial I created if it makes sense (https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/views/citizens/savings_and_investments/_attribute.html.erb)
Feel free to move it to `/shared/` and rename it if you do use it

app/forms/citizens/other_assets_form.rb
Kalagan: That's good too

Gemfile
tmckernan: Why do you need this gem if your not using assigns or assert_template in your request spec? Also we are using request specs to test the behaviour. Both assert_template and assigns assert the implementation, not the behaviour.

db/migrate/20181206142708_create_other_asset_declaration.rb
stephenrichards: I thought that it would allow money values of more than 2 decimal places to be stored, but testing shows that the default is to truncate it to 2, so happy to remove that.

spec/factories/other_assets_declarations.rb
stephenrichards: I think so.  You can't create an OtherAssetsDeclaration without a LegalAidApplication existing first.  Unless you know of a better way to do it?

reggieb: Have you tried just:
```ruby
FactoryBot.define do
  factory :other_assets_declaration do
    legal_aid_application
  end
end
```

Gemfile
stephenrichards: It's needed for `expect(response).to render_template(xxx)`

app/forms/citizens/other_assets_form.rb
```diff
@@ -0,0 +1,104 @@
...
+      ALL_ATTRIBUTES
+        .map { |attribute| send(attribute) }
+        .compact
+        .each { |value| value.delete!(',') }
+    end
+
+    def all_second_home_values_present_if_any_are_present
+      return unless any_second_home_value_present?
+
+      %i[second_home_value second_home_mortgage second_home_percentage].each do |attr|
```
reggieb: Could `SECOND_HOME_ATTRIBUTES` be used here.

app/forms/citizens/other_assets_form.rb
```diff
@@ -0,0 +1,104 @@
...
+        errors.add(attr, :blank) if __send__(attr).blank?
```
reggieb: I think `send` and `__send__` are the same thing in this object. So I think we should be consistent and use one in the object code.

app/views/citizens/other_assets/_second_home_conditional_checkbox.html.erb
```diff
@@ -0,0 +1,24 @@
+
+<%
+  check_box_attribute = 'check_box_second_home'
+  conditional_div_id = 'conditional_second_home'
+  hint = I18n.translate("helpers.hint.#{check_box_attribute}", default: nil)
+  value1 = model.second_home_value
+  value2 = model.second_home_mortgage
+  value3 = model.second_home_percentage
```
reggieb: Can more meaningful names be used here rather than `value(n)`. `sh_value`, `sh_mortgage`, `sh_percentage` perhaps if the full names are really problematic.

app/views/citizens/other_assets/show.html.erb
```diff
@@ -0,0 +1,30 @@
+<% content_for :navigation do %>
+  <%= link_to 'Back', citizens_additional_accounts_path, class: 'govuk-back-link', id: 'back' %>
```
reggieb: `t('generic.back')` rather than `'Back'` for the link text.

app/controllers/citizens/other_assets_controller.rb
```diff
@@ -0,0 +1,43 @@
+module Citizens
+  class OtherAssetsController < BaseController
+    def show
+      declaration
+      @form = Citizens::OtherAssetsForm.new(current_params)
```
reggieb: Actually, the latest version of `BaseForm` has fixed the issue with using `...Form.new(model: model)` so my original comment stands. You can replace this line with:
```ruby
@form = Citizens::OtherAssetsForm.new(model: legal_aid_application)
```
And remove the `current_params` params

PR #126, opened 2018-12-11
AP-239 Citizen capital 2a savings and investments

Kalagan:
[AP-239](https://dsdmoj.atlassian.net/browse/AP-239)

Create a new page: `/citizens/savings_and_investment`.
At the top, it shows the true layer accounts if there are any.

The page contains 7 checkbox/inputs with the same pattern:
Click on the checkbox to reveal the text input.
These amounts are saved in a new model : `savings_amount`

image

image

## Checklist

### DIFF
--- a/app/controllers/citizens/savings_and_investments_controller.rb
+++ b/app/controllers/citizens/savings_and_investments_controller.rb
@@ -1,6 +1,6 @@
 module Citizens
   class SavingsAndInvestmentsController < BaseController
-    before_action :bank_accounts, :back_link
+    helper_method :bank_accounts, :back_link
 
     def show
       @form = SavingsAmounts::SavingsAmountsForm.new(current_params)
@@ -18,8 +18,12 @@ module Citizens
 
     private
 
+    def bank_accounts
+      @bank_accounts ||= legal_aid_application.applicant.bank_providers.flat_map(&:bank_accounts)
+    end
+
     def back_link
-      @back_link ||= citizens_own_home_path
+      citizens_own_home_path
 
       # TODO
       # 1a for people that answered No to 1a
@@ -33,10 +37,6 @@ module Citizens
 
     def check_box_attributes
       @check_box_attributes ||= SavingsAmounts::SavingsAmountsForm::CHECK_BOXES_ATTRIBUTES
-    end
-
-    def bank_accounts
-      @bank_accounts ||= legal_aid_application.applicant.bank_providers.flat_map(&:bank_accounts)
     end
 
     def current_params
--- a/app/controllers/citizens/savings_and_investments_controller.rb
+++ b/app/controllers/citizens/savings_and_investments_controller.rb
@@ -1,0 +1,58 @@
+module Citizens
+  class SavingsAndInvestmentsController < BaseController
+    before_action :bank_accounts, :back_link
+
+    def show
+      @form = SavingsAmounts::SavingsAmountsForm.new(current_params)
+    end
+
+    def update
+      @form = SavingsAmounts::SavingsAmountsForm.new(form_params.merge(model: savings_amount))
+
+      if @form.save
+        render plain: 'Navigate to question 3a. Other capital assets'
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def back_link
+      @back_link ||= citizens_own_home_path
+
+      # TODO
+      # 1a for people that answered No to 1a
+      # 1d for people that answered either Yes on 1a and No on 1d
+      # 1e for people that answered either Yes on 1a and Yes on 1d
+    end
+
+    def attributes
+      @attributes ||= SavingsAmounts::SavingsAmountsForm::ATTRIBUTES
+    end
+
+    def check_box_attributes
+      @check_box_attributes ||= SavingsAmounts::SavingsAmountsForm::CHECK_BOXES_ATTRIBUTES
+    end
+
+    def bank_accounts
+      @bank_accounts ||= legal_aid_application.applicant.bank_providers.flat_map(&:bank_accounts)
+    end
+
+    def current_params
+      savings_amount.attributes.symbolize_keys.slice(*attributes)
+    end
+
+    def savings_amount
+      @savings_amount ||= legal_aid_application.savings_amount || legal_aid_application.build_savings_amount
+    end
+
+    def legal_aid_application
+      @legal_aid_application ||= LegalAidApplication.find(session[:current_application_ref])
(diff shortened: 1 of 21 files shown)


app/controllers/citizens/savings_and_investments_controller.rb
```diff
@@ -0,0 +1,58 @@
+module Citizens
+  class SavingsAndInvestmentsController < BaseController
+    before_action :bank_accounts, :back_link
```
reggieb: This isn't doing anything, as the method are lazy loading when first called.

app/controllers/citizens/savings_and_investments_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def show
+      @form = SavingsAmounts::SavingsAmountsForm.new(current_params)
```
reggieb: I think if you pass the `savings_amount` in as model, you don't have to do the more complicated `current_params` construction:
```ruby
SavingsAmounts::SavingsAmountsForm.new(model: savings_amount)
```

app/forms/savings_amounts/savings_amounts_form.rb
```diff
@@ -0,0 +1,55 @@
...
+      isa
+      cash
+      other_person_account
+      national_savings
+      plc_shares
+      peps_unit_trusts_capital_bonds_gov_stocks
+      life_assurance_endowment_policy
+    ].freeze
+
+    CHECK_BOXES_ATTRIBUTES = ATTRIBUTES.map { |attribute| "check_box_#{attribute}".to_sym }
```
reggieb: Should the resultant array be frozen?

app/forms/savings_amounts/savings_amounts_form.rb
```diff
@@ -0,0 +1,55 @@
...
+      end
+    end
+
+    def add_validations
+      ATTRIBUTES.each do |attribute|
+        check_box_attribute = "check_box_#{attribute}".to_sym
+        singleton_class.class_eval { validates attribute, numericality: { greater_than_or_equal_to: 0 }, allow_blank: true }
+        singleton_class.class_eval { validates attribute, presence: true, if: proc { |a| a.send(check_box_attribute).present? } }
+      end
+    end
```
reggieb: I'm not sure about this, but as `ATTRIBUTES` and all the methods called are defined at the class level, I think you can do this:
```
 module SavingsAmounts
  class SavingsAmountsForm

  ATTRIBUTES = %i[
      isa
      cash
      other_person_account
      national_savings
      plc_shares
      peps_unit_trusts_capital_bonds_gov_stocks
      life_assurance_endowment_policy
    ].freeze

  ATTRIBUTES.each do |attribute|
    check_box_attribute = "check_box_#{attribute}".to_sym
    validates attribute, numericality: { greater_than_or_equal_to: 0 }, allow_blank: true
    validates attribute, presence: true, if: proc { |a| a.send(check_box_attribute).present? }
  end
```
Note that `ATTRIBUTES.each` has to come after the constant is defined.

As well as being less code, this will have the advantage of only being called once (when the singleton `SavingsAmounts::SavingsAmountsForm` is first loaded), rather than when each form instance is initiated. (I'm a little concerned that at the moment, your code will update the singleton class every time an instance of the form is initiated, which may have unexpected consequences)

app/views/citizens/savings_and_investments/_attribute.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  checked = value.present? || model.send(check_box_attribute).present? ? 'checked' : ''
+%>
+
+<div class="govuk-checkboxes__item">
+  <%= form.check_box check_box_attribute, { class: 'govuk-checkboxes__input', 'data-aria-controls' => attribute, checked: checked }, 'true', '' %>
+  <%= form.label check_box_attribute, class: 'govuk-label govuk-checkboxes__label' %>
+  <% if hint %>
+  <span class="govuk-hint govuk-checkboxes__hint">
+    <%= hint %>
+  </span>
```
reggieb: Lines 11 to 14 could be replaced with:
```erb
<%= govuk_hint(hint, class: 'govuk-checkboxes__hint') if hint %>
```

app/views/citizens/savings_and_investments/show.html.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<%= render partial: 'accounts_summary', locals: { bank_accounts: @bank_accounts } if @bank_accounts.any? %>
+
+<%= form_with(model: @form,
+              scope: :savings_amount,
+              url: citizens_savings_and_investment_path,
+              method: :patch,
+              local: true) do |form| %>
+
+              <%= render partial: 'shared/forms/error_summary', locals: { model: @form } %>
```
reggieb: This line should be indents to match the beginning of the previous declaration (so two spaced in from the `form_with` line), rather than inline with the parameters passed into `form_with`

app/controllers/citizens/savings_and_investments_controller.rb
Kalagan: But they are not called anywhere

Kalagan: Although I should use
```
    def back_link
      @back_link = citizens_own_home_path
```
instead of 
```
    def back_link
      @back_link ||= citizens_own_home_path
```
because these methods only serve to populate the @ variables so they can be used in the page

reggieb: OK - my bad .... though an alternative solution would be to replace `before_action` with `helper_method`.
```ruby
helper_method :bank_accounts, :back_link
```
That would make `bank_accounts` and `back_link` available to call from the views.

app/controllers/citizens/savings_and_investments_controller.rb
Kalagan: I tried and it doesn't throw any error but it doesn't prepopulate the attributes with the values already present in the `savings_amount `

app/controllers/citizens/savings_and_investments_controller.rb
Kalagan: 👍 I'll use `helper_method`

app/controllers/citizens/savings_and_investments_controller.rb
reggieb: Hmm. I'll have to play - but away from this PR. The `current_params` pattern is used everywhere else so is fine here too.

PR #128, opened 2018-12-12
Add font type for currency unit

willc-work:
Incorrect font was being used for the currency unit in the property value page

### DIFF
--- a/app/assets/stylesheets/govuk-currency-input.scss
+++ b/app/assets/stylesheets/govuk-currency-input.scss
@@ -28,6 +28,7 @@ Used with:
 
 @media (min-width: 40.0625em) {
   .govuk-currency-input__inner__unit {
+    font-family: Arial, Helvetica, sans-serif;
     font-size: 19px;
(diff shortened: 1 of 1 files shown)


Kalagan:
I think we can now use `form.govuk_text_field` with `input_prefix: £`
Like on this page: https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/views/citizens/percentage_homes/show.html.erb#L19
Or my recent PR: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/126/files#diff-d33f90903cb3f9e021f96709ac5eb493R14

This is a feature I merged last week.

It uses this CSS file https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/assets/stylesheets/govuk-prefix-input.scss
Which has `font-family: "nta", Arial, sans-serif;`.
It's the font used for other text in the app.
How did you come up with ` ont-family: Arial, Helvetica, sans-serif;` ?

PR #129, opened 2018-12-12
Refactor property value page to use helper method for showing currenc…

willc-work:
Refactor the property value form to use the helper method for text fields that inserts a prefix. The existing version was not showing the currency unit in the correct font.

### DIFF
--- a/app/views/citizens/property_values/show.html.erb
+++ b/app/views/citizens/property_values/show.html.erb
@@ -13,7 +13,7 @@
 
       <%= govuk_fieldset_header t('.field_set_header'), padding_below: 6 %>
         <% input_error_class = govuk_error_class(form.object.errors[:property_value]) %>
-      <%= form.govuk_text_field :property_value, input_prefix: '£', class: "govuk-input govuk-!-width-one-third #{input_error_class}" %>
+      <%= form.govuk_text_field :property_value, input_prefix: t('currency.gbp'), class: "govuk-input govuk-!-width-one-third #{input_error_class}" %>
     <% end %>
 
       <%= govuk_submit_button t('generic.continue') %>
--- a/app/views/citizens/property_values/show.html.erb
+++ b/app/views/citizens/property_values/show.html.erb
@@ -12,8 +12,7 @@
         ) do %>
 
       <%= govuk_fieldset_header t('.field_set_header'), padding_below: 6 %>
-        <% input_error_class = govuk_error_class(form.object.errors[:property_value]) %>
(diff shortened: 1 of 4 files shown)


app/views/citizens/property_values/show.html.erb
```diff
@@ -12,13 +12,8 @@
...
 
       <%= govuk_fieldset_header t('.field_set_header'), padding_below: 6 %>
-      <%= form.label :property_value, t('.label'), class: 'govuk-label' %>
-      <%= govuk_hint t('.hint'), id: :property_value_hint %>
-      <%= govuk_error_message form.object.errors[:property_value].first %>
-      <%= govuk_currency_input_wrapper do %>
         <% input_error_class = govuk_error_class(form.object.errors[:property_value]) %>
-        <%= form.text_field :property_value, class: "govuk-input govuk-!-width-one-third govuk-currency-input__inner__input #{input_error_class}" %>
-      <% end %>
+      <%= form.govuk_text_field :property_value, input_prefix: '£', class: "govuk-input govuk-!-width-one-third #{input_error_class}" %>
```
tmckernan: `<%= form.govuk_text_field :property_value, input_prefix: '£', class: "govuk-input govuk-!-width-one-third #{input_error_class}" %>`
```suggestion
      <%= form.govuk_text_field :property_value, input_prefix: t('currency. gbp'), class: "govuk-input govuk-!-width-one-third #{input_error_class}" %>
```

Kalagan: You shouldn't need to pass `input_error_class`. It should automatically figure out if there is an error

app/views/citizens/property_values/show.html.erb
```diff
@@ -12,13 +12,7 @@
         ) do %>
 
       <%= govuk_fieldset_header t('.field_set_header'), padding_below: 6 %>
-      <%= form.label :property_value, t('.label'), class: 'govuk-label' %>
-      <%= govuk_hint t('.hint'), id: :property_value_hint %>
```
Kalagan: Can we remove `'.hint'` and `'.label'` from the locale file?
I think they are on lines  319 and 320

db/schema.rb
```diff
@@ -143,10 +143,10 @@
     t.boolean "open_banking_consent"
     t.datetime "open_banking_consent_choice_at"
     t.string "own_home"
-    t.decimal "percentage_home"
     t.decimal "property_value", precision: 10, scale: 2
-    t.decimal "outstanding_mortgage_amount"
+    t.decimal "percentage_home"
     t.string "shared_ownership"
+    t.decimal "outstanding_mortgage_amount"
```
Kalagan: I think this file shouldn't be modified.
If you revert it to the version in master and use `rails db:schema:load` that might fix your db (not sure)

PR #147, opened 2018-12-14
Make forms more consistent, and simplify initializating with existing objects

reggieb:
This PR: 
## Moves the responsibility of gathering existing attributes, from the controller to the form
That is, this controller code:
```ruby
def show
  @form = SomeForm.new(current_params)
end

def current_params
  model.attributes.symbolize_keys.slice(:an_attribute, :another_attribute)
end
```
Is replaced with:
```ruby
def show
  @form = SomeForm.new(model: model)
end
```
This means the complicated bit (deciding where to get the data (existing or passed in attribute)) is defined once (in `BaseForm`), and not in every controller.

## Moves responsibility for selecting the model from the form, to the controller

Each form now just accepts the object it applies to, and it relies on the controller to choose/gather the object and pass it to the form. So parameters passed to the form to allow it to select an object have been removed. That is, instead of passing in the id of the `LegalAidApplication` to the form for it to gather that, and then from that get the target object (e.g. `Applicant`), the controller gathers the target object and passes that directly to the form.

That simplified a lot of the form logic and remove a lot of the differences between forms.

## Made namespaces more consistent

The namespace reflects the object the form is applied to. So a form that modifies an `Applicant` is in the `Applicants` name space. This was the case for some, but not all forms.

### DIFF
--- a/app/controllers/citizens/own_homes_controller.rb
+++ b/app/controllers/citizens/own_homes_controller.rb
@@ -1,13 +1,11 @@
 module Citizens
   class OwnHomesController < BaseController
     def show
-      @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
-      @form = Citizens::OwnHomeForm.new(current_params)
+      @form = LegalAidApplications::OwnHomeForm.new(model: legal_aid_application)
     end
 
     def update
-      @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
-      @form = Citizens::OwnHomeForm.new(form_params)
+      @form = LegalAidApplications::OwnHomeForm.new(form_params)
 
       if @form.save
         if @form.own_home == 'no'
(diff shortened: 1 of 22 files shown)


app/controllers/citizens/own_homes_controller.rb
```diff
@@ -2,12 +2,12 @@ module Citizens
   class OwnHomesController < BaseController
     def show
       @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
```
tmckernan: This line can be removed as it has been replaced by legal_aid_application method

app/controllers/citizens/own_homes_controller.rb
```diff
@@ -2,12 +2,12 @@ module Citizens
...
-      @form = Citizens::OwnHomeForm.new(current_params)
+      @form = LegalAidApplications::OwnHomeForm.new(model: legal_aid_application)
     end
 
     def update
       @legal_aid_application = LegalAidApplication.find(session[:current_application_ref])
```
tmckernan: Same here, this line can be removed as it has been replaced by legal_aid_application method

PR #159, opened 2018-12-18
Refactor Check your answer 'redirect' mechanism

Kalagan:
I had a go at changing the way the "Check your answer" flow works.
By passing a `redirect_path` parameter around rather than using the state of the `legal_aid_application`.
Personally, I think it makes things simpler and we don't have to worry about the state.
Let me know what you think.

## Checklist

### DIFF
--- a/app/controllers/concerns/providers/steppable.rb
+++ b/app/controllers/concerns/providers/steppable.rb
@@ -45,17 +45,17 @@ module Providers
       }
     }.freeze
 
-    PATHS_NOT_REQUIRING_LEGAL_AID_APPLICATION_INSTANCE = [
-      :providers_legal_aid_applications_path
+    PATHS_NOT_REQUIRING_LEGAL_AID_APPLICATION_INSTANCE = %i[
+      providers_legal_aid_applications_path
     ].freeze
 
-    CHECK_YOUR_ANSWERS_STEP = :providers_legal_aid_application_check_provider_answers_path
-
-    CONTROLLERS_THAT_GO_FORWARD_NORMALLY_DURING_REVIEW = [:address_lookups].freeze
+    MULTI_STEPS_CONTROLLERS = %i[address_lookups].freeze
 
     included do
       # Define @back_step_url in controller to over-ride behaviour
       def back_step_url
+        return redirect_path if redirect_path
+
         raise "back step not found for controller name #{controller_name}" unless current_back_method || @back_step_url
 
         @back_step_url ||= send(current_back_method, with_legal_aid_application_if_needed(current_back_method))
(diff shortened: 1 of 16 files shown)


reggieb:
As discussed, my problem with this solution is that if the return is not immediate (for example address modification occurs over two pages - four actions) how do you pass the return information through these steps. This solution uses hidden fields in the affected forms. I think that spreads the logic around too much, in that at the moment only `Steppable` needs to make decisions based on the journey type (normal or check-your-answers), whereas in this solution each form in a multi-page modification, needs to make a decision - each of those forms needs the modification.

In summary, I think this is a perfectly valid way of handling the issue. I just don't think it is better than the current solution.

Having said that, I am hugely biased having written the current solution.

stephenrichards:
I think I agree with Rob that passing hidden fields around spreads the love to widely.

Personally, I'd be in favour of a solution the was system wide, and driven by a yaml config file.  The config file could specify the controller path, give the default Back and Forward paths, and then a series of "predicate methods" which if true, yielded another path.  That way, we have all the details for what goes where in one place, and just one object responsible for determining where to go.

Thoughts?

reggieb:
The main section of `Steppable` is a hash - so could be easily represented via yaml if you want to go that way.

Kalagan:
@stephenrichards ,
Yes I completely agree for the main flow and your predicate methods idea. Not sure if we can define logic in a YAML file though. Right now I'm working on https://dsdmoj.atlassian.net/browse/AP-251
And I'm thinking of building a hash like this:
```
STEPS_FLOW = {
  ...
  other_capital: {
    back: :savings_investments,
    forward: ->(application) { application.own_home_yes? ? :capital_restrictions : :check_answers }
  }
  ...
```

And then another hash to define the paths of each steps:

```
STEP_PATHS: {
  ...
  other_capital: ->(application) { ........_path(application) },
  capital_restrictions: ->(_application) { ........._path }
  ...
}
```

But for the check_your_answer logic, we need something else

reggieb:
Rather than two constants, could we do:
```ruby
STEPS_FLOW = {
  ...
  other_capital: {
    path: ->(application) { ........_path(application) },
    back: :savings_investments,
    forward: ->(application) { application.own_home_yes? ? :capital_restrictions : :check_answers }
  }
```
Also if we're defining procs within the definitions (which I don't have a problem with), then I think we should keep with using hashes rather than moving to using yaml.

reggieb:
Another consideration - is now the time to refactor the citizens journey so that the current legal aid application is passed via the url (as in the providers journey) rather than a session key. Otherwise, we'll have to do a fairly large rejig of the citizens version of steppable later.

Kalagan:
Do we necessarily need to do that?

PR #160, opened 2018-12-18
Ap 220 provider does your client own their home

reggieb:
[Jira AP-220](https://dsdmoj.atlassian.net/browse/AP-220)

This PR: 

- splits the citizens own home view into two: `show` and a `form` partial 
- uses the form in a new providers controller
- adds `Providers::OwnHomesController`
- changes `legal_aid_application.own_home` to use an `enum`


config/locales/en.yml
```diff
@@ -254,6 +254,13 @@ en:
         happen_next_heading: What happens next
         happen_next_text: "You will be sent the results of your client's financial assessment so you can confirm they are eligible for legal aid."
         back_button: "Back to your applications"
+    own_homes:
+      show:
+        h1-heading: Does your client own the home that they live in?
+        own_home:
```
mpw5: This section is the same as the citizen version starting on line 311. Is there a way of reusing that rather than defining everything twice? (Not a blocker...)

reggieb: If you look - the heading is different. "Does your client own the home that they live in?" rather than  "Do you own the home that you live in?".

reggieb: Following discussion, it was agreed that having all the translations for a provider page, with the provider section of the translation file was less likely to cause confusion, and therefore the current format should be kept.

mpw5: 👍

app/controllers/providers/own_homes_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def current_params
+      legal_aid_application.attributes.symbolize_keys.slice(:own_home)
```
Kalagan: Didn't you make some modification to avoid this but instead pass the model to the form?

app/views/citizens/own_homes/_form.html.erb
```diff
@@ -0,0 +1,36 @@
...
+            <%= form.label 'own_home_owned_outright', action_t('.own_home.owned_outright'), class: 'govuk-label govuk-radios__label' %>
+          </div>
+          <div class='govuk-radios__item'>
+            <%= form.radio_button 'own_home', 'no', class: "govuk-radios__input #{input_error_class}" %>
+            <%= form.label 'own_home_no', action_t('.own_home.no'), class: 'govuk-label govuk-radios__label' %>
+          </div>
+        </div>
+      </div>
+
+      <div class="govuk-!-padding-bottom-6"></div>
```
Kalagan: If you rebase from master you should be able to use the new `govuk_collection_radio_buttons` compoment. you could simplify this file a lot:
```
  <%= link_to t('generic.back'), citizens_consent_path(id: session[:current_application_ref]), class: 'govuk-back-link', id: 'back' %>
<% end %>

<% content_for(:page_title) { t('.h1-heading') } %>

<%= render partial: 'shared/forms/error_summary', locals: { model: @form } %>

<div class="govuk-grid-row">
    <%= form_with(model: @form, url: form_path, method: :patch, local: true) do |form| %>

      <%
        options = [
          { value: :mortgage,       label: action_t('.own_home.mortgage') },
          { value: :owned_outright, label: action_t('.own_home.owned_outright') },
          { value: :no,             label: action_t('.own_home.no') }
        ]
      %>

      <%= form.govuk_collection_radio_buttons(:own_home, options, :value, :label, title: content_for(:page_title)) %>

      <div class="govuk-!-padding-bottom-6"></div>

      <%= form.submit t('generic.continue'), id: 'continue', class: 'govuk-button' %>
    <% end %>
  </div>
</div>

```

app/controllers/providers/own_homes_controller.rb
reggieb: I'm an idiot.

app/views/citizens/own_homes/_form.html.erb
reggieb: Thank you!

config/environments/development.rb
```diff
@@ -58,5 +58,5 @@
 
   config.x.logs_faraday_response = true
 
-  Rails.application.routes.default_url_options[:host] = ENV['HOST'] || 'localhost:3002'
+  Rails.application.routes.default_url_options[:host] = ENV['HOST'] || "localhost:#{ENV.fetch('PORT', 3002)}"
```
reggieb: Fixes a personal annoyance - I run with `PORT=3000` in `.env` and have got tired of always having to correct the start citizen journey url.

config/locales/en.yml
```diff
@@ -228,6 +223,25 @@ en:
       heading_1: Apply for Legal Aid
       welcome_home: Welcome Home
   providers:
+    start:
```
reggieb: This section clashed with a section I'd added in a rebase - I thought I'd fix that by taking the opportunity to moving it to the start of the `providers` section as it's the start or that journey.

reggieb:
@Kalagan & @mpw5 I think I've address your concerns. Can you take another look please.

config/locales/en.yml
Kalagan: For the actual field labels, a 3rd option would be to use a `&` variable like use there:
https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/config/locales/en.yml#L47

```
    attributes:
      address: &address_attrs
        address_line_one: 'Building and street'
        address_line_two: ''
        city: 'Town or city'
        county: 'County'
        postcode: 'Postcode'
....
....
      applicants/address_form:
        <<: *address_attrs
```

reggieb: That's the best compromise

app/views/citizens/own_homes/_form.html.erb
```diff
@@ -0,0 +1,21 @@
...
+        options = [
+          { value: :mortgage,       label: action_t('.own_home.mortgage') },
+          { value: :owned_outright, label: action_t('.own_home.owned_outright') },
+          { value: :no,             label: action_t('.own_home.no') }
+        ]
+      %>
+
+      <%= form.govuk_collection_radio_buttons(:own_home, options, :value, :label, title: content_for(:page_title)) %>
+
+      <div class="govuk-!-padding-bottom-6"></div>
```
tmckernan: can you please move this form to shared/forms/_own_homes_form?

reggieb: Done

app/views/citizens/own_homes/show.html.erb
```diff
@@ -4,17 +4,4 @@
 
 <% content_for(:page_title) { t('.h1-heading') } %>
 
-<%= render partial: 'shared/forms/error_summary', locals: { model: @form } %>
```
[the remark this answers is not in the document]
reggieb: Are you sure or perhaps you're looking at old version. This is a screen shot of an error:

providers_own_home_error

However, thank you for making me look at the errors page as it's forced me to rethink the way translation files are working.

config/locales/en.yml
reggieb: However, my solution doesn't work on error, because on error the `show` template is rendered in the `update` action - kicking out the translation path.

I've reverted it all to use the default path the form file syntax.

app/views/citizens/own_homes/show.html.erb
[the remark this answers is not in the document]
willc-work: must have been an older version, apologies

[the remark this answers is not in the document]
reggieb: No problem

PR #168, opened 2018-12-19
AP-251 Fix citizen capital flow

Kalagan:
[AP-251](https://dsdmoj.atlassian.net/browse/AP-251)

Implementation of a service that defines within a hash the flow of pages with symbols, or predicates when there is a fork in the flow.
Each hash value defines a specific page with 3 values: `:path`, `:back`, `:forward`.

`:path` defines the path of the page
`:back` and `:forward` define which page go before and after

For example :
```ruby
  STEPS_FLOW = {
    ...
    property_values: {
      path: :citizens_property_value_path,
      back: :own_homes,
      forward: ->(application) { application.own_home_mortgage? ? :outstanding_mortgages : :shared_ownerships }
    },
    ..
  }
```

In this example, the page previous to the `:property_values` page is always `:own_homes` so we can use a symbol.
However, the next page should go to the `:outstanding_mortgages` page if the applicant owns their home with a mortgage, or the `:shared_ownerships` page if they don't.
In that case, we can use a predicate to figure out which is the next step.

This PR also includes a `Flowable` concern which sits between the controller and the `PageFlowService`.

This concern includes a `back_path` method similar to the one in `Steppable` plus a `go_forward` that can be used in all controllers when the form is submitted and will either redirect to the appropriate next page or, if the next page is not yet implemented, will render in plain text the path of the expected next page (see screenshot below).

image

## Checklist


app/services/page_flow_service.rb
```diff
@@ -0,0 +1,90 @@
+class PageFlowService
```
reggieb: Like it.

app/services/page_flow_service.rb
```diff
@@ -0,0 +1,90 @@
...
+      forward: ->(application) { application.shared_ownership? ? :percentage_homes : :savings_and_investments }
+    },
+    percentage_homes: {
+      path: :citizens_percentage_home_path,
+      back: :shared_ownerships,
+      forward: :savings_and_investments
+    },
+    savings_and_investments: {
+      path: :citizens_savings_and_investment_path,
+      back: lambda do |application|
```
reggieb: I think you can:
```ruby
back: -> do |application|
```
For consistency sake.

app/services/page_flow_service.rb
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }.freeze
+
+  def initialize(legal_aid_application, current_step)
+    @legal_aid_application = legal_aid_application
+    @current_step = current_step
+  end
+
+  def back_path
+    STEPS_FLOW[back_step][:path]
```
reggieb: Would `dig` be good here?
```ruby
STEP_FLOW.dig back_step, :path
```

app/services/page_flow_service.rb
```diff
@@ -0,0 +1,90 @@
...
+  def forward_step
+    @forward_step ||= step(:forward)
+  end
+
+  def step(direction)
+    raise "Step #{current_step} is not defined" unless STEPS_FLOW[current_step]
+
+    step = STEPS_FLOW[current_step][direction]
+
+    raise "#{direction.to_s.humanize} step of #{current_step} is not defined" unless step
```
reggieb: I'd suggest using `dig` and then just doing the check once:
```ruby
step =  STEPS_FLOW.dig current_step, direction
raise "#{direction.to_s.humanize} step of #{current_step} is not defined" unless step
```

app/services/page_flow_service.rb
```diff
@@ -0,0 +1,90 @@
...
+
+    step.is_a?(Symbol) ? step : step.call(legal_aid_application)
```
reggieb: As calling `call` occurs because the step is a `Proc` rather than because it is not a `Symbol` I'd suggest:
```ruby
return step unless step.is_a?(Proc)
step.call(legal_aid_application)
```

app/views/citizens/other_assets/show.html.erb
```diff
@@ -1,5 +1,5 @@
 <% content_for :navigation do %>
-  <%= link_to t('generic.back'), citizens_additional_accounts_path, class: 'govuk-back-link', id: 'back' %>
+  <%= link_to t('generic.back'), back_path, class: 'govuk-back-link', id: 'back' %>
```
reggieb: `link_to t('generic.back'), back_path, class: 'govuk-back-link', id: 'back'`

This code is repeated in at least 6 places. I'd think a helper method makes sense, along the lines of `provider_back_link`. In fact, if the two journeys used the same method name form the "Back url/path", the same helper method could be used across both journeys. It may even make sense to make the default behaviour of `content_for :navigation` to call this back link helper method.

spec/models/legal_aid_application_spec.rb
```diff
@@ -173,4 +173,64 @@
...
+  describe '#own_home?' do
+    context 'legal_aid_application.own_home is nil' do
+      before { legal_aid_application.update!(own_home: nil) }
+      it 'returns false' do
+        expect(legal_aid_application.own_home?).to eq(false)
+      end
+    end
+
+    context 'legal_aid_application.own_home is "no"' do
+      before { legal_aid_application.update!(own_home: 'no') }
```
reggieb: FYI - as `own_home` is an enum you can do this: `legal_aid_application.own_home_no!` to set `own_home` as 'no'

However, i'd suggest a better way of doing this is to use the existing traits (less database calls) to define `legal_aid_application`:
```ruby
context 'legal_aid_application.own_home is "no"' do
  let(:legal_aid_application) { create :legal_aid_application, :without_own_home }
  it 'returns false' do
    expect(legal_aid_application.own_home?).to eq(false)
  end
end
```

spec/models/legal_aid_application_spec.rb
```diff
@@ -173,4 +173,64 @@
...
+      it 'returns false' do
+        expect(legal_aid_application.own_home?).to eq(false)
+      end
+    end
+
+    context 'legal_aid_application.own_home is not "no"' do
+      before { legal_aid_application.update!(own_home: :mortgage) }
```
reggieb: Again - use `let(:legal_aid_application) { create :legal_aid_application, :with_own_home_mortgaged}`

spec/models/legal_aid_application_spec.rb
```diff
@@ -173,4 +173,64 @@
...
+    end
+
+    context 'no home, savings or assets' do
+      it 'returns nil' do
+        expect(legal_aid_application.own_capital?).to eq(nil)
+      end
+    end
+
+    context 'own home' do
+      before { legal_aid_application.update!(own_home: :owned_outright) }
```
reggieb: Can traits be added to the factory - rather than using `update!`

spec/services/page_flow_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      before { stub_const("#{described_class}::STEPS_FLOW", own_homes: {}) }
+
+      it 'raises an error' do
+        expect { subject.back_path }.to raise_error("Back step of #{current_step} is not defined")
+      end
+    end
+  end
+
+  describe '#forward_path' do
+    context 'forward_path has no logic' do
```
reggieb: OK - this is really picky but ... This context describes the default behaviour, and therefore I don't think it should be in a context block. Personally, I think there should usually be a default behaviour (no context), and then contexts should be used to describe behaviour away from the default. If nothing else that will simplify the failure text (which for this spec will be something like: this spec failed because "#forward_path forward_path has no logic")

spec/services/page_flow_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      let(:current_step) { :percentage_homes }
+      let(:expected_forward_path) { :citizens_savings_and_investment_path }
+
+      it 'returns forward path' do
+        expect(subject.forward_path).to eq(expected_forward_path)
+      end
+    end
+
+    context 'forward_path has logic' do
```
reggieb: "forward_path" is already defined in the `describe` block, so using it here results in perhaps a confusing failure message. So I'd suggest:
```ruby
context 'with logic` do
```
As then the failure message will be prepended "#forward_path with logic ..."
Yes - I am still being picky.

spec/services/page_flow_service_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+
+    context 'forward step is not defined' do
+      let(:current_step) { :own_homes }
+      before { stub_const("#{described_class}::STEPS_FLOW", own_homes: {}) }
+
+      it 'raises an error' do
+        expect { subject.forward_path }.to raise_error("Forward step of #{current_step} is not defined")
+      end
+    end
```
reggieb: Again - I don't think the error handling needs to be too clever. It should only occur through miscoding and/or during the build of a new step.

app/models/concerns/capital.rb
```diff
@@ -0,0 +1,15 @@
...
+  included do
+    def positive?
+      amount_attributes.values.any? { |v| v.present? && v.positive? }
+    end
+
+    private
+
+    def amount_attributes
+      attributes.except('id', 'legal_aid_application_id', 'updated_at', 'created_at')
+    end
```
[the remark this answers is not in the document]
Kalagan: Yes, I wouldn't mind using his implementation instead. Mine and his are quite similar (using a blacklist of attributes rather than a whitelist)

app/services/page_flow_service.rb
Kalagan: By default, Rubocop wants `lambda do |x|` for multiline lambdas.
(https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/Lambda)
But I agree with you and I'm changing our Rubocop rules to always use `->(x)` instead of `lambda do |x|`

PR #169, opened 2018-12-20
Ap223 provider outstanding mortgage

willc-work:
Adds the outstanding mortgage form to the provider journey and refactors the citizen form

[Link to story](https://dsdmoj.atlassian.net/browse/AP-223)

Used outstanding_mortgage form on citizen journey and refactored using Stephen's approach for the property value form so that translations are journey specific. 

Added the form to the steppable module - still needs to be defined with correct steps 

## Checklist

### DIFF
--- a/app/views/citizens/outstanding_mortgages/show.html.erb
+++ b/app/views/citizens/outstanding_mortgages/show.html.erb
@@ -1,36 +1,9 @@
-<%= render 'shared/forms/error_summary', model: @form %>
+<% content_for(:page_title) { t('.h1-heading') } %>
 
-<fieldset class="govuk-fieldset">
-  <div class="govuk-!-padding-bottom-2"></div>
+<% content_for :navigation do %>
+    <%= link_to t('generic.back'), citizens_property_value_path, class: 'govuk-back-link', id: 'back' %>
+<% end %>
 
-  <div class="govuk-grid-row">
-    <div class="govuk-grid-column-two-thirds">
-      <%= form_with(
-            model: @form,
-            url: citizens_outstanding_mortgage_path,
-            method: :patch,
-            local: true
-          ) do |form| %>
-        <div class="govuk-!-padding-bottom-2"></div>
+<%= render partial: 'shared/forms/error_summary', locals: { model: @form, attribute_prefix: :legal_aid_application_ } %>
 
-        <%= govuk_form_group(
-              show_error_if: form.object.errors[:outstanding_mortgage_amount].present?,
-              input: :outstanding_mortgage_amount,
-              hint_id: :outstanding_mortgage_amount_hint
-            ) do %>
-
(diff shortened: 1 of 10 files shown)


app/views/citizens/outstanding_mortgages/show.html.erb
```diff
@@ -1,36 +1,9 @@
...
-        <div class="govuk-grid-row">
-          <div class="govuk-grid-column-two-thirds">
-            <%= form.submit t('generic.continue'), id: 'continue', class: 'govuk-button govuk-button' %>
-         </div>
-       </div>
-      <% end %>
-    </div>
-  </div>
-</fieldset>
+<%= render partial: '/shared/forms/outstanding_mortgage_form', locals: { model: @form, url: citizens_outstanding_mortgage_path } %>
```
stephenrichards: I think the whole page after the error summary should be wrapped in a two thirds column inside a grid row, so something like:

```
<div class="govuk-grid-row">
    <!-- h1 heading (if not in form partial) -->
    <!-- render partial for form -->
  </div>
</div>```

File: config/locales/en.yml [hunk 2]
```diff
@@ -341,6 +340,12 @@ en:
           nino: National Insurance number
           address: Correspondence address
           email: Email address
+    outstanding_mortgages:
+      show:
+        h1-heading: What is the outstanding mortgage?
+      field_set_header: What is the outstanding mortgage on your client's home?
+      outstanding_mortgage_amount: Enter the outstanding mortgage amount
+      hint: ' '
```
mpw5: Should there be some hint text here? If not, can the line be removed?

willc-work: I need to confirm with Simon so its a placeholder for now. I was also testing to see if this would insert a blank hint to give me some spacing😜

app/views/citizens/outstanding_mortgages/show.html.erb
willc-work: thanks, fixed in both citizen and provider view now.

config/locales/en.yml
mpw5: 👍

app/views/citizens/outstanding_mortgages/show.html.erb
```diff
@@ -5,5 +5,8 @@
 <% end %>
 
 <%= render partial: 'shared/forms/error_summary', locals: { model: @form, attribute_prefix: :legal_aid_application_ } %>
-
-<%= render partial: '/shared/forms/outstanding_mortgage_form', locals: { model: @form, url: citizens_outstanding_mortgage_path } %>
+  <div class="govuk-grid-row">
```
mpw5: Looks like this is isn't aligned quite right - there's some whitespace that doesn't need to be there, the whole <div> block can be left-aligned (the providers version is correct).

PR #176, opened 2019-01-03
AP-250 A list of legal aid application with links to current positions

reggieb:
[Jira AP-250](https://dsdmoj.atlassian.net/browse/AP-250)

# What

- Records the current controller on each legal aid application view in the providers journey.
- Updates the list of legal aid application in  `provider/legal_aid_applications#index` with a link to the current journey position for each application in the list.

# Notes

- The goal is for the list to only be of applications belonging to the logged in provider. As authentication is not ready, the list is of the first ten legal aid applications.
- I've refactored `Providers::Steps` so that I could use that functionality to provide the link to the current step, and to being it closer to Julien's pattern used in the Citizen journey (more work needed here, but I hope this refactor will make it easier to bring the two together - that is, use the same techniques for both journeys.)

legal_aid_applications_index


reggieb:
AP-238 contains the styling of the application list - so I'll incorporate that into this PR

app/views/providers/legal_aid_applications/_legal_aid_applications.html.erb
```diff
@@ -0,0 +1,30 @@
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <table class="govuk-table">
+      <caption class="govuk-table__caption">Current applications</caption>
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header" scope="col">Client's name</th>
```
Kalagan: Should we store these in the locale file?

reggieb: Good point!!

config/locales/en.yml
```diff
@@ -606,3 +608,6 @@ en:
     gbp: £
     eur: €
     usd: $
+  time:
+    formats:
+      date: '%a, %d %b %Y'
```
stephenrichards: This seems to be a date format under neath time, and the date and time formats are already defined in lines 2-7 of this file.

reggieb: Poor merging on my part - thank you. I've corrected that.

PR #177, opened 2019-01-03
AP-238 list provider applications

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-238)

Listed all legal Aid applications on the Provider home page.

There is no link yet to open an individual application - am expecting this to be done as part of AP-250.

Also, until the provider sign on is merged, this page lists ALL applications; this will later be changed to list just those belonging to the provider that the current user is a member of.

## Checklist


reggieb:
Nice.

PR #178, opened 2019-01-03
AP-246 Applicant Check Answers

Kalagan:
## AP-246 Applicant Check Answers

[AP-246](https://dsdmoj.atlassian.net/browse/AP-246)

- Adding the "Check answers" page at the end of the flow.
- It is reusing some code from the providers's check answers page
- This task display the right information with the "change" link to right page.
- But it doesn't implement the back-and-forth page change. This will be done with [AP-258](https://dsdmoj.atlassian.net/browse/AP-258)

## Checklist


app/models/legal_aid_application.rb
```diff
@@ -1,4 +1,5 @@
-class LegalAidApplication < ApplicationRecord
+# TODO: Refactor this class to make it smaller
+class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLength
```
stephenrichards: I think we need a `# rubocop:enable Metrics/ClassLength` at the bottom of this class

Kalagan: You don't need one if `rubocop:disable` is inline
https://rubocop.readthedocs.io/en/latest/configuration/#disabling-cops-within-source-code (bottom of page)

Kalagan: I rebased from `master` and used your code

PR #183, opened 2019-01-07
Improve error reporting of BenefitChecker API

Kalagan:
Improve error reporting of BenefitChecker API by including more information about the error.
Recently, the API raised an error because the `surname` passed in parameter was only one character.
However, the error we saw in Sentry was simply **"Savon::SOAPFault: (soapenv:Server.generalException)"**.
With this PR, the error includes the real error message **"Error in request parameter 'Surname'"**

## Checklist

### DIFF
--- a/spec/services/benefit_check_service_spec.rb
+++ b/spec/services/benefit_check_service_spec.rb
@@ -59,8 +59,20 @@ RSpec.describe BenefitCheckService do
     context 'when the API raises an error' do
       let(:last_name) { 'SERVICEEXCEPTION' }
 
+      it 'raises a detailed error' do
+        expect { subject.call }.to raise_error(described_class::ApiError, /Service unavailable/)
+      end
+    end
+
+    context 'when the API times out' do
+      before do
+        savon_client = double(:savon_client)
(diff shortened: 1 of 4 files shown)


spec/services/benefit_check_service_spec.rb
```diff
@@ -59,8 +59,16 @@
     context 'when the API raises an error' do
       let(:last_name) { 'SERVICEEXCEPTION' }
 
+      it 'raises a detailed error' do
+        expect { subject.call }.to raise_error(described_class::ApiError, /Service unavailable/)
+      end
+    end
+
+    context 'when the API times out' do
+      before { allow_any_instance_of(Savon::Client).to receive(:call).and_raise(Net::ReadTimeout) }
```
reggieb: Can you do this:
```ruby

before do
  savon_client = double(:savon_client)
  allow(savon_client).to receive(:call).and_raise(Net::ReadTimeout)
  allow(subject).to receive(:soap_client).and_return(savon_client)
end
```
Uglier, but avoids using `allow_any_instance_of` (if my code works). Not totally sure this is an improvement but worth suggesting.

Kalagan: Yes it's a bit better

PR #185, opened 2019-01-08
AP-248 provider log out mark up

reggieb:
[Jira AP-248](https://dsdmoj.atlassian.net/browse/AP-248)

Adds two links to the header that will appear if the provider is logged in. These links will provide log out and a link to the providers user information.

## Tasks
- [x] Create markup
- [ ] Integrate with provider authentication (once [AP-98](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/158) completed)


reggieb:
Note that to get the desired behaviour, I had to move the `provider_root` outside the provider authentication scope. That is, anyone can see that page - log in is only prompted on clicking "Start" on that page.

I've also added some comments regarding `provider_root` as that should be externally hosted eventually.

spec/lib/id_p_settings_adapter_spec.rb
```diff
@@ -3,7 +3,9 @@
 RSpec.describe IdPSettingsAdapter do
   describe 'IdPSettingsAdapter.mock_saml?' do
     it 'return false' do
-      expect(IdPSettingsAdapter.mock_saml?).to eq false
+      mock_saml_setting = Rails.configuration.x.laa_portal[:mock_saml]
+      expected = mock_saml_setting.present? ? ActiveRecord::Type::Boolean.new.cast(mock_saml_setting) : false
+      expect(IdPSettingsAdapter.mock_saml?).to eq expected
```
Kalagan: You could test the different results with
```ruby
allow(Rails.configuration.x.laa_portal).to receive(:mock_saml).and_return('true')
```
and
```ruby
allow(Rails.configuration.x.laa_portal).to receive(:mock_saml).and_return('false')
```

spec/lib/id_p_settings_adapter_spec.rb
```diff
@@ -3,7 +3,9 @@
 RSpec.describe IdPSettingsAdapter do
   describe 'IdPSettingsAdapter.mock_saml?' do
     it 'return false' do
-      expect(IdPSettingsAdapter.mock_saml?).to eq false
+      mock_saml_setting = Rails.configuration.x.laa_portal[:mock_saml]
```
Kalagan: I think `laa_portal.mock_saml` works instead of `laa_portal[:mock_saml]`

reggieb: That's better - thank you - though implementing your other suggestion removes this line.

PR #209, opened 2019-01-15
Remove unused API code

mpw5:
## Remove legacy API code

There is some legacy code from when this application was an API. This
is no longer needed as all functionality has been brought into the app.

This commit removes the controllers and associated tests.

## Checklist


app/controllers/api_controller.rb
```diff
@@ -1,11 +1,2 @@
 class ApiController < ActionController::API
-  rescue_from ActiveRecord::RecordNotFound do
-    render json: {}, status: :not_found
-  end
-
-  protected
-
-  def render_400(errors)
-    render json: { errors: ErrorsSerializer.new(errors) }, status: :bad_request
-  end
```
reggieb: While we are keeping at least one AP call - should this be kept?

tmckernan:
@mpw5  can you remove these routes

PR #211, opened 2019-01-15
AP-270 Clearer subheading on Check answers pages

reggieb:
[JIra AP-270](https://dsdmoj.atlassian.net/browse/AP-270)

The main change is that when there are multiple answers to a "question" the answers are listed below the question--link line.

To achieve this the logic that decides which pattern will be used has been moved from the partial to the helper - and then each option has been given a dedicated partial.

To accommodate the more complicated currency items, a third item type has been added.

The result look like this:
check_you_answer_demo


reggieb:
Small tweak to display correctly when there are no capital entries with a set of items (here - no savings)
check_your_answers_nil_savings

PR #213, opened 2019-01-16
Add clean_up_ecr script

mpw5:
## What

AWS ECR repositories have a limit of 1000 images. When this limit is reached no further builds are possible.

This script will delete all but the last 30 days of images, allowing builds to resume.

To run it, first follow the instructions in https://dsdmoj.atlassian.net/wiki/spaces/ATPPB/pages/1377992881/ECR+Repository+Maintenance to connect to the ECR instance.

(The code for this was stolen shamelessly from https://gist.github.com/dmerrick/503655db7ff90b70c2d15fc9dd83b45e)

## Checklist

### DIFF
--- a/bin/clean_up_ecr.rb
+++ b/bin/clean_up_ecr.rb
@@ -1,0 +1,25 @@
+require 'json'
+
+repo = 'laa-apply-for-legal-aid/applyforlegalaid-service'
+delete_if_older_than = 30 # days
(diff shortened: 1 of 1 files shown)


Kalagan:
In the future, it might be interesting to have CircleCI do it automatically. For every job or we could add it in our nightly job

PR #218, opened 2019-01-18
AP-305 Contact Us page

Kalagan:
## Contact Us page

[AP-305](https://dsdmoj.atlassian.net/browse/AP-305)

Contact Us page with phone and email details

## Checklist


app/views/contacts/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%= content_for(:page_title) %></h1>
+
+    <p class="govuk-body"><%= t('.page_header') %></p>
+
+    <p class="govuk-body">
+      <%= t('.case_inquiries.header') %>
+      <br />
```
reggieb: I think we should be using GOV-UK spacing elements rather than `br` tags in our markup. See https://design-system.service.gov.uk/styles/spacing/

app/views/layouts/_footer.html.erb
```diff
@@ -6,8 +6,7 @@
         <h2 class="govuk-visually-hidden"><%= t('layouts.application.footer.support') %></h2>
         <ul class="govuk-footer__inline-list">
           <li class="govuk-footer__inline-list-item">
-            <a class="govuk-footer__link" href="mailto:apply@digital.justice.gov.uk">
-              <!-- TO DO add in correct link/email -->
+            <a class="govuk-footer__link" href="/contact">
```
reggieb: Using a route helper method and `link_to` would be preferable especially as paths starting with a forward slash can be problematic if the site gets hosted in a sub-domain in the future.
```erb
<%= link_to t('layouts.application.footer.contact'), contact_path, class: 'govuk-footer__link' %>
```
I think `contact_path` will be the correct route helper method.

features/providers/step_definitions/civil_journey_steps.rb
```diff
@@ -108,6 +108,10 @@
   expect(page).to have_content('Check your answers')
 end
 
+Then('I should be on the {string} page') do |title|
```
reggieb: There are a number of steps defined with a very similar structure to this. In fact you use one on line 241 of `civil_application_journey.feature`: "Then I should be on the Applicant page".

I'd suggest a slight rewording that I think will make it a little clearer what is being match here:
```
Then('I should be on a page showing {string}') do |title|
```
The line in the feature then becomes:
```
Then I should be on a page showing "Contact us"
```
We could probably also do with a Tech Debt ticket to remove all steps using the page specific form, and replace them with your more generic version.

spec/requests/contact_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      expect(response.body).to include(I18n.translate('contacts.show.h1-heading'))
+      expect(response.body).to include(I18n.translate('contacts.show.page_header'))
+      expect(response.body).to include(I18n.translate('contacts.show.case_inquiries.header'))
+      expect(response.body).to include(I18n.translate('contacts.show.case_inquiries.phone.text'))
+      expect(response.body).to include(I18n.translate('contacts.show.case_inquiries.phone.phone_number'))
+      expect(response.body).to include(I18n.translate('contacts.show.technical_support.header'))
+      expect(response.body).to include(I18n.translate('contacts.show.technical_support.email.text'))
+      expect(response.body).to include(I18n.translate('contacts.show.technical_support.email.email_address'))
+      expect(response.body).to include(I18n.translate('contacts.show.technical_support.phone.text'))
+      expect(response.body).to include(I18n.translate('contacts.show.technical_support.phone.phone_number'))
```
reggieb: Do we really need all these checks?

Kalagan: maybe not...
Which ones would you keep?

Kalagan:
@reggieb I agree with all your comments.
Sorry for merging it while you were reviewing it.
No need to create a tech debt ticket. I'll create a new PR with all the changes (once I'm done with my current ticket).

spec/requests/contact_spec.rb
reggieb: I'd suggest the ones that are important are the phone number and email address.

app/views/contacts/show.html.erb
Kalagan: But according to the prototype https://legal-aid-substantive.herokuapp.com/footer/contact,
we don't want any spacing. Just a new line

reggieb: OK

PR #219, opened 2019-01-18
AP-291 Use h1 for page titles

reggieb:
Standardising pages so that they have a more consistent layout and use the h1 or fieldset heading as the page title. In doing so I have had to modify most pages and ensure they follow a standard pattern.

This work has included:

- Making sure that back links are defined in either `STEPS` (Provider Journey) or `PageFlowService` (Citizen Journey) where possible
- Fixed an issue that prevented Citizens going back to the first page in their journey via back links.

The main change is to update Citizen and Provider pages so that they use common templates. This is done via a couple of helper methods `citizen_page` and `provider_page` respectively. So now if you add a new page in the provider journey like this:
```erb
<%= provider_page page_title: t('.page_title') do %>
  <p>Some content</p>
<% end %>
```
The page will be set up with a page title, heading, and back link, and the content will be display within a two-thirds column. See `PageTemplateHelper` for information about how this behaviour can be modified.


Kalagan:
image
LEFT is this PR
RIGHT is master

According to gov.uk Design system, we want the left one:
https://design-system.service.gov.uk/components/radios/
image

Kalagan:
Although some pages in master (`/shared_ownership`, `/percentage_home`) are already following the pattern from the left and your PR is not breaking anything.

Kalagan:
Simon just confirmed to me that **Check your answers** pages should be 2/3rd.
So all good 👍

reggieb:
@Kalagan I believe the current position is that if a form consists of a single input, on error the red bar should extend past the heading. If there are multiple inputs then the red bar should only extend next to the input that has failed.

I hope what I have done is flexible enough to allow both. I tried to match the current pages - so where the header was within a fieldset and used an h1 `govuk-fieldset__heading` to display the header, I the **form** template is used. Where that pattern didn't work I was able to use one of the other templates to match the current behaviour.

PR #221, opened 2019-01-18
Changes to UAT deploy process

stephenrichards:
## Changes to UAT deploy process

This PR makes two changes to the UAT deploy process:
* Enables the deploy to UAT to be kicked off without waiting for the test suite to successfully complete
* Checks to see if there are more than 10 releases on the UAT environment, and deletes the oldest until there are only 10 remaining

## Checklist


bin/uat_deploy
```diff
@@ -37,7 +42,8 @@ deploy() {
                 --set image.repository="$IMG_REPO" \
                 --set image.tag="${CIRCLE_SHA1}" \
                 --set ingress.hosts="{$RELEASE_HOST}" \
-                --set replicaCount="1"
+                --set replicaCount="1" \
+                --timeout 60
```
Kalagan: the default one is 300, isn't it? Did you mean to reduce it from 5 minutes to 1 minute?

PR #232, opened 2019-01-23
AP-288 Mk II Relax National Insurance Test validation in some environments

reggieb:
A re-write of https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/206 to fix issues with merged secrets files


app/models/applicant.rb
```diff
@@ -51,6 +51,17 @@ def validate_national_insurance_number
     return if national_insurance_number.blank?
 
     normalise_national_insurance_number
+    return if test_level_validation? && known_test_ninos.include?(national_insurance_number)
+
     errors.add(:national_insurance_number, :invalid) unless NINO_REGEXP =~ national_insurance_number
   end
+
+  # These Nation Insurance Numbers are used for testing but do not match NINO_REGEXP
```
Kalagan: "Nation Insurance Numbers"? 
Isn't it "National Insurance Numbers"?

helm_deploy/apply-for-legal-aid/templates/deployment.yaml
```diff
@@ -158,6 +158,11 @@ spec:
               secretKeyRef:
                 name: apply-for-legal-aid-elasticache-instance-output
                 key: auth_token
+          - name: TEST_LEVEL_NINO_VALIDATION
+            valueFrom:
+              secretKeyRef:
+                name: {{ template "apply-for-legal-aid-uat.fullname" . }}
```
Kalagan: I think it should be `"apply-for-legal-aid.fullname"` instead of `"apply-for-legal-aid-uat.fullname"`

reggieb:
Thank you @Kalagan - I've made those changes

reggieb:
Following further testing I've found that the original solution didn't work - mainly because `Applicants::BasicDetailsForm` was duplicating many of the validations. So I've removed the duplication and made this object the sole owner of these validations.

Also due to the problems I had trying to deploy a new environment dedicated environment variable, I've used an existing one (`LAA_PORTAL_MOCK_SAML`) to identify when the app is in staging or UAT.

app/models/applicant.rb
```diff
@@ -12,45 +12,11 @@ class Applicant < ApplicationRecord
...
-      dob_year: date_of_birth.strftime('%Y'),
-      dob_month: date_of_birth.strftime('%m'),
-      dob_day: date_of_birth.strftime('%d')
-    }
-  end
-
-  private
-
-  def validate_date_of_birth
-    errors.add(:date_of_birth, :invalid) if date_of_birth.present? && (date_of_birth < Date.new(1900, 0o1, 0o1) || date_of_birth > Date.today)
```
Kalagan: we don't need this anymore?

app/models/applicant.rb
```diff
@@ -12,45 +12,11 @@ class Applicant < ApplicationRecord
...
     email
   end
 
   def full_name
     "#{first_name} #{last_name}".strip
   end
-
-  def dob
-    {
-      dob_year: date_of_birth.strftime('%Y'),
```
Kalagan: we don't need this method anymore?

app/models/applicant.rb
reggieb: That's now done solely by `BasicDetailsForm` lines 15-21.

app/models/applicant.rb
reggieb: It seems not. I had it in, and when the tests were run it was showing as no test coverage. I removed it, test coverage went to 100% and nothing broke.

app/models/applicant.rb
Kalagan: Ah ok I didn't realise it was duplicated

PR #233, opened 2019-01-24
AP-260 Merits Assessment - Has client received legal help for the matter?

Kalagan:
## What

[AP-260](https://dsdmoj.atlassian.net/browse/AP-260)

- Add `text_area` to `FormBuilder`
- new page: Merits Assessment - Has client received legal help for the matter?

## Checklist

### DIFF
--- a/app/controllers/providers/client_received_legal_helps_controller.rb
+++ b/app/controllers/providers/client_received_legal_helps_controller.rb
@@ -3,6 +3,7 @@ module Providers
     include ApplicationDependable
     include Steppable
     include SaveAsDraftable
+    before_action :authorize_legal_aid_application
 
     def show
       @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(model: merits_assessment)
@@ -31,6 +32,10 @@ module Providers
 
     def client_received_legal_help_params
       params.require(:merits_assessment).permit(:client_received_legal_help, :application_purpose)
+    end
+
+    def authorize_legal_aid_application
+      authorize legal_aid_application
     end
   end
 end
--- a/app/controllers/providers/client_received_legal_helps_controller.rb
+++ b/app/controllers/providers/client_received_legal_helps_controller.rb
@@ -1,0 +1,36 @@
+module Providers
+  class ClientReceivedLegalHelpsController < BaseController
+    include ApplicationDependable
+    include Steppable
+    include SaveAsDraftable
+
+    def show
+      @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(model: merits_assessment)
+    end
+
+    def update
+      @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(client_received_legal_help_params.merge(model: merits_assessment))
+
+      if @form.save
+        # TODO: remove this condition once next step is implemented
+        if params.key?(:continue_button)
+          render plain: 'Placeholder: Are the proceedings, for which funding is being sought, currently, before the court?'
+          return
+        end
+        continue_or_save_draft
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def merits_assessment
+      @merits_assessment ||= legal_aid_application.merits_assessment || legal_aid_application.build_merits_assessment
+    end
+
+    def client_received_legal_help_params
+      params.require(:merits_assessment).permit(:client_received_legal_help, :application_purpose)
+    end
+  end
+end
--- a/app/forms/base_form.rb
+++ b/app/forms/base_form.rb
@@ -102,7 +102,7 @@ module BaseForm
 
       attrs.map(&:to_s).each do |method|
         model_value = model_attributes[method]
-        instance_variable_set(:"@#{method}", model_value) if model_value && attributes[method].nil?
+        instance_variable_set(:"@#{method}", model_value) if !model_value.nil? && attributes[method].nil?
       end
     end
   end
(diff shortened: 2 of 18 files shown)


app/forms/base_form.rb
```diff
@@ -102,7 +102,7 @@ def set_instance_variables_for_attributes_if_not_set_but_in_model(attrs:, model_
 
       attrs.map(&:to_s).each do |method|
         model_value = model_attributes[method]
-        instance_variable_set(:"@#{method}", model_value) if model_value && attributes[method].nil?
+        instance_variable_set(:"@#{method}", model_value) if !model_value.nil? && attributes[method].nil?
```
Kalagan: I had to make this change for boolean.
When `model_value` is `false` and `attributes[method]` is `nil`

app/forms/merits_assessments/client_received_legal_help_form.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    before_validation :clear_application_purpose
+    validates :client_received_legal_help, presence: true
+    validates :application_purpose, presence: true, if: proc { |form| form.client_received_legal_help.to_s == 'false' }
+
+    private
+
+    def clear_application_purpose
+      application_purpose.clear if client_received_legal_help.to_s == 'true'
+    end
```
reggieb: I wonder if this would be a little cleaner:
```ruby
validates :application_purpose, presence: true, unless: :client_received_legal_help?

private

def client_received_legal_help?
  client_received_legal_help.to_s == 'true'
end

def clear_application_purpose
  application_purpose.clear if client_received_legal_help?
end
```
It means that both the clearing process and the validation use the same comparison logic and handles instances where `client_received_legal_help` is neither 'true' nor 'false'.

app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -5,13 +5,14 @@ class FormBuilder < ActionView::Helpers::FormBuilder # rubocop:disable Metrics/C
     # https://guides.rubyonrails.org/configuring.html#configuring-action-view
     ActionView::Base.field_error_proc = proc { |html_tag| html_tag.html_safe }
 
-    CUSTOM_OPTIONS = %i[label input_prefix field_with_error hint title inline].freeze
+    CUSTOM_OPTIONS = %i[label input_prefix field_with_error hint title inline legacy_method].freeze
```
reggieb: I wonder if `text_input` would be a better label. I'm not sure what these methods are a legacy of.

spec/requests/providers/client_received_legal_helps_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+        end
+
+        it 'updates the record' do
+          expect(legal_aid_application.merits_assessment.reload.client_received_legal_help).to eq(client_received_legal_help)
+          expect(legal_aid_application.merits_assessment.reload.application_purpose).to eq(application_purpose)
+        end
+
+        # TODO: remove this test and implement next one when next page is known
+        it 'returns next page placeholder' do
+          expect(response.body).to include('Placeholder: Are the proceedings, for which funding is being sought, currently, before the court?')
```
reggieb: I don't think there is a need to match the whole text. `include('Placeholder: Are the proceedings')` is enough, and means the specs won't fail if someone removes a comma.

spec/requests/providers/client_received_legal_helps_spec.rb
```diff
@@ -0,0 +1,119 @@
+require 'rails_helper'
+
+RSpec.describe 'provider own home requests', type: :request do
```
mpw5: It looks like this was copied over from another spec 😉. The description should probably be changed.

db/migrate/20190123151514_add_application_purpose_to_nerits_assessments.rb
```diff
@@ -0,0 +1,5 @@
+class AddApplicationPurposeToNeritsAssessments < ActiveRecord::Migration[5.2]
```
willc-work: the name of this migration refers to 'Nerits' rather than 'Merits' not a huge issue at all and maybe more work to change than is worth

app/lib/govuk_elements_form_builder/form_builder.rb
Kalagan: Yes that's better

app/forms/merits_assessments/client_received_legal_help_form.rb
Kalagan: The code is much cleaner but it doesn't work perfectly.
We want to do `validates :application_purpose, presence: true` only client_received_legal_help has been selected to "No".
So if the user didn't select anything, we don't want to trigger that validation.

i.e.:
We don't want to show the second error message
image

app/controllers/providers/client_received_legal_helps_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Providers
+  class ClientReceivedLegalHelpsController < BaseController
+    include ApplicationDependable
+    include Steppable
+    include SaveAsDraftable
+
+    def show
+      @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(model: merits_assessment)
```
tmckernan: why have you not restrict this application, to be view by this provider only?

Kalagan: I forgot. It's done now :)

tmckernan: 👍

app/forms/merits_assessments/client_received_legal_help_form.rb
reggieb: :+1: Fair enough.

PR #235, opened 2019-01-24
Remove 'Signed in successfully' message

mpw5:
## What

[AP-317](https://dsdmoj.atlassian.net/browse/AP-317)

When a user signs in via the portal, devise sets a flash message which is displayed to the user. They don't need to see this.

Remove it by removing the message wording from the config/locales/devise.en.yml file.

This will allow other flash messages set by devise to still work as expected (if we need them - if we decide later that we don't then they can be removed by a separate change).

## Checklist


reggieb:
Fix is now on staging.

Before:
before_flash_message_mod

After
after_flash_message_mod

PR #238, opened 2019-01-25
AP-261 - Merits Assessment: Proceedings Before The court

mpw5:
## What

[AP-261](https://dsdmoj.atlassian.net/browse/AP-261)

Add the Merits Assessment page "Are the proceedings, for which funding is being sought, currently, before the court?"

This is based on Julien's [PR](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/233) for AP-260, to make use of his new text_area FormBuilder component and to allow the two new pages to be integrated correctly.

### DIFF
--- a/app/controllers/providers/client_received_legal_helps_controller.rb
+++ b/app/controllers/providers/client_received_legal_helps_controller.rb
@@ -13,11 +13,6 @@ module Providers
       @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(client_received_legal_help_params.merge(model: merits_assessment))
 
       if @form.save
-        # TODO: remove this condition once next step is implemented
-        if params.key?(:continue_button)
-          render plain: 'Placeholder: Are the proceedings, for which funding is being sought, currently, before the court?'
-          return
-        end
         continue_or_save_draft
       else
         render :show
--- a/app/controllers/providers/proceedings_before_the_courts_controller.rb
+++ b/app/controllers/providers/proceedings_before_the_courts_controller.rb
@@ -1,0 +1,42 @@
+module Providers
+  class ProceedingsBeforeTheCourtsController < BaseController
+    include ApplicationDependable
+    include Steppable
+    include SaveAsDraftable
+    before_action :authorize_legal_aid_application
+
+    def show
+      @form = MeritsAssessments::ProceedingsBeforeTheCourtForm.new(model: merits_assessment)
+    end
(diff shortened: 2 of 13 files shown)


app/controllers/providers/client_received_legal_helps_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    before_action :authorize_legal_aid_application
+
+    def show
+      @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(model: merits_assessment)
+    end
+
+    def update
+      @form = MeritsAssessments::ClientReceivedLegalHelpForm.new(client_received_legal_help_params.merge(model: merits_assessment))
+
+      if @form.save
```
reggieb: A suggestion. If you add this private method:
```ruby
def form
  @form ||= MeritsAssessments::ClientReceivedLegalHelpForm.new(model: merits_assessment)
end
```
I think the code of both the `show` and `update` methods can be simplified a lot:

```ruby
def show
  form
end

def update
  if form.update(client_received_legal_help_params)
    continue_or_save_draft
  else
    render :show
  end
end
```

app/controllers/providers/proceedings_before_the_courts_controller.rb
```diff
@@ -0,0 +1,42 @@
+module Providers
+  class ProceedingsBeforeTheCourtsController < BaseController
+    include ApplicationDependable
+    include Steppable
+    include SaveAsDraftable
+    before_action :authorize_legal_aid_application
+
+    def show
+      @form = MeritsAssessments::ProceedingsBeforeTheCourtForm.new(model: merits_assessment)
```
reggieb: Again, consider using:
```ruby
def form
  @form ||= MeritsAssessments::ProceedingsBeforeTheCourtForm.new(model: merits_assessment)
end
```

app/controllers/providers/client_received_legal_helps_controller.rb
Kalagan: But the code of `show` and `update` is not exactly the same.
the form in `update` also needs the `params`

```ruby
# show
MeritsAssessments::ClientReceivedLegalHelpForm.new(model: merits_assessment)

# update
MeritsAssessments::ClientReceivedLegalHelpForm.new(client_received_legal_help_params.merge(model: merits_assessment))
```

PR #239, opened 2019-01-25
Ap 265 client declaration

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-265)

Adds the client merits declaration page to the application. 
Uses a boolean value for `:clients_merits_declaration` however this is always going to be `true` as there is no option for a user to not agree

## Checklist


app/controllers/providers/merits_declarations_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    include Steppable
+
+    before_action :authorize_legal_aid_application
+
+    def show
+      @form = MeritsAssessments::MeritsDeclarationForm.new(model: merits_assessment)
+    end
+
+    def update
+      @form = MeritsAssessments::MeritsDeclarationForm.new(merits_declaration_params.merge(model: merits_assessment))
```
Kalagan: I wonder if we even need a form and some params here.
Because there is no validation needed, no condition. If the user clicks on the button in the page, we always need to set `client_merits_declaration` to `true`
I reckon this would be enough:
```ruby
def update
  merits_assessment.update!(client_merits_declaration: true)
  # TO DO add correct redirect here when it is implemented
  # redirect_to next_step_url
  render plain: 'Placeholder: Merits check your answers page will go here'
end
```

And in the page you can simplify the submit button:
```ruby
<%= form.button t('generic.continue'), class: 'govuk-button' %>
```

Other devs, what do you think?

db/schema.rb
```diff
@@ -152,10 +152,10 @@
     t.boolean "open_banking_consent"
     t.datetime "open_banking_consent_choice_at"
     t.string "own_home"
-    t.decimal "property_value", precision: 10, scale: 2
     t.decimal "percentage_home"
-    t.string "shared_ownership"
+    t.decimal "property_value", precision: 10, scale: 2
     t.decimal "outstanding_mortgage_amount"
+    t.string "shared_ownership"
```
Kalagan: No big deal but ideally your PR shouldn't change the order of columns.
I think whenever you run a migration, `schema.rb` is modified to match the order of columns in your DB.
The order of columns in your local DB is different probably because you were testing some not merged PRs.
You can reset your DB to match `schema.rb` with the command `rails db:schema:load`

app/controllers/providers/merits_declarations_controller.rb
reggieb: I agree. If there is no validation, there is no need for a Form object.

In fact I think you can simplify this even further by removing the form on the page and replacing it with:
```erb
<%= button_to t('generic.continue'), path_to_action, method: :patch, class: 'govuk-button' %>
```

PR #244, opened 2019-01-28
AP 304 feedback page

reggieb:
[Jira AP-304](https://dsdmoj.atlassian.net/browse/AP-304)

- Adds a `Feedback` object to store user feedback submission and information about their Browser, OS, and whether they came from the Providers or Citizens journey. 
- Adds a Feedbacks controller with `new` `create` and `show` actions
- Stores url that user left to access feedback so they can get back there easily
- Sends an email to `apply-for-legal-aid@digital.justice.gov.uk` with feedback data
- Updated the form builder so that title text sizes can be modified.
- Added a method to make it easier to build a set of radio buttons for an object's enum property
- Updated header and footer "feedback" links to point at `feedbacks#new`

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,11 +1,29 @@
 class FeedbackController < ApplicationController
   def new
-    @feedback = Feedback.new
+    feedback
   end
 
   def create
-    @feedback = Feedback.new
-    @feedback.errors.add(:improvment_suggestion, 'this is an error')
-    render :new
+    if feedback.update(feedback_params)
+      redirect_to feedback
+    else
+      render :new
+    end
+  end
+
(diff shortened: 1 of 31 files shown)


app/controllers/feedback_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def feedback
+    @feedback ||= Feedback.new(
+      source: source,
+      os: browser.platform.name,
+      browser: browser.name,
+      browser_version: browser.full_version
+    )
+  end
```
stephenrichards: I don't think we are capturing the actual page they were on when they clicked the feedback link (although I see from the ticket that isn't down as a requirement, but it is something we discussed).

PR #246, opened 2019-01-29
AP-279 Move legal proceedings page

Kalagan:
## What

[AP-279](https://dsdmoj.atlassian.net/browse/AP-279)

Move legal proceedings page to between the address selection on the check answers page

## Checklist

### DIFF
--- a/app/controllers/providers/proceedings_types_controller.rb
+++ b/app/controllers/providers/proceedings_types_controller.rb
@@ -6,6 +6,7 @@ module Providers
     # GET /provider/applications/:legal_aid_application_id/proceedings_type
     def show
       authorize @legal_aid_application
+      @back_step_url = back_step_path unless legal_aid_application.checking_answers?
       proceeding_types
     end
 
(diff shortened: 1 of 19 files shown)


app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -33,5 +34,11 @@ def permitted_params
     def proceeding_types
       @proceeding_types ||= ProceedingType.all
     end
+
+    def back_step_path
+      return providers_legal_aid_application_address_selection_path if applicant.address&.lookup_used?
+
+      providers_legal_aid_application_address_path
+    end
```
reggieb: You shouldn't have to create a new method. Just over-ride the `back_step_url` method.
```ruby
def show
  authorize @legal_aid_application
  proceeding_types
end
   .....
def back_step_url
  super unless legal_aid_application.checking_answers?
  return providers_legal_aid_application_address_selection_path if applicant.address&.lookup_used?

  providers_legal_aid_application_address_path
end
```
As the method is already defined as a helper method in `Steppable`, `back_step_path` will be available directly in the views.

config/locales/en.yml
```diff
@@ -494,10 +494,10 @@ en:
     check_provider_answers:
       index:
         title: Check your answers
-        section_1:
+        section_proceeding:
           heading: Scope of legal aid
           proceeding: Proceeding
-        section_2:
+        section_client:
```
reggieb: Much better

PR #247, opened 2019-01-29
AP-264 Add prospects of success

tmckernan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-264)

-new page: Merits Assessment - What are the prospects of success?

## Checklist


spec/requests/providers/success_prospects_spec.rb
```diff
@@ -0,0 +1,115 @@
+require 'rails_helper'
+
+RSpec.describe Providers::EstimatedLegalCostsController, type: :request do
+  let(:legal_aid_application) { create :legal_aid_application, :with_applicant }
+  let(:secure_id) { legal_aid_application.generate_secure_id }
```
willc-work: this line doesn't seem to be required

app/assets/javascripts/success_prospects.es6
```diff
@@ -0,0 +1,17 @@
+$(function() {
+  $('#success-prospect-text-area').hide();
+  var option_name = $("input[type=radio][name='merits_assessment[success_prospect]']:checked").val()
```
reggieb: As this files uses ES6, I think `var` should be replace with `let`

app/assets/javascripts/success_prospects.es6
```diff
@@ -0,0 +1,17 @@
...
+    displayTextArea(option_name)
+  }
+
+  $('.success_prospect_option').click(function() {
+    displayTextArea($(this).val());
+  });
+});
+
+function displayTextArea(option_name) {
+  var id = "#conditional-"+option_name;
```
reggieb: `let`

app/assets/javascripts/success_prospects.es6
```diff
@@ -0,0 +1,17 @@
...
+  if (typeof(option_name) != "undefined" && option_name != 'better') {
+    displayTextArea(option_name)
+  }
+
+  $('.success_prospect_option').click(function() {
+    displayTextArea($(this).val());
+  });
+});
+
+function displayTextArea(option_name) {
```
reggieb: The behaviour of this function is specific to `success-prospect` where as the name implies it can be used on any text area. I'd suggest the function name reflect the tight scope of the function: 
```javascript
function showSuccessProspectTextArea(option_name) {
```

app/assets/javascripts/success_prospects.es6
tmckernan: changed

app/assets/javascripts/success_prospects.es6
tmckernan: changed

app/assets/javascripts/success_prospects.es6
tmckernan: changed

app/models/merits_assessment.rb
```diff
@@ -1,3 +1,22 @@
...
+    _prefix: true
+  )
+
+  def list_success_prospects
+    MeritsAssessment.success_prospects.drop(1).map { |k| k }
+  end
+
+  def first_success_prospect
+    MeritsAssessment.success_prospects.keys.first
+  end
```
reggieb: I have a little problem with the naming of methods in this Model. They imply that what is significant is the position of elements in the enum, and do not convey the intention of the class. I would suggest this:
```ruby
class MeritsAssessment < ApplicationRecord
  belongs_to :legal_aid_application

  enum(
    success_prospect: {
      likely: '50% or better'.freeze,
      marginal: 'Marginal 45 - 50%'.freeze,
      borderline: 'borderline'.freeze,
      uncertain: 'uncertain'.freeze,
      poor: 'poor'.freeze
    },
    _prefix: true
  )

  def prospect_likely_to_succeed
    :likely
  end

  def prospects_unlikely_to_succeed
    @prospects_unlikely_to_succeed ||= self.class.success_prospects.except(prospect_likely_to_succed.to_s)
  end
end
```
Obviously that will require a rewrite to replace the calls to `list_success_prospects` and `first_success_prospect`, and the use of `better` elsewhere in the code.

app/forms/merits_assessments/success_prospect_form.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    form_for MeritsAssessment
+
+    attr_accessor :success_prospect, :success_prospect_details
+
+    before_validation :clear_success_prospect_details
+    validates :success_prospect, presence: true
+    validates :success_prospect_details, presence: true, unless: proc { |form| form.success_prospect.to_s == model.first_success_prospect.to_s }
+
+    delegate :list_success_prospects, :first_success_prospect, to: :model
```
reggieb: To delegate to the model class, you should be able to do this:
```ruby
delegate :prospect_likely_to_succeed, :prospects_unlikely_to_succeed : to: :model_class

File: app/models/merits_assessment.rb (hunk 5 above)
Comment (reggieb) [reply app/models/merits_assessment.rb]: Sorry - Looking at this again. The methods are specific to the class level object - not the instance. Therefore `prospects_unlikely_to_succeed` and `prospect_likely_to_succeed` should be class methods. Otherwise `prospects_unlikely_to_succeed` will be unnecessarily redetermined each time an instance is created, when it only needs to be done once. So:
```ruby
class MeritsAssessment < ApplicationRecord
  belongs_to :legal_aid_application

  # defining before enum so it can be used in the enum
  def self.prospect_likely_to_succeed
    :likely
  end

  enum(
    success_prospect: {
      prospect_likely_to_succeed => '50% or better'.freeze,
      marginal: 'Marginal 45 - 50%'.freeze,
      borderline: 'borderline'.freeze,
      uncertain: 'uncertain'.freeze,
      poor: 'poor'.freeze
    },
    _prefix: true
  )

  def self.prospects_unlikely_to_succeed
    @prospects_unlikely_to_succeed ||= success_prospects.except(prospect_likely_to_succed.to_s)
  end
end
```
It also allows `prospect_likely_to_succeed` to be used in the enum construction - DRYing things up a little more.

PR #250, opened 2019-01-30
Ap 295 display selected legal proceeding

reggieb:
[AP-295](https://dsdmoj.atlassian.net/browse/AP-295)

This PR changes the behaviour of `proceeding_types` page. Now:

- When a Provider selects a Proceeding Type from the search results it appears in the list of selected Proceeding Types.
- They can then either: select another Proceeding Type, remove a selected one, or click on Continue to move to the next page
- If they click Continue without having first selected a Proceeding Type, they see an error and are prompted to select one.

### DIFF
--- a/app/assets/stylesheets/providers/proceeding_types.scss
+++ b/app/assets/stylesheets/providers/proceeding_types.scss
@@ -1,6 +1,9 @@
+@import "govuk-frontend/settings/all";
+@import "govuk-frontend/helpers/all";
+
 .proceeding-item {
   display: none;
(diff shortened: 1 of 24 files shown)


reggieb:
Hold off the reviews - rebasing is more complicated than I hoped (@Kalagan got his PR in first :smiley:)

reggieb:
Should be OK to review now. :crossed_fingers:

reggieb:
I must read the specs. more carefully.

PR #252, opened 2019-01-31
Refactor exisiting lists to use the same partial

willc-work:
The existing detailed_list partial was not suitable for all list types, this creates a new list_item partial that is more flexible allowing new list items to be easily added or removed without needing to amend the code in the view.

The list_item partial checks the locales (en.yml) file for the relevant items. Instead of using:

```
item_1: First list item
item_2: Second list item
item_3: Another list item for good measure
```
You can now use:

```
list: |
  First list item
  Second list item
  Another list item for good measure
```
Items in the list can be added or removed without needing to change anything in the view. 

The detailed_list partial has also been amended to use this partial.

## Checklist

### DIFF
--- a/app/views/citizens/information/show.html.erb
+++ b/app/views/citizens/information/show.html.erb
@@ -4,7 +4,7 @@
   <h2 class="govuk-heading-m"><%= t('.list.title') %></h2>
 
   <p class="govuk-body">
-    <%= render partial: 'shared/forms/list_items', locals: { translation_path: 'citizens.information.show.list' } %>
+    <%= list_from_translation_path 'citizens.information.show.list' %>
   </p>
 
   <%= link_to t('generic.continue'), citizens_consent_path, class: 'govuk-button govuk-button', id: 'continue' %>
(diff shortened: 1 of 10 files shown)


app/views/citizens/information/show.html.erb
```diff
@@ -4,11 +4,7 @@
   <h2 class="govuk-heading-m"><%= t('.list.title') %></h2>
 
   <p class="govuk-body">
-    <ul class="govuk-list govuk-list--bullet">
-      <li> <%= t('.list.item_1') %> </li>
-      <li> <%= t('.list.item_2') %> </li>
-      <li> <%= t('.list.item_3') %> </li>
-    </ul>
+    <%= render partial: 'shared/forms/list_items', locals: { translation_path: 'citizens.information.show.list' } %>
```
reggieb: Could you please wrap this up in a helper method (in `ApplicantionHelper` for example). Something like:
```ruby
def list_from_translation_path(translation_path)
  render 'shared/forms/list_items', translation_path: translation_path
end
```
Then the code in the views will be a lot easier to read. For example here it would be:
```erb
<%= list_from_translation_path 'citizens.information.show.list' %>
```

PR #254, opened 2019-01-31
AP-328 Refactor translation files

stephenrichards:
## Refactor Translation files into smaller self-contained units

[Link to story](https://dsdmoj.atlassian.net/browse/AP-328)

- Split the large `config/locales/en.yml` file into several smaller files and moved to the `config/en` directory
- Moved other `en` files into that directory
- Alphabetised the keys (this is easily done with `i18n-tasks normalize` for future reference)
- Created `config/i18n-tasks.yml` file to tell `i18n-tasks` where to look for translations and rules to follow when normalizing translation files
- Created `config/initializers/i18n.rb` to append the new files to the `I18n.load_path`

## Notes to developers

Use the `i18n-tasks normalize -p` command!  This will:
- move the keys into the correct file according to the rules in `config/i18n-tasks.yml:26`
- re-order the keys into alphabetical order

## Checklist


config/i18n-tasks.yml
```diff
@@ -0,0 +1,145 @@
...
+            devise,
+            forms,
+            generic,
+            helpers,
+            layouts,
+            model_enum_translations,
+            providers,
+            shared}.*', 'config/locales/%{locale}/\1.yml']
+    ## Catch-all default:
+     - config/locales/%{locale}/default.yml
```
reggieb: Can this be done dynamically? For example this gets all the file names in the current locales:
```ruby
Dir[Rails.root.join('config/locales/*.yml')].map{|p| File.basename(p)}
```
There is an option to use an erb version of `config/i18n-tasks.yml` so it looks like you can mix ruby into the YAML: `config/i18n-tasks.yml.erb`

PR #256, opened 2019-02-01
AP-287 Applicant capital: Keep fields in sync

reggieb:
[Jira AP-287](https://dsdmoj.atlassian.net/browse/AP-287)

Added a call back to `LegalAidApplication` that looks for changes to `own_home` and `shared_ownership` and resets fields that are invalidated by the change.

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -1,5 +1,5 @@
 # TODO: Think about how we refactor this class to make it smaller
-class LegalAidApplication < ApplicationRecord
+class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLength
   include TranslatableModelAttribute
   include LegalAidApplicationStateMachine # States are defined here
 
@@ -21,6 +21,7 @@ class LegalAidApplication < ApplicationRecord
 
   before_create :create_app_ref
   before_save :set_open_banking_consent_choice_at
+  before_save :keep_attributes_in_sync
 
   attr_reader :proceeding_type_codes
   validate :proceeding_type_codes_existence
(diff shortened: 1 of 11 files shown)


app/models/legal_aid_application.rb
```diff
@@ -116,4 +117,31 @@ def create_app_ref
...
+    self.property_value = nil
+    self.outstanding_mortgage_amount = nil
+    self.shared_ownership = nil
+    self.percentage_home = nil
+  end
+
+  def reset_to_match_own_home_mortgage
+    return unless own_home_changed? && own_home_mortgage?
+
+    self.outstanding_mortgage_amount = nil
```
Kalagan: should the condition be `!own_home_mortgage?` instead of `own_home_mortgage` ?

spec/factories/legal_aid_applications.rb
```diff
@@ -71,13 +71,15 @@
     end
 
     trait :with_everything do
+      after(:create) do |legal_aid_appliction|
+        legal_aid_appliction.update(outstanding_mortgage_amount: Faker::Number.decimal.to_d)
```
Kalagan: typo `legal_aid_appliction`

spec/factories/legal_aid_applications.rb
```diff
@@ -71,13 +71,15 @@
...
+      end
```
Kalagan: Why do we need to use this `after(:create)` instead of the line you removed?

spec/factories/legal_aid_applications.rb
reggieb: Code removed in refactor

spec/factories/legal_aid_applications.rb
reggieb: Because one callback was resetting it under some circumstances. To be honest, it's a smell that the refactor removed.

app/models/legal_aid_application.rb
reggieb: I think refactor has fixed the logic.

app/services/keep_in_sync.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def initialize(legal_aid_application)
+    @legal_aid_application = legal_aid_application
+  end
+
+  def sync
+    when_own_home_no
+    when_own_home_owned_outright
+    when_shared_ownership_no
+    legal_aid_application.save
```
Kalagan: I suggest `.save!` instead of `.save`

PR #259, opened 2019-02-04
AP 325 Fix validation

AbdirizakObsiye2018:
## AP-325 Fix and standardise currency validations

[Link to story](https://dsdmoj.atlassian.net/browse/AP-325)

Previously, currency validations were inconsistent and allowed too many decimals.
Have created a new currency validator, which is used on all forms.
In addition, CurrencyCleaner will remove currency signs and commas if the string is a valid value.
If there are errors, the original String that was input by the user is redisplayed back on the form.

## Checklist

### DIFF
--- a/app/forms/base_form.rb
+++ b/app/forms/base_form.rb
@@ -67,7 +67,7 @@ module BaseForm
     def save
       return false unless valid?
 
-      model.attributes = assignable_attributes
+      model.attributes = clean_attributes(assignable_attributes)
       model.save(validate: false)
     end
 
@@ -96,6 +96,10 @@ module BaseForm
     end
 
     private
+
+    def clean_attributes(hash)
+      hash.transform_values { |v| v.is_a?(String) ? v.tr('£,', '') : v }
+    end
 
     def set_blanks_to_nil
       self.class.locally_assigned.each do |attr|
--- a/app/forms/base_form.rb
+++ b/app/forms/base_form.rb
@@ -76,6 +76,10 @@ module BaseForm
       []
     end
 
+    def attributes_to_clean
+      []
+    end
+
     def assignable_attributes
       exclude_attrs = exclude_from_model + [:model]
       attributes.except(*exclude_attrs.map(&:to_s))
@@ -98,7 +102,9 @@ module BaseForm
     private
 
     def clean_attributes(hash)
-      hash.transform_values { |v| v.is_a?(String) ? v.tr('£,', '') : v }
+      hash.each_with_object({}) do |(k, v), new_hash|
+        new_hash[k] = k.to_sym.in?(attributes_to_clean) ? v.to_s.tr('£,', '') : v
+      end
     end
 
     def set_blanks_to_nil
--- a/app/lib/govuk_elements_form_builder/form_builder.rb
+++ b/app/lib/govuk_elements_form_builder/form_builder.rb
@@ -233,7 +233,7 @@ module GovukElementsFormBuilder
     def error_tag(attribute, options)
       return unless error?(attribute, options)
 
-      message = object.errors[attribute].join(', ')
+      message = object.errors[attribute].first
       return unless message.present?
 
       content_tag(:span, message, class: 'govuk-error-message', id: "#{attribute}-error")
--- a/app/services/currency_cleaner.rb
+++ b/app/services/currency_cleaner.rb
@@ -1,5 +1,5 @@
 class CurrencyCleaner
-  COMMA_STRIPPING_REGEX = /,(?=\d{3}\b)/.freeze
+  THOUSANDS_SEPARATOR_REGEX = /,(?=\d{3}\b)/.freeze
   LEADING_POUND_SIGN_REGEX = /^£/.freeze
   PURE_NUMERIC_REGEX = /^-?\d+\.?\d+?$/.freeze
 
@@ -17,6 +17,6 @@ class CurrencyCleaner
   private
 
   def remove_commas
-    @original_value.sub(LEADING_POUND_SIGN_REGEX, '').gsub(COMMA_STRIPPING_REGEX, '')
+    @original_value.sub(LEADING_POUND_SIGN_REGEX, '').gsub(THOUSANDS_SEPARATOR_REGEX, '')
   end
 end
--- a/app/services/currency_cleaner.rb
+++ b/app/services/currency_cleaner.rb
@@ -1,0 +1,22 @@
+class CurrencyCleaner
+  COMMA_STRIPPING_REGEX = /,(?=\d{3}\b)/.freeze
+  LEADING_POUND_SIGN_REGEX = /^£/.freeze
+  PURE_NUMERIC_REGEX = /^-?\d+\.?\d+?$/.freeze
+
+  def initialize(value)
+    @original_value = value
+  end
+
+  # will remove all commas if valid i.e. followed by 3 digits
+  # if invalid, the original String is returned
+  def clean
+    clean_value = remove_commas
+    PURE_NUMERIC_REGEX.match?(clean_value) ? clean_value : @original_value
+  end
+
+  private
+
+  def remove_commas
+    @original_value.sub(LEADING_POUND_SIGN_REGEX, '').gsub(COMMA_STRIPPING_REGEX, '')
+  end
+end
--- a/app/services/currency_cleaner.rb
+++ b/app/services/currency_cleaner.rb
@@ -9,7 +9,7 @@ class CurrencyCleaner
 
   # will remove all commas if valid i.e. followed by 3 digits
   # if invalid, the original String is returned
-  def clean
+  def call
     clean_value = remove_commas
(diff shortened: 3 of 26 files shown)


reggieb:
Firstly, clean up is done before validation and as the clean up removes `£`, the validation patterns should not need to include that character.

I'd strongly recommend that you have a look at [Custom Validator in the Rails Guides](https://guides.rubyonrails.org/active_record_validations.html#custom-validators). This provides a standardised way of enhancing validation. Also look at the [`numericality` validation option](https://guides.rubyonrails.org/active_record_validations.html#numericality) that will ensure that a number is numerical.

So using numericality validation makes the `valid_number_pattern` redundant. So what is needed is one of two options:

1. Make the 2 digit pattern available to validation (which your update does) so it can be used like this:
```ruby
validates :a_currency_field, numericality: true, format: { with: only_2_decimals }
```
2. Add a custom validator
It might be possible to make a new validator that inherits from Rails' [`NumericalityValidator`](https://github.com/rails/rails/blob/6a1b7985602c5bfab4c8875ca9bf0d598e063a65/activemodel/lib/active_model/validations/numericality.rb) (thereby getting all the numericality functionality for free). I think something like this will work:
```ruby
module ActiveModel
  module Validations
    class IsCurrencyValidator < NumericalityValidator
      def validate_each(record, attr_name, value)
        super
        unless value == value.truncate(2)
         record.errors.add(attr_name, :too_many_decimals)
        end
    end
  end
end
```

Then remove the number pattern check from the clean up:
```ruby
   def clean_up_currency(*attr_names)
      attr_names.each do |attr_name|
        field_value = __send__(attr_name)
        field_value.delete!(',£') if field_value.is_a?(String)
      end
    end
```

And the final form (and model) code becomes:
```ruby
before_validation { clean_up_currency :a_currency_field }
validates :a_currency_field, is_currency: true
```
Oh. And I should also add that I should have read my own advice before I wrote some of the similar code I've added to the repo :disappointed:

spec/forms/outstanding_mortgage_form_spec.rb
```diff
@@ -75,14 +75,14 @@
       end
 
       context 'with negative numbers' do
-        let(:amount) { Faker::Number.negative }
+        let(:amount) { "'#{Faker::Number.negative}'" }
```
reggieb:
```ruby
let(:amount) { Faker::Number.negative.to_s }
```

spec/validators/currency_validator_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      dummy.val2 = 'qwe'
+      expect(dummy).to be_invalid
+      expect(dummy.errors[:val1]).to eq ['is not a number']
+      expect(dummy.errors[:val2]).to eq ['is not a number']
+      expect(dummy.errors[:val3]).to be_empty
+    end
+
+    it 'errors if given negative number' do
+      dummy.val1 = '-1234'
+      dummy.val2 = '1234'
```
reggieb: I'd suggest using Faker to generate these numbers. It's possible that `1234` is a special case. It's better practice to randomly generate test numbers.
```ruby
dummy.val1 = Faker::Number.negative.to_s
dummy.val2 = Faker::Number.to_s

File: app/services/currency_cleaner.rb [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+  # if invalid, the original String is returned
+  def clean
+    clean_value = remove_commas
+    PURE_NUMERIC_REGEX.match?(clean_value) ? clean_value : @original_value
+  end
+
+  private
+
+  def remove_commas
+    @original_value.sub(LEADING_POUND_SIGN_REGEX, '').gsub(COMMA_STRIPPING_REGEX, '')
```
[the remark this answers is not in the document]
AbdirizakObsiye2018: The regex has changed for COMMA_STRIPPING_REGEX and thus no longer need '\1' . Please see the currency_cleaner_spec.rb file.

```
rb(main):001:0> COMMA_STRIPPING_REGEX = /,(?=\d{3}\b)/.freeze
=> /,(?=\d{3}\b)/
irb(main):002:0> val = "123,456,364"
=> "123,456,364"
irb(main):003:0> val.gsub(COMMA_STRIPPING_REGEX, '')
=> "123456364"
irb(main):004:0> val2 = "12,35544,66"
=> "12,35544,66"
irb(main):005:0> val.gsub(COMMA_STRIPPING_REGEX, '')
=> "123456364"```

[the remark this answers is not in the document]
Comment (reggieb) [reply app/services/currency_cleaner.rb]: Ah! Positive look ahead. Neat. :+1:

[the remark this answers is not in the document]
Comment (AbdirizakObsiye2018) [reply app/services/currency_cleaner.rb]: In the regex `?=` is a positive lookahead, a type of zero-width assertion. What it’s saying is that the captured match must be followed by whatever is within the parentheses but that part isn’t captured. Therefore, only the commas are matched if and only if they are followed by three digits. Therefore, only the commas will be replaced.

File: app/services/currency_cleaner.rb [hunk 4]
```diff
@@ -0,0 +1,22 @@
+class CurrencyCleaner
+  COMMA_STRIPPING_REGEX = /,(?=\d{3}\b)/.freeze
+  LEADING_POUND_SIGN_REGEX = /^£/.freeze
+  PURE_NUMERIC_REGEX = /^-?\d+\.?\d+?$/.freeze
```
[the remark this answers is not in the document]
AbdirizakObsiye2018: I'd prefer to leave it as an instance method.

app/forms/base_form.rb
```diff
@@ -87,6 +87,10 @@ def attributes
 
     private
 
+    def clean_attributes(hash)
+      hash.each { |_k, v| v.tr!('£,', '') if v.is_a?(String) }
```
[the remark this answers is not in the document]
stephenrichards: It gives different results:
`hash = {:a=>"£123,456.00", :b=>"", :c=>false}
hash.each { |_k, v| v.tr!('£,', '') if v.is_a?(String) } #=> {:a=>"123456.00", :b=>"", :c=>false}
hash.transform_values { |v| v.tr('£,', '') if v.is_a?(String) } #=>  {:a=>"123456.00", :b=>"", :c=>nil}
`

spec/validators/currency_validator_spec.rb
AbdirizakObsiye2018: no, we're testing specific Strings.

app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -233,7 +233,7 @@ def hint_tag(attribute, options)
     def error_tag(attribute, options)
       return unless error?(attribute, options)
 
-      message = object.errors[attribute].join(', ')
+      message = object.errors[attribute].first
```
Kalagan: Any reason why we just take the first error?

app/services/currency_cleaner.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(value)
+    @original_value = value
+  end
+
+  # will remove all commas if valid i.e. followed by 3 digits
+  # if invalid, the original String is returned
+  def clean
```
Kalagan: At the moment, almost all of our services with only one public method follow the pattern of calling that method `call`.
I think it would be good to keep following that pattern.
For example:
`CurrencyCleaner.call` instead of `CurrencyCleaner.clean`
`DataImporter.call` instead of `DataImporter.import`
`EmailSender.call` instead of `EmailSender.send`
etc...

app/views/shared/forms/_error_summary.html.erb
```diff
@@ -6,9 +6,9 @@
     </h2>
     <div class="govuk-error-summary__body">
       <ul class="govuk-list govuk-error-summary__list">
-        <% model.errors.each do |attribute, message| %>
+        <% model.errors.keys.each do |attribute| %>
           <li>
-            <%= link_to message, "##{attribute_prefix}#{attribute}" %>
+            <%= link_to model.errors[attribute].first, "##{attribute_prefix}#{attribute}" %>
```
Kalagan: Similar question than the one on `FormBuilder`:
Isn't more useful to show all error messages rather than only the first one of each attribute?

app/lib/govuk_elements_form_builder/form_builder.rb
reggieb: If you look at most of the designs, there is a consistent effort to avoid multiple errors per attribute. Mostly that's done via the form, but often it's done with custom validation (it why some forms have more complicated validation than would normally be used). 

This is a simpler solution.

Kalagan: Ok. Fair enough. I was thinking of the story with the file upload I'm working on at the moment.
For example you could potentially have these 2 errors for the same attribute:
- The selected file must be a DOC, DOCX, RTF, [anything else we accept..], JPG, BMP, PNG, TIF or PDF
- The selected file must be smaller than [xxx]MB

I thought it would be usefull to show both

PR #262, opened 2019-02-05
AP-347 Restrict applications to creator

tmckernan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-347)

Update `providers/applications` to only display the current provider applications.

## Checklist

### DIFF
--- a/app/controllers/providers/legal_aid_applications_controller.rb
+++ b/app/controllers/providers/legal_aid_applications_controller.rb
@@ -5,8 +5,8 @@ module Providers
 
     # GET /provider/applications
     def index
-      # TODO: update to return only current Provider's applications when provider can be identified
-      @applications = LegalAidApplication.limit(10)
+      # TODO: Add pagination at some point
+      @applications = current_provider.legal_aid_applications.latest.limit(10)
     end
 
     # POST /provider/applications
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -28,6 +28,8 @@ class LegalAidApplication < ApplicationRecord
 
   delegate :full_name, to: :applicant, prefix: true, allow_nil: true
 
+  scope :latest, -> { order(created_at: :desc) }
+
   enum(
     own_home: {
       no: 'no'.freeze,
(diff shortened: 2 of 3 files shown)


app/controllers/providers/legal_aid_applications_controller.rb
```diff
@@ -5,8 +5,7 @@ class LegalAidApplicationsController < BaseController
 
     # GET /provider/applications
     def index
-      # TODO: update to return only current Provider's applications when provider can be identified
-      @applications = LegalAidApplication.limit(10)
+      @applications = LegalAidApplication.where(provider: current_provider).limit(10)
```
reggieb: I'd suggest:
```ruby
@applications = current_provider.legal_aid_applications.limit(10)
```

Kalagan: I would also add a TODO to add pagination at some point

spec/requests/providers/legal_aid_applications_spec.rb
```diff
@@ -4,6 +4,8 @@
   describe 'GET /providers/applications' do
     let!(:legal_aid_application) { create :legal_aid_application }
     subject { get providers_legal_aid_applications_path }
+    let(:provider) { legal_aid_application.provider }
+    let(:other_provider) { create(:provider) }
```
Kalagan: No big deal but it would make sense to have `subject` below the 3 `let`s rather than in the middle of them

spec/requests/providers/legal_aid_applications_spec.rb
```diff
@@ -32,6 +34,17 @@
         end
       end
     end
+
+    context 'when login as other provider is authenticated' do
```
Kalagan: Maybe I'm wrong but this sentence seems a bit weird to me.
Maybe "when logged in as another provider" or "when another provider is authenticated"?

app/controllers/providers/legal_aid_applications_controller.rb
Kalagan: Also, which 10 applications are fetched? Ideally, it should be the 10 most recent.
`current_provider.legal_aid_applications.order(created_at: :desc).limit(10)`
But maybe it is already the case by default?

app/models/legal_aid_application.rb
```diff
@@ -28,6 +28,8 @@ class LegalAidApplication < ApplicationRecord
 
   delegate :full_name, to: :applicant, prefix: true, allow_nil: true
 
+  scope :latest, -> { order(created_at: :desc) }
```
reggieb: Nice. I've wondered about adding a `default_scope` to set the ordering based on created at, but that can be problematic, so this is probably the best compromise.

PR #264, opened 2019-02-06
AP 338 update save as draft behaviour

reggieb:
[Jira AP-338](https://dsdmoj.atlassian.net/browse/AP-338)

Improves the save as Draft process so that partially submitted forms (as long as the parts are valid) are saved.

- Add controller concern `Draftable` - that marks or unmarks `legal_aid_application` as `draft` and adds methods to handle "Save as Draft"
- Add `save_as_draft` to base form to allow modified validation of objects when in draft mode. Also adds a `draft?` flag that can be used to modify validation behaviour when in draft mode.
- Fixes some pages where clicking "Save as Draft" had no effect.
- Update the Flow Service so that it only controls the normal flow and doesn't have to know about "Save as Draft"
- Refactor next action buttons by adding wrapper helper methods

Note with this update there should be three possible outcomes from clicking "Save as Draft":

1. Form contains some valid entries -> Valid entries saved to object, application marked as 'draft' redirected to `providers/applications`
2. Form contains only blank entries -> Nothing saved to object, application marked as 'draft' redirected to `providers/applications`
3. Form contains some invalid entries -> Nothing saved to object, form displayed with errors.

Also after selecting "Save as Draft", the "View" link in `providers/applications` returns the user to the form with the entries made so far, present (nothing needed to be added to this PR to achieve this).


app/forms/merits_assessments/client_received_legal_help_form.rb
```diff
@@ -6,14 +6,19 @@ class ClientReceivedLegalHelpForm
...
+      if: proc { |form| !form.draft? && form.client_received_legal_help.to_s == 'false' }
+    )
 
     private
 
     def clear_application_purpose
       application_purpose.clear if client_received_legal_help.to_s == 'true'
     end
+
+    before_validation :clear_application_purpose
```
Kalagan: any reason why you moved this all the way down there? I usually see these closer from the top

reggieb: I have no idea. I'll move it back to the top.

PR #269, opened 2019-02-07
FIX : Wrong Sidekiq instance picking up job

Kalagan:
I noticed that the pulling of data from TrueLayer was not working in UAT because all UAT envs share the same redis instance and therefore, any UAT sidekiq instance could pick up a job started from any other UAT envs.
I'm adding the host (i.e.: `fix-sidekiq-applyforlegalaid-uat.apps.cloud-platform-live-0.k8s.integration.dsd.io`) as a namespace

## Checklist

### DIFF
--- a/config/initializers/sidekiq.rb
+++ b/config/initializers/sidekiq.rb
@@ -2,16 +2,17 @@ require 'sidekiq'
 require 'sidekiq-status'
 
 redis_url = "rediss://:#{ENV['REDIS_PASSWORD']}@#{ENV['REDIS_HOST']}:6379" if ENV['REDIS_HOST'].present? && ENV['REDIS_PASSWORD'].present?
+namespace = ENV.fetch('HOST', 'laa-apply')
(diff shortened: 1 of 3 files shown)


config/initializers/sidekiq.rb
```diff
@@ -2,16 +2,17 @@
 require 'sidekiq-status'
 
 redis_url = "rediss://:#{ENV['REDIS_PASSWORD']}@#{ENV['REDIS_HOST']}:6379" if ENV['REDIS_HOST'].present? && ENV['REDIS_PASSWORD'].present?
+namespace = ENV.fetch('HOST', nil)
```
reggieb: Should the namespace default to something if not defined by `ENV['HOST']`? `'apply'` perhaps?

Kalagan: Yes, you're right. So there is always a namespace and it is more predictable.

PR #271, opened 2019-02-08
AP-339 Implement latest GOV-UK check answers styling and fix long emails

reggieb:
[Jira AP-339](https://dsdmoj.atlassian.net/browse/AP-339)

Unfortunately this was a little like attacking a Hydra. Each action seemed to need multiple actions.

So in the end:
- Implemented [latest GOV-UK Check your answers styling](https://design-system.service.gov.uk/patterns/check-answers/). As there are many check your answers pages - they all needed to be updated.
- To get the new styling, `govuk-frontend` needed to be upgraded.
- Refactored `check_answer_link` to work with new styling - in the end split it into three. One to handle single entries, a second for lists of items, and another for groups of currency entries.
- This almost fixed the email issue, but with very long emails there was still an issue. Adding a `min-width` to `.govuk-summary-list__key` fixed that.

### DIFF
--- a/app/helpers/check_answers_helper.rb
+++ b/app/helpers/check_answers_helper.rb
@@ -1,4 +1,8 @@
 module CheckAnswersHelper
+  # Creates one dictionary list item - so needs to be used within a `govuk-summary-list` `dl`:
+  #     <dl class="govuk-summary-list govuk-!-margin-bottom-9">
+  #       <%= check_answer_link ..... %>
+  #     </dl>
   def check_answer_link(url:, question:, answer:, name:)
     render(
       'shared/check_answers/item',
@@ -9,6 +13,7 @@ module CheckAnswersHelper
     )
   end
 
+  # Creates both the outer `dl` and the inner list items
   def check_answer_list(url:, question:, answers:, name:)
     render(
       'shared/check_answers/items',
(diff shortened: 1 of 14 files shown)


app/helpers/check_answers_helper.rb
```diff
@@ -1,17 +1,31 @@
...
     render(
-      partial: "shared/check_answers/#{partial}",
-      locals: { url: url, question: question, answer: answer, name: name, is_long: is_long }
+      'shared/check_answers/item',
+      name: name,
+      url: url,
+      question: question,
+      answer: answer
+    )
+  end
```
reggieb: The main reason I split the group methods below, out of `check_answer_link` was because they need to be used differently.  `check_answer_link` provides one items within a list - and therefore needs enclosing `dl` tags. So:
```erb
<dl class="govuk-summary-list govuk-!-margin-bottom-9">
  <%= check_answer_link ..... %>
</dl>
```

The two list methods below create both the items and the `dl` that contains them, so don't need to be enclosed within a `dl` tag

app/views/citizens/own_homes/show.html.erb
```diff
@@ -1,3 +1,3 @@
-<%= citizen_page page_title: t('.h1-heading') do %>
+<%= citizen_page page_title: t('.h1-heading'), template: :basic do %>
```
reggieb: Fix for a problems spotted while navigating the pages worked on.

PR #277, opened 2019-02-11
Ap 319 statement of case upload

Kalagan:
[AP-319](https://dsdmoj.atlassian.net/browse/AP-319)

- Add `file_field` to `form_builder`
- refactor `form_builder` to use the Rspec `shared_examples_for` pattern to avoid duplication
- Add file upload to the statement of case page.

I investigated using https://github.com/ministryofjustice/mojfile-uploader .
However, after talking with its main developer (Jesus Laiz), we agreed that it is overkill and that the simplest solution for us is to use Rails default upload (ActiveStorage, that I use in this PR) or gems like Shrine or CarrierWave and then do file conversion and malware check in Sidekiq workers.

With ActiveStorage we could implement direct uploads (file goes directly from user's computer to S3) but cloud platform's S3 cannot be setup for this yet. I raised an issue with them: https://github.com/ministryofjustice/cloud-platform/issues/636

## Checklist


features/support/steps_helper.rb
```diff
@@ -21,3 +21,7 @@
 Then('I fill {string} with {string}') do |field, value|
   fill_in(field.parameterize(separator: '_'), with: value)
 end
+
+Then('I upload a pdf file') do
+  attach_file('Upload a file', Rails.root + 'spec/fixtures/files/lorem_ipsum.pdf')
```
reggieb: I'd suggest:
```ruby
attach_file('Upload a file', Rails.root.join('spec/fixtures/files/lorem_ipsum.pdf'))
```

spec/requests/providers/statement_of_case_spec.rb
```diff
@@ -50,20 +50,20 @@
 
   describe 'PATCH providers/proceedings_before_the_court' do
     subject { patch providers_legal_aid_application_statement_of_case_path(legal_aid_application), params: params.merge(submit_button) }
-    let(:entered_text) { Faker::Lorem.paragraph(3) }
-
+    let(:statement) { Faker::Lorem.paragraph(3) }
+    let(:original_file) { Rack::Test::UploadedFile.new(Rails.root + 'spec/fixtures/files/lorem_ipsum.pdf', 'application/pdf') }
```
reggieb: Could we wrap this in a method defined in `rails_helper.rb` ? Something like:
```ruby
def rack_test_file(path, content_type = "text/plain", binary = false)
  path = Rails.root.join(path)
  Rack::Test::UploadedFile.new path, content_type, binary
end
```
Then here the code would be:
```ruby
let(:original_file) {  rack_test_file 'spec/fixtures/files/lorem_ipsum.pdf', 'application/pdf' }
```

spec/lib/govuk_elements_form_builder/form_builder_spec.rb
```diff
@@ -9,55 +9,37 @@ class TestHelper < ActionView::Base; end
...
+      div = tag.parent
       expect(div.name).to eq('div')
       expect(div.classes).to include('govuk-form-group')
     end
 
     it 'includes a label' do
-      label = input.parent.at_css('label')
-
       expect(label.classes).to include('govuk-label')
       expect(label[:for]).to eq('email')
```
reggieb: This shared example would be a lot more flexible if every instance of `'email'` was replaced with `attribute`. Note that `attribute` is already used in the hint sections of this shared example.

spec/lib/govuk_elements_form_builder/form_builder_spec.rb
```diff
@@ -91,136 +79,106 @@ class TestHelper < ActionView::Base; end
...
+        expect(tag.classes).to include(custom_class)
       end
     end
 
-    context 'has an input_prefix option' do
-      let(:prefix) { '£' }
-      let(:params) { [:email, input_prefix: prefix] }
+    context 'pass a label parameter with text and size' do
+      let(:custom_label) { "Your client's email" }
+      let(:params) { [:email, label: { text: custom_label, size: :m }] }
```
reggieb: Again - this ties the shared example to only test an "email" attributes. So I'd suggest:
```ruby
let(:params) { [attribute.to_sym, label: { text: custom_label, size: :m }] }
```

spec/lib/govuk_elements_form_builder/form_builder_spec.rb
```diff
@@ -70,19 +52,25 @@ class TestHelper < ActionView::Base; end
       let(:params) { [:email, label: custom_label] }
```
reggieb: Another one
```ruby
let(:params) { [attribute.to_sym, label: custom_label] }
```

spec/lib/govuk_elements_form_builder/form_builder_spec.rb
```diff
@@ -70,19 +52,25 @@ class TestHelper < ActionView::Base; end
...
 
       it 'shows the custom label' do
-        label = input.parent.at_css('label')
-
         expect(label.classes).to include('govuk-label')
         expect(label[:for]).to eq('email')
```
reggieb:
```ruby
expect(label[:for]).to eq(attribute)
```

spec/lib/govuk_elements_form_builder/form_builder_spec.rb
```diff
@@ -91,136 +79,106 @@ class TestHelper < ActionView::Base; end
...
-        expect(input.classes).to include(custom_class)
+        expect(tag.classes).to include(custom_class)
       end
     end
 
-    context 'has an input_prefix option' do
-      let(:prefix) { '£' }
-      let(:params) { [:email, input_prefix: prefix] }
+    context 'pass a label parameter with text and size' do
+      let(:custom_label) { "Your client's email" }
```
reggieb: This could just be random text:
```ruby
let(:custom_label) { Faker::Lorem.sentence }
```

spec/lib/govuk_elements_form_builder/form_builder_spec.rb
```diff
@@ -91,136 +79,106 @@ class TestHelper < ActionView::Base; end
...
-        expect(input.previous_element.content).to eq(prefix)
-        expect(input.previous_element.name).to eq('span')
-        expect(input.previous_element.classes).to contain_exactly('govuk-prefix-input__inner__unit')
-        expect(input.parent.name).to eq('div')
-        expect(input.parent.classes).to contain_exactly('govuk-prefix-input__inner')
-        expect(input.parent.parent.name).to eq('div')
-        expect(input.parent.parent.classes).to contain_exactly('govuk-prefix-input')
+      it 'shows the custom label' do
+        expect(label.classes).to include('govuk-label')
+        expect(label[:for]).to eq('email')
```
reggieb: `attribute`

Kalagan:
@reggieb, thanks for reviewing. I made all your suggested changes

PR #280, opened 2019-02-12
Improve text layout on check merits answers screen

mpw5:
## What

[AP-361](https://dsdmoj.atlassian.net/browse/AP-361)

On a number of merits questions a provider can enter text explanations. It is possible that a user will enter multiple paragraphs of text, especially on the Statement of Case screen. At present when these responses are displayed on the check your answers screen line-breaks are ignored which means that multiple paragraphs are displayed as a single block of text, making it hard to read.

To improve this, use the `simple_format` method to display text on the check your answers screen as it was entered by the user.

Note: there are other existing issues with the look and feel of this screen which will be addressed by a separate story to be created by the designer.

## Checklist

### DIFF
--- a/app/views/providers/check_merits_answers/show.html.erb
+++ b/app/views/providers/check_merits_answers/show.html.erb
@@ -16,7 +16,7 @@
         ) %>
   </dl>
 
-  <div class='govuk-body'><%= @merits.application_purpose %></div>
+  <div class='govuk-body'><%= simple_format @merits.application_purpose %></div>
 
   <dl class="govuk-summary-list govuk-!-margin-bottom-9">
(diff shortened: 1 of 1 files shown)


app/views/providers/check_merits_answers/show.html.erb
```diff
@@ -54,9 +54,9 @@
           question: t('.items.prospects_of_success'),
           answer: t("shared.forms.success_prospect.success_prospect_item.#{@merits.success_prospect}")
         ) %>
-  </dv>
+  </dl>
```
reggieb: Ah! you spotted that too! :smile:

PR #284, opened 2019-02-15
Ap 330 check malware

Kalagan:
## What

[AP-330](https://dsdmoj.atlassian.net/browse/AP-330)

Adding malware validation on the statement of case file upload. 
For now it is fairly simple but I think we can improve it with ulterior tech debt tasks:

TODO:
- move `clamav` to its own container.
- do the malware check in sidekiq. This is a bit tricky because we still need to show **"The selected file contains a virus"** to the user. Therefore, the page would still need to wait for the result of the scan before moving on.

Ideally I think we could even create a completely separate web service that we could make available to other .gov services

## Checklist


app/controllers/providers/statement_of_cases_controller.rb
```diff
@@ -6,21 +6,19 @@ class StatementOfCasesController < ProviderBaseController
...
+      @form = StatementOfCases::StatementOfCaseForm.new(statement_of_case_params.merge(model: statement_of_case))
+
+      render :show unless save_continue_or_draft(@form)
     end
 
     private
 
     def statement_of_case_params
-      params.require(:statement_of_case).permit(:statement, :original_file)
+      params.require(:statement_of_case).permit(:statement, :original_file).merge(provider_uploader: current_provider)
```
reggieb: Is it worth just doing one `merge` on the `statement_of_case_params`. So here:
```ruby
params.require(:statement_of_case).permit(:statement, :original_file).merge(provider_uploader: current_provider, model: statement_of_case)
```
And remove the merge in the update action.

PR #286, opened 2019-02-17
Add a back to applications link to the access denied page

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-351)

Add a return to applications link to the access denied page so a provider can exit the page and return to where they can search for the application they need.

## Checklist

### DIFF
--- a/app/views/shared/access_denied.html.erb
+++ b/app/views/shared/access_denied.html.erb
@@ -5,6 +5,8 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= t('.heading_1') %></h1>
     <p class="govuk-body"><%= t('.permission_text') %></p>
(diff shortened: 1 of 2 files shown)


app/views/shared/access_denied.html.erb
```diff
@@ -5,5 +5,6 @@
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl"><%= t('.heading_1') %></h1>
     <p class="govuk-body"><%= t('.permission_text') %></p>
+    <p class="govuk-body"><a href="/providers/applications"><%= t('.back_to_apps_link') %></a></p>
```
reggieb: Can you please use a `link_to` and the correct route helper:
```irb
<p class="govuk-body">
  <%= link_to t('.back_to_apps_link'), providers_legal_aid_applications_path %>
</p>
```
Hard coding a route relative to root can cause problems in the long term - specifically if the app is hosted within a sub-domain.

PR #292, opened 2019-02-18
Ap 336 provider capital cya design

mpw5:
## What

[AP-336](https://dsdmoj.atlassian.net/browse/AP-336)

Bring the provider check passported answers page in line with the current design. Borrows from the citizen version of this page, where these changes have already been made.

* User new helper methods created for the corresponding citizen change
* Add new merits header and text to the view
* Correct minor wording issues
* Add missing translations for additional properties
* Add specs and feature tests
* Remove unused helper methods and partials

## Checklist


spec/requests/providers/check_passported_answers_spec.rb
```diff
@@ -92,13 +93,35 @@
...
         it 'does not display percentage owned' do
           expect(response.body).not_to include(number_to_percentage(application.percentage_home, precision: 2))
           expect(response.body).not_to include('Percentage')
         end
       end
 
+      context 'applicant does not have any savings' do
+        let(:application) { create :legal_aid_application, :with_everything, :with_no_savings, :answers_checked }
+        it 'displays that no savings have been declared' do
+          expect(response.body).to include(I18n.translate('.generic.none_declared'))
```
reggieb: The `t` alias of `translate` is used elsewhere, so I would suggest:
```
I18n.t('.generic.none_declared')
```

spec/requests/providers/check_passported_answers_spec.rb
```diff
@@ -42,6 +42,7 @@
         expect(response.body).to include('Savings')
         expect(response.body).to include('assets')
         expect(response.body).to include('restrictions')
+        expect(response.body).not_to include(I18n.translate('.generic.none_declared'))
```
reggieb: This is a really important line of code. Without it the specs at 103, 110, and 117 have very limited value - in that this line tests the normal behaviour, and 103, 110, 117 contexts test how this behaviour is changed by modification to the application choices.

Therefore I would suggest this expectation is moved out to its own `it` clause, and that contexts 103, 110, 117 are moved higher within the file so that they immediately follow it.

With that change the specs will effectively read:

- This is the normal behaviour
- In this context x that behaviour changes
- In this context y that behaviour changes
- In this context z that behaviour changes

app/views/shared/check_answers/_one_link_section.html.erb
```diff
@@ -1,5 +1,8 @@
-
-<div class="govuk-grid-row" id="app-check-your-answers__<%= name %>">
+<% if answer_hash.present? %>
+  <div class="govuk-grid-row" id="app-check-your-answers__<%= name %>">
+<% else %>
+  <div class="govuk-grid-row" id="app-check-your-answers__<%= name %>_header">
+<% end %>
```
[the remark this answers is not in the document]
mpw5: Thanks, that's much neater than mine 👍

PR #296, opened 2019-02-19
Improve malware scanner

Kalagan:
## What

Improve the call to the malware scan `clamdscan`.
Before, if there was a problem with the command itself, it would no raise an error but fail silently and give a false positive result.
Now, it will raise an error.
And `Open3.capture3` seems to be a prefered way to run system commands with which you can get every outputs (stdout, stderr, and the exit code)

## Checklist

### DIFF
--- a/app/services/malware_scanner.rb
+++ b/app/services/malware_scanner.rb
@@ -1,9 +1,11 @@
 class MalwareScanner
+  ClamdscanError = Class.new(StandardError)
+
   def self.call(file_path)
     new(file_path).call
   end
 
-  COMMAND = 'clamdscan --fdpass --no-summary'.freeze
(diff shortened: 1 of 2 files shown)


app/services/malware_scanner.rb
```diff
@@ -17,16 +19,25 @@ def call
     MalwareScanResult.new(
       uploader: uploader,
       virus_found: virus_found?,
-      scan_result: scan_result,
+      scan_result: scan_result.stdout.strip,
```
reggieb: I think the strip should be used within the object constructor rather than on its use. So here:
```ruby
scan_result: scan_result.stdout
```
And below:
```ruby
OpenStruct.new(
  stdout: stdout.strip,
  status: status
)

PR #297, opened 2019-02-19
AP-185 Prep: Create underlying model structure

reggieb:
[Jira AP-185](https://dsdmoj.atlassian.net/browse/AP-185)

This PR creates the models that will be used within AP-185:

- **TransactionType** There are a small number of transaction types split into two operation types (credit and debit). A population method (with matching seeding action) creates the required transaction types.
- **LegalAidApplicationTransactionType** Joins transaction types to legal aid applications

It also adds a `belongs_to :transaction_type` to `BankTransaction`.

So with this in place you can do the following:

```ruby
TransactionType.populate

bank_transaction = BankTransaction.first
transaction_type = TransactionType.first
bank_transaction.transaction_type = transaction_type

legal_aid_application = LegalAidApplication.first
debit_transaction_type = TransactionType.debits.first
credit_transaction_type = TransactionType.credits.first

legal_aid_application.transaction_types = [debit_transaction_type, credit_transaction_type]
legal_aid_application.transaction_types.debits # => [debit_transaction_type]
legal_aid_application.transaction_types.credits # => [credit_transaction_type]
```


README.md
```diff
@@ -84,7 +84,9 @@ VCR_RECORD_MODE=all bin/rake
 
 The deployment is triggered on all builds in [CircleCI](https://circleci.com/gh/ministryofjustice/laa-apply-for-legal-aid) but requires approval to the desired environment.
 
-**NOTE:** **git-crypt** is required to store secrets required for **uat**, **staging** and **production** environments. To be able to modify those secrets, **git-crypt** needs to be set up according to the following [guide](https://ministryofjustice.github.io/cloud-platform-user-docs/03-other-topics/001-git-crypt-setup/#git-crypt).
+**NOTE:** **git-crypt** is required to store secrets required for **uat**, **staging** and **production** environments.
+To be able to modify those secrets, **git-crypt** needs to be set up according to the following
+[guide](https://ministryofjustice.github.io/cloud-platform-user-docs/03-other-topics/001-git-crypt-setup/#git-crypt).
```
reggieb: Just tidying up (spitting the text into separate lines so it's easier to read within an editor)

app/models/legal_aid_application.rb
```diff
@@ -18,6 +18,8 @@ class LegalAidApplication < ApplicationRecord
   has_one :statement_of_case
   has_many :legal_aid_application_restrictions
   has_many :restrictions, through: :legal_aid_application_restrictions
+  has_many :legal_aid_application_transaction_types
```
Kalagan: is this line needed if we never call it in the code?

app/models/transaction_type.rb
```diff
@@ -0,0 +1,29 @@
...
+    ]
+  }.freeze
+
+  scope :debits, -> { where(operation: :debit) }
+  scope :credits, -> { where(operation: :credit) }
+
+  def self.populate
+    existing = pluck(:name).map(&:to_sym)
+    NAMES.each do |operation, names|
+      (names - existing).each { |name| create!(name: name, operation: operation) }
```
Kalagan: you could also use `.find_or_create!` instead of fetching the existing ones. Just a suggestion. Feel free to ignore

app/models/transaction_type.rb
```diff
@@ -0,0 +1,29 @@
+class TransactionType < ApplicationRecord
```
Kalagan: Usually we would define
```ruby
enum operation: {
  credit: 'credit'.freeze,
  debit: 'debit'.freeze
}
```
in this file and it would prevent creating a `TransactionType` with an operation different than `debit` or `credit`.
Not sure if we can reuse the enum from `BankTransaction` like `enum operation: BankTransaction::operation`

app/models/transaction_type.rb
reggieb: The problem with using `find_or_create` in the loop is that it will make a SQL call (to see if one exists) on each iteration of the loop. Doing the `pluck` means only SQL to determine existing items. And the only SQL made after that being to create the items that don't exist.

Kalagan: Ok it makes sense

app/models/transaction_type.rb
reggieb: I didn't do that because the operations are effectively defined by `NAMES` - they are the keys! Defining them again in an `enum` would mean they were defined twice.

app/models/legal_aid_application.rb
reggieb: Yes - without it the next line won't work.

app/models/transaction_type.rb
reggieb: I could do:
```ruby
enum operation: NAMES.keys.each_with_object({}) {|k,h| h[k] = k.to_s.freeze}
```

PR #298, opened 2019-02-20
AP-324 convert statement of case file to PDF

Kalagan:
## What

[AP-324](https://dsdmoj.atlassian.net/browse/AP-324)

- When a statement of case file is uploaded, convert it to a pdf in a background worker
- `config.active_storage.service` was duplicated in the config files so I removed the duplication

## Checklist


app/services/active_storage_pdf_converter.rb
```diff
@@ -0,0 +1,34 @@
+class ActiveStoragePdfConverter
+  def self.call(active_storage_file)
+    new(active_storage_file).call
+  end
```
reggieb: :smiley:

app/services/statement_of_cases/pdf_converter.rb
```diff
@@ -0,0 +1,35 @@
...
+    def converted_file
+      ActiveStoragePdfConverter.call(statement_of_case.original_file)
+    end
+
+    def original_file_base_name
+      File.basename(original_file_name, File.extname(original_file_name))
+    end
+
+    def original_file_name
+      @original_file_name ||= statement_of_case.original_file.filename.to_s
```
reggieb: Is 'filename' going to ever be anything other than a string ... and if it's not a string shouldn't that cause an error rather than just returning a string representation of that not-a-string thing.

docker/run
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+/etc/init.d/clamav-daemon start
```
reggieb: Has that slipped in from another PR?

app/services/statement_of_cases/pdf_converter.rb
Kalagan: `filename` is actually an instance of `ActiveStorage::Filename` (https://api.rubyonrails.org/classes/ActiveStorage/Filename.html).
But I just realised that it has `.base` and `.extension_with_delimiter` methods so I should be able to use that instead of `File.basename` and `File.extname` on line 28 of this file

docker/run
Kalagan: Kind of, I was trying to fix a bug in which `clamav-daemon` doesn't restart  when the container restart. I could write it in another PR

reggieb: No problem - just making sure it was something that would cause a error on its own.

Kalagan:
superseded by https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/321

PR #300, opened 2019-02-21
Ap342 change links

willc-work:
Amends the page flow so that when a user wants to change a capital answer from the check your answers page, they go via the restrictions page before returning to the check your answers page.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-342)

amended citizen capital flow, updated cucumber tests to accommodate the new flow and updated affected tests.

## Checklist


spec/requests/citizens/percentage_home_spec.rb
```diff
@@ -49,5 +49,22 @@
...
+
+    context 'while checking answers' do
+      before { application.check_citizen_answers! }
+
+      it 'redirects to the "restrictions" page' do
+        subject
+        expect(response).to redirect_to(citizens_restrictions_path)
+      end
+    end
+    # context 'while checking answers' do
```
mpw5: Can this commented code be deleted?

features/providers/step_definitions/civil_journey_steps.rb
```diff
@@ -128,6 +128,13 @@
   end
 end
 
+Given('I select Check Your Answers Change link for {string}') do |field_name|
+  field_name.downcase!
+  field_name.gsub!(/\s+/, '_')
+  # Used where there are duplicated id fields
+  find("#app-check-your-answers__#{field_name}:nth-child(1)").click_link('Change')
```
mpw5: Should there be elements with duplicated ids? I think ids should be unique on the page. It might be better to change the way the html is generated to prevent the duplication, rather than add a workaround here.

PR #301, opened 2019-02-21
Ap 185 identify types of income pages

reggieb:
[Jira AP-185](https://dsdmoj.atlassian.net/browse/AP-185)

Adds a page that allow a Citizen to identify types of income


app/views/citizens/identify_types_of_incomes/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <%= govuk_hint t('.no_income_message'), class: 'govuk-!-padding-bottom-5' %>
+
+    <%= transaction_type_check_boxes(
+          form: form,
+          transaction_types: TransactionType.credits,
+          hints: { benefits: t('.benefits_examples') }
+        ) %>
+
+    <%= next_action_buttons(
+          show_draft: local_assigns.key?(:show_draft) ? show_draft : false,
```
Kalagan: why do we need this logic? Shouldn't it be always `false` (which is the default value I believe) because we are on the citizen side?

reggieb: Yes - you are right. I'll change it.

reggieb: In fact, as the default for `next_action_buttons` is to set it false, I don't need a `show_draft` clause here at all!

PR #303, opened 2019-02-22
AP 247 provider capital introduction

willc-work:
add the capital introduction page to the start of the capital flow questions.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-247)

updated the provider flow to direct the provider to this page if the client is passported, otherwise they dont see this page. 

## Checklist

### DIFF
--- a/app/views/providers/capital_introductions/show.html.erb
+++ b/app/views/providers/capital_introductions/show.html.erb
@@ -1,11 +1,11 @@
 <%= page_template page_title: t('.h1-heading') do %>
-  <%= simple_format t('.text_1') %>
+  <%= simple_format t('.does_your_client') %>
 
-  <%= list_from_translation_path 'providers.capital_introductions.show' %>
+  <%= list_from_translation_path 'providers.capital_introductions.show.about_the_client' %>
 
-  <%= simple_format t('.text_2') %>
+  <%= simple_format t('.client_to_provide') %>
 
-  <%= list_from_translation_path 'providers.capital_introductions.show.list_2' %>
+  <%= list_from_translation_path 'providers.capital_introductions.show.items_to_be_provided' %>
 
   <div class="govuk-!-padding-bottom-4"></div>
 
--- a/app/views/providers/capital_introductions/show.html.erb
+++ b/app/views/providers/capital_introductions/show.html.erb
@@ -1,0 +1,18 @@
+<%= page_template page_title: t('.h1-heading') do %>
+  <%= simple_format t('.text_1') %>
+
+  <%= list_from_translation_path 'providers.capital_introductions.show' %>
+
+  <%= simple_format t('.text_2') %>
+
+  <%= list_from_translation_path 'providers.capital_introductions.show.list_2' %>
+
+  <div class="govuk-!-padding-bottom-4"></div>
+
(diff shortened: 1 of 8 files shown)


app/views/providers/capital_introductions/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= page_template page_title: t('.h1-heading') do %>
+  <%= simple_format t('.text_1') %>
```
reggieb: I think this would be more readable if the text was labelled more meaningfully (see translation file)

config/locales/en/providers.yml
```diff
@@ -214,3 +214,18 @@ en:
...
+          has any other ways to contribute towards legal costs
+        list_2:
+          list: |
+            a statement of case
+            information about any attempts to settle
+            cost estimates
+            the prospects of success
+        h1-heading: Before you continue
+        text_1: "You'll need to tell us if your client:"
+        text_2: "You'll also need to provide:"
```
reggieb: `list_2`, `text_1` and `text_2` could be made meaningful and in that way it would easier to read the erb markup and understand what is being presented on the page. I would suggest renaming these elements to:

- `list_2` => `items_to_be_provided`
- `text_1` => `does_your_client`
- `text_2` => `client_to_provide`

Also you could add a meaningful grouping label to the first list:
```yml
     show:
        about_the_client:
          list: |
            owns a home or has a mortgage (and their values)
            has any savings, investments or other capital (and their values)
            has any other ways to contribute towards legal costs
```

spec/requests/providers/capital_introductions_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        login_as provider
+        subject
+      end
+
+      it 'returns http success' do
+        expect(response).to have_http_status(:ok)
+        expect(response.body).to include('Before you continue')
+      end
+
+      context 'when the Continue button is pressed' do
```
reggieb: This is a completely different spec - the fact that the subject has changed tells us that. So rather being a change of context within the GET action spec, this needs its own PATCH describe block outside of the GET describe block.

config/locales/en/providers.yml
willc-work: the view is definitely much clearer with your suggestions

spec/requests/providers/capital_introductions_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        subject { patch providers_legal_aid_application_capital_introduction_path(legal_aid_application) }
+        let(:submit_button) { { continue_button: 'Continue' } }
+
+        it 'redirects to next page' do
+          subject
+          expect(response).to redirect_to(providers_legal_aid_application_own_home_path(legal_aid_application))
+        end
+      end
+    end
+  end
```
reggieb: I think you've done that in a rush. Try this:
```ruby
describe 'PATCH /providers/applications/:id/capital_introduction' do
  subject { patch providers_legal_aid_application_capital_introduction_path(legal_aid_application) }

  before do
    login_as provider
  end

  it 'redirects to next page' do
    expect(subject).to redirect_to(providers_legal_aid_application_own_home_path(legal_aid_application))
  end
end
```

PR #311, opened 2019-02-26
HTLM tidy up as per ticket.

MoJ-Longbeard:
Including change to CSS as per ticket comment in preparation for HTML
change in comments which isn't done here.

## What

[AP-360](https://dsdmoj.atlassian.net/browse/AP-360)

- Moved `<meta charset="utf-8" />` to the top of the `<head>` tag.
- Removed `type='text/javascript'` from the `<script>` tag.
- Changed the CSS to not be reliant on `<span class="user-info">` as this tag will need to be removed.

## Checklist


mpw5:
FYI - I've pushed an additional [commit](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/311/commits/c0e860c9411ca33480707bf000e8c3323f9de7ca) to this branch which I discussed with @MoJ-Longbeard  earlier. It removes the code from `app/helpers/application_helper.rb` which generates `<span class="user-info">`. Malcolm has made the necessary changes so that the page works correctly without that tag in another commit in this PR.

PR #312, opened 2019-02-26
Ap 341 Add super user for staging maintenance

reggieb:
[Jira AP-341](https://dsdmoj.atlassian.net/browse/AP-341)

Adds:
- `admin` area of app
- first admin controller, with index as admin root. That is, `/admin` points to `admin/legal_aid_application#index`
- view of most recent Legal Aid Applications
- facility to delete all applications (if in Admin Reset Enabled mode)
- facility to set Admin Reset Enabled mode via environment variable - (so that disabled on production instance)
- `AdminUser` with facility to generate one via seeds
- Restricted access to `/admin` to logged in admin users via Devise (using username/password)
- log out for admin user when in `/admin`


app/views/admin_users/sessions/new.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= page_template page_title: t('.heading'), back_link: :none do %>
+
+  <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
+    <div class="govuk-form-group">
+      <%= f.label :username, class: 'govuk-label' %>
+      <%= f.text_field :username, for: 'width-15', autofocus: true, class: 'govuk-input govuk-input--width-20' %>
```
Kalagan: what does the `for: 'width-15'` do?

db/seeds/admin_users.rb
```diff
@@ -0,0 +1,6 @@
+if Rails.configuration.x.admin_portal.password.present?
+  AdminUser.find_or_create_by(username: 'apply_maintenance') do |admin_user|
+    admin_user.password = Rails.configuration.x.admin_portal.password
+    admin_user.email = 'apply-for-legal-aid@digital.justice.gov.uk'.freeze
```
Kalagan: Should we define `'apply_maintenance'` and `'apply-for-legal-aid@digital.justice.gov.uk'` as constants in the `AdminUser` model?

reggieb: I don't think so - as I'm not sure how persistent they need to be. At the moment, I think they are just values for an initial instance of `AdminUser` so relevant to seeds, but quite possibly no relevance to the model going forward.

I could be wrong.

app/views/admin_users/sessions/new.html.erb
reggieb: Ah - good spot - some code I put in by mistake and forgot to remove.

db/seeds/admin_users.rb
Kalagan: fair enough!

config/locales/en/admin.yml
```diff
@@ -0,0 +1,19 @@
+---
```
stephenrichards: If the `admin` keys are being separated out, then you should add the key to `config/i18n-tasks.yml` so that if we ever run `i18n-tasks normalize -p`, it will keep the `admin` keys in a separate file.

reggieb: I've modified `config/i18n-tasks.yml` following the [README](https://github.com/glebm/i18n-tasks#pattern-router), so that we no longer need to manually add new filenames to the config file.

stephenrichards: Nice!  But it does mean  that one file is created for each level-1 key, so 11 new files are created.  Are we happy to leave it like that and run `18n-tasks normalize -p` sometime in the future, or run it now and check the new files into github?

I'm easy either way.

reggieb: I ran `i18n-tasks normalize` to check my modification worked. - this PR includes the modifications made by running that command after I made the change to the config..

reggieb: Just had another play `i18n-tasks normalize -p` does add new files. Personally, I'd prefer a system that does require me to keep updating the configuration, even if that does generate more files.

PR #313, opened 2019-02-26
AP-383 identify salary - sorting and select all

Kalagan:
## What

[AP-383](https://dsdmoj.atlassian.net/browse/AP-383)

Adding features to the bank transactions page: 
- select all
- ability to sort columns
- Removing check box next to items already selected in other categories

**select all** and the **sort** are mostly copied from Malcolm's code on the prototype

## Checklist

### DIFF
--- a/app/assets/javascripts/table-select-all.es6
+++ b/app/assets/javascripts/table-select-all.es6
@@ -1,0 +1,32 @@
+$(document).ready(function() {
+  if ($('.select-clear-all').length) {
+    function resetSelectClearLink(table) {
+      const link = table.find('.select-clear-all');
+      link.removeClass('select-all clear-all');
+
+      if (table.find('input:checked').length) {
+        link.addClass('clear-all').text(link.data('copy-clear'));
+      }
+      else {
+        link.addClass('select-all').text(link.data('copy-select'));
+      }
+    }
+
+    $('.select-clear-all').click(function() {
+      const table = $(this).parents('table');
+      if($(this).hasClass('clear-all')) {
+        table.find('input:checked').prop('checked', false);
+      } else if ($(this).hasClass('select-all')) {
+        table.find('input:not(:checked)').prop('checked', true);
+      }
+      resetSelectClearLink($(this).parents('table'))
+      return false;
+    });
+
+    $('table input').click(function() {
+      resetSelectClearLink($(this).parents('table'))
+    });
+
+    resetSelectClearLink($('.select-clear-all').parents('table'))
+  }
+});
--- a/app/assets/javascripts/table-sort.es6
+++ b/app/assets/javascripts/table-sort.es6
@@ -1,0 +1,78 @@
+$(document).ready(function() {
+  if ($('table.sortable')) {
+    let table;
+    $('th.sort')
+      .wrapInner('<span class="sortable-column" title="sort this column"/>')
+      .each(function(index) {
+        const th = $(this),
+          thIndex = index + 1;
+        let inverse = false;
+
+        th.click(() => {
+          table = th.parents('table');
+          th.parent().children().removeClass('header-sort-asc header-sort-desc')
+          th.addClass(inverse ? 'header-sort-asc' : 'header-sort-desc')
+
+          table.find('td').filter(function() {
+            return $(this).index() === thIndex;
+          }).sortElements((a, b) => (
+            $(a).sortableValue() > $(b).sortableValue() ?
+              inverse ? -1 : 1 :
+              inverse ? 1 : -1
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/table-select-all.es6
```diff
@@ -0,0 +1,31 @@
...
+    function resetSelectClear(table) {
+      const link = table.find('.select-clear-all');
+      link.removeClass('select-all clear-all');
+
+      if (table.find('input:checked').length) {
+        link.addClass('clear-all').text(link.data('copy-clear'));
+      }
+      else {
+        link.addClass('select-all').text(link.data('copy-select'));
+      }
```
reggieb: I think this means that the select-all option is only available if no check boxes are checked. That is, if you check one check box, you can only clear all. Is that correct?

app/assets/javascripts/table-select-all.es6
```diff
@@ -0,0 +1,31 @@
...
+      if($(this).hasClass('clear-all')) {
+        table.find('input:checked').trigger('click');
+      } else if ($(this).hasClass('select-all')) {
+        table.find('input:not(:checked)').trigger('click');
+      }
+      return false;
+    });
+
+    $('table input').click(function() {
+      resetSelectClear($(this).parents('table'))
```
reggieb: If I read this correctly, if any check box is clicked, all the check boxes are reset even though only one check box has changed.

app/assets/javascripts/table-sort.es6
```diff
@@ -0,0 +1,78 @@
+$(document).ready(function() {
+  if ($('table.sortable')) {
+    let table;
+    $('th.sort')
+      .wrapInner('<span class="sortable-column" title="sort this column"/>')
+      .each(function() {
+        const th = $(this),
+          thIndex = th.index();
```
reggieb: I think the[ jQuery function `each`](http://api.jquery.com/jquery.each/) passes the index and current element to the inner function, so I don't think you should have to grab them again.:
```javascript
.each(function(thIndex, th) {
```

app/assets/javascripts/table-sort.es6
```diff
@@ -0,0 +1,78 @@
...
+        let inverse = false;
+
+        th.click(() => {
```
reggieb: Why is the ES6 syntax for a function call being used here and not elsewhere?

app/assets/javascripts/table-select-all.es6
Kalagan: Yes, that's correct

app/assets/javascripts/table-select-all.es6
Kalagan: `resetSelectClear` doesn't touch the checboxes. It only changes the class and copy of the "select all / clear all" link. Maybe the name of that method is confusing...

app/assets/javascripts/table-sort.es6
Kalagan: good point, I'll try that

app/assets/javascripts/table-sort.es6
Kalagan: If you use ES6's `=>`, you don't have access to `this` inside the function.
And most other functions in this code are using `this`

app/assets/javascripts/table-select-all.es6
reggieb: OK

app/views/citizens/transactions/show.html.erb
```diff
@@ -8,29 +8,43 @@
...
       <tbody class="govuk-table__body">
         <%= form.collection_check_boxes :transaction_ids, @bank_transactions, :id, :description do |builder| %>
           <tr class="govuk-table__row">
             <th class="govuk-table__cell">
-              <div class="govuk-checkboxes__item">
-                <% checked = builder.object.transaction_type_id == @transaction_type.id %>
-                <%= builder.check_box(class: 'govuk-checkboxes__input', checked: checked) %>
-                <label class="govuk-label govuk-checkboxes__label"></label>
-              </div>
+              <% if builder.object.transaction_type_id == nil || builder.object.transaction_type_id == @transaction_type.id %>
```
reggieb: Would this work here:
```erb
<% if builder.object.transaction_type_id ||= @transaction_type.id %>
```

Kalagan: I think that's different.
your code assigns the transaction_type we want to select to all transactions that are not assigned to any transaction type yet

reggieb: Sorry - you're right.

PR #315, opened 2019-02-27
AP-385 Using ClamAV inside a container and re-enabling it

Kalagan:
## What

[AP-385](https://dsdmoj.atlassian.net/browse/AP-385)

Setting up ClamAV daemon inside a separate container.
On our main container, we still need to install it but we don't need to start it as a service or to update the virus database.
We use clamdscan with `clamdscan -c config/clamd.container.conf --stream`
```
# config/clamd.container.conf
TCPSocket 3310
TCPAddr localhost
```
It means the file is scanned by ClamAV on the other container.

For local dev, you can set `LOCAL_CLAMAV=true` in your `.env` to keep using the `clamdscan` from your computer.

## Checklist


.circleci/config.yml
```diff
@@ -30,6 +31,12 @@ references:
         sudo pip3 install awscli
         $(aws ecr get-login --region eu-west-1 --no-include-email)
 
+  install_clamav: &install_clamav
+    run:
+      name: Install Clamav Client
+      command: |
```
reggieb: Do you need the pipe if there is only one command?
```yml
command: sudo apt-get install clamav-daemon -y
```

Kalagan: good spot

PR #316, opened 2019-02-28
AP-73 Analytics

AbdirizakObsiye2018:
Set up Google Analytics

[Link to ticket](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-73)

Google Analytics (GA) gave a script, which I placed inside a partial to be rendered in the application.html.erb
It is only included when the environment is production or development ( Analytics on the development environment will be deleted once this feature has passed UAT).

The GA tracking id is placed inside environment variables, allowing for different tracking id's depending on the environment.

Excluded linters on the google analytics partial because it gave the below error:
- `erb interpolation in javascript tag must call '(...).to_json'`

## Checklist


app/views/layouts/_google_analytics.html.erb
```diff
@@ -0,0 +1,9 @@
+<% if Rails.env.production? ||  Rails.env.development? %>
```
reggieb: Do we really want to enable this in development? Yes while initially testing it, but once we're happy it's working - production only I think.

reggieb: Actually, would it be even easier just to use the presence of the environment variable?
```erb
<% if ENV['GOOGLE_ANALYTICS_TRACKING_ID'].present? %>
```

AbdirizakObsiye2018: Done. Thanks, good suggestion.

PR #321, opened 2019-03-01
AP-324 Convert statement of case files to PDF

Kalagan:
## What

[AP-324](https://dsdmoj.atlassian.net/browse/AP-324)

Re-work of PR https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/298 because statement of cases can now have several files

- When statement of case files are uploaded, convert them to a pdf in background workers
- `config.active_storage.service` was duplicated in the config files so I removed the duplication

## Checklist


app/services/pdf_converter.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def pdf_file
+    PdfFile.find_or_create_by!(original_file_id: original_file_id)
+  end
+
+  def converted_file
+    return downloaded_file if original_attachment.content_type == 'application/pdf'
+
+    file = Tempfile.new
+    Libreconv.convert(downloaded_file.path, file.path)
```
reggieb: Should there be a `file.close` here after the convert call?

spec/requests/providers/statement_of_case_spec.rb
```diff
@@ -45,18 +46,18 @@
...
       {
-        statement_of_case: {
-          statement: entered_text,
-          original_files: [original_file]
-        }
+        statement: entered_text,
+        original_files: [original_file]
       }
     end
+    let(:params) { { statement_of_case: params_statement_of_case }.merge(submit_button).merge(upload_button) }
```
reggieb: As I don't think there is a scenario where the two button are pressed together ... rather than `.merge(submit_button).merge(upload_button)` I'd suggest:
```rspec
let(:draft_button) { { draft_button: 'Save as draft' } }
let(:upload_button) { { upload_button: 'Upload' } }
let(:button_clicked) { { } }
let(:params) { { statement_of_case: params_statement_of_case }.merge(button_clicked) }
```
Then in the spec where draft button behaviour is tested:
```rspec
let(:button_clicked) { draft_button }
```
And where the upload button is used:
```rspec
let(:button_clicked) { upload_button }
```

spec/requests/providers/statement_of_case_spec.rb
```diff
@@ -45,18 +46,18 @@
...
+
+    subject { patch providers_legal_aid_application_statement_of_case_path(legal_aid_application), params: params }
 
     context 'when the provider is authenticated' do
```
reggieb: I know you haven't touched this, but while you're in the spec ... this context is irrelevant. I would delete it.

spec/requests/providers/statement_of_case_spec.rb
```diff
@@ -66,20 +67,42 @@
...
+
       context 'Continue button pressed' do
         context 'model has no files attached previously' do
           context 'file is empty and text it empty' do
-            before { params[:statement_of_case].delete(:original_files) }
-            let(:entered_text) { '' }
+            let(:params_statement_of_case) do
+              {
+                statement: ''
+              }
```
reggieb: In the context `context 'text is empty but file is present' do` the condition where `statement` is `''` is achieved by `let(:entered_text) { '' }`. I think the same pattern could be used here to keep consistency.

app/services/pdf_converter.rb
Kalagan: yes you're right

PR #325, opened 2019-03-04
AP-367 Add destroy uploaded file action to statement of case

reggieb:
[Jira AP-367](https://dsdmoj.atlassian.net/browse/AP-367)

- Add an destroy action to the Statement of Case controller. 
- Adds a form to each file in the original files list - that on submission submits the current original file id to the destroy action
- Restyles the submit buttons to look like links
- Moves the start of the form on the show page - so that the destroy forms are not nested within the main page forms.

### DIFF
--- a/app/views/providers/statement_of_cases/_uploaded_files.html.erb
+++ b/app/views/providers/statement_of_cases/_uploaded_files.html.erb
@@ -15,13 +15,13 @@
       <td class="govuk-table__cell no-wrap"><%= number_to_human_size(file.blob.byte_size) %></td>
       <td class="govuk-table__cell"><span class="govuk-tag"><%= t('.uploaded') %></span></td>
       <td class="govuk-table__cell">
-        <%= form_tag(
+        <%= button_to(
+              t('.delete'),
               providers_legal_aid_application_statement_of_case_path(@legal_aid_application),
-              method: :delete
-            ) do %>
-          <%= hidden_field_tag :original_file_id, file.id %>
-          <%= submit_tag t('.delete'), class: 'button-as-link' %>
-        <% end %>
+              method: :delete,
+              class: 'button-as-link',
+              params: { original_file_id: file.id }
+            ) %>
       </td>
     </tr>
   <% end %>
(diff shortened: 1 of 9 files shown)


app/views/providers/statement_of_cases/_uploaded_files.html.erb
```diff
@@ -14,9 +14,16 @@
...
       <td class="govuk-table__cell"><span class="govuk-tag"><%= t('.uploaded') %></span></td>
-      <td class="govuk-table__cell"><%= t('.delete') %></td>
+      <td class="govuk-table__cell">
+        <%= form_tag(
+              providers_legal_aid_application_statement_of_case_path(@legal_aid_application),
+              method: :delete
+            ) do %>
+          <%= hidden_field_tag :original_file_id, file.id %>
+          <%= submit_tag t('.delete'), class: 'button-as-link' %>
+        <% end %>
```
Kalagan: This should work as well:
```ruby
        <%= button_to(
              t('.delete'),
              providers_legal_aid_application_statement_of_case_path(@legal_aid_application),
              method: :delete,
              class: 'button-as-link',
              params: { original_file_id: file.id }
        %>
```

spec/requests/providers/statement_of_case_spec.rb
```diff
@@ -265,4 +265,58 @@
...
+    end
+
+    it 'redirects to show' do
+      subject
+      expect(request).to redirect_to(providers_legal_aid_application_statement_of_case_path(legal_aid_application))
+    end
+
+    it 'deletes the file' do
+      expect { subject }.to change { ActiveStorage::Attachment.count }.by(-1)
+      expect(ActiveStorage::Attachment.exists?(original_file.id)).to be(false)
```
Kalagan: It might be worth also checking `ActiveStorage::Blob.count`

app/views/providers/statement_of_cases/_uploaded_files.html.erb
reggieb: Yes - that's better.

PR #327, opened 2019-03-06
AP-395 Upload multiple files - JavaScript Enhancements

Kalagan:
## What

[AP-395](https://dsdmoj.atlassian.net/browse/AP-395)

Paired with Abdi
Implement file upload and file delete via JavaScript on statement of case page

## Checklist


app/assets/javascripts/statement_of_case_file_upload.es6
```diff
@@ -0,0 +1,84 @@
...
+            <td class="govuk-table__cell"></td>
+            <td class="govuk-table__cell no-wrap">
+              <img src="${image_loading_small}" class="small-loading-image"/>
+              ${copy_uploading}
+            </td>
+            <td class="govuk-table__cell"></td>
+          </div>
+        `
+      ))
+      uploaded_files_table_container.find('table').show().find('tbody').append(rows.join(''))
```
reggieb: Would this work:
```javascript
uploaded_files_table_container.find('table tbody').append(rows.join(''))
```

app/views/layouts/application.html.erb
```diff
@@ -78,6 +78,7 @@
     </div>
 
     <%= render partial: 'layouts/footer' %>
+    <%= render partial: 'layouts/javascript_variables' %>
```
reggieb: If these variables are only used on one page, should this be at the bottom of that page.

app/assets/javascripts/statement_of_case_file_upload.es6
Kalagan: We also need to call `show` on the table. When the statement of case has no file yet, the table is in the page but hidden. That way, when we add the first file, we just need to show the table and a row

PR #329, opened 2019-03-06
Add limitations page into the provider flow

willc-work:
Add a page into the provider flow to show the limitations on what work can be done/claimed for. 

[Link to story](https://dsdmoj.atlassian.net/browse/AP-402)

Moved the proceeding_type partials into the shared folder and amended them to make them re-useable. Added the new page into the provider flow. At present if you change the proceeding type from the check your answers page you are redirected back to check your answers, this behaviour is not mentioned on the ticket so I have not changed it but am seeking clarification from designers.

## Checklist


app/views/providers/limitations/show.html.erb
```diff
@@ -0,0 +1,35 @@
+<%= page_template(
+      page_title: t('.h1-heading')
+    ) %>
+
+<% if @legal_aid_application.proceeding_types.present? %>
```
reggieb: If no proceeding types are present ....... when can this occur? 

If it does occur, how does the user move on from this page as the "continue" button is within the conditional block?

app/views/providers/limitations/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h2 class="govuk-heading-m">
+        <%= t('.h2-heading', count: pluralize(@legal_aid_application.proceeding_types.size, t('.h2-heading'))) %>
```
reggieb: As the translation at `h2-heading` is just "Proceeding" with no sign of a `count` variable, this line should just be:
```erb
<%= t('.h2-heading' %>
```

app/views/providers/limitations/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      </h2>
+      <div class="selected-proceeding-types">
+        <%= render partial: 'shared/forms/proceedings_types/proceeding_type', collection: @legal_aid_application.proceeding_types, locals: { translation_path: 'providers.limitations.show' } %>
+      </div>
+
+      <div class="govuk-!-padding-bottom-6"></div>
+
+      <p class="govuk-body">
+        <%= t('.form_of_service.html') %> <%= t('.form_of_service_type') %> <%= content_tag(:br) %>
```
reggieb: ```<%= content_tag(:br) %>``` :scream: 

Just `<br>`

Having said that, there is almost always a better way of separating markup into separate lines. This looks like a [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) to me. So:
```
<dl>
  <dt><%= t('.form_of_service.html') %></dt>
  <dd><%= t('.form_of_service_type') %></dd>

  <dt><%= t('.incurrable_costs.html') %></dt>
  <dd><%= t('.incurrable_costs_value') %></dd>
  ....
```
Then use CSS (and perhaps some gov-uk classes) to get the formatting right

app/views/providers/limitations/show.html.erb
```diff
@@ -0,0 +1,35 @@
+<%= page_template(
+      page_title: t('.h1-heading')
+    ) %>
```
reggieb: I don't think needs to be multi-line.

app/views/providers/limitations/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      </h2>
+      <div class="selected-proceeding-types">
+        <%= render partial: 'shared/forms/proceedings_types/proceeding_type', collection: @legal_aid_application.proceeding_types, locals: { translation_path: 'providers.limitations.show' } %>
+      </div>
+
+      <div class="govuk-!-padding-bottom-6"></div>
+
+      <p class="govuk-body">
+        <%= t('.form_of_service.html') %> <%= t('.form_of_service_type') %> <%= content_tag(:br) %>
+        <%= t('.incurrable_costs.html') %> <%= t('.incurrable_costs_value') %> <%= content_tag(:br) %>
```
Kalagan: Using `<br>` for spacing is not a great practice I believe.
Furthermore, `content_tag(:br)` generates two `br` tags like that: `<br></br>` (which is kind of hacky).
I'd suggest using spacing like `<div class="govuk-!-padding-bottom-6"></div>` instead.

I was trying a few things and this doesn't need extra spacing and looks very close to the prototype.
```html
      <div class="govuk-body">
        <p><%= t('.form_of_service.html') %> <%= t('.form_of_service_type') %></p>
        <p><%= t('.incurrable_costs.html') %> <%= t('.incurrable_costs_value') %></p>
        <p><%= t('.allowable_work.html') %> <%= t('.body') %></p>
      </div>
```

You could add a bit of spacing between the `p` tags:
```html
      <div class="govuk-body">
        <p><%= t('.form_of_service.html') %> <%= t('.form_of_service_type') %></p>
        <div class="govuk-!-padding-bottom-1"></div>
        <p><%= t('.incurrable_costs.html') %> <%= t('.incurrable_costs_value') %></p>
        <div class="govuk-!-padding-bottom-1"></div>
        <p><%= t('.allowable_work.html') %> <%= t('.body') %></p>
      </div>
```

app/views/providers/limitations/show.html.erb
```diff
@@ -0,0 +1,35 @@
+<%= page_template(
+      page_title: t('.h1-heading')
+    ) %>
+
+<% if @legal_aid_application.proceeding_types.present? %>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
```
Kalagan: if you use the block version of the template (`<%= page_template(page_title: t('.h1-heading')) do %>`) around everything, you wouldn't need these 2 divs

app/views/shared/forms/proceedings_types/_proceeding_type.html.erb
```diff
@@ -8,7 +8,7 @@
   <div class="govuk-grid-column-one-third control-column">
     <p class="govuk-body govuk-!-margin-bottom-1 govuk-!-margin-top-4">
       <%= link_to(
-            t('generic.remove'),
+            t("#{translation_path}.link_title"),
```
reggieb: I'd suggest this as being a little cleaner:
```ruby
t('.link_title', scope: translation_path)
```
See https://guides.rubyonrails.org/i18n.html#looking-up-translations

app/views/providers/limitations/show.html.erb
Kalagan: Not a fan of this. I wonder if we should redirect to the previous page in that case.
With something like this in the controller:
```ruby
    def show
      unless legal_aid_application.proceeding_types.present?
        redirect_to providers_legal_aid_application_proceedings_types_path(legal_aid_application)
      end
    end
```

app/views/shared/forms/proceedings_types/_proceeding_type.html.erb
Kalagan: clever 👍

config/locales/en/providers.yml
```diff
@@ -239,3 +240,17 @@ en:
     success_prospects:
       show:
         h1-heading: What are the prospects of success?
+    limitations:
+      show:
+        h1-heading: What you're covered for
+        h2-heading: Proceeding
+        body: As to proceedings under Part IV Family Law Act 1996 limited to all steps up to and including obtaining and serving a final order and in the event of breach leading to the exercise of a power of arrest to representation on the consideration 	of the breach by the court (but excluding applying for a warrant of arrest, if not attached, and representation in contempt proceedings).
```
reggieb: Break this over multiple lines. If you use the YAML `>` operator the line breaks will be removed when rendered in the HTML markup. Personally I'd be tempted to use `|` as then the text will be more readable in the markup as well - it line breaks will disappear when rendered in the browesr. So:
```yaml
body: >
  As to proceedings under Part IV Family Law Act 1996 limited to all steps up to and
  including obtaining and serving a final order and in the event of breach leading to 
  the exercise of a power of arrest to representation on the consideration of the 
  breach by the court (but excluding applying for a warrant of arrest, if not attached, 
  and representation in contempt proceedings).
```
See https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

config/locales/en/providers.yml
```diff
@@ -239,3 +240,17 @@ en:
...
+        link_title: Change
+        form_of_service:
+          html: "<b>Form of service: </b>"
```
Kalagan: I don't have a strong opinion about this but it feels a bit weird to have some HTML in here.

reggieb: This is bad. You're mixing styling into translation when there is no need to.

Get the original mark up correct (if you don't like the description list I've suggested, then put a span around these elements - and do the styling with CSS).

Oh and `strong` would be better than `b` tags here. See https://stackoverflow.com/a/271755

app/views/providers/limitations/show.html.erb
reggieb: If you go for Julien's paragraph markup, I'd suggest either:
```erb
<p><strong><%= t('.form_of_service.html') %></strong> <%= t('.form_of_service_type') %></p>
```
Or use spans and style them with CSS rather than having the styling defined in the translation file.

app/views/providers/limitations/show.html.erb
reggieb: Julien's paragraphs works too if you prefer.

app/views/providers/limitations/show.html.erb
willc-work: the if block is not required so I have removed it. I had thought that it was possible to create an application without a proceeding type.

app/views/providers/limitations/show.html.erb
willc-work: I used <strong> with Julians suggestion. it seemed a bit much to add css just to bold these items.

PR #331, opened 2019-03-07
AP-408 Enter details of latest incident

reggieb:
[JIRA AP-408](https://dsdmoj.atlassian.net/browse/AP-408)

The main thing this does is add a new page where the provider can enter the date and details of an incident. The most complicated part of the task was getting the form to work - specifically getting the date assembly working (mostly reworking the way date of birth is handled in the applicants form).

I also added a couple of rspec helper methods ( so we don't need to keep updating existing specs when the flow changes). 
- **`provider_draft_endpoint`** returns the url a provider should be sent to on Saving as Draft. I've updated `Providers::Draftable` so that this endpoint is defined in one place - this method using that definition.
- **`flow_forward_path`** looks at the last `request.path` and uses that to determine what the next forward url should be via flows. 

I've updates a couple of other specs to make sure that the way I'd coded these helpers was generic.

### DIFF
--- a/spec/support/flow_helpers.rb
+++ b/spec/support/flow_helpers.rb
@@ -15,4 +15,8 @@ module FlowHelpers
     )
     flow.forward_path
   end
+
+  def provider_draft_endpoint
+    __send__(Providers::Draftable::ENDPOINT)
+  end
 end
--- a/spec/support/flow_helpers.rb
+++ b/spec/support/flow_helpers.rb
@@ -1,0 +1,18 @@
+module FlowHelpers
+  # Takes the last request path and uses it to determine the next step from flows
+  # Makes some assumptions so may not work with every request
+  def flow_forward_path
+    path = request.path
+    parts = path.split('/').select(&:present?)
+    journey = parts.first.to_sym
+    step_index = journey == :citizens ? 1 : 3
+    current_step = parts[step_index].pluralize.to_sym
+    legal_aid_application = journey == :citizens ? LegalAidApplication.new : LegalAidApplication.find(parts[2])
+    flow = Flow::BaseFlowService.flow_service_for(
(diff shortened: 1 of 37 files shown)


spec/support/flow_helpers.rb
```diff
@@ -0,0 +1,22 @@
+module FlowHelpers
+  # Takes the last request path and uses it to determine the next step from flows
+  # Makes some assumptions so may not work with every request
+  def flow_forward_path
+    path = request.path
+    parts = path.split('/').select(&:present?)
+    journey = parts.first.to_sym
+    step_index = journey == :citizens ? 1 : 3
+    current_step = parts[step_index].pluralize.to_sym
+    legal_aid_application = journey == :citizens ? LegalAidApplication.new : LegalAidApplication.find(parts[2])
```
Kalagan: Did you try using `Rails.application.routes.recognize_path` ?
This seems to work:
```ruby
  def flow_forward_path
    path_details = Rails.application.routes.recognize_path(request.path, method: request.method)
    controller_details = path_details[:controller].split('/')
    controller_module = controller_details.first.to_sym
    controller_name = controller_details.last.to_sym
    legal_aid_application_id = path_details[:legal_aid_application_id]
    Flow::BaseFlowService.flow_service_for(
      controller_module,
      legal_aid_application: legal_aid_application_id ? LegalAidApplication.find(legal_aid_application_id) : LegalAidApplication.new,
      current_step: controller_name
    ).forward_path
  end
```

reggieb: That's nice!

reggieb:
Note to self: Check status of https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/334 before merging.

PR #334, opened 2019-03-08
Ap 268 Provider Starts Merits Application for non-passported applications

AbdirizakObsiye2018:
Provider Starts Merits Application for non-passported applications

[link to ticket](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-268)

Changed the flow so that after a Citizen/Applicant completes their means assessment, the provider_step will be changed to :merits_start which links to the start of merits application page (client_received_legal_helps)

This means that on the index page of the legal aid applications (providers/applicants), the provider will be redirected to the client_received_legal_helps page after they click a application that has status "Means test completed"

## Checklist


app/controllers/citizens/check_answers_controller.rb
```diff
@@ -9,6 +9,7 @@ def index
 
     def continue
       legal_aid_application.complete_means! unless legal_aid_application.means_completed?
+      legal_aid_application.update!(provider_step: :merits_start)
```
Kalagan: This line and the new code in `providers_helper.rb` are kind of duplications.
They both say : 
`When the means are completed, the providers next step is the start of the merits flow`.
This logic shouldn't be in 2 places.

Actually it would make sense to update `provider_step` in the state machine:
```ruby
# app/models/concerns/legal_aid_application_state_machine.rb

      event :complete_means do
        transitions from: :checking_citizen_answers, to: :means_completed,
                    after: -> do
                      CleanupCapitalAttributes.call(self)
                      update!(provider_step: :client_received_legal_helps)
                    end
        transitions from: :checking_passported_answers, to: :means_completed
      end
```

or 

```ruby
      event :complete_means do
        transitions from: :checking_citizen_answers, to: :means_completed,
                    after: -> do
                      CleanupCapitalAttributes.call(self)
                      update!(provider_step: Flow::Flows::ProviderMerits::STEPS[:merits_start])
                    end
        transitions from: :checking_passported_answers, to: :means_completed
      end
```

AbdirizakObsiye2018: Thanks. Good idea and now it's implemented.

app/models/concerns/legal_aid_application_state_machine.rb
```diff
@@ -50,7 +50,10 @@ module LegalAidApplicationStateMachine
 
       event :complete_means do
         transitions from: :checking_citizen_answers, to: :means_completed,
-                    after: -> { CleanupCapitalAttributes.call(self) }
+                    after: -> do
+                      CleanupCapitalAttributes.call(self)
+                      update!(provider_step: :client_received_legal_helps)
```
Kalagan: The should be sufficient for the feature to work. Do you still need the changes to the other files?

AbdirizakObsiye2018: Done thanks

reggieb:
Note that if https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/331 gets merged into master before this one, the code here will need a slight modification - as in that PR the start of Merits flow changes to `details_latest_incidents`

PR #339, opened 2019-03-11
AP-375 Show selected transaction on summary of income types page

reggieb:
[Jira AP-375](https://dsdmoj.atlassian.net/browse/AP-375)

- Updates model associations so that can easily get from an application to the bank transactions associated with it.
- Adds a scope to get a hash with transaction types as keys, and the array of bank transactions of that type as the values
- Adds a table of selected bank transactions to each transaction type section of `citizens/income_summary`
- Adds an action that disassociates a bank transaction from a transaction type and uses that as an endpoint for "remove" "links" (that are in fact buttons, but look like links) 

income_summary_with_bank_transactions

### DIFF
--- a/app/controllers/citizens/bank_transactions_controller.rb
+++ b/app/controllers/citizens/bank_transactions_controller.rb
@@ -1,0 +1,17 @@
+module Citizens
+  class BankTransactionsController < ApplicationController
+    include ApplicationFromSession
+    before_action :authenticate_applicant!
+
+    def remove_transation_type
+      bank_transaction.update transaction_type: nil
+      redirect_back fallback_location: citizens_identify_types_of_income_path
+    end
+
+    private
+
+    def bank_transaction
+      @bank_transaction ||= BankTransaction.find(params[:id])
+    end
+  end
+end
--- a/app/controllers/citizens/bank_transactions_controller.rb
+++ b/app/controllers/citizens/bank_transactions_controller.rb
@@ -4,7 +4,7 @@ module Citizens
     before_action :authenticate_applicant!
 
     def remove_transation_type
-      bank_transaction.update transaction_type: nil
+      bank_transaction.update! transaction_type: nil
       redirect_back fallback_location: citizens_identify_types_of_income_path
     end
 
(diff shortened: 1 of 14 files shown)


app/controllers/citizens/bank_transactions_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Citizens
+  class BankTransactionsController < ApplicationController
+    include ApplicationFromSession
+    before_action :authenticate_applicant!
+
+    def remove_transation_type
+      bank_transaction.update transaction_type: nil
```
Kalagan: `.update!` ?
😄

app/controllers/citizens/transactions_controller.rb
```diff
@@ -31,7 +31,7 @@ def reset_selection
     end
 
     def set_selection
-      BankTransaction
+      bank_transactions
         .where(id: transaction_ids_to_select)
         .update_all(transaction_type_id: transaction_type.id)
```
Kalagan: Would this be even better?
```ruby
bank_transactions
  .where(id: selected_transaction_ids)
  .update_all(transaction_type_id: transaction_type.id)
```

And we can get rid of the `transaction_ids_to_select` method

spec/requests/citizens/check_answers_spec.rb
```diff
@@ -68,7 +68,6 @@
     context 'applicant does not own home' do
       let(:legal_aid_application) { create :legal_aid_application, :with_everything, :without_own_home }
       it 'does not display property value' do
-        expect(response.body).not_to include(number_to_currency(legal_aid_application.property_value, unit: '£'))
```
Kalagan: what was wrong with these?

reggieb: The variance of values is not great enough. That is, occasionally there is another value on the page that has the same amount. So you get a false failure. The spec already checks the section isn't there. By adding a check that NOWHERE ON THE PAGE is there the same value, the spec is just being made more likely to cause an erroneous result.

app/controllers/citizens/bank_transactions_controller.rb
reggieb: doh!

PR #340, opened 2019-03-12
Tech debt: Step into journeys to improve cucumber performance

reggieb:
Steps that jumps into a journey by working through a large sequence of previous steps, have been replaced with steps that build the application and associated objects first and then step straight into journey at the appropriate point.

On my system, this PR changed the running time of the `cucumber` command:

Before: real (4 min 16 sec), user (1 min 53 sec)
After: real (2 min 7 sec), user (0 min 56 sec)

So time taken to complete the cucumber tests has been halved.

As the new system uses existing factory traits - there is now some randomness in attribute values. So some feature test that assumed a certain set of attributes are present before the changes made by the specific feature test, have been modified (and some removed). 

The basic assumptions I've used are:

- Feature test should test flow rather than specific behaviour (which should be tested in the request specs). For example, when checking answers, that when clicking on a change link, the user is taken on a flow that returns them to the checking answers page, without concentrating on how each page in that flow behaves.

While modifying these tests I've found a couple of bugs that I've fixed:

- There was a dependency in the `other_assets_declaration` page, that the user had previously visited another page.
- The no link items page created two elements with the same id.

### DIFF
--- a/features/citizens/citizens.feature
+++ b/features/citizens/citizens.feature
@@ -85,7 +85,8 @@ Feature: Citizen journey
 
   @javascript
   Scenario: I want to change property details via the check your answers page
-    Given I complete the citizen journey as far as check your answers
+    Given I have completed an application
+    And I complete the citizen journey as far as check your answers
     And I click Check Your Answers Change link for 'Own home'
     Then I should be on a page showing 'Do you own the home that you live in?'
     Then I click 'Continue'
@@ -103,13 +104,11 @@ Feature: Citizen journey
     Then I should be on a page showing 'Check your answers'
     And the answer for 'Own home' should be 'Yes, with a mortgage or loan'
     And the answer for 'Property value' should be '£500,000.00'
-    And the answer for 'Outstanding mortgage' should be '£100,000.00'
-    And the answer for 'Shared ownership' should be 'Yes, a partner or ex-partner'
-    And the answer for 'Percentage home' should be '50.00%'
 
   @javascript
   Scenario: I want to remove property details via the check your answers page
-    Given I complete the citizen journey as far as check your answers
+    Given I have completed an application
+    And I complete the citizen journey as far as check your answers
     And I click Check Your Answers Change link for 'Own home'
     Then I should be on a page showing 'Do you own the home that you live in?'
     Then I choose 'No'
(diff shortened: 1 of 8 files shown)


features/providers/check_your_answers.feature
```diff
@@ -146,9 +146,6 @@ Feature: Checking answers backwards and forwards
       Then I should be on a page showing 'Check your answers'
       And the answer for 'Own home' should be 'Yes, with a mortgage or loan'
       And the answer for 'Property value' should be '£500,000.00'
-      And the answer for 'Outstanding mortgage' should be '£100,000.00'
-      And the answer for 'Shared ownership' should be 'Yes, a partner or ex-partner'
-      And the answer for 'Percentage home' should be '50.00%'
```
reggieb: I've removed these three check as the test an assumed previous configuration, rather than the actions specified in feature test. That is, the feature test specifically sets a Property value, but does not touch the values that I've removed.

features/providers/check_your_answers.feature
```diff
@@ -161,24 +158,22 @@ Feature: Checking answers backwards and forwards
       And the answer for 'Own home' should be 'No'
 
     @javascript @vcr
-    Scenario: I want to remove savings via the capital check your answers page
+    Scenario: I want to view savings via the capital check your answers page
       Given I complete the passported journey as far as capital check your answers
       And I click Check Your Answers Change link for 'Savings and investments'
       Then I should be on a page showing 'Does your client have any savings and investments?'
-      Then I deselect 'Cash savings'
+      Then I click 'Continue'
```
reggieb: That a particular saving can be deselected via a page entry is a test of that page and is done via a request spec. What needs to be tested here is the flow behaviour - that is, that a user can get to this page from check your answers and return successfully.

features/citizens/citizens.feature
```diff
@@ -85,7 +85,8 @@ Feature: Citizen journey
 
   @javascript
   Scenario: I want to change property details via the check your answers page
-    Given I complete the citizen journey as far as check your answers
+    Given I have completed an application
+    And I complete the citizen journey as far as check your answers
```
reggieb: Done in two steps so that subsequent steps can modify the created application before stepping into the page.

PR #344, opened 2019-03-13
AP-398 Hide Menu button for Citizens

Obsiye:
[link to ticket](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-398)

I created an application helper method for the menu button ( similar to the existing home_link)
This is because in that method, it returns nothing if there's nothing in the user_header_link
(which should only be visible for providers).

## Checklist

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -19,6 +19,12 @@ module ApplicationHelper
     t ".#{[*controller, lazy_t].join('.')}", *args
   end
 
+  def menu_button
+    button_tag(t('generic.menu'),
+               type: 'button', role: 'button', class: 'govuk-header__menu-button js-header-toggle',
+               aria: { controls: 'navigation', label: t('generic.toggle_navigation') })
+  end
+
   def back_link(text: t('generic.back'), path: back_path, method: nil)
     return unless path
 
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -18,7 +18,7 @@
     </div>
     <div class="govuk-header__content">
       <%= home_link %>
-      <button type="button" role="button" class="govuk-header__menu-button js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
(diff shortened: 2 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -19,6 +19,14 @@ def controller_t(lazy_t, *args)
     t ".#{[*controller, lazy_t].join('.')}", *args
   end
 
+  def menu_button
+    return if user_header_link.nil?
```
reggieb: I think the decision to show the menu button belongs in the view. So I'd suggest removing this link and in the view doing:
```erb
<%= menu_button if user_header_link.nil? %>
```

app/helpers/application_helper.rb
```diff
@@ -19,6 +19,14 @@ def controller_t(lazy_t, *args)
...
+
+    button_tag(t('generic.menu'),
+               type: 'button', role: 'button', class: 'govuk-header__menu-button js-header-toggle',
+               aria: { controls: 'navigation', label: 'Show or hide Top Level Navigation' })
```
reggieb: `'Show or hide Top Level Navigation'` should be in a translation file.

app/helpers/application_helper.rb
Obsiye: Thanks 👍 

I did <%= menu_button unless user_header_link.nil? %> because it only shows for providers

Kalagan: would `<%= menu_button unless user_header_link %>` work?

reggieb: Personally - I prefer `unless user_header_link` but some devs coming from other code languages struggle with that bit of Ruby sugar and prefer the more explicit `.nil?` check. So I'm happy with either.

Kalagan: Abdi, I see you just merged this PR with `<%= menu_button unless user_header_link.nil? %>`
Which is the opposite of the suggestion from Rob and I
If `unless user_header_link.nil?` is the right condition, then `if user_header_link` would be simpler

PR #345, opened 2019-03-13
Ap 426 remove council tax

mpw5:
## What

[AP-426](https://dsdmoj.atlassian.net/browse/AP-426)

Changes to the 'identify types of outgoing' and 'outgoings summary' screens:

- The main change is the removal of council tax as a category (I've added an `archived_at` column to the `transaction_types` table in the database to facilitate this)

- There are also a couple of minor wording changes and the addition of a hint

## Checklist

### DIFF
--- a/app/models/transaction_type.rb
+++ b/app/models/transaction_type.rb
@@ -14,15 +14,15 @@ class TransactionType < ApplicationRecord
     ],
     debit: %i[
       rent_or_mortgage
-      council_tax
       child_care
       maintenance_out
       legal_aid
     ]
   }.freeze
 
-  scope :debits, -> { where(operation: :debit) }
-  scope :credits, -> { where(operation: :credit) }
+  scope :active, -> { where(archived_at: nil) }
+  scope :debits, -> { active.where(operation: :debit) }
+  scope :credits, -> { active.where(operation: :credit) }
 
   def self.populate
     populate_records
@@ -36,6 +36,8 @@ class TransactionType < ApplicationRecord
         transaction_type.update! sort_order: start_number
       end
     end
+
+    TransactionType.active.where.not(name: TransactionType::NAMES.values.flatten).update(archived_at: Time.now)
   end
 
   def label_name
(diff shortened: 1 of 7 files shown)


db/data/20190313120752_delete_council_tax_from_transaction_types.rb
```diff
@@ -0,0 +1,9 @@
+class DeleteCouncilTaxFromTransactionTypes < ActiveRecord::Migration[5.2]
+  def up
+    TransactionType.find_by(name: 'council_tax')&.destroy
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
reggieb: Why is this irreversible? If the up destorys, then all the down has to do is create:
```
def down
  TransactionType.create!(name: 'council_tax', operation: 'debit')
  TransactionType.populate_records # to update sorting.
end
```

Kalagan:
I also agree with Rob's feedback.
However, I'm not keen on the way his suggestion for the `populate` method set all transaction_types to archived before de-archiving the active one.

I would rather only archive the non active one.
With a query like this: 
```ruby
TransactionType.where.not(name: TransactionType::NAMES.values.flatten).update!(archived_at: Time.now)`
```

reggieb:
Add an `active` to that call so that only transaction types that are not already archived are updated
```
TransactionType.active.where.not(name: TransactionType::NAMES.values.flatten).update!(archived_at: Time.now)
```

mpw5:
Thanks @reggieb, @Kalagan. I've removed the original work and added changes based on your suggestions.

app/models/transaction_type.rb
```diff
@@ -36,6 +36,8 @@ def self.populate_records
         transaction_type.update! sort_order: start_number
       end
     end
+
+    TransactionType.active.where.not(name: TransactionType::NAMES.values.flatten).update(archived_at: Time.now)
```
Kalagan: I would be nice to have this covered by unit tests

Kalagan: Something like this:
```ruby
    # spec/models/transaction_type_spec.rb
    context 'when some transactions types are being archived' do
      let!(:old_transaction_type) { create :transaction_type, name: :council_tax }

      it 'archives old transaction_type' do
        ...
      end

      it 'does not archive still existing transaction_types' do
        ...
      end
    end
```

PR #357, opened 2019-03-15
AP-392 Amend merits assessment flow for MVP

AbdirizakObsiye2018:
Amend merits assessment flow for MVP

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-392)

I added a conditional forward step in provider_merits.rb under respondents. This is because the provider will not see client_received_legal_helps and proceedings_before_the_courts for passported and non-passported applicants that have the MVP proceeding type.

Also, only show  client_received_legal_helps and proceedings_before_the_courts headings in the check your merits answers page, if the case is not the MVP proceeding type.

## Checklist


app/models/legal_aid_application.rb
```diff
@@ -105,6 +105,10 @@ def other_assets?
     other_assets_declaration.present? && other_assets_declaration.positive?
   end
 
+  def proceeding_type?(type)
+    proceeding_types.any? { |proceeding_type| proceeding_type[:ccms_matter] == type }
```
Kalagan: proceeding_type is an object, not a hash, so you can use `proceeding_type.ccms_matter` instead of `proceeding_type[:ccms_matter]`

app/services/flow/flows/provider_merits.rb
```diff
@@ -9,7 +9,7 @@ class ProviderMerits < FlowSteps
         },
         respondents: {
           path: ->(application) { urls.providers_legal_aid_application_respondent_path(application) },
-          forward: :client_received_legal_helps,
+          forward: ->(application) { application.proceeding_type?('Domestic Abuse') ? :statement_of_cases : :client_received_legal_helps },
```
Kalagan: Testing the value of a string like that doesn't feel robust.
Looking at the data, it looks like there is also a `ccms_matter_code` column so I suggest we use that instead.
And to write it in a constant. In the `ProceedingType` class would make sense:
```ruby
DOMESTIC_ABUSE_CCMS_MATTER_CODE = 'MINJN'.freeze
```

app/views/providers/check_merits_answers/show.html.erb
```diff
@@ -53,27 +53,29 @@
...
-    <%= check_answer_link(
-          name: :proceedings_currently_before_court,
-          url: providers_legal_aid_application_proceedings_before_the_court_path,
-          question: t('.items.proceedings_currently_before_court'),
-          answer: yes_no(@merits.proceedings_before_the_court)
-        ) %>
-  </dl>
-
-  <div class='govuk-body'><%= simple_format @merits.details_of_proceedings_before_the_court %></div>
+  <% unless @legal_aid_application.domestic_abuse? %>
```
reggieb: As there is quite a lot of markup between here and the `end` declaration, I'd suggest that you put it into a partial, and then make rendering the partial conditional. Something like:
```
<%= render('merits_check_answers', merits: @merits) unless @legal_aid_application.domestic_abuse? %>
```

spec/factories/proceeding_types.rb
```diff
@@ -8,5 +8,9 @@
     ccms_category_law_code { 'Category law code' }
     ccms_matter { 'Matter' }
     ccms_matter_code { 'Matter code' }
+
+    trait :domestic_abuse do
+      ccms_matter_code { 'MINJN' }
```
reggieb: Should you use the constant here? 
```ruby
ccms_matter_code { ProceedingType::DOMESTIC_ABUSE_CCMS_MATTER_CODE }
```

app/views/providers/check_merits_answers/show.html.erb
AbdirizakObsiye2018: changed merits_check_answers to non_domestic_abuse_answers

PR #361, opened 2019-03-18
Add merits start page

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-434)

Create screen at start of merits flow to inform provider of the
information they will need to enter.

- add controller and route
- add view
- amend flows
- add text to locale file
- create spec
- amend feature test

## Checklist

### DIFF
--- a/app/views/providers/start_merits_assessments/show.html.erb
+++ b/app/views/providers/start_merits_assessments/show.html.erb
@@ -1,0 +1,13 @@
+<%= page_template page_title: t('.h1-heading') do %>
+  <%= simple_format t('.do_merits_of_case_qualify') %>
+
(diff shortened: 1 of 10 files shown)


app/views/providers/start_merits_assessments/show.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= page_template page_title: t('.h1-heading') do %>
+  <%= simple_format t('.text') %>
```
reggieb: Can you please use meaningful names for your translation links. It makes the code a lot more readable, and content easier to identify. Something like `:do_merits_qualify_case` here for example.

mpw5: I've updated both the state machine and the locales file.

PR #362, opened 2019-03-18
AP 412 show full date range of transactions

reggieb:
[Jira AP-412](https://dsdmoj.atlassian.net/browse/AP-412)

Records and uses a specific time period for the bank transactions pulled from True Layer.

The start and finish times are now calculated by and stored on the Legal Aid Application. To accommodate this I've modified the `BankDataImportService` so that the application is passed to it rather than the applicant.

I have also refactored the process so that the Omniauth controller now just authenticates and gathers the authentication cert and timeout. These are stored against the applicant, and are stored in a SecureData object.

The role of triggering the request to gather the account data is now taken by the accounts controller.


app/controllers/citizens/accounts_controller.rb
```diff
@@ -4,9 +4,14 @@ class AccountsController < BaseController
     before_action :authenticate_applicant!
 
     def index
-      return if worker_working?
+      return if !legal_aid_application.transactions_gathered? && worker_working?
```
Kalagan: Does it mean the scenario in which the applicant adds several banks doesn't work anymore?

Kalagan: instead, you could set `session[:worker_id]` in the previous controller and remove `worker_id` from the session in this controller if the job is done

```ruby
session[:worker_id] = nil unless worker_working?
```

app/models/applicant.rb
```diff
@@ -21,4 +22,21 @@ def email_address
   def full_name
     "#{first_name} #{last_name}".strip
   end
+
+  def store_true_layer_token(token:, expires:)
+    data = { token: token, expires: expires }
+    update!(true_layer_secure_data_id: SecureData.create_and_store!(data))
```
Kalagan: Do we need to use `SecureData` for that?
At the moment, `BankDataImportService` saves these in clear in `BankProvider`

app/models/legal_aid_application.rb
```diff
@@ -44,6 +43,13 @@ class LegalAidApplication < ApplicationRecord
     _prefix: true
   )
 
+  def bank_transactions
+    set_transaction_period unless transaction_period_start_at? && transaction_period_finish_at?
+    applicant.bank_transactions.where(
+      happened_at: transaction_period_start_at..transaction_period_finish_at
```
Kalagan: I think this filtering could be left out. And we could trust TrueLayer that they returned us the period we asked

app/workers/import_bank_data_worker.rb
```diff
@@ -3,12 +3,17 @@ class ImportBankDataWorker
...
-      applicant: Applicant.find(applicant_id),
-      token: token,
-      token_expires_at: token_expires_at
+      legal_aid_application: legal_aid_application
     )
     store errors: command.errors.to_a.flatten.to_json unless command.success?
   end
+
+  def legal_aid_application
+    @legal_aid_application ||= LegalAidApplication.find(legal_aid_application_id)
```
Kalagan: Unless `legal_aid_application_id` is called from outside this class, this and the `attr_reader` is a bit useless in my opinion. Is this just to avoid using a variable in the `perform` method?
Is the `@legal_aid_application ||=` memoization pattern useful here?

app/controllers/citizens/accounts_controller.rb
reggieb: Or I could clear this flag in the "do you have another bank account" action

app/models/applicant.rb
reggieb: Those tokens allow you to access someones bank account details. I think we should make it difficult for someone to just grab tokens from the database and start trawling the API for bank data.

app/models/legal_aid_application.rb
reggieb: You're assuming the applicant won't have multiple applications. I thought the filtering could be added at minimal cost, and would prevent inconsistencies.

app/controllers/concerns/backable.rb
```diff
@@ -18,8 +28,14 @@ def back_path
...
 
     def update_page_history
+      return if self.class.skip_back_history_actions.include?(action_name.to_sym)
       return unless request.request_method_symbol == :get
 
+      process_history
+      remove_old_history
+    end
+
+    def process_history
```
reggieb: Split out this process logic because rubocop was complaining about the length of the `update_page_history` method

app/controllers/citizens/accounts_controller.rb
reggieb: For the record: This issues was addresses by splitting the accounts action into two - one to gather the bank transactions, and another to display them.

app/models/applicant.rb
reggieb: I've modified `BankProvider` so that it stores the id of the applicant's `SecureData`. That will stop people being able to access users data from keys pulled direct from a database dump, and still allow the keys to be retrieved in the future if that is needed.

app/views/citizens/income_summary/index.html.erb
```diff
@@ -1,6 +1,12 @@
 <%= page_template page_title: t('.page_heading'), template: :form do %>
 
-  <p class="govuk-body"><%= t('.add_all_income') %></p>
+  <h2 class="govuk-heading-m">
+    <%= t(
+          '.add_all_income',
+          start_at: l(@legal_aid_application.transaction_period_start_at.to_date, format: :long_date),
+          finish_at: l(@legal_aid_application.transaction_period_start_at.to_date, format: :long_date)
```
mpw5: This feedback's a bit late, sorry! I'm only looking at this because there's a similar change to be made to the outgoing payments screen and wanted to get an idea how to approach it, but I spotted a small bug.

This line is displaying the `transaction_period_start_at` date when it should be the `transaction_period_finish_at` date, ie:

```finish_at: l(@legal_aid_application.transaction_period_finish_at.to_date, format: :long_date)```

reggieb: Good spot - thank you

PR #366, opened 2019-03-18
AP-388 Extract Sidekiq to a dedicated pod

Kalagan:
## What

[AP-388](https://dsdmoj.atlassian.net/browse/AP-388)

- Move Sidekiq to its own pod
  - Rename `deployment.yaml` to `deployment_web.yaml`
  - Add `deployment_worker.yaml`
  - Move all env variable definitions to `_envs.tpl` because they are the same for the web app and the worker
  - Add the `sidekiq_alive` gem so we can have k8s probes for the worker
- A few other small refactors
  - remove `replicaCount` params from the commands run by CircleCI to use the ones in the value files instead
  - Kerin advised that the `values*.yaml` files should only be used for values that are different depending on the environment. So I removed a few of them

## Checklist


helm_deploy/apply-for-legal-aid/templates/deployment_web.yaml
```diff
@@ -0,0 +1,88 @@
...
+          ports:
+            - name: clamav
+              containerPort: 3310
+              protocol: TCP
+          resources:
+{{ toYaml .Values.clamav.resources | indent 12 }}
+        - name: metrics
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          command: ['sh', '-c', "bundle exec prometheus_exporter"]
```
reggieb: Just out of interest - is there a reason why this `command` can use a string with spaces, where other's seem to have to split the command words. Would this work:
```
command: ['sh', '-c', "bundle", "exec", "prometheus_exporter"]
```
And more interestingly - would this work:
```
command: ['sh -c bundle exec prometheus_exporter']
```
Because if that works lets stop messing around with arrays and just pass the commands in as a single string.

Kalagan: Weird but neither `['sh', '-c', 'bundle', 'exec', 'prometheus_exporter']` or `['sh -c bundle exec prometheus_exporter']` works. It needs to be `['sh', '-c', 'bundle exec prometheus_exporter']`

reggieb: Weird in deed. Thank you for trying.

PR #369, opened 2019-03-19
AP-428 Update client details page inputs and text styling

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-428)

Created a helper for input labels in govuk style and applied it to the relevant fields
amend input widths
Remove unused translations

## Checklist

### DIFF
--- a/app/views/providers/applicants/show.html.erb
+++ b/app/views/providers/applicants/show.html.erb
@@ -13,18 +13,18 @@
       ) do |form| %>
 
     <div class="govuk-!-padding-bottom-2"></div>
-    <%= govuk_bold_label t('.full_name') %>
-    <%= form.govuk_text_field :first_name, class: 'govuk-input govuk-!-width-three-quarters' %>
-    <%= form.govuk_text_field :last_name, class: 'govuk-input govuk-!-width-three-quarters' %>
+    <%= form.govuk_label(:fullname, t('.full_name')) %>
+    <%= form.govuk_text_field :first_name, class: 'govuk-input govuk-!-width-three-quarters', 'aria-label' => 'first name' %>
+    <%= form.govuk_text_field :last_name, class: 'govuk-input govuk-!-width-three-quarters', 'aria-label' => 'last name' %>
 
     <div class="govuk-!-padding-bottom-4"></div>
 
     <%= date_input_fields prefix: :dob, field_name: :date_of_birth, form: form %>
 
     <div class="govuk-!-padding-bottom-6"></div>
-    <%= govuk_bold_label t('.nino_label') %>
+    <%= form.govuk_label(:national_insurance_number, t('.nino_label')) %>
     <%= form.govuk_text_field :national_insurance_number, label: nil, class: 'govuk-input govuk-input--width-10' %>
-    <%= govuk_bold_label t('.email_label') %>
+    <%= form.govuk_label(:email, t('.email_label')) %>
     <%= form.govuk_text_field :email, label: nil, class: 'govuk-input govuk-!-width-three-quarters' %>
 
     <div class="govuk-!-padding-bottom-2"></div>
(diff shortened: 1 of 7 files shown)


app/views/providers/applicants/show.html.erb
```diff
@@ -13,18 +13,19 @@
...
 
     <div class="govuk-!-padding-bottom-4"></div>
 
     <%= date_input_fields prefix: :dob, field_name: :date_of_birth, form: form %>
 
     <div class="govuk-!-padding-bottom-6"></div>
-
-    <%= form.govuk_text_field :national_insurance_number, class: 'govuk-!-width-one-half' %>
-    <%= form.govuk_text_field :email, class: 'govuk-!-width-three-quarters' %>
+    <%= govuk_bold_label t('.nino_label') %>
```
reggieb: The field the label is being applied to is `:national_insurance_number`. So:
```erb
<%= govuk_label(:national_insurance_number, t('.nino_label')) %>
```

app/views/providers/applicants/show.html.erb
```diff
@@ -13,18 +13,19 @@
...
+    <%= form.govuk_text_field :national_insurance_number, label: nil, class: 'govuk-input govuk-input--width-10' %>
+    <%= govuk_bold_label t('.email_label') %>
```
reggieb: Again, the field is `:email` so:
```erb
<%= govuk_label(:email, t('.email_label')) %>
```
I'll assume you can modify the others the same way

app/views/providers/applicants/show.html.erb
```diff
@@ -13,18 +13,19 @@
       ) do |form| %>
 
     <div class="govuk-!-padding-bottom-2"></div>
-
-    <%= form.govuk_text_field :first_name, class: 'govuk-!-width-one-half' %>
-    <%= form.govuk_text_field :last_name, class: 'govuk-!-width-one-half' %>
+    <%= govuk_bold_label t('.full_name') %>
```
reggieb: OK - this is a difficult one and you may have to do something strange here. Basically, you have one label for two fields. Looking at [this Stackoverflow answer](https://stackoverflow.com/a/31274644), I think the best option would be to use a span and add `arel-label` entries for the inputs:
```
<%= content_tag :span, t('.full_name'), class: 'govuk-fieldset__legend govuk-fieldset__legend--m') %>
<%= form.govuk_text_field :first_name, class: 'govuk-input govuk-!-width-three-quarters', arel_label: :first_name %> 
<%= form.govuk_text_field :last_name, class: 'govuk-input govuk-!-width-three-quarters', arel_label: :last_name: :name %>
```
I think the `arel_label` options will be convert to `arel-label` in the output but check and if not you'll need to use the `'arel-label' => :last_name` form.

reggieb: Also this may be a simpler style to use for the first line if you prefer:
```html
<span class="govuk-fieldset__legend govuk-fieldset__legend--m"><%= t('.full_name') %></span>
```

app/views/providers/applicants/show.html.erb
```diff
@@ -13,18 +13,19 @@
       ) do |form| %>
 
     <div class="govuk-!-padding-bottom-2"></div>
-
-    <%= form.govuk_text_field :first_name, class: 'govuk-!-width-one-half' %>
-    <%= form.govuk_text_field :last_name, class: 'govuk-!-width-one-half' %>
+    <%= form.govuk_label(:fullname, t('.full_name')) %>
```
reggieb: As I said before, this is problematic. Have a look at the HTML `label` docs:
> The `<label>` tag defines a label for a `<button>`, `<input>`, `<meter>`, `<output>`, `<progress>`, `<select>`, or `<textarea>` element.
In other words the label has to match an input element. There is no `full_name` input element. Full name is entered via TWO fields `first_name` and `last_name`. So you need to treat this as just information provided to the user - almost a heading - that has been styled to look like a label. Therefore I'd suggest:
```
<span class="govuk-fieldset__legend govuk-fieldset__legend--m"><%= t('.full_name') %></span>
```

PR #370, opened 2019-03-20
Ap 386 remove transactions without reload

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-386)

Use javascript to allow a user to remove bank transactions that they have identified as income without refreshing the screen.

- Adds 'remote: true' to the partial that generates the 'Remove' link to so that an ajax request is send when clicked.
- Adds a javascript file to handle the ajax response and remove the appropriate row of the table.
- Corrects a typo in 'remove_transaction_type'.

## Checklist

### DIFF
--- a/app/assets/javascripts/bank_transactions.es6
+++ b/app/assets/javascripts/bank_transactions.es6
@@ -1,0 +1,13 @@
+$(document).ready(function(){
+    document.body.addEventListener('ajax:success', function(event) {
+      var row = $(event.srcElement).parents("tr");
+      row.css("height",row.height()); //we fix the row height to stop it collapsing straight away
+      row.children("td").css("display","none"); //we remove the row innards to allow the row to collapse to zero height
+      row.slideUp(); //invoke collapse
+    })
+
+    document.body.addEventListener('ajax:error', function(event){
+      console.log("Remove failed for unknown reason.  Page reloading.") //placeholder for any error message
+      location.reload(); //reloads page - because if it has failed, the row might've been removed behind the scenes
+    });
+});
--- a/app/assets/javascripts/bank_transactions.es6
+++ b/app/assets/javascripts/bank_transactions.es6
@@ -5,6 +5,12 @@ $(document).ready(function(){
       row.css("height",row.height()); //we fix the row height to stop it collapsing straight away
       row.children("td").css("display","none"); //we remove the row innards to allow the row to collapse to zero height
       row.slideUp(); //invoke collapse
+
+      if (!row.siblings("tr").filter(function() {return $(this).css('display') !== 'none';}).length) { //there are no remaining siblings - all rows have been removed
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/bank_transactions.es6
```diff
@@ -0,0 +1,13 @@
+$(document).ready(function(){
+    document.body.addEventListener('ajax:success', function(event) {
```
Kalagan: This event will be triggered for all `'ajax:success'` events. I suggest we restrict to only the "remove transactions" events by adding a condition like so:
```javascript
document.body.addEventListener('ajax:success', function(event) {
  if (!event.srcElement.classList.contains('remote-remove-transaction')) return
```
same for the `'ajax:error'` event.

And add that class to the form generated by `button_to` in the view with the `form_class` attribute : 
```ruby
<%= button_to(
      t('.remove_link'),
      remove_transaction_type_citizens_bank_transaction_path(bank_transaction),
      method: :patch,
      class: 'button-as-link',
      form_class: 'remote-remove-transaction',
      remote: true
    ) %>
```

app/controllers/citizens/bank_transactions_controller.rb
```diff
@@ -3,7 +3,7 @@ class BankTransactionsController < ApplicationController
     include ApplicationFromSession
     before_action :authenticate_applicant!
 
-    def remove_transation_type
+    def remove_transaction_type
       bank_transaction.update! transaction_type: nil
       respond_to do |format|
         format.html do
```
Kalagan: This code doesn't seem to work correctly.
The response of the AJAX call is the HTML one, not the JSON :
image

Replacing the `format.json` with `format.js` seems to fix the problem

Kalagan: You would have to change the specs like so:
image

app/assets/javascripts/bank_transactions.es6
mpw5: Doh. Of course, I should have realised that. Thanks.

app/controllers/citizens/bank_transactions_controller.rb
mpw5: 👍

spec/requests/citizens/bank_transactions_spec.rb
```diff
@@ -32,17 +32,23 @@
...
-    context 'with JSON request' do
-      let(:headers) { { 'ACCEPT' => 'application/json' } }
+    context 'with ajax request' do
+      subject do
+        patch(
+          remove_transaction_type_citizens_bank_transaction_path(bank_transaction),
+          headers: headers,
+          xhr: true
+        )
+      end
```
Kalagan: Another to way to implement this:
image

I think it's a bit nicer as it's more obvious that the only difference is that the request is `XHR`

PR #373, opened 2019-03-21
AP-436 Use sample bank transactions instead of TrueLayers' data

Kalagan:
## What

[AP-436](https://dsdmoj.atlassian.net/browse/AP-436)

- Added a settings page
- Added a link to it in the admin page
- Added a "Mock TrueLayer Data" to the settings page.
- Modify the TrueLayer API client so if the flag above is turned on, the sample data is used.

Settings page:
image

## Checklist

### DIFF
--- a/app/models/setting.rb
+++ b/app/models/setting.rb
@@ -1,0 +1,9 @@
+class Setting < ApplicationRecord
+  def self.mock_true_layer_data?
+    setting.mock_true_layer_data?
+  end
+
+  def self.setting
+    Setting.first || Setting.create!
+  end
+end
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -1,0 +1,29 @@
+module Admin
+  class SettingsController < ApplicationController
+    before_action :authenticate_admin_user!
+
+    def show
+      @form = Settings::SettingForm.new(model: setting)
+    end
+
+    def update
+      @form = Settings::SettingForm.new(form_params.merge(model: setting))
+
+      if @form.save
+        redirect_to admin_settings_path, notice: 'Settings have been updated'
+      else
+        render :show
+      end
+    end
+
+    private
+
+    def form_params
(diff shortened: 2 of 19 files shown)


app/models/setting.rb
```diff
@@ -0,0 +1,9 @@
+class Setting < ApplicationRecord
+  def self.mock_true_layer_data?
+    setting.mock_true_layer_data?
+  end
+
+  def self.setting
+    Setting.first || Setting.new.tap(&:save!)
```
reggieb: Would `Setting.create!` work here rather than `Setting.new.tap(&:save!)`

app/services/true_layer/api_client.rb
```diff
@@ -16,10 +16,14 @@ def account_holders
     end
 
     def accounts
+      return SimpleResult.new(value: SampleData::ACCOUNTS) if mock_true_layer_data?
```
reggieb: I wonder if it might be worth separating the mock behaviour into a separate object. Something like this:
```ruby
module TrueLayer
  class ApiClientMock < ApiClient
    def accounts
      SimpleResult.new(value: SampleData::ACCOUNTS)
    end

    def transactions(account_id, date_from, date_to)
      SimpleResult.new(value: SampleData::TRANSACTIONS)
    end

    def account_balance(account_id)
      SimpleResult.new(value: SampleData::BALANCES) if mock_true_layer_data?
    end
  end
end
```
Then in `TrueLayer::BankDataImportService`:
```ruby
    def api_client_class
      return ApiClientMock if Setting.mock_true_layer_data?
      ApiClient
    end

    def api_client
      @api_client ||= api_client_class.new(token)
    end
```

app/models/setting.rb
Kalagan: Yes. Somehow I thought `create!` returned a boolean

app/services/true_layer/api_client.rb
Kalagan: Great idea

PR #374, opened 2019-03-21
AP-296 Check Your Answer is Read Only for submitted applicant

AbdirizakObsiye2018:
Ap 296 Check Your Answers page is Read Only for applications that are provider_submitted.

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-296)

I created read_only variable in the controller which is true when the state is provider_submitted. This means that the check your answers page for provider can be read only on the condition that the application is submitted.

Also, When a provider clicks link 'View' on an application that has state provider_submitted, they'll be directed to the read only version of the check your answers page. 

Created a helper method for getting the html value of a button.

Updated the provider_step when the provider submits an application and is on the application_confirmation page, so when they view the application from the providers/applicants page, they'll end up on the check your answers and not the application_confirmation page. 

## Checklist

### DIFF
--- a/app/controllers/providers/application_confirmations_controller.rb
+++ b/app/controllers/providers/application_confirmations_controller.rb
@@ -1,6 +1,10 @@
 module Providers
   class ApplicationConfirmationsController < ProviderBaseController
-    def show
+    def show; end
+
+    private
+
+    def set_legal_aid_application
       legal_aid_application.update!(provider_step: :check_provider_answers)
     end
   end
--- a/app/controllers/providers/application_confirmations_controller.rb
+++ b/app/controllers/providers/application_confirmations_controller.rb
@@ -1,5 +1,7 @@
 module Providers
   class ApplicationConfirmationsController < ProviderBaseController
-    def show; end
+    def show
+      legal_aid_application.update!(provider_step: :check_provider_answers)
+    end
   end
 end
--- a/app/views/providers/check_provider_answers/index.html.erb
+++ b/app/views/providers/check_provider_answers/index.html.erb
@@ -1,11 +1,11 @@
 <%= page_template(
-      page_title: t('.title'),
+      page_title: @read_only ? '' : t('.title'),
       back_link: {
-        path: reset_providers_legal_aid_application_check_provider_answers_path,
-        method: :post
+        path: @read_only ? providers_legal_aid_applications_path : reset_providers_legal_aid_application_check_provider_answers_path,
+        text: @read_only ? t('generic.home') : t('generic.back'),
+        method: @read_only ? :get : :post
       }
     ) do %>
-
   <h2 class="govuk-heading-m"><%= t '.section_client.heading' %></h2>
 
   <dl class="govuk-summary-list govuk-!-margin-bottom-9">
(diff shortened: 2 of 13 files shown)


app/controllers/providers/application_confirmations_controller.rb
```diff
@@ -1,5 +1,7 @@
 module Providers
   class ApplicationConfirmationsController < ProviderBaseController
-    def show; end
+    def show
+      legal_aid_application.update!(provider_step: :check_provider_answers)
```
reggieb: The provider step is updated as a before action (see `ApplicationDependable`, which is included into `ProviderBaseController`). So with your code here `provider_step` will be written to twice in this action. A better solution would be to overwrite the `set_legal_aid_application` method in this controller:
```ruby
def show; end

private

def set_legal_aid_application
  legal_aid_application.update!(provider_step: :check_provider_answers)
end
```

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -1,11 +1,11 @@
 <%= page_template(
-      page_title: t('.title'),
+      page_title: @read_only ? '' : t('.title'),
```
reggieb: Are you sure this page should have no title in read only? Should there be a read only version of the page title?

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -1,11 +1,11 @@
...
       back_link: {
-        path: reset_providers_legal_aid_application_check_provider_answers_path,
-        method: :post
+        path: @read_only ? providers_legal_aid_applications_path : reset_providers_legal_aid_application_check_provider_answers_path,
+        text: @read_only ? t('generic.home') : t('generic.back'),
+        method: @read_only ? :get : :post
```
reggieb: I think this would be clearer:
```erb
<%
  read_only_back_link = {
      path: providers_legal_aid_applications_path,
      text: t('generic.home'),
      method: :get
    }
  back_link = {
      path: reset_providers_legal_aid_application_check_provider_answers_path,
      text: t('generic.back'),
      method: :post
    }  
%>
<%= page_template(
      page_title: @read_only ? '' : t('.title'),
      back_link: @read_only ? read_only_back_link : back_link
  ) %>

File: app/views/providers/check_provider_answers/index.html.erb [hunk 4]
```diff
@@ -61,15 +67,16 @@
...
-  <div class="govuk-!-padding-bottom-9"></div>
 
   <%= next_action_buttons_with_form(
-        url: continue_providers_legal_aid_application_check_provider_answers_path,
-        method: :patch,
-        show_draft: true
+        url: @read_only ? providers_legal_aid_applications_path : continue_providers_legal_aid_application_check_provider_answers_path,
+        method: @read_only ? :get : :patch,
+        show_draft: @read_only ? false : true,
+        continue_button_text: @read_only ? t('generic.back_to_your_applications') : t('generic.continue')
```
reggieb: Again - rather than making every option react to the same tertiary operator, it would be clearer to separate the options into two versions. So:
```erb
<%
  next_button_options = { .... }
  read_only_next_button_options = { ... }
%>
<%= next_action_buttons_with_form(
  @read_only ? read_only_next_button_options : next_button_options
) %>

File: app/views/shared/page_templates/_default_page_template.html.erb [hunk 5]
```diff
@@ -8,7 +8,9 @@
         <%= content_for(:page_title) %>
       </h3>
     <% else %>
-      <h1 class="govuk-heading-xl"><%= content_for(:page_title) %></h1>
+      <% if content_for(:page_title) %>
```
reggieb: Use `content_for?` (with a `?`) for conditionals:
```erb
<% if content_for?(:page_title) %>
```

features/providers/step_definitions/civil_journey_steps.rb
```diff
@@ -298,6 +330,20 @@
   expect(page).to have_content('Other assets')
 end
 
+Then('I am on the read only version of the check your answers page') do
+  expect(page).to have_content('Home')
+  expect(page).not_to have_content('change')
+  expect(page).not_to have_content('Continue')
+  expect(page).not_to have_content('Back')
+  expect(page).not_to have_content('Check your answers')
```
reggieb: This feels link you are testing how the page is behaving rather than you checking that this is the right page. By adding so many checks, this makes the condition much more fragile (a content change to any of those text entries could cause a failure).

If you need to test that when the page is in Read-Only mode, that the page displays the context as you have set out here, then do that in a request spec and not in the cucumber feature.

app/controllers/providers/application_confirmations_controller.rb
AbdirizakObsiye2018: Thanks, Implementing it now. 👍

app/views/providers/check_provider_answers/index.html.erb
AbdirizakObsiye2018: That makes a lot more sense, thanks. 👍

app/views/providers/check_provider_answers/index.html.erb
AbdirizakObsiye2018: Spoke to the designer and they said it would stay blank because they haven't come up with a title yet and they didn't want to block this ticket. There's also having different heading styles which will be designed later.

app/views/providers/check_provider_answers/index.html.erb
AbdirizakObsiye2018: 👍

features/providers/step_definitions/civil_journey_steps.rb
AbdirizakObsiye2018: Changed to the below because the read only page currently has no page title. therefore, I needed to find something within the page to prove that i've landed on the correct page.

Added:
  expect(page).to have_content('Home')
  expect(page).not_to have_css('.change-link')

PR #376, opened 2019-03-22
AP-392 Merits-Assessment-Flow

AbdirizakObsiye2018:
Merits-Assessment-Flow is correct to current requirements

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-392)

- Skip over two pages in the provider merits flow for passported and non-passported applications. This is because these pages are not needed now but could be used in the future for certain proceeding types. The pages are  proceedings_before_the_courts  and client_received_legal_helps.

- Remove the mentions of client has not received legal help and proceedings before the courts in the check merits answers page.

- Created tests for Merits-Assessment-Flow to check that the flow is correct and updated tests to not include proceedings_before_the_courts  and client_received_legal_helps pages.

## Checklist

### DIFF
--- a/app/services/flow/flows/provider_merits.rb
+++ b/app/services/flow/flows/provider_merits.rb
@@ -14,11 +14,11 @@ module Flow
         },
         respondents: {
           path: ->(application) { urls.providers_legal_aid_application_respondent_path(application) },
-          forward: :client_received_legal_helps,
+          forward: :statement_of_cases,
           check_answers: :check_merits_answers
         },
         client_received_legal_helps: {
-          path: ->(application) { urls.providers_legal_aid_application_client_received_legal_help_path(application) },
+          # path: ->(application) { urls.providers_legal_aid_application_client_received_legal_help_path(application) },
           forward: :proceedings_before_the_courts,
           check_answers: :check_merits_answers
         },
--- a/app/services/flow/flows/provider_merits.rb
+++ b/app/services/flow/flows/provider_merits.rb
@@ -18,7 +18,10 @@ module Flow
           check_answers: :check_merits_answers
         },
         client_received_legal_helps: {
-          # path: ->(application) { urls.providers_legal_aid_application_client_received_legal_help_path(application) },
+          # TODO: Should be removed when this page is being used.
+          # :nocov:
+          path: ->(application) { urls.providers_legal_aid_application_client_received_legal_help_path(application) },
+          # :nocov:
           forward: :proceedings_before_the_courts,
           check_answers: :check_merits_answers
         },
--- a/features/providers/civil_application_journey.feature
+++ b/features/providers/civil_application_journey.feature
@@ -439,7 +439,6 @@ Feature: Civil application journeys
     Then I fill "Bail conditions set details" with "Foo bar"
     Then I click "Continue"
     And I should not see "Client received legal help"
-    And I should not see "Proceedings currently before court"
     Then I should be on a page showing "Statement of case"
     Then I fill "Statement" with "Statement of case"
     Then I upload a pdf file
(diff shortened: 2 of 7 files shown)


features/providers/civil_application_journey.feature
```diff
@@ -404,14 +438,8 @@ Feature: Civil application journeys
...
-    Then I should be on a page showing "Has your client received legal help for the matter?"
-    Then I choose "No"
-    Then I fill "Application purpose" with "because I can't afford it"
-    Then I click "Continue"
-    Then I should be on a page showing "Are the proceedings, for which funding is being sought, currently, before the court?"
-    Then I choose "Yes"
-    Then I fill "Details of proceedings before the court" with "The case is already in the court"
-    Then I click "Continue"
+    And I should not see "Client received legal help"
+    And I should not see "Proceedings currently before court"
```
reggieb: Do we really need two checks here that we are not on a particular page - surely one would be fine.

spec/requests/providers/respondents_spec.rb
```diff
@@ -81,7 +81,7 @@
 
     it 'redirects to the next page' do
       subject
-      expect(response).to redirect_to(providers_legal_aid_application_client_received_legal_help_path)
+      expect(response).to redirect_to(providers_legal_aid_application_statement_of_case_path)
```
reggieb: You could use `flow_forward_path` here:
```ruby
expect(response).to redirect_to(flow_forward_path)
```

app/services/flow/flows/provider_merits.rb
```diff
@@ -14,11 +14,11 @@ class ProviderMerits < FlowSteps
         },
         respondents: {
           path: ->(application) { urls.providers_legal_aid_application_respondent_path(application) },
-          forward: :client_received_legal_helps,
+          forward: :statement_of_cases,
           check_answers: :check_merits_answers
         },
         client_received_legal_helps: {
-          path: ->(application) { urls.providers_legal_aid_application_client_received_legal_help_path(application) },
+          # path: ->(application) { urls.providers_legal_aid_application_client_received_legal_help_path(application) },
```
reggieb: An entry in flows should always have a path. I don't think this should be commented out.

AbdirizakObsiye2018: I had to comment it out because the code coverage was less than 100% for this. There's no rspec test for this because that page is not used within the app, however, it may be used in the future ( after speaking to the interaction designer).

reggieb: In general, it not a good idea to comment sections of codes out, because they may be used later. It seems like a good idea, but almost always just leads to lots of commented code being left if the code base.

If you are absolutely sure this needs to be commented out - then comment out the whole lot (`client_received_legal_helps` key and its value) - just the `path` part of the value.

Kalagan: You could exclude this line from the code coverage report: https://github.com/colszowka/simplecov#ignoringskipping-code

app/services/flow/flows/provider_merits.rb
```diff
@@ -14,11 +14,13 @@ class ProviderMerits < FlowSteps
         },
         respondents: {
           path: ->(application) { urls.providers_legal_aid_application_respondent_path(application) },
-          forward: :client_received_legal_helps,
+          forward: :statement_of_cases,
           check_answers: :check_merits_answers
         },
         client_received_legal_helps: {
+          # :nocov:
```
reggieb: Sorry to be a pain - but could you also add a `TODO:` comment stating when the `nocov` should be removed.

AbdirizakObsiye2018: No problemo 👍

PR #377, opened 2019-03-22
AP-413 Add styling to wrap text on provider profile

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-413)

The 'role' field on the provider profile page can contain a long list of roles which does not currently wrap.

This commit adds a css class to force the text to wrap. It also changes the way the comma-separated list of roles is displayed by adding a space after each comma, which will prevent the list wrapping mid-word.

## Checklist

### DIFF
--- a/app/views/providers/providers/show.html.erb
+++ b/app/views/providers/providers/show.html.erb
@@ -14,7 +14,7 @@
       </tr>
       <tr class="govuk-table__row">
         <th class="govuk-table__header" scope="row"><%= t('.role') %></th>
-        <td class="govuk-table__cell"><%= @provider.roles %></td>
+        <td class="govuk-table__cell wrap-text"><%= @provider.roles&.gsub(/,(?=\S)/, ', ') %></td>
       </tr>
       <tr class="govuk-table__row">
         <th class="govuk-table__header" scope="row"><%= t('.account_number') %></th>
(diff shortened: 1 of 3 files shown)


spec/requests/providers/providers_spec.rb
```diff
@@ -19,5 +19,6 @@
 
   it 'displays provider data' do
     expect(response.body).to include(provider.username)
+    expect(response.body).to include(provider.roles.gsub(/,(?![ ])/, ', '))
```
reggieb: I don't think it's a good idea to test something with exactly the same code as being implemented. I'd suggest:
```rspec
context 'with unspaced roles' do
  let(:provider) { create :provider, roles: 'foo,bar' }
  
  it 'should display roles with spaces do
    expect(response.body).to include('foo, bar')
  end
end
```

app/views/providers/providers/show.html.erb
```diff
@@ -14,7 +14,7 @@
       </tr>
       <tr class="govuk-table__row">
         <th class="govuk-table__header" scope="row"><%= t('.role') %></th>
-        <td class="govuk-table__cell"><%= @provider.roles %></td>
+        <td class="govuk-table__cell wrap-text"><%= @provider.roles.gsub(/,(?![ ])/, ', ') %></td>
```
reggieb: I'd suggest this match as being clearer: ,(?=\S)
```ruby
@provider.roles.gsub(/,(?=\S)/, ', ')
```
With `[ ]` it's not really clear what's between the brackets. `\S` is more explicit (anything that is not white space)

PR #359, opened 2019-03-15
AP-399 Fix address field width for mobile devices

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-399)

Fixed issue where the address selection field extended beyond the window on mobile devices.
Remove unnecessary divs to match govuk styling requirements

## Checklist

Before you ask people to review this PR:

- [x] Tests and rubocop should be passing: `bundle exec rake`
- [x] Github should not be reporting conflicts; you should have recently run `git rebase master`.
- [x] There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- [x] The PR description should say what you changed and why, with a link to the JIRA story.
- [x] You should have looked at the diff against master and ensured that nothing unexpected is included in your changes.
- [x] You should have checked that the commit messages say why the change was made.

PR #386, opened 2019-03-25
AP-399 Correction of address dropdown width fix

reggieb:
Previous fix included the correct fix, but also introduced a change that broke header styles of other pages

[Jira Ticket AP-399](https://dsdmoj.atlassian.net/browse/AP-399)

### DIFF
--- a/app/views/providers/address_selections/show.html.erb
+++ b/app/views/providers/address_selections/show.html.erb
@@ -1,15 +1,15 @@
-<%= form_with(
-      model: @form,
-      scope: :address_selection,
-      url: providers_legal_aid_application_address_selection_path(@legal_aid_application),
-      method: :patch,
-      local: true
-    ) do |f| %>
+<%= page_template(
+      page_title: t('forms.address_selection.heading'),
+      template: :form
+    ) do %>
 
-  <%= page_template(
-        page_title: t('forms.address_selection.heading'),
-        template: :form
-      ) do %>
+  <%= form_with(
+        model: @form,
+        scope: :address_selection,
+        url: providers_legal_aid_application_address_selection_path(@legal_aid_application),
+        method: :patch,
+        local: true
+      ) do |f| %>
 
     <div class="govuk-!-padding-bottom-8"></div>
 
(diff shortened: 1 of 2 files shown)


app/views/providers/address_selections/show.html.erb
```diff
@@ -1,15 +1,15 @@
...
+      template: :form
+    ) do %>
+
+  <%= form_with(
+        model: @form,
+        scope: :address_selection,
+        url: providers_legal_aid_application_address_selection_path(@legal_aid_application),
+        method: :patch,
+        local: true
+      ) do |f| %>
```
reggieb: This reverses a change made in https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/359

app/views/providers/address_selections/show.html.erb
```diff
@@ -39,7 +39,13 @@
...
       <% end %>
       <% input_error_class = @form.errors[:address].any? ? 'govuk-select--error' : '' %>
-      <%= f.select(:lookup_id, @addresses.collect { |a| [a.full_address, a.lookup_id] }, { include_blank: "#{@addresses.size} addresses found" }, class: "govuk-select govuk-!-width-full #{input_error_class}", id: :address) %>
+      <%= f.select(
+            :lookup_id,
+            @addresses.collect { |a| [a.full_address, a.lookup_id] },
+            { include_blank: "#{@addresses.size} addresses found" },
+            class: "govuk-select govuk-!-width-full #{input_error_class}",
+            id: :address
+          ) %>
```
reggieb: The select code includes the fix: Adding class `govuk-!-width-full`. All that I've changed is to split the select code onto multiple lines so that it's easier to read.

app/views/shared/page_templates/_form_page_template.html.erb
```diff
@@ -1,10 +1,14 @@
...
-  </legend>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+        <h1 class="govuk-fieldset__heading">
+          <%= content_for(:page_title) %>
+        </h1>
+      </legend>
+    </div>
+  </div>
```
reggieb: This corrects the change in https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/359 that broke headings on other pages.

app/views/providers/address_selections/show.html.erb
MoJ-Longbeard: Does this change keep the structure of the HTML that #359 fixed?  
i.e.
`<form> <fieldset> <legend> <h1>`

reggieb: No it reverses that .... but the #359 fix only corrects that on one page. It leaves every other instance of the form page template behaving the same way. 

This issue needs to be raised in a separate ticket and fixing everywhere it occurs -which probably means a rethink of the way the form page template is used.

MoJ-Longbeard: Yes, okay.  That sounds sensible.

PR #388, opened 2019-03-25
AP-451 Handle benefit checker known issue

reggieb:
[Jira AP-451](https://dsdmoj.atlassian.net/browse/AP-451)

Benefit Checker fails with single letter surnames. This is handled here in the controller by:

- detecting the known issue
- creating a `BenefitCheckResult` with the result: "skipped:known_issue" - which registers as a negative result
- redirects to the next step in the flow


app/controllers/providers/check_benefits_controller.rb
```diff
@@ -1,11 +1,24 @@
 module Providers
   class CheckBenefitsController < ProviderBaseController
     def index
+      return set_negative_result_and_go_forward if know_issue_prevents_benefit_check?
```
willc-work: is this supposed to be ```know_issue``` as in 'we know this issue exists' or should it be ```known_issue``` to match up with the 'skipped:known_issue' result ?

app/controllers/providers/check_benefits_controller.rb
```diff
@@ -1,11 +1,24 @@
...
+    private
+
+    def known_issue_prevents_benefit_check?
+      applicant.last_name.length == 1
+    end
+
+    def set_negative_result_and_go_forward
+      legal_aid_application.create_benefit_check_result(result: 'skipped:known_issue')
+      go_forward
+    end
```
Kalagan: The `.create` would fail in the unlikely (but possible) scenario in which `legal_aid_application.benefit_check_result` already exist.
(And would fail silently!)

Would this be a bit better?
```ruby
    def set_negative_result_and_go_forward
      benefit_check_result.update!(result: 'skipped:known_issue')
      go_forward
    end

    def benefit_check_result
      @benefit_check_result ||= legal_aid_application.benefit_check_result || legal_aid_application.build_benefit_check_result
    end
```

reggieb: As `legal_aid_application.benefit_check_result` is a `has_one` relationship, I think in theory all that would happen is the new one would be added to the database along side an existing one. Normally the new one would take over as the one returned by the `has_one`. However, as we're using uuids for the id, it's pot luck which one would become the one attached to the application on reload.

Kalagan: Alright!
ActiveRecord is smarter than I thought. I just assumed it would fail but didn't test...
Thanks for testing and for adding the `!`

PR #392, opened 2019-03-27
AP-463 Setup spectre client

reggieb:
[Jira AP-463](https://dsdmoj.atlassian.net/browse/AP-463)

I've created this PR so that people can see how relatively easy it is to set up cucumber to output screen shots and send them to a [Spectre](https://github.com/wearefriday/spectre) server.

However, I think there are a number of issues with running this on CI:

- The [Spectre app is no longer actively supported](https://github.com/wearefriday/spectre/issues/50) - the main development seems to have stopped three years ago. Things like the auto-creation of an instance on Heroku no longer seem to work. The Ruby version it uses is current fixed at 2.3.3.
- _(see comment below)_ How do you generate the base line set of images. You could perhaps create a rake task that rolled the changes back to master, ran the cucumber tests, then reapplied the changes and ran them again - but things like migrations and gem changes would make that difficult.
- _(see comment below)_  I can't see a way to delete old "projects" from the current web interface. Without that, the store could become very large with a lot of redundant images being stored. We could modify the app - but that means maintaining another app.

I wonder if other people have found issues with using it continually which is why it's no longer actively maintained.

I think it would be quite a handy tool to run locally. It would be easy to set the spectre url and enable/skip the screen shots via an environment variable, and you could manage the stored screen shots manually.

In my testing I ran Spectre locally at `localhost:3100`, and used the following command to just run the cucumber feature that took the snap shots:

```bash
cucumber features/providers/civil_application_journey.feature:11
```

Here are a few screen shots of the results. Note that in the live app, clicking on an image opens it up and allows you to zoom in to a reasonable degree.

**A summary of the runs I've triggered**
Note that the app seems to automatically store just the last 5 runs (in effect the results of last five times that the cucumber command was run in this branch)
localhost_3100_projects_laa_apply_for_legal_aid_application_suites_ap-463-spectre-trial-setup (1)

**Details of a particular run showing a difference**
In this run I removed a `t` from a I18n call in the view
localhost_3100_projects_laa_apply_for_legal_aid_application_suites_ap-463-spectre-trial-setup_runs_10

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -76,6 +76,7 @@ group :development, :test do
   # Available in dev env for generators
   gem 'factory_bot_rails'
   gem 'rspec-rails', '~> 3.8'
+  gem 'spectre_client', git: 'git@github.com:wearefriday/spectre_client.git'
 end
 
 group :development do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,3 +1,10 @@
+GIT
+  remote: git@github.com:wearefriday/spectre_client.git
+  revision: c55325a3a13b9a80c20985a73caae512e3778c3f
+  specs:
+    spectre_client (0.1.87)
+      rest-client (~> 2.0)
+
 GEM
   remote: https://rubygems.org/
   specs:
@@ -152,6 +159,8 @@ GEM
       ruby-saml (~> 1.7)
     diff-lcs (1.3)
     docile (1.3.1)
+    domain_name (0.5.20180417)
+      unf (>= 0.0.5, < 1.0.0)
     dotenv (2.7.1)
     dotenv-rails (2.7.1)
       dotenv (= 2.7.1)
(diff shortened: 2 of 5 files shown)


reggieb:
I've noticed on failure, when you drill into the comparison there is a "Set as baseline" button. That does what you'd expect (and hope) it would. Defines a new baseline - that is, sets the "failure" image as the new base line.

Further - the base line seems to match on the "name", and goes across "Suite names" (which I have set as the current branch name). That is, I did the following:

1. Created a new branch from the current one and switched to that branch
2. ran the cucumber test - that created a new suite - with the base line being that from the other suite (branch).
3. I modified a view and ran the cucumber test again
4. This failed in the second suite. So I then set that image as base line
5. I returned to the original suite - the base line now matched that in the second suite.

In other words we can use "Set as baseline" to manage the base line across suites.

This together with the fact that the app seems to only keep 5 test results per suite may be a solution to my concerns about base lines and image storage.

reggieb:
Following discussion with @stephenrichards it was decided that hosting this solution would add more work that the problem demands. Therefore this is being closed.

PR #396, opened 2019-03-28
AP-462 Changed text of button

MoJ-Longbeard:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-462)

Changed text of submit button to continue button on check_passported_answers

## Checklist

### DIFF
--- a/app/views/providers/check_passported_answers/show.html.erb
+++ b/app/views/providers/check_passported_answers/show.html.erb
@@ -17,6 +17,6 @@
         url: continue_providers_legal_aid_application_check_passported_answers_path(@legal_aid_application),
         method: :patch,
         show_draft: true,
-        continue_button_text: t('generic.submit')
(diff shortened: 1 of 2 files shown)


app/views/providers/check_passported_answers/show.html.erb
```diff
@@ -17,6 +17,6 @@
         url: continue_providers_legal_aid_application_check_passported_answers_path(@legal_aid_application),
         method: :patch,
         show_draft: true,
-        continue_button_text: t('generic.submit')
+        continue_button_text: t('generic.continue')
```
Kalagan: The default value of this parameter is `t('generic.continue')`. As you can see on the method:
https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/helpers/flow_helper.rb#L6
So you can just ommit this parameter

PR #404, opened 2019-03-29
AP-468 HostingEnvironment class to determine where we are running

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-468)

Created a HostingEnvironment class to determine which environment we're running in.

Within this class, there is the .env class method to return the environment name in symbol format.
Also, created methods: test?, development?, uat?, staging? and live? that return true if it's in said environment.
A runtime error is set off if the environment url is unknown.

Lastly, created tests.

## Checklist

### DIFF
--- a/config/initializers/hosting_environment.rb
+++ b/config/initializers/hosting_environment.rb
@@ -1,0 +1,1 @@
+require "#{Rails.root}/lib/hosting_environment.rb"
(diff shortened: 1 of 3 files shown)


Kalagan:
Sorry I didn't have time to review this in time but why don't we use a ENV variable instead?
For example `ENV['APPLY_ENV']`
Then we can set the value in each `values-*.yaml` files.

This implementation relies on the URL which is not very reliable in my opinion.
`applyforlegalaid.apps.cloud-platform-live-0.k8s.integration.dsd.io`.

Also like Mark said, `production` makes more sense than `live` I think.

Kalagan:
Also, what is the use case of this? Why do we need to know if we are in UAT, staging or prod?

config/initializers/hosting_environment.rb
```diff
@@ -0,0 +1 @@
+require "#{Rails.root}/lib/hosting_environment.rb"
```
Kalagan: I think you don't need to add this if you put `HostingEnvironment` in `/app/lib` instead of `/lib`

PR #406, opened 2019-03-29
Ap 429 expire applicant link

willc-work:
The secure link sent to the client needs to expire either after 7 days or when they client completes the citizen means assessment journey.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-429)

Added an expire_at value to the secure_data used to generate the url. This is checked in the controller and the 'Expired page' is shown if it is passed the expired date (set to 8 days from issue so that it expires at midnight at the end of the 7th day.

Added a completed_at DateTime field to legal_aid_applications, this is set when the state of the applications transitions from ```checking_citizen_answers``` to ```means_completed``` Although the if block in the controller just checks this field has a value I thought a datetime would be more useful than a true/false value.

Added tests for the new navigation and changes to the generate_secure_id method.

## Checklist

### DIFF
--- a/app/controllers/citizens/legal_aid_applications_controller.rb
+++ b/app/controllers/citizens/legal_aid_applications_controller.rb
@@ -4,15 +4,10 @@ module Citizens
     # User passes in the Secure Id at the start of the journey. If login succeeds, they
     # are redirected to index where the first page is displayed.
     def show
-      sign_out current_provider if provider_signed_in?
       secure_id = params[:id]
+      @result = LegalAidApplication.find_by_secure_id!(secure_id)
 
-      legal_aid_application = LegalAidApplication.find_by_secure_id!(secure_id)
-
-      session[:current_application_id] = legal_aid_application.id
-
-      sign_applicant_in_via_devise(legal_aid_application.applicant)
-      redirect_to citizens_legal_aid_applications_path
+      check_for_expired_url
     rescue ActiveRecord::RecordNotFound
       # TODO: Handle failure
       # TODO: Modify Devise failures to handle failure to authenticate with project styled pages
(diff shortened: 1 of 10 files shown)


app/models/legal_aid_application.rb
```diff
@@ -54,12 +54,17 @@ def bank_transactions
...
+      SimpleResult.new(error: :expired)
+    else
+      SimpleResult.new(value: find_by!(secure_data[:legal_aid_application]))
+    end
   end
 
   def generate_secure_id
     SecureData.create_and_store!(
       legal_aid_application: { id: id },
+      expired_at: Time.current.beginning_of_day + 8.days,
```
reggieb: `8` is a magic number. As this is an important configuration setting it really should be in a constant and declared near the top of this model:
```ruby
SECURE_ID_DAYS_TO_EXPIRE = 8
```
Then here:
```
expired_at: Time.current.beginning_of_day + SECURE_ID_DAYS_TO_EXPIRE.days,
```

app/controllers/citizens/legal_aid_applications_controller.rb
```diff
@@ -4,28 +4,53 @@ class LegalAidApplicationsController < BaseController
...
-      legal_aid_application = LegalAidApplication.find_by_secure_id!(secure_id)
-
-      session[:current_application_ref] = legal_aid_application.id
-
-      sign_applicant_in_via_devise(legal_aid_application.applicant)
-      redirect_to citizens_legal_aid_applications_path
+      if find_legal_aid_application.error == :expired
+        expired
+      elsif application.completed_at.present?
+        completed
```
reggieb: If an `if` statement get to the point where you need to use an `elsif`, that's a smell! There is usually a better way of doing it. Here I'd suggest:
```ruby
return expired if find_legal_aid_application.error == :expired
return completed if application.completed_at.present?

start_applicant_flow
```

Kalagan:
@reggieb thanks, I made the changes

PR #423, opened 2019-04-04
Ap 370 create test applications

AbdirizakObsiye2018:
Ap 370 create test applications

[Link to story](https://dsdmoj.atlassian.net/browse/AP-370)

Created a button on the admin page to create fake applications with many different states/ at various parts of the application process ( around 7 applications which I create for each provider in said testing environment).

- I have created a service class that creates a batch of applications for usage in testing within UAT/Staging. 
- Created Factory bot traits for the test applications.
- Created a request test.

## Checklist


spec/factories/legal_aid_applications.rb
```diff
@@ -145,5 +149,48 @@
       transaction_period_start_at { 3.months.ago.beginning_of_day }
       transaction_period_finish_at { Time.now.beginning_of_day }
     end
+
+    trait :test_application_initiated do
```
reggieb: The `test_application_` prefix is too verbose and assumes these traits will only be useful in the current context. I'd suggest replacing it with `at_` so: `at_initiated`, `at_checking_answers`. In tests we can then do:
```ruby
let(:legal_aid_application) { create :legal_aid_application, :at_checking_passported_answers }
```
I'd also check to see if any duplication can be removed. For example, I expect specs currently using the trait `:mean_completed` could use the new `at_means_completed` and therefore the spec could be updated and the now redundant trait removed.

app/views/admin/legal_aid_applications/index.html.erb
```diff
@@ -6,6 +6,15 @@
 
   <p><%= link_to t('.edit_settings'), admin_settings_path, class: 'govuk-link' %></p>
 
+  <% if HostingEnvironment.staging? || HostingEnvironment.uat? || HostingEnvironment.development? %>
```
reggieb: :man_facepalming:

config/routes.rb
```diff
@@ -22,7 +22,8 @@
 
   namespace :admin do
     root to: 'legal_aid_applications#index'
-    resources :legal_aid_applications, only: %i[index destroy] do
+    resources :legal_aid_applications, only: %i[index update destroy] do
+      patch :create_test_applications, on: :collection
```
reggieb: That probably should be a `POST` - a resource is not being amended, new ones are being created.

spec/requests/admin/legal_aid_applications_spec.rb
```diff
@@ -37,6 +37,20 @@
     end
   end
 
+  describe 'PATCH /admin/legal_aid_applications/create_test_applications' do
+    subject { patch create_test_applications_admin_legal_aid_applications_path }
+    let(:count) { 1 }
+
+    it 'creates test legal_aid_applications' do
+      expect { subject }.to change { LegalAidApplication.count }.to be > 1
```
reggieb: Why are you being so vague here - you know exactly how many new items should be created:
```
number_new = TestApplicationCreationService::APPLICATION_TEST_TRAITS.size
expect { subject }.to change { LegalAidApplication.count }.by(number_new)
```

app/services/test_application_creation_service.rb
```diff
@@ -0,0 +1,40 @@
...
+    test_application_initiated
+    test_application_checking_answers
+    test_application_checking_passported_answers
+    test_application_answers_checked
+    test_application_provider_submitted
+    test_application_means_completed
+    test_application_chekcing_merits_answers
+  ].freeze
+
+  PROVIDERS = Provider.all
```
reggieb: The constant `PROVIDERS` isn't used anywhere. Remove it.

app/services/test_application_creation_service.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def self.call
+    new.call
+  end
+
+  def initialize
+    @providers = Provider.all
```
reggieb: This should not be in the initializer. Create a `providers` method and use that:
```ruby
def providers
  Provider.all
end

def call
  providers.each do |provider|
```

spec/factories/legal_aid_applications.rb
Kalagan: I suggested Abdi to use traits with `test_application_ `, but `at_` is much better and concise 👍

PR #425, opened 2019-04-05
AP-467 Add pagination to providers home page

reggieb:
[Jira AP-467](https://dsdmoj.atlassian.net/browse/AP-467)

Adds pagination to the providers page using the [Pagy gem](https://github.com/ddnexus/pagy).

I've got everything working as defined in the spec, except on page 4 the links look like:

pagy-start-sequence

After that (page 5 onwards) they match the spec:

pagy-mid-sequence

### DIFF
--- a/app/controllers/providers/legal_aid_applications_controller.rb
+++ b/app/controllers/providers/legal_aid_applications_controller.rb
@@ -1,11 +1,17 @@
 module Providers
   class LegalAidApplicationsController < ProviderBaseController
+    include Pagy::Backend
     legal_aid_application_not_required!
+
+    DEFAULT_PAGE_SIZE = 20
 
     # GET /provider/applications
     def index
-      # TODO: Add pagination at some point
-      @applications = current_provider.legal_aid_applications.latest.limit(10)
+      @pagy, @legal_aid_applications = pagy(
+        current_provider.legal_aid_applications.latest,
+        items: params.fetch(:page_size, DEFAULT_PAGE_SIZE),
+        size: [1, 1, 1, 1] # control of how many elements shown in page info
+      )
     end
 
     # POST /provider/applications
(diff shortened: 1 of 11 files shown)


app/views/providers/legal_aid_applications/_legal_aid_applications.html.erb
```diff
@@ -26,5 +26,11 @@
         </tbody>
       <% end %>
     </table>
+    <% if local_assigns[:pagy] %>
```
Kalagan: will this condition ever be false?

app/controllers/providers/legal_aid_applications_controller.rb
```diff
@@ -1,11 +1,17 @@
...
 
+    DEFAULT_PAGE_SIZE = 20
+
     # GET /provider/applications
     def index
-      # TODO: Add pagination at some point
-      @applications = current_provider.legal_aid_applications.latest.limit(10)
+      @pagy, @legal_aid_applications = pagy(
+        current_provider.legal_aid_applications.latest,
+        items: params.fetch(:page_size, DEFAULT_PAGE_SIZE),
```
Kalagan: will `params[: page_size]` ever be present?

app/views/providers/legal_aid_applications/_legal_aid_applications.html.erb
reggieb: I'm conscious that this is partial that could be used elsewhere. I thought it safer to put this in place.

app/controllers/providers/legal_aid_applications_controller.rb
reggieb: It's very difficult to test the functionality without it. For example, to test that the pagination links look like this:

pagy-start-sequence

You'd have to create something in the order of 160 applications.

Kalagan: ah ok that makes sense

PR #426, opened 2019-04-05
AP-502 Set sample bank transactions in a CSV

Kalagan:
## What

[AP-502](https://dsdmoj.atlassian.net/browse/AP-502)

In addition to modifying the sample bank transactions, this PR moves them from a hash to a CSV file.
That way, subsequent changes will be much easier. We can just download the CSV from google spreadsheet.

I implemented it so the same transactions generate always the same `transaction_id` so they are not duplicated in the database if the user goes through the TrueLayer process more than once

## Checklist


app/services/true_layer/sample_transactions_csv_parser.rb
```diff
@@ -0,0 +1,54 @@
...
+  class SampleTransactionsCsvParser
+    TRANSACTIONS_CSV = 'db/sample_data/bank_transactions.csv'.freeze
+
+    def self.call
+      new.call
+    end
+
+    def call
+      @transaction_ids = []
+      CSV.parse(csv_file).map do |row|
```
reggieb: If you use `CSV.read` you can pass in the file path directly without first converting to a string in `csv_file`:
```ruby
CSV.read(Rails.root.join(TRANSACTIONS_CSV))
```

reggieb: Also if you add a header row to the CSV:
```csv
date,description,amount_out,amount_in 
27/09/2018,Cash Bnkm Cooperative,£40.00,
28/09/2018,Tombola.co.uk betting betting,£10.00,
28/09/2018,McDonalds Kilburn,£8.48,
```
You can do this:
```ruby
CSV.foreach(Rails.root.join(TRANSACTIONS_CSV), headers: true) do |row|
  p row['description']
end
```
It makes the code here a little more verbose, but makes the data handling more flexible because:

- The columns no longer need to be in exactly the same order as you've defined
- There can be extra columns that are ignored.
- The data itself is more informative in that they now have headers that describe each column.

reggieb: The later being particular important to clarify which row is `amount_out` and which `amount_in`

Kalagan: Thanks! I'll do that

spec/services/true_layer/api_client_mock_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      expect(result).to eq(second_result)
+    end
+
+    it 'returns unique transaction_ids' do
+      transaction_ids = result.pluck(:transaction_id)
+      expect(transaction_ids.uniq).to eq(transaction_ids)
+    end
+
+    context 'with known data' do
+      let(:csv_file) { 'spec/fixtures/db/sample_data/bank_transactions.csv' }
```
reggieb: I think you can make your sample smaller (if it can do 2 it can do 22), and make them more varied. I'd suggest something like:
```csv
date,description,amount_out,amount_in
01/12/2020,something with some spaces,1234.56,
12/01/2010,simple,,"£1,234.56"
```
So just two but quite different:

1. date that could be parsed incorrectly with a large month, description with spaces, debit in a simple format without limiters
2. date that could be parsed incorrectly with a large day, one word description, credit in a format with `£` and `,`.

PR #428, opened 2019-04-05
AP-322 Better reference numbers

Kalagan:
## What

[AP-322](https://dsdmoj.atlassian.net/browse/https://dsdmoj.atlassian.net/browse/AP-322)

Generating random and unique reference numbers.

Also changing the session variable from `session[:current_application_ref]` to `session[:current_application_id]` because it contains the id of the application, not `application_ref` so it's confusing

## Checklist


spec/services/reference_number_creator_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    context 'first generated numbers are taken' do
+      let(:existing_reference_number) { 'L-AAA-AAA' }
+      let(:other_existing_reference_number) { 'L-BBB-BBB' }
+      let(:third_reference_number) { 'L-CCC-CCC' }
+
+      before do
+        (create :legal_aid_application).update!(application_ref: existing_reference_number)
+        (create :legal_aid_application).update!(application_ref: other_existing_reference_number)
```
reggieb: Creating and then updating is unnecessary I think as the `create_app_ref` is irrelevant as the ref is being supplied. As the update is not required, the creation can be done in a `let!` rather than using a before. So:
```ruby
let!(:legal_aid_application) { create :legal_aid_application, application_ref: existing_reference_number }
let!(:other_legal_aid_application) { create :legal_aid_application, application_ref: other_existing_reference_number }
```

Kalagan: I tried your solution but the problem is that `application_ref` get overwritten with a random reference number because of the `before_create :create_app_ref` callback

reggieb: Then doesn't `legal_aid_application#create_app_ref` need updating so that it doesn't behave that way (as the current implementation has the potential to create a lot of unexpected behaviour):
```ruby
  def create_app_ref
    self.application_ref = SecureRandom.uuid unless application_ref.present?
  end
```

Kalagan: Yes good idea

reggieb: And of course it should now be:
```ruby
self.application_ref = ReferenceNumberCreator.call unless application_ref.present?
```

PR #431, opened 2019-04-08
Ap 499 generate ccms soap packets

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-499)

Initial attempt at creating classes to create soap requests and parse responses.

This PR creates several classes  in the CCMS module namespace that deal with creating the XML SOAP packet and interacting with CCMS in order to submit an applications (note the use of Client rather than Applicant in order to tie up with CCMS nomenclature):

- `ReferenceDataRequestor` - send XML in order to request a unique reference for a claim
- `ReferenceDataParser` - parse the response from the above class
- `ClientSearchRequestor` - send XML to query whether or not the applicant already exists on CCMS
- `ClientSearchParser` - parse the response from the above class
- `CreateClientRequestor` - send XML to add a new applicant to CCMS
- `CreateClientStatusRequestor` - send XML to query the status of a previous Create Client request
- `CreateClientStatusParser` - parse the response from the above request
- `AddCaseRequestor` - Generate the XML to add a new case to CCMS
- `AttributeValueGenerator` - a helper class to generate values for the AddCaseRequestor

`Nokogiri::XML::Builder` is used to generate the XML payloads.  The `Savon` soap gem uses `Gyoko` under the hood to generate XML from ruby hashes, but the structure of the AddCase payload is too complex and inconsistent (specifically in how arrays are encoded) to use.

As yet, we don't have connectivity with the test instance of CCMS, so no communication is carried out - we are just testing the structure of the XML payloads.

This PR has no impact on the way the system works at the moment - it is just a building block for the future.

## Checklist


app/services/ccms/base_requestor.rb
```diff
@@ -0,0 +1,55 @@
...
+          xml.__send__('soap:Body') { soap_body(xml) }
+        end
+      end
+    end
+
+    def soap_header(xml)
+      xml.__send__('ns1:Security') do
+        xml.__send__('ns1:UsernameToken') do
+          xml.__send__('ns1:Username', ENV['SOAP_CLIENT_USERNAME'])
+          xml.__send__('ns1:Password', 'Type' => ENV['SOAP_CLIENT_PASSWORD_TYPE']) { xml.text ENV['SOAP_CLIENT_PASSWORD'] }
```
Kalagan: `SOAP_CLIENT_...` is a bit too generic because we are using other SOAP APIs.

app/services/ccms/base_requestor.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+      end
+    end
+
+    def ns3_header_rq(xml)
+      xml.__send__('ns3:TransactionRequestID', transaction_request_id)
+      xml.__send__('ns3:Language', 'ENG')
+      xml.__send__('ns3:UserLoginID', ENV['USER_LOGIN'])
+      xml.__send__('ns3:UserRole', ENV['USER_ROLE'])
```
Kalagan: Ditto. A bit too generic. I suggest the env variables have CCMS in their names

spec/services/ccms/add_case_requestor_spec.rb
```diff
@@ -0,0 +1,261 @@
+# add_case_requestor_spec.rb
```
Kalagan: remove this comment?

PR #433, opened 2019-04-09
AP-449 Add applicant error pages for when link does not work

reggieb:
[Jira AP-449](https://dsdmoj.atlassian.net/browse/AP-449)

Adds an **errors** controller and uses it to display errors for "page not found", "assessment already completed" and "link expired".

Then uses error pages as endpoints for errors at start of Citizen journey.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,12 @@
 class ApplicationController < ActionController::Base
   layout 'application'
   include Backable
+
+  rescue_from ActiveRecord::RecordNotFound, with: :page_not_found
+
+  private
+
+  def page_not_found
+    redirect_to error_path(:page_not_found)
+  end
 end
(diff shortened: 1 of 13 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,4 +1,12 @@
...
   layout 'application'
   include Backable
+
+  rescue_from ActiveRecord::RecordNotFound, with: :page_not_found
+
+  private
+
+  def page_not_found
+    redirect_to error_path(:page_not_found)
+  end
```
reggieb: As "page not found" is a generic error, I've hooked that in at the lowest level - so this is the default error anywhere in the app for `ActiveRecord::RecordNotFound`.

reggieb:
On hold at the moment - as some functionality (identifying expiry) relies on code in https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/406

reggieb:
#406 is now in master. So have update this PR to use it's functionality. So now ready to review.

PR #436, opened 2019-04-10
AP-482 Fix markup issues in providers/applicant

reggieb:
[Jira AP-482](https://dsdmoj.atlassian.net/browse/AP-482)

Corrects some errors in the mark up of the providers/applicant page.

- Modified the page template so that the margin below the heading can be set easily. Using the form page template results in a 15px margin below the heading, whereas the default page template heading has 30px below the head.
- Put the applicant name fields within a fieldset and put the legend within a legend tag rather than a span, and again within the new fieldset.
- Removed the duplication of the class "govuk-input" within the name input fields.
- Used the `govuk_text_field` `label` option to add a correctly format label *within the form group* which then allowed me to remove the labels that were rendering outside the form group.

When I started this I thought I would continue through and replace all the uses of the form page template, but the number of issues on this single page made this more complicated than I thought. So I have not completed this beyond modify the page template used by admin/settings.

### DIFF
--- a/app/views/shared/page_templates/_default_page_template.html.erb
+++ b/app/views/shared/page_templates/_default_page_template.html.erb
@@ -8,9 +8,7 @@
         <%= content_for(:page_title) %>
       </h3>
     <% else %>
-      <% if content_for?(:page_title) %>
-        <h1 class="govuk-heading-xl"><%= content_for(:page_title) %></h1>
-      <% end %>
+      <%= page_heading page_heading_options %>
     <% end %>
     <%= content %>
   </div>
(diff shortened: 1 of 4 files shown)


app/views/shared/page_templates/_default_page_template.html.erb
```diff
@@ -8,9 +8,7 @@
         <%= content_for(:page_title) %>
       </h3>
```
reggieb: I think new `page_heading` method can be used here, but I didn't make that change in this PR but rather have postponed that change to when a page including a panel is being updated, so that the change can be properly tested.

reggieb:
I thought I'd see if I could refactor some of the `PageTemplateHelper.page_template` complication into a service object, but it's too complicated. Specifically because the functionality requires calling `capture` and `content_for` which I couldn't get to load nicely within the service object.

So this is ready for review again.

PR #476, opened 2019-05-01
AP 532 Applicant needs new link notification

reggieb:
[JIRA AP-532](https://dsdmoj.atlassian.net/browse/AP-532)

This PR changes the response to someone trying to start the Citizen journey with an expired link. They are now sent to a page where they are prompted to send a request for a new link.

## The changes:

- A refactor of the existing mailer so that I could reuse code in the a new mailer.
- Change the citizens/legal_aid_applications_controller response to an expired request to redirect to a new end point.
- Removed the error/link_expired endpoint as it is now redundant
- Moved the `LegalAidApplication.find_by_secure_id!` functionality to a service so that I could reuse the functionality a little more easily.
- Created `Citizens::ResendLinkRequestsController` with `show` and `update` actions to handle the new request link functionality. 
- Created `ResendLinkRequestMailer` to use a [new Notify template](https://www.notifications.service.gov.uk/services/163de143-2d34-4454-b1fc-2f41624f2771/templates/a577582b-3e60-44a4-885c-d4b12bf23958).
- Moved `support_email_address` definition to `config/application` so that it's set in the same place as other global settings and can be easily updated later to be over-ridden with an environment variable if needed. It needed to be in a common location now that it is used across multiple mailers.

### DIFF
--- a/spec/services/secure_application_finder_spec.rb
+++ b/spec/services/secure_application_finder_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe SecureApplicationFinder, type: :service do
       expect(subject.legal_aid_application).to eq(legal_aid_application)
     end
 
-    it 'has no errors' do
+    it 'has :expired error' do
       expect(subject.error).to eq(:expired)
     end
   end
--- a/spec/services/secure_application_finder_spec.rb
+++ b/spec/services/secure_application_finder_spec.rb
@@ -1,0 +1,33 @@
+require 'rails_helper'
+
+RSpec.describe SecureApplicationFinder, type: :service do
+  let(:legal_aid_application) { create :legal_aid_application }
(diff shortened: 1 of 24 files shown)


spec/services/secure_application_finder_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  context 'when expired' do
+    let(:expired_at) { 1.hour.ago }
+
+    it 'does not find the application' do
+      expect(subject.legal_aid_application).to eq(legal_aid_application)
+    end
+
+    it 'has no errors' do
+      expect(subject.error).to eq(:expired)
```
Kalagan: -`it 'has no errors' do`
+`it 'has an :expired error' do`

PR #479, opened 2019-05-02
AP-559 CCMS Submission - initial orchestration work

mpw5:
## What

[AP-559](https://dsdmoj.atlassian.net/browse/AP-559)

This is the initial work to orchestrate the exchange of data between Apply and CCMS in order to inject a case. It comprises three sub-stories:

[AP-573 - Create tables](https://dsdmoj.atlassian.net/browse/AP-573)
[AP-574 - Obtain case ref](https://dsdmoj.atlassian.net/browse/AP-574)
[AP-575 - Get applicant reference](https://dsdmoj.atlassian.net/browse/AP-574)

* Create `submission` and `submission_history` models and migrations. The submission model contains a state machine to control the progress of a case through the necessary steps.

* Add functionality to `submission` to obtain a ccms_case_reference from CCMS, persist it to the database, and update the state.

* Add functionality to `submission` to obtain a search CCMS for an applicant, and if (a) one or more applicants are returned, obtain the applicant_ccms_reference of the first applicant in the response, persist it to the database, and update the state, or if (b) no applicants are returned, initiate the creation of a new applicant in CCMS, and update the state.

* Refactor of various `_requestor` and `_parser` classes to accommodate the above and improve naming.

## Checklist


spec/services/ccms/applicant_add_requestor_spec.rb
```diff
@@ -15,14 +31,14 @@ module CCMS
...
-          requestor = described_class.new
+          requestor = described_class.new(applicant)
           expect(requestor.transaction_request_id).to start_with '20190102030405123456'
         end
       end
     end
 
     def expected_xml
-      File.read("#{File.dirname(__FILE__)}/data/expected_create_client_status_request.xml")
+      File.read("#{File.dirname(__FILE__)}/data/applicant_add_request.xml")
```
reggieb: There needs to be some consistency in how the data files are accessed. In other specs the file has been located relative to `Rails.root`. Here the files directory is being used (which be problematic if spec is moved).

Also, I think data files should be at spec root rather than nested within the specs. So move them to `spec/data/ccms`

As this is being used so much, I'd strongly recommend that this is wrapped in a method. Something like this:

Add a new file `spec/support/file_helpers.rb` with:
```ruby
def cms_data_from_file(filename)
  path = Rails.root.join 'spec/data/ccms', filename
  File.read path
end
```
Then this method can be used in specs. So for example, this method could be replaced with:
```ruby
let(:expected_xml) { cms_data_from_file 'applicant_add_request.xml' }
```

spec/models/ccms/submission_spec.rb
```diff
@@ -0,0 +1,280 @@
...
+          expect {
+            sub.process!
+          }.to raise_error RuntimeError, 'Unknown state'
+        end
+      end
+
+      context 'initialised state' do
+        let(:submission) { create :submission, :initialised, legal_aid_application: legal_aid_application }
+        context 'operation successful' do
+          let(:response) { File.read(File.join(Rails.root, 'spec', 'services', 'ccms', 'data', 'reference_data_response.xml')) }
```
reggieb: Instead of `File.read(File.join(Rails.root, 'spec', 'services', 'ccms', 'data', 'reference_data_response.xml'))`

You could: `File.read(Rails.root.join 'spec', 'services', 'ccms', 'data', 'reference_data_response.xml')`

Though as the elements are all static I think this is cleaner:

```ruby
File.read(Rails.root.join 'spec/services/ccms/data/reference_data_response.xml')
```

reggieb: However, even better use
```ruby
cms_data_from_file 'reference_data_response.xml'
```
 as suggested in another comment.

spec/models/ccms/submission_spec.rb
```diff
@@ -0,0 +1,280 @@
...
+      end
+    end
+
+    # private methods tested here because they are mocked out above
+    #
+    describe '#reference_data_requestor' do
+      it 'only instantiates one copy of the ReferenceDataRequestor' do
+        sub = Submission.new
+        requestor1 = sub.__send__(:reference_data_requestor)
+        requestor2 = sub.__send__(:reference_data_requestor)
```
reggieb: Why is `__send__` being used with simple static content? 

Also within spec variables should be defined via `let`

I'd suggest:
```ruby
let(:requestor1) { sub.reference_data_requestor }
let(:requestor2) { sub.reference_data_requestor }
```

spec/models/ccms/submission_spec.rb
```diff
@@ -0,0 +1,280 @@
...
+        expect(requestor1).to be_instance_of(ReferenceDataRequestor)
+        expect(requestor1.object_id).to eq requestor2.object_id
+      end
+    end
+
+    describe '#applicant_search_requestor' do
+      it 'only instantiates one copy of the ApplicantSearchRequestor' do
+        sub = Submission.new(legal_aid_application: legal_aid_application)
+        requestor1 = sub.__send__(:applicant_search_requestor)
+        requestor2 = sub.__send__(:applicant_search_requestor)
```
reggieb: As above

spec/services/ccms/applicant_search_requestor_spec.rb
```diff
@@ -1,11 +1,19 @@
...
 module CCMS
-  RSpec.describe AddCaseStatusRequestor do
+  RSpec.describe ApplicantSearchRequestor do
+    let(:applicant) do
+      double Applicant,
+             first_name: 'lenovo',
+             last_name: 'hurlock',
+             date_of_birth: Date.new(1969, 1, 1),
+             national_insurance_number: 'YS327299B'
+    end
```
reggieb: Again - why not use a factory!

spec/services/ccms/applicant_search_response_parser_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+module CCMS
+  RSpec.describe ApplicantSearchResponseParser do
+    describe 'parsing applicant search responses' do
+      let(:no_results_response_xml) { File.read("#{File.dirname(__FILE__)}/data/applicant_search_response_no_results.xml") }
+      let(:one_result_response_xml) { File.read("#{File.dirname(__FILE__)}/data/applicant_search_response_one_result.xml") }
+      let(:multiple_results_response_xml) { File.read("#{File.dirname(__FILE__)}/data/applicant_search_response_multiple_results.xml") }
+
+      it 'raises if the transaction_request_ids dont match' do
+        expect {
+          parser = described_class.new('20190301030405987654', no_results_response_xml)
```
reggieb: `parser` should be defined in a `let` and what is the significance of '20190301030405987654'?

app/models/ccms/submission.rb
```diff
@@ -0,0 +1,140 @@
+module CCMS
+  class Submission < ApplicationRecord # rubocop:disable Metrics/ClassLength
```
reggieb: This model is too big. There are two steps that should be made to reduce it's size:

1. Strip the state machine functionality out into a concern (see `LegalAidApplicationStateMachine`
2. Split the complicated methods into services.

app/models/ccms/submission.rb
```diff
@@ -0,0 +1,140 @@
...
+        create_history(aasm_state, :applicant_submitted)
+        submit_applicant!
+      else
+        create_failure_history(aasm_state, response)
+        fail!
+      end
+    rescue StandardError => e
+      create_failure_history(aasm_state, e)
+      fail!
+    end
```
reggieb: `obtain_case_reference`, `obtain_applicant_reference`, `add_applicant` are complicated methods that should be split out into separate services.

As well as reducing the complication of this class and these methods, I expect that it would be a lot easier to test the behaviour without having to do so much mocking in the specs.

app/models/ccms/submission.rb
```diff
@@ -0,0 +1,140 @@
...
+    def create_failure_history(from_state, error)
+      SubmissionHistory.create submission: self,
+                               from_state: from_state,
+                               to_state: :failed,
+                               success: false,
+                               details: error.is_a?(Exception) ? format_exception(error) : error
+    end
+
+    def format_exception(error)
+      "#{error.class}\n#{error.message}\n#{error.backtrace.join("\n")}"
```
reggieb: As `join` is being used here anyway, I'd suggest:
```ruby
[error.class, error.message, error.backtrace].flatten.join("\n")
```

app/services/ccms/applicant_add_requestor.rb
```diff
@@ -1,5 +1,11 @@
 module CCMS
-  class CreateClientRequestor < BaseRequestor
+  class ApplicantAddRequestor < BaseRequestor
+    def initialize(applicant)
+      @applicant = applicant
+      @address = applicant.address
+      super()
+    end
```
reggieb: I'd use `attr_reader` to make these accessible within the rest of class rather than passing the variable around. Perhaps even use delegate:
```ruby
attr_reader :applicant
delegate :address, to: :applicant

def initialize(applicant)
  @applicant = applicant
  super
end
```

app/services/ccms/applicant_add_requestor.rb
```diff
@@ -30,13 +36,13 @@ def client(xml)
...
-      xml.__send__('ns4:Surname', 'Hurlock')
-      xml.__send__('ns4:FirstName', 'lenovo')
+      xml.__send__('ns4:Surname', @applicant.last_name)
+      xml.__send__('ns4:FirstName', @applicant.first_name)
     end
 
     # this is all mandatory: we don't hold any of this data except date of birth
     def personal_information(xml)
-      xml.__send__('ns5:DateOfBirth', '1969-01-01')
+      xml.__send__('ns5:DateOfBirth', @applicant.date_of_birth.strftime('%Y-%m-%d'))
```
reggieb: If this is the format CCMS likes dates I'd suggest adding the following to `config/initializers/date_formats.rb`

```ruby
ccms_formats = {
  cmms_date: '%Y-%m-%d'
}

Time::DATE_FORMATS.merge! ccms_formats
Date::DATE_FORMATS.merge! ccms_formats
```
Then here:
```ruby
xml.__send__('ns5:DateOfBirth', @applicant.date_of_birth.to_s(:ccms_date))
```

app/services/ccms/applicant_add_requestor.rb
```diff
@@ -52,10 +58,10 @@ def contacts(xml)
     end
 
     def address(xml)
-      xml.__send__('ns4:AddressLine1', '102 Petty France')
-      xml.__send__('ns4:City', 'London')
+      xml.__send__('ns4:AddressLine1', @address.address_line_one + ' ' + @address.address_line_two)
```
reggieb: I'd suggest the combination of lines one and two of the address, is moved to `Address`, where handling on line two absence can be easily added. Something like:
```ruby
def first_lines
  [address_line_one, address_line_two].select(&:present?).join(' ')
end
```
Then here:
```ruby
xml.__send__('ns4:AddressLine1', @address.address_first_lines)
```

app/services/ccms/applicant_search_requestor.rb
```diff
@@ -1,5 +1,10 @@
 module CCMS
-  class ClientSearchRequestor < BaseRequestor
+  class ApplicantSearchRequestor < BaseRequestor
+    def initialize(applicant)
+      @applicant = applicant
+      super()
```
reggieb: `super`. There is no need for the `()` in ruby.

spec/factories/submissions.rb
```diff
@@ -0,0 +1,13 @@
+FactoryBot.define do
+  factory :submission, class: CCMS::Submission do
+    legal_aid_application { create :legal_aid_application }
+
+    trait :initialised do
+      aasm_state { 'initialised' }
+    end
```
reggieb: As default, won't the state by definition be set at `initialised`? If so - this is redundant.

app/models/ccms/submission.rb
```diff
@@ -0,0 +1,140 @@
+module CCMS
+  class Submission < ApplicationRecord # rubocop:disable Metrics/ClassLength
+    include AASM
+
+    self.table_name = :ccms_submissions
```
reggieb: Having to fix the table name is often a smell. It suggest the model namespace isn't configured correctly, or the class name is poor.

app/models/ccms/submission_history.rb
```diff
@@ -0,0 +1,7 @@
+module CCMS
+  class SubmissionHistory < ::ApplicationRecord
+    self.table_name = :submission_histories
```
reggieb: Why is the submissions table prefixed with `ccms_` and this table isn't?

spec/services/ccms/applicant_add_requestor_spec.rb
```diff
@@ -1,11 +1,27 @@
...
+             pretty_postcode: 'SW1H 9AJ'
+    end
+
+    let(:applicant) do
+      double Applicant,
+             address: address,
+             first_name: 'lenovo',
+             last_name: 'Hurlock',
+             date_of_birth: Date.new(1969, 1, 1)
+    end
```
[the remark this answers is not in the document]
mpw5: We need the applicant and address to have specific values, so they can be compared to the expected content in the test xml files. Using factories will give us random values (unless I'm missing something?)

spec/models/ccms/submission_spec.rb
mpw5: `reference_data_requestor` is a private method. `sub.reference_data_requestor` fails with a scoping error.

(Whether it should be private in the first place is another question...)

I've moved the definitions to a `let` block though.

spec/models/ccms/submission_spec.rb
mpw5: As above

spec/services/ccms/applicant_search_requestor_spec.rb
mpw5: As above

app/services/ccms/applicant_search_requestor.rb
mpw5: I think in this instance we do need the (). Unlike `initialize` in this class, the `initialize` method in `BaseRequestor` doesn't take any arguments. Calling `super` attempts to pass `applicant` through to the parent, which causes an error. `super()` doesn't pass the argument.

spec/services/ccms/applicant_add_requestor_spec.rb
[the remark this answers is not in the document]
reggieb: You can define the attributes when using factories in exactly the same way:
```
let(:applicant) do
  create(
    :applicant, 
    address: address, 
    first_name: 'lenovo', 
    last_name: 'Hurlock', 
    date_of_birth: Date.new(1969, 1, 1)
  )
end
```
The values set in the factory definitions just set the defaults. If you specify attributes in the `create ` (or `build` etc.) call, the values passed in will over-ride those set in the factory definition.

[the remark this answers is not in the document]
mpw5: Ah, I see. I'll change it.

mpw5:
@reggieb - I think I've (finally) addressed all your comments. If you're able to give it another look over that would be appreciated. I'm aware that there is a *lot* going on in this PR now so if it makes it easier to review I could pull some of it out into another PR (especially the last commit, which refactors the `submission` class and spec into a number of services).

@stephenrichards - it might be good if you could take a look at this too

app/services/ccms/obtain_applicant_reference_service.rb
```diff
@@ -0,0 +1,29 @@
+module CCMS
+  class ObtainApplicantReferenceService < BaseSubmissionService
+    def call
+      tx_id = applicant_search_requestor.transaction_request_id
+      response = applicant_search_requestor.call
+      parser = ApplicantSearchResponseParser.new(tx_id, response)
+      process_records(parser)
+    rescue StandardError => e
```
reggieb: See earlier comments about rescuing `StandardError`. Here be monsters.

app/services/ccms/obtain_case_reference_service.rb
```diff
@@ -0,0 +1,19 @@
+module CCMS
+  class ObtainCaseReferenceService < BaseSubmissionService
+    def call
+      tx_id = reference_data_requestor.transaction_request_id
+      response = reference_data_requestor.call
+      submission.case_ccms_reference = ReferenceDataResponseParser.new(tx_id, response).parse
+      create_history(submission.aasm_state, :case_ref_obtained)
+      submission.obtain_case_ref!
+    rescue StandardError => e
```
reggieb: Here be monsters

spec/models/ccms/submission_spec.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_helper'
+
+module CCMS
+  RSpec.describe Submission do
+    let(:sub) { create :submission }
```
reggieb: Is `submission` really that hard to type/read? Use `submission`:
```ruby
let(:submission) { create :submission }
```

spec/models/ccms/submission_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    describe '#process' do
+      let(:history) { SubmissionHistory.find_by(submission_id: submission.id) }
+
+      context 'invalid state' do
+        it 'raises if state is invalid' do
+          sub.aasm_state = 'xxxxx'
+          expect {
+            sub.process!
+          }.to raise_error RuntimeError, 'Unknown state'
```
reggieb:
```ruby
}.to raise_error CcmsError, 'Unknown state'
```

spec/models/ccms/submission_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        let(:obtain_case_reference_service_double) { ObtainCaseReferenceService.new(sub) }
+        it 'calls the obtain_case_reference service' do
+          expect(ObtainCaseReferenceService).to receive(:new).with(sub).and_return(obtain_case_reference_service_double)
+          expect(obtain_case_reference_service_double).to receive(:call)
+          sub.process!
+        end
+      end
+
+      context 'case_ref_obtained state' do
+        let(:sub) { create :submission, :case_ref_obtained }
```
reggieb: `submission` !!!!

spec/services/ccms/add_applicant_service_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+        subject.call
+        expect(submission.aasm_state).to eq 'failed'
+      end
+
+      it 'records the error in the submission history' do
+        expect { subject.call }.to change { CCMS::SubmissionHistory.count }.by(1)
+        expect(history.from_state).to eq 'case_ref_obtained'
+        expect(history.to_state).to eq 'failed'
+        expect(history.success).to be false
+        expect(history.details).to match(/RuntimeError/)
```
reggieb: `/CcmsError/`

PR #481, opened 2019-05-03
AP531 Bail conditions content on /respondent page

AbdirizakObsiye2018:
Bail conditions content on /respondent page 

For the question 'Have bail conditions been set?' on the respondent page, the label for its text area should change depending on whether the user selects yes or no as an answer to this question.

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-531)

1. Added custom data attributes to store the two labels for Bail conditions. 
2. Using JavaScript in the front-end to change the label depending on the option selected for bail_conditions_set ('yes', 'no').
3. Update the error text and the error summary list depending on the label.

## Checklist

### DIFF
--- a/app/assets/javascripts/respondent_detail.es6
+++ b/app/assets/javascripts/respondent_detail.es6
@@ -1,0 +1,31 @@
+$(document).ready(() => {
+  if ($("#bail_conditions_set").length) {
+    const bailConditionOptions = "input[type = radio][name = 'respondent[bail_conditions_set]']";
+    $(bailConditionOptions).on("change", setBailConditionsLabel);
+    setBailConditionsLabel();
+    if($('#bail_conditions_set_details-error').length) updateError();
+  };
+
+  function setBailConditionsLabel() {
+    const bailConditionsLabel = "label[for='bail_conditions_set_details']";
+    $(bailConditionsLabel).text($('#bail_conditions_set_details').attr(`data-bail-conditions-${selectedBailOption()}`));
+  }
+
+  function selectedBailOption(){
+    let optionChecked = "input[type = radio][name = 'respondent[bail_conditions_set]']:checked";
+    let optionValue = $(optionChecked).val() == 'false' ? 'no' : 'yes'
+    return optionValue
+  }
+
+  function updateError(){
+    if ($('#bail_conditions_set_details-error').length && selectedBailOption() == 'no'){
+      $('#bail_conditions_set_details-error').text($('#bail_conditions_set_details').attr(`data-bail-conditions-no`));
+      updateErrorSummary();
+    }
+  }
+
+  function updateErrorSummary(){
+    const errorSummaryListText = $(document.body.querySelectorAll('*[href="#bail_conditions_set_details"]')[0]);
+    errorSummaryListText.text($('#bail_conditions_set_details').attr(`data-bail-conditions-no`));
+  }
+});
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/respondent_detail.es6
```diff
@@ -0,0 +1,31 @@
+$(document).ready(() => {
+  if ($("#bail_conditions_set").length) {
```
Kalagan: I suggest having this condition wrap all the code. Otherwise the functions will be created on all pages and could conflict with other functions. Especially `updateError` which has a rather generic name.

app/assets/javascripts/respondent_detail.es6
```diff
@@ -0,0 +1,31 @@
...
+    const bailConditionOptions = "input[type = radio][name = 'respondent[bail_conditions_set]']";
+    $(bailConditionOptions).on("change", setBailConditionsLabel);
+    setBailConditionsLabel();
+    if($('#bail_conditions_set_details-error').length) updateError();
+  };
+
+  function setBailConditionsLabel() {
+    const bailConditionsLabel = "label[for='bail_conditions_set_details']";
+    $(bailConditionsLabel).text($('#bail_conditions_set_details').attr(`data-bail-conditions-${selectedBailOption()}`));
```
Kalagan: You should be able to use
```javascript
.data(`bail-conditions-${selectedBailOption()}`)
```
instead of
```javascript
.attr(`data-bail-conditions-${selectedBailOption()}`)
```

app/assets/javascripts/respondent_detail.es6
```diff
@@ -0,0 +1,31 @@
+$(document).ready(() => {
+  if ($("#bail_conditions_set").length) {
+    const bailConditionOptions = "input[type = radio][name = 'respondent[bail_conditions_set]']";
+    $(bailConditionOptions).on("change", setBailConditionsLabel);
```
Kalagan: Wouldn't it be useful to also change the error message?
```javascript
    $(bailConditionOptions).on("change", () => {
      setBailConditionsLabel()
      updateError()
    })
```

app/assets/javascripts/respondent_detail.es6
```diff
@@ -0,0 +1,31 @@
...
+  };
+
+  function setBailConditionsLabel() {
+    const bailConditionsLabel = "label[for='bail_conditions_set_details']";
+    $(bailConditionsLabel).text($('#bail_conditions_set_details').attr(`data-bail-conditions-${selectedBailOption()}`));
+  }
+
+  function selectedBailOption(){
+    let optionChecked = "input[type = radio][name = 'respondent[bail_conditions_set]']:checked";
+    let optionValue = $(optionChecked).val() == 'false' ? 'no' : 'yes'
```
Kalagan: you can use `const` here instead of `let` because these variables are not changed

app/assets/javascripts/respondent_detail.es6
AbdirizakObsiye2018: Good suggestion. I was just apprehensive to add that (changing error message) at the time.

app/views/providers/respondents/show.html.erb
```diff
@@ -26,9 +26,13 @@
       questions.each do |question|
         question.show_details_when = nil if question.attribute == :bail_conditions_set
       end
-    %>
 
-    <%= render partial: 'respondent_detail', collection: questions, as: :question, locals: { form: form } %>
+      bail_conditions_titles = {
+        'yes' => t('.bail_conditions_set_yes'),
+        'no' => t('.bail_conditions_set_no')
```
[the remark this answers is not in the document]
reggieb: Also symbols use less memory as they are immutable. See https://medium.com/@lcriswell/ruby-symbols-vs-strings-248842529fd9

PR #492, opened 2019-05-10
AP-404 accessible tables

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-404)

Added in new IDs for every data cell.  
Added in a new HTML property `aria-labelled-by` which referenced the ID each data cell in the row.
Added in the data inside the `label` tag, in a visually-hidden `span`.

## Checklist


app/assets/javascripts/table-sort.es6
```diff
@@ -2,6 +2,9 @@ $(document).ready(function() {
   if ($('table.sortable')) {
     let table;
     $('th.sort')
+      .addClass("js-sortable")
+        //this class used to style as links, if JS not enabled, titles won't look clickable.
```
reggieb: Very minor suggestion - but usually I'd expect a comment to be above the item it's concerning. So:
```javascript
//this class used to style as links, if JS not enabled, titles won't look clickable.
.addClass("js-sortable")
```
Or to make it really clear put it on the same line:
```javascript
.addClass("js-sortable") //this class used to style as links, if JS not enabled, titles won't look clickable.
```

app/assets/javascripts/table-sort.es6
```diff
@@ -27,6 +30,12 @@ $(document).ready(function() {
 
           return false;
         });
+        th.keyup(function(ev) {
+          if (ev.which==13 || ev.which==32)  {
+            //on space or return, the column is sorted
```
reggieb: Again - this comment refers to the `if` so should be above that statement.

app/assets/javascripts/table-sort.es6
```diff
@@ -24,9 +25,33 @@ $(document).ready(function() {
             return this.parentNode;
           });
           inverse = !inverse;
-
+
```
reggieb: The way this is coded is far more complicated than it needs to be. Basically, the code above this comment is creating some output and putting it into the page, and the code below is gathering that data again from the page and parsing it to work out the same logic.

For example: The parent `if` statements between lines 31 and 47 are basically saying "is there a tag with the class set so it sorts like ...." - and not just in a restricted part of the page - but the whole page. It then uses the result of the searches to make decisions on what the screen reader message should be.

When: 

- The code already knows which html element is being used to sort the table. It's defined above as `th`.
- The code already knows which direction the sort is occurring (ascending/descending). It's defined by `inverse` (and see my other comment about using that to define `sortDirection`).
- The first part of the screen reader message will be the text content of element `th`. It should always be `th.text()`.

So: 

What's needed is to work out the data type: "numeric", "alphabetic", "description". At the moment that's done by testing for one of three HTML classes. It would be much easier if a `th` element being numerical was assigned using a `data-` attribute rather than a class. So:
```html
<th data-sort-type="numeric">1</th>
<th data-sort-type="alphabetic">2</th>
<th data-sort-type="description">3</th>
```
With that in place, the sort type will be returned by `th.dataset.sort-type`

We now have the starting text (`th.text()`), the sort direction (`sortDirection`) and the sort type (`th.dataset.sort-type`). We next need to workout the ending text based on sort direction and type. We can use a matrix for that:
```javascript
endText = {
  asc: {
    numberic: 'from big to small',
    alphabetic: 'in reverse order',
    description: 'in <em>ascending</em> order'
  },
  desc: {
    numberic: 'from small to big',
    alphabetic: 'in alphabetical order',
    description: 'in <em>descending</em> order'
  }
}
```
With all that in place, we can define the construction of the screen reader message once:
```javascript
$("#screen-reader-messages").html("Sorted by " + th.text() + " " + endText[sortDirection][th.dataset.sort-type]);
```

reggieb: Oh - and this line will need to be moved to after the screen reader message has been set:
```javascript
inverse = !inverse;
```

MoJ-Longbeard: Changes made.  Except:
`th.dataset.sort-type` is throwing an error, so I am using `th.attr("data-sort-type")`.
`th.text()` contains the "click to sort by" message, so I'm using `th.find(".aria-sort-description").text()`
Thanks.

app/assets/javascripts/table-sort.es6
```diff
@@ -23,10 +38,19 @@ $(document).ready(function() {
             // parentNode is the element we want to move
             return this.parentNode;
           });
+          
+          //this adds a message to the message div, stating what it is sorted by
+          $("#screen-reader-messages").html("Sorted by " + th.find(".aria-sort-description").text() + endText[sortDirection][th.attr("data-sort-type")]);
```
reggieb: I think there should be a space between the text coming in from the `aria-sort-description` and the end text.

app/assets/javascripts/table-sort.es6
```diff
@@ -1,18 +1,33 @@
...
+            numeric: ' from big to small',
+            alphabetic: ' in <em>reverse</em> order',
+            undefined: ' in <em>ascending</em> order' //for when the data-sort-type is not set
+          },
+          desc: {
+            numeric: ' from small to big',
+            alphabetic: ' in <em>alphabetical</em> order',
+            undefined: ' in <em>descending</em> order' //for when the data-sort-type is not set
+          }
+        }
```
reggieb: I'd move the definition of `endText` outside the `each` loop - as it is the same for every iteration - and only needs to be defined once. So put it straight after `let table;`

app/assets/javascripts/table-sort.es6
MoJ-Longbeard: I have added a space to the strings in end text - do you think it should be separate?

app/views/providers/transactions/show.html.erb
```diff
@@ -15,32 +15,43 @@
...
           </th>
-          <th class="govuk-table__header sort header-sort-asc table-combine_right_if_narrow" scope="col"><%= t('.col_date') %><span class="table-hidden_right_title"> &amp; <%= t('.col_description') %></span></th>
-          <th class="govuk-table__header sort" scope="col"><%= t('.col_description') %></th>
-          <th class="govuk-table__header govuk-table__header--numeric sort" scope="col"><%= t('.col_amount') %></th>
         </tr>
       </thead>
       <tbody class="govuk-table__body">
         <%= form.collection_check_boxes :transaction_ids, @bank_transactions, :id, :description do |builder| %>
-          <tr class="govuk-table__row">
+          <tr class="govuk-table__row" area-hidden=true>
```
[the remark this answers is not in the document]
MoJ-Longbeard: Yes!  :grimacing:

app/assets/javascripts/table-sort.es6
MoJ-Longbeard: Made separate.

PR #501, opened 2019-05-13
AP 587 Retry failed email notifications

reggieb:
[JIRA AP-587](https://dsdmoj.atlassian.net/browse/AP-587)

This PR addresses 2 issues.

### Failed emails not being retried by Sidekiq
There were two things preventing Sidekiq handling failed email sends:

1. When mail send was being retried, Sidekiq was not finding the module `NotifyTemplateMethods` that was mixed into the mailers. Adding a `require_relative` pointing at the module file fixed that.
2. The mailer method `deliver_later` was not raising errors on connection failure. Changing this to the bang version fixed that: `deliver_later!`

With these changes in place when I disconnected my computer from the network and triggered a client journey start email, it was held in Sidekiq and retried until the connection was re-enabled.

email_retry_enabled

### Unknown step causing providers list of applications page to crash
While working on this bug I found another bug that I've also fixed here. If the code base changed so that a step that an application was associated with was removed, adding a link to the applications "current" location caused an error. I've made the following changes to fix that:

1. Changed the step path not found error raised by the flow process to use a specific error rather than a standard error.
2. Caught that error in the link creation helper and replaced it with a link to the start of the provider journey for that application.
3. Refactored the start of the provider journey so that it was defined by a `Flow::KeyPoint`.

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -7,7 +7,8 @@ class FeedbackController < ApplicationController
 
   def create
     feedback.update(feedback_params)
-    FeedbackMailer.notify(feedback).deliver_later if feedback_submitted?
+    # Must use bang version `deliver_later!` or failures won't be retried by sidekiq
+    FeedbackMailer.notify(feedback).deliver_later! if feedback_submitted?
 
     redirect_to feedback
   end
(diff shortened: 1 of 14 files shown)


reggieb:
No problems on UAT and reviewed by muliple devs - so merging

PR #504, opened 2019-05-15
AP-576 - CCMS -Poll add_applicant requests

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-576)

This change adds functionality to the CCMS orchestration process to poll the status of the creation of an applicant in CCMS. It:

- amends the add_applicant_service so that the transaction_id sent in the add request is recorded against the submission, so that it can be included in the poll request
- adds a poll count to the submission record which is incremented for each poll. An error is thrown when a limit (10) is reached
- adds a new check_applicant_status_service to handle the call to CCMS and process the response
- fixes minor issues with existing tests/test data

## Checklist

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2019_05_01_123416) do
+ActiveRecord::Schema.define(version: 2019_05_14_142921) do
+
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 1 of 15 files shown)


db/schema.rb
```diff
@@ -279,26 +282,26 @@
...
+    t.decimal "second_home_percentage"
+    t.decimal "timeshare_value"
+    t.decimal "land_value"
+    t.decimal "jewellery_value"
+    t.decimal "vehicle_value"
+    t.decimal "classic_car_value"
+    t.decimal "money_assets_value"
+    t.decimal "money_owed_value"
+    t.decimal "trust_value"
     t.datetime "created_at", null: false
```
stephenrichards: How did the precision come to be removed from these fields in the schema?

Kalagan: This file contains lots of unnecessary change. It happens when migrations are not run in the same order.
@mpw5 , if you want to reset your local DB to match `db/schema.rb`, you can run `bin/rails db:schema:load db:seed` (it will remove all data). And then re-run your migration

mpw5: Thanks both. Fixed now.

PR #505, opened 2019-05-15
Ap 584 Declared Income Links

AbdirizakObsiye2018:
On the means summary page the declared income and outgoings links go to income summary and outgoings summary pages where the save and continue returns back to the means summary page.

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-584)

1. Update means_summaries/show page links for the View/change declared outgoings and View/change income outgoings to route to the outgoings summary and income summary page.
3. The forward step of the income_summary is means_summary if the application is in the state checking answers. This stops it from continuing to the outgoings_summary page and forwards it to the means summary. 
5. Add a feature test to show that the routing is correct

## Checklist

- [] You should have checked that the commit messages say why the change was made.


app/services/flow/flows/provider_capital.rb
```diff
@@ -72,7 +72,9 @@ class ProviderCapital < FlowSteps
         },
         income_summary: {
           path: ->(application) { urls.providers_legal_aid_application_income_summary_index_path(application) },
-          forward: :outgoings_summary
+          forward: ->(application) do
+            application.checking_answers? ? :means_summaries : :outgoings_summary
+          end
```
Kalagan: instead of changing the `forward` key, you can add the `check_answers` key:
```ruby
  check_answers: :means_summaries
```
The logic you are adding is already supported by the flow service:
if the application is in a "check answers" state and a `check_answers` key is defined, then use that.
Otherwise, use the `forward` key.
(https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/services/flow/base_flow_service.rb#L29)

features/providers/civil_application_journey.feature
```diff
@@ -328,6 +328,23 @@ Feature: Civil application journeys
...
+    Then I should be on a page showing 'Your client has completed their financial assessment'
+    Then I click 'Continue'
+    Then I should be on a page showing "Your client's income"
+    Then I click 'Save and continue'
+    Then I should be on a page showing "Your client's regular payments"
+    Then I click 'Save and continue'
+    Then I should be on a page showing "Check your answers"
+    Then I click link "View/change declared income"
+    Then I should be on a page showing "Your client's income"
+    Then I click link "Back"
```
Kalagan: shouldn't we be testing if I click on `'Save and continue'` here instead?

app/services/flow/flows/provider_capital.rb
AbdirizakObsiye2018: awesome I didn't know that

features/providers/civil_application_journey.feature
AbdirizakObsiye2018: Yes definitely X D thank you . I at one point was trying to test the ability to change an income and hence left that there without.

PR #506, opened 2019-05-16
AP-577 - CCMS Add Case Orchestration

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-577)

Add functionality to submission.process! for submissions in the
applicant_ref_obtained state:

- additions to existing submission model and spec
- new field to store the case_add_transaction_id in the database so it
can be used for polling later
- new add_case service to handle api call and response parsing

## Checklist

### DIFF
--- a/app/services/ccms/add_case_service.rb
+++ b/app/services/ccms/add_case_service.rb
@@ -1,12 +1,11 @@
 module CCMS
   class AddCaseService < BaseSubmissionService
     def call
-      @response = case_add_requestor.call
       if case_add_response_parser.success?
         submission.case_add_transaction_id = case_add_requestor.transaction_request_id
         create_history('applicant_ref_obtained', submission.aasm_state) if submission.submit_case!
       else
-        handle_failure(@response)
+        handle_failure(response)
       end
     rescue CcmsError, StandardError => e # TODO: Replace `StandardError` with list of known expected errors
       handle_failure(e)
@@ -19,7 +18,11 @@ module CCMS
     end
 
     def case_add_response_parser
-      @case_add_response_parser ||= CaseAddResponseParser.new(case_add_requestor.transaction_request_id, @response)
+      @case_add_response_parser ||= CaseAddResponseParser.new(case_add_requestor.transaction_request_id, response)
+    end
+
+    def response
+      @response ||= case_add_requestor.call
     end
   end
 end
--- a/app/services/ccms/add_case_service.rb
+++ b/app/services/ccms/add_case_service.rb
@@ -1,0 +1,25 @@
+module CCMS
+  class AddCaseService < BaseSubmissionService
+    def call
+      @response = case_add_requestor.call
+      if case_add_response_parser.success?
+        submission.case_add_transaction_id = case_add_requestor.transaction_request_id
+        create_history('applicant_ref_obtained', submission.aasm_state) if submission.submit_case!
+      else
+        handle_failure(@response)
+      end
+    rescue CcmsError, StandardError => e # TODO: Replace `StandardError` with list of known expected errors
+      handle_failure(e)
+    end
+
+    private
+
+    def case_add_requestor
+      @case_add_requestor ||= CaseAddRequestor.new(submission.legal_aid_application.applicant)
+    end
+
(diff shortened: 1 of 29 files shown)


app/services/ccms/add_case_service.rb
```diff
@@ -0,0 +1,26 @@
+module CCMS
+  class AddCaseService < BaseSubmissionService
+    def call
+      response = case_add_requestor.call
+      if case_add_response_parser(response).success?
+        submission.case_add_transaction_id = case_add_requestor.transaction_request_id
+        create_history(submission.aasm_state, :case_submitted)
```
reggieb: Would this work? :- 
```ruby
create_history(submission.aasm_state, :case_submitted) if submission.submit_case!
```
That way the history will only be created if the case submission is successful.

app/services/ccms/add_case_service.rb
```diff
@@ -0,0 +1,26 @@
...
+      handle_failure(e)
+    end
+
+    private
+
+    def case_add_requestor
+      @case_add_requestor ||= CaseAddRequestor.new(submission.legal_aid_application.applicant)
+    end
+
+    def case_add_response_parser(response)
```
reggieb: There are two scenarios here:
1. The `response` will change and therefore needs to be passed in each time. In which case it shouldn't be cached with `@case_add_response_parser ||=`
2. The `response` is set once in which case it does not need to be passed into this method as it is already available to this instance method.

app/services/ccms/add_case_service.rb
mpw5: Good point. That will work, with a minor change to the values passed to `create_history`, as the status is now being updated before the history record is created, rather than afterwards.

That pattern is used in several other classes, so I'll push a separate commit changing them too.

app/services/ccms/add_case_service.rb
mpw5: It's scenario 2. I'll change it.

mpw5:
Thanks @reggieb. I've pushed changes for your suggestions.

app/services/ccms/case_add_response_parser.rb
```diff
@@ -9,6 +9,10 @@ def parse
       extracted_status
     end
 
+    def success?
+      parse == 'Success'
+    end
```
reggieb: I really don't like this generic `parse` method pattern. Every object that needs to use one of these parsers has to know that it first needs to call the `parse` method, before it calls the method to retrieve the desired data - or like this, you end up with code logic that is obscure as to it's meaning because the word `parse` tell the reader nothing about the nature of the data being retrieved.

The purpose of the `parse` method seems to be to ensure that an id check is made before any data is gathered. Why not refactor it to expose the data method through the parse method. For example, if this was added to `BaseResponseParser`:

```ruby
def parse(data_method)
  raise CcmsError, 'Invalid transaction request id' unless extracted_id_matches_request_id?
  send(data_method)
end
```
`extracted_id_matches_request_id?` could be defined in `BaseResponseParser` with each child class configuration defining its specific "transaction id path" either as now with `TRANSACTION_ID_PATH` or with a class method (which would not require constants to be overridden).

Then here the `success?` method could be rewritten as:
```ruby
def success?
  parse(:extracted_status) == 'Success'
end
```
And now the code itself explains what it's doing.

spec/services/ccms/add_case_service_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  # private method tested here because it is mocked out above
+  #
+  describe '#case_add_requestor' do
+    let(:service_double) { CCMS::AddCaseService.new(submission) }
+    let(:requestor1) { service_double.__send__(:case_add_requestor) }
+    let(:requestor2) { service_double.__send__(:case_add_requestor) }
+    it 'only instantiates one copy of the CaseAddRequestor' do
+      expect(requestor1).to be_instance_of(CCMS::CaseAddRequestor)
+      expect(requestor1.object_id).to eq requestor2.object_id
```
reggieb: Could this be simplified to:
```ruby
expect(requestor1).to eq(requestor2)
```

PR #508, opened 2019-05-16
AP 569 update types of income pages

reggieb:
[Jira AP-569](https://dsdmoj.atlassian.net/browse/AP-569)

This add a None Selected checkbox to the Update Types of Income Page in the Citizen Journey.

JavaScript is used to deselect selected checkboxes when the None Selected checkbox is checked.

An error is added if none of the checkboxes is selected.

### DIFF
--- a/app/assets/javascripts/checkbox_control.es6
+++ b/app/assets/javascripts/checkbox_control.es6
@@ -1,12 +1,49 @@
 $(function() {
+
+  /*
+    Usage:
+      Required: A group of checkboxes,
+        and another control checkbox outside the group
+
+      Enclose the checkbox group with a div of class "deselect-group"
+        and set it's "data-deselect-ctrl" to the jquery identifier for the control checkbox
+  */
   $('.deselect-group').each(function() {
     let container = $(this);
     let control = $(container.data('deselect-ctrl'));
-    if(control.length) {
-      control.change( function () {
-        if(this.checked) {
-          container.find("input:checkbox").prop("checked", !this.checked);
-        };
+    let checkboxMemory = [];
+    let checkboxes = container.find("input:checkbox");
+    let controlExists = control.length
+
+    if(controlExists) {
+      /*
+        Monitor changes of the control
+        For each checkbox within the container,
+        If control is selected:
+          remember how checkbox was set
+          and set checkbox to false
+        If deselected:
+          Set each checkbox based on its remembered state
+      */
+      control.change( function() {
+        let controlChecked = this.checked;
+        checkboxes.each(function(index) {
+          let checkbox = $(this);
+          if(controlChecked) {
+            checkboxMemory[index] = this.checked;
+            checkbox.prop("checked", false);
+          } else {
+            checkbox.prop("checked", checkboxMemory[index]);
+          };
+        });
+      });
+
+      /*
+        Monitor changes to the checkboxes within the container.
+        If one is selected, deselect the control.
+      */
+      checkboxes.change( function() {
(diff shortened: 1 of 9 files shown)


app/helpers/gov_uk_form_helper.rb
```diff
@@ -21,11 +21,24 @@ def govuk_form_group(show_error_if: nil, hint_id: nil, input: nil, &block)
...
+  #   <%= govuk_fieldset_header 'My Heading' %>
+  #
+  # or
+  #
+  #   <%= govuk_fieldset_header do %>
+  #     <h1 class="govuk-fieldset__heading">My Heading</h1>
+  #   <% end %>
+  #
+  # Both result in the same output
+  def govuk_fieldset_header(text = nil, padding_below: nil, &block)
```
reggieb: I added the wrapping functionality so that both heading (h1 and h2) required on the new page, could be contained within the fieldset legend.

spec/requests/citizens/identify_types_of_incomes_spec.rb
```diff
@@ -69,6 +77,33 @@
...
+      it 'redirects to the next step' do
+        expect(subject).to redirect_to(flow_forward_path)
+      end
+
+      context 'and application has transactions' do
+        let(:legal_aid_application) do
+          create :legal_aid_application, :with_applicant, transaction_types: income_types
+        end
+
+        it 'does not add transaction types to the application' do
```
Kalagan:
```suggestion
        it 'removes transaction types from the application' do
```
Would this make more sense?

app/assets/javascripts/checkbox_control.es6
```diff
@@ -0,0 +1,50 @@
...
+        and set it's "data-deselect-ctrl" to the jquery identifier for the control checkbox
+  */
+  $('.deselect-group').each(function() {
+    let container = $(this);
+    let control = $(container.data('deselect-ctrl'));
+    let checkboxMemory = [];
+    let checkboxes = container.find("input:checkbox");
+    let controlExists = control.length
+
+    if(controlExists) {
```
Kalagan: would a return early statement be better?
```javascript
if(!control.length) return
```
I think the `controlExists` variable is unnecessary and using `control.length` is clear enough but feel free to keep it

app/assets/javascripts/checkbox_control.es6
```diff
@@ -0,0 +1,50 @@
...
+
+      Enclose the checkbox group with a div of class "deselect-group"
+        and set it's "data-deselect-ctrl" to the jquery identifier for the control checkbox
+  */
+  $('.deselect-group').each(function() {
+    let container = $(this);
+    let control = $(container.data('deselect-ctrl'));
+    let checkboxMemory = [];
+    let checkboxes = container.find("input:checkbox");
+    let controlExists = control.length
```
Kalagan: This is the only line where `;` is missing.
But `;` is actually not necessary so you can remove them completely

app/assets/javascripts/checkbox_control.es6
reggieb: My lack of familiarity with JavaScript idioms. `thing.length` returning `nil` if `thing` doesn't exist on page always makes me have to think "what's that doing?".

I hadn't realised you could do `if(something)) return`. I'll try that now.

app/assets/javascripts/checkbox_control.es6
reggieb: Old dog and all that :). I'll add the one rather than remove the many.

PR #518, opened 2019-05-20
AP-592 add other type of payment for provider

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-592)

Implement the **Identify types of outgoings** page on the provider flow.
Move the view to a partial in order to reuse the html from the applicant flow

## Checklist

### DIFF
--- a/app/controllers/providers/identify_types_of_outgoings_controller.rb
+++ b/app/controllers/providers/identify_types_of_outgoings_controller.rb
@@ -5,7 +5,7 @@ module Providers
     def show; end
 
     def update
-      return continue_or_draft if none_selected? || transactions_added
+      return continue_or_draft if none_selected? || transactions_added || draft_selected?
 
       legal_aid_application.errors.add :base, :none_selected
       render :show
--- a/app/controllers/providers/identify_types_of_outgoings_controller.rb
+++ b/app/controllers/providers/identify_types_of_outgoings_controller.rb
@@ -1,6 +1,42 @@
 module Providers
   class IdentifyTypesOfOutgoingsController < ProviderBaseController
-    def show
+    before_action :authorize_legal_aid_application
+
+    def show; end
+
+    def update
+      return continue_or_draft if none_selected? || transactions_added
+
+      legal_aid_application.errors.add :base, :none_selected
+      render :show
+    end
+
+    private
+
+    def legal_aid_application_params
+      params.require(:legal_aid_application).permit(transaction_type_ids: [])
+    end
+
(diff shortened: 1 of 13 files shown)


app/controllers/providers/identify_types_of_outgoings_controller.rb
```diff
@@ -1,6 +1,42 @@
 module Providers
   class IdentifyTypesOfOutgoingsController < ProviderBaseController
-    def show
+    before_action :authorize_legal_aid_application
+
+    def show; end
+
+    def update
+      return continue_or_draft if none_selected? || transactions_added
```
reggieb: If the user selects "save as draft" without selecting anything, should they see an error? I assumed not. This version of this line behaves that way:
```ruby
return continue_or_draft if draft_selected? || none_selected? || transactions_added
```

app/views/shared/forms/_identify_types_of_outgoings_form.html.erb
```diff
@@ -0,0 +1,36 @@
...
+      <div class="govuk-checkboxes__item">
+        <%= check_box_tag :none_selected, 'true', false, class: 'govuk-checkboxes__input' %>
+        <%= label_tag :none_selected, t('.none_selected'), class: 'govuk-label govuk-checkboxes__label' %>
+      </div>
+    </div>
+
+  <% end %>
+
+  <%= next_action_buttons(
+        show_draft: local_assigns.key?(:show_draft) ? show_draft : false,
```
reggieb: Hmmm. Perhaps that's the way I should have done it in my PR. I've created a helper method that checks which journey we're in, and uses that to determine whether the draft button is required: `show_draft: current_journey?(:providers)`.

Kalagan: I think both approaches are good. Maybe yours is better actually...
You also had to deal with locales different depending on the journey.
I used this pattern because it is used everywhere else:
image

app/controllers/providers/identify_types_of_outgoings_controller.rb
Kalagan: Yes, I agree with you. I'll add that.

app/views/shared/forms/_identify_types_of_outgoings_form.html.erb
reggieb: Keep yours the way it is. I'll update mine to match the current code base.

PR #523, opened 2019-05-22
AP-510 new "interruption panel" page - restyling of check_benefits

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-510)

- New SCSS file for interruption panel styles.
- Changed HTML of `providers/check_benefits/index.html.erb`.
- Changed text in `providers.yml` as per ticket.


config/locales/en/providers.yml
```diff
@@ -62,16 +62,17 @@ en:
...
           title: must complete a financial assessment
+          paragraph: 'This is because they do not receive benefits that qualify for legal aid.'
+          next_1: Your client will need to answer questions about their financial situation.
+          next_2: 'We''ll ask you to give them access to the online assessment which they can complete at home or at another location.'
         positive_result:
-          body: Your client receives benefits that qualify for legal aid.
-          next: Your client will have to declare whether they have any savings, own a home or anything else that can pay for legal aid.
-          title: receives a qualifying benefit
+          title: receives benefits that qualify for legal aid
+          paragraph: 'You''ll need to:'
```
willc-work: Would this be better as ```You will need to``` instead of ```You'll```?  GDS suggests avoiding contractions

MalcolmVonMoJ: This is one for Jim.  He, Justin, Vero etc are all looking at this ticket again.  So there might be some changes like that coming...

Kalagan: FYI, if you use double quotes you won't have to escape the apostrophe.
`paragraph: "You'll need to:"`

MalcolmVonMoJ: They have re-looked at this and no changes.  

You can also see Jim is happy to go ahead with the current wording. So I guess this contraction is okay.

MalcolmVonMoJ: I was just copying previous notation of single quotes and escaping that is used elsewhere.

app/views/providers/check_benefits/index.html.erb
```diff
@@ -1,31 +1,39 @@
-<%= content_for(:panel) do %>
-  <div class="govuk-panel govuk-panel--confirmation">
-    <h2 class="govuk-panel__title">
+<div class="interruption-panel">
+  <%= page_template page_title: '', column_width: 'full'  do %>
+<%
+# blank title - title dealt with below
```
reggieb: `page_template` passes `page_title` to `content_for(:page_title)`. This block has two main functions:

1. It is used in to populate the return from the helper method `page_heading`, which itself is used in the _default_ page template to add an `<h1>` tag at the top of the page.
2. It populates the page title in the page layout: `HTML <head> <title>`

By passing an empty string into "page_title" you've effectively removed the unique HTML page title from this page.

If you want to alter where the page heading is displayed on the page, use the _basic_ template, and then insert `page_heading` at the location where you want it. 

I'm not quite sure what your main heading is. I think it is `@legal_aid_application.applicant.full_name + t("#{result_namespace}.title")`

However, that's not a good way to use a translation. In other languages, the sentence may not start with the person's name. For (an obscure example) "Fred Blogs must complete a financial assessment" in Cebuano is "Ang Fred Blogs kinahanglang makompleto ang pinansyal nga pagtasa". So it better to pass the name into the translation, so that it can be placed into the correct part of the sentence. The translation code becomes:
```
title: "%{name} must complete a financial assessment"
```
And it's used like this:
```erb
t("#{result_namespace}.title", name: @legal_aid_application.applicant.full_name)
```

So here:
```erb
<%= page_template(
          page_title: t("#{result_namespace}.title", name: @legal_aid_application.applicant.full_name), 
          column_width: 'full',
          template: :basic
        ) do %>
```
Then the `<h1>` on lines 6-9 is replaced with:
```erb
<%= page_heading(size: 'l') %>
```

app/views/providers/check_benefits/index.html.erb
```diff
@@ -1,31 +1,39 @@
...
-        show_draft: true,
-        continue_button_text: t('generic.continue')
-      ) %>
-<% end %>
+    </h1>
+    <div class="govuk-grid-column-two-thirds">
+      <% if result_namespace == '.negative_result' %>
+        <p class="govuk-body">
+          <%= t "#{result_namespace}.paragraph" %>
+        </p>
```
reggieb: So the paragraph appears above the `<h2>` on a negative result, and below it on a positive result. That seems a little weird and adds quite a lot of complication to the page.

config/locales/en/providers.yml
reggieb: YAML automatically converts the value to a string so you don't need put the text into quotes unless there is a special character in the string (for example `:`) which YAML has problems parsing. So:
```yaml
next_2: We'll ask you to give them access to the online assessment which they can complete at home or at another location.
paragraph: "You'll need to:"
```

app/views/providers/check_benefits/index.html.erb
MalcolmVonMoJ: This was my way of reducing the number of variables in `providers.yml`.  If that isn't important, I can make them separate things.

app/views/providers/check_benefits/index.html.erb
MalcolmVonMoJ: Thanks, will re-do.

app/views/providers/check_benefits/index.html.erb
reggieb: Nope - now I'm totally confused as to what you are doing.

Let me be a little clearer. The code above - lines 12-14 (that only appears if `result_namespace == '.negative_result'`):
```erb
<p class="govuk-body">
  <%= t "#{result_namespace}.paragraph" %>
</p>
```
is exactly the same as the code between lines 20-22 in a section that only appears if `result_namespace == '.positive_result'`

As the `.next_heading` code is below the first case, and above the second, and assuming the only `result_namespace` values are `'.positive_result'` and `.negative_result` all this effectively does is move the heading below or above the paragraph code respectively.

If as I expect, the heading can be above the paragraph in both conditions this whole section (lines 11-30) can be simplified to:
```erb
<h2 class="govuk-heading-m">
  <%= t '.next_heading' %>
</h2>
<p class="govuk-body">
  <%= t "#{result_namespace}.paragraph" %>
</p>
<% if result_namespace == '.positive_result' %>
  <ul class="govuk-list govuk-list--bullet">
    <li><%= t "#{result_namespace}.next_1" %></li>
    <li><%= t "#{result_namespace}.next_2" %></li>
  </ul>
<% else %>
  <p class="govuk-body"><%= t "#{result_namespace}.next_1" %></p>
  <p class="govuk-body"><%= t "#{result_namespace}.next_2" %></p>
<% end %>
```

MalcolmVonMoJ: No, the _next heading_ is above the paragraph in one situation and below in the other situation.  

The paragraphs have different text, I have called them both _paragraph_ to reduce the number of variables in the yml file. 

I'll re-work this all tomorrow, splitting them into two variables, and hopefully it'll be clearer then.  Will let you know when it's done for you to look at again.

config/locales/en/providers.yml
```diff
@@ -62,16 +62,17 @@ en:
...
+          title: "%{name} must complete a financial assessment"
+          next_1: Your client will need to answer questions about their financial situation.
+          next_2: 'We''ll ask you to give them access to the online assessment which they can complete at home or at another location.'
         positive_result:
-          body: Your client receives benefits that qualify for legal aid.
-          next: Your client will have to declare whether they have any savings, own a home or anything else that can pay for legal aid.
-          title: receives a qualifying benefit
+          title: "%{name} receives benefits that qualify for legal aid"
+          next_1: 'answer questions about your client''s property, savings and other assets'
+          next_2: provide details of the case
```
reggieb: Use keys that describe the text content rather than generic names - it is then much clearer in the markup what is being passed in. So I'd suggest:
```yaml
what_happens_next: What happens next
because_not_receiving_benefits: This is because they do not receive benefits that qualify for
legal aid.
provider_needs_to: "You'll need to:"
negative_result:
  must_complete_assessment: "%{name} must complete a financial assessment"
  must_answer_financial_questions: Your client will need to answer questions about their financial situation.
  give_access_to_assessment: We'll ask you to give them access to the online assessment which they can complete at home or at another location.
positive_result:
  citizen_received_benefits: "%{name} receives benefits that qualify for legal aid"
  must_answer_saving_questions: answer questions about your client's property, savings and other assets
  must_provide_case_details: provide details of the case
```
With those keys someone looking at the markup, won't need to refer to the translation file all the time to get a gist of what is being displayed on the page.

features/providers/step_definitions/civil_journey_steps.rb
```diff
@@ -346,8 +346,12 @@
   expect(page).to have_content("Enter your client's home address")
 end
 
-Then('I am on the benefit check results page') do
-  expect(page).to have_content("Your client's tax and benefits status")
+Then('I am on the benefit check results page and the benefits qualify') do
+  expect(page).to have_content('receives benefits that qualify for legal aid')
+end
```
reggieb: Rather than create page specific step definitions there is already a step definition that would more clearly expresses what is happening here: "Then I should be on a page showing"

So remove these two step definitions, and in the features when you want to check the citizen received benefits use:
```
Then I should be on a page showing "receives benefits that qualify for legal aid"
```
And when the benefits don't qualify:
```
Then I should be on a page showing "must complete a financial assessment"
```

config/locales/en/providers.yml
MalcolmVonMoJ: I'm not sure I should change `title` as this is used in the page template, and if I am using completely different keys for `positive_result` and `negative_result` then there's no need for that distinction.  

So I have changed the global keys as you suggested, and changed the keys with positive/negative differences to keys that make sense in both situations.

features/providers/step_definitions/civil_journey_steps.rb
MalcolmVonMoJ: I'm not quite following this I'm afraid.  
I changed it like this as there was no common text between the two pages now, aside from something too generic like "what happens next".  This seemed akin to removing the test completely, so we opted for this approach, splitting up the two distinct pages.

features/providers/civil_application_journey.feature
```diff
@@ -217,8 +216,7 @@ Feature: Civil application journeys
     Then I click 'Save and continue'
     Then I should be on a page showing 'Check your answers'
     Then I click 'Save and continue'
-    Then I am on the benefit check results page
-    Then I see a notice saying that the citizen does not receive benefits
+    Then I am on the benefit check results page and the benefits do not qualify
```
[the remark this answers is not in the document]
MalcolmVonMoJ: Okay.  I understand now.  Thanks.  

There are a few other one-line tests like that in the `civil_journey_steps.rb` file, perhaps they could be similarly cleaned up.  

Made those changes and am committing now.

PR #524, opened 2019-05-22
AP-617 Create object to store and return thresholds

reggieb:
[Jira AP-617](https://dsdmoj.atlassian.net/browse/AP-617)

Creates a generic class that allows data to be retrieved from a YAML file. Then uses a sub-class of that class to retrieve thresholds set in `config/ccms/threasholds.yml`

With the current configuration you can do this:

```
2.5.3 :001 > CCMS::Threshold.threshold :gross_income_upper
 => {:children=>{:one=>1, :two=>3, :many=>7}} 
2.5.3 :002 > CCMS::Threshold.threshold :capital_upper
 => 10000 
```


reggieb:
Just realised I've done this in the wrong app! Pillock me!

PR #527, opened 2019-05-24
AP-578 - Poll CCMS case creation

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-578)

Add a new service to poll CCMS to establish the success of a request to create a case in CCMS.

## Checklist

### DIFF
--- a/app/services/ccms/check_case_status_service.rb
+++ b/app/services/ccms/check_case_status_service.rb
@@ -1,0 +1,33 @@
+module CCMS
+  class CheckCaseStatusService < BaseSubmissionService
+    def call
+      tx_id = case_add_status_requestor.transaction_request_id
+      submission.poll_count += 1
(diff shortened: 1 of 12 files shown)


app/services/ccms/check_case_status_service.rb
```diff
@@ -0,0 +1,33 @@
+module CCMS
+  class CheckCaseStatusService < BaseSubmissionService
+    def call
+      tx_id = case_add_status_requestor.transaction_request_id
+      submission.poll_count += 1
+      parser = CaseAddStatusResponseParser.new(tx_id, response)
```
stephenrichards: I think we should have separate poll counts for each type of polling, i.e. add_client, add_case, and in the future upload documents (although we'll probably want to do that on a per-file basis and store it in the hash of file upload details)

PR #542, opened 2019-05-29
Ap 543 financial page content

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-543)

Adds the users email to the 'about financial assessments' page and includes a change link.
The change link goes to a new page that allows thew user to update the email address.

## Checklist


app/views/providers/about_the_financial_assessments/show.html.erb
```diff
@@ -1,22 +1,30 @@
...
   </h2>
 
   <p>
-    <%= t '.section_1.text' %>
+    <%= t '.section_2.text' %>
   </p>
 
+  <%= check_answer_email_link(
+        name: :email,
+        url: providers_legal_aid_application_update_email_address_path(anchor: :email), # currently not using anchor may need fo rnew page
```
Kalagan: I don't understand this comment `currently not using anchor may need fo rnew page`
This link IS using an anchor.
Is a full stop missing?

app/views/providers/update_email_addresses/show.html.erb
```diff
@@ -0,0 +1,28 @@
...
+      ) do |form| %>
+
+    <div class="govuk-!-padding-bottom-6"></div>
+
+    <%= form.govuk_text_field(
+          :email,
+          label: { text: t('providers.applicants.show.email_label'), size: :m },
+          hint: t('helpers.hint.applicant.email'), size: :m,
+          class: 'govuk-input govuk-!-width-three-quarters'
+        ) %>
```
Kalagan: `govuk_text_field` by default grabs the label from `activemodel.attributes.applicant.email` and the hint from `helpers.hint.applicant.email`. And they happen to be exactly what you need.

Also, by default the class `govuk-input` is always applied.

So you can just write:
```ruby
    <%= form.govuk_text_field(
          :email,
          label: { size: :m },
          class: 'govuk-!-width-three-quarters'
        ) %>
```

config/routes.rb
```diff
@@ -99,6 +99,7 @@
       end
       resources :restrictions, only: %i[index create] # as multiple restrictions
       resource :about_the_financial_assessment, only: %i[show update]
+      resource :update_email_address, only: %i[show update]
```
Kalagan: I would just call this route `email_address` instead of `update_email_address`. It feels more REST/RoR.
And it also fits with the other routes.
For example we have `property_value`, not `update_property_value`,
`applicant`, not `update_applicant`, etc...

app/services/flow/flows/provider_start.rb
```diff
@@ -48,6 +48,10 @@ class ProviderStart < FlowSteps
           path: ->(application) { urls.providers_legal_aid_application_about_the_financial_assessment_path(application) },
           forward: :application_confirmations
         },
+        update_email_addresses: {
+          path: ->(application) { urls.providers_legal_aid_application_update_email_address_path(application) },
+          forward: :application_confirmations
```
Kalagan: According to the Jira ticket, `forward` should be `about_the_financial_assessments`

features/providers/civil_application_journey.feature
```diff
@@ -294,6 +294,23 @@ Feature: Civil application journeys
...
+    Then I click 'Continue'
+    Then I should be on a page showing 'Does your client use online banking?'
+    Then I choose "Yes"
+    Then I click 'Save and continue'
+    Then I am on the About the Financial Assessment page
+    And I click Check Your Answers Change link for 'Email'
+    Then I should be on a page showing 'Email address'
+    Then I fill 'email' with 'test@test.com'
+    Then I click 'Save and continue'
+    Then I am on the application confirmation page
```
Kalagan: As my previous comment, I think the next page should be the `About the online financial assessment
` page.
In that case, you could check that the email has been updated:
`Then I am on the About the Financial Assessment page`
`Then I should be on a page showing 'test@test.com'`

spec/requests/providers/update_email_addresses_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'rails_helper'
+
+RSpec.describe 'update client email address before application confirmation', type: :request do
+  let(:application) { create(:legal_aid_application) }
+  let(:application_id) { application.id }
+  let(:provider) { application.provider }
+
+  describe 'GET /providers/applications/:legal_aid_application_id/update_email_address' do
+    subject { get "/providers/applications/#{application_id}/update_email_address" }
+
```
Kalagan: You could add this:
```ruby
    context 'when the provider is not authenticated' do
      before { subject }
      it_behaves_like 'a provider not authenticated'
    end
```

app/controllers/providers/update_email_addresses_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def applicant
+      legal_aid_application.applicant || legal_aid_application.build_applicant
+    end
+
+    def form_params
+      merge_with_model(applicant) do
+        params.require(:applicant).permit(
+          :first_name, :last_name, :dob_day, :dob_month, :dob_year,
+          :national_insurance_number, :email
```
Kalagan: only the `email` is sent so you can remove the other ones.

app/views/providers/update_email_addresses/show.html.erb
```diff
@@ -0,0 +1,28 @@
+<%= page_template(
+      page_title: t('.page_title'),
+      back_link: { text: t('generic.back') },
```
Kalagan: `t('generic.back')` is the default value so you can remove this line

app/views/providers/about_the_financial_assessments/show.html.erb
```diff
@@ -1,22 +1,30 @@
...
-    <%= t '.section_1.heading' %>
+    <%= t '.section_2.heading' %>
   </h2>
 
   <p>
-    <%= t '.section_1.text' %>
+    <%= t '.section_2.text' %>
   </p>
 
+  <%= check_answer_email_link(
```
Kalagan: I don't think this link requires a dedicated method and a new partial in the shared folder. It's only used once.
In my opinion, you could insert the content of the partial directly here.
But if you want to create a partial, I recommend adding it in the same folder and calling it directly.
Like Rob did on the `respondent` page for example:
https://github.com/ministryofjustice/laa-apply-for-legal-aid/tree/master/app/views/providers/respondents

app/views/shared/check_answers/_answer_no_question.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <div class="govuk-grid-column-two-thirds">
+    <h2 class="govuk-body"><%= answer %></h2>
+  </div>
+    <div class="govuk-grid-column-one-third govuk-summary-list--no-border align-text-right">
+      <p>
+        <%= link_to(url, class: 'govuk-link change-link') do %>
+          <%= t('generic.change') %>
+      </p>
+    </div>
+  <% end %>
```
Kalagan: The two closing tags `</p>` and `</div>` are in the middle of the ruby block.

spec/requests/providers/update_email_addresses_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+        subject
+        expect(response.body).to include(I18n.translate('providers.applicants.show.email_label'))
+        expect(unescaped_response_body).to include(I18n.translate('helpers.hint.applicant.email'))
+      end
+    end
+  end
+
+  describe 'PATCH /providers/applications/:legal_aid_application_id/update_email_address' do
+    subject { patch "/providers/applications/#{application_id}/update_email_address", params: params }
+
```
Kalagan: You could also add
```ruby
    context 'when the provider is not authenticated' do
      before { subject }
      it_behaves_like 'a provider not authenticated'
    end
```
here

spec/requests/providers/update_email_addresses_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+    let(:provider) { application.provider }
+    let(:params) do
+      {
+        applicant: {
+          first_name: 'John',
+          last_name: 'Doe',
+          national_insurance_number: 'AA 12 34 56 C',
+          dob_year: '1981',
+          dob_month: '07',
+          dob_day: '11',
```
Kalagan: This doesn't reflect the behaviour of the page. There is only the `email`

features/providers/civil_application_journey.feature
```diff
@@ -294,6 +294,23 @@ Feature: Civil application journeys
...
+  Scenario: I want to change email address from the about financial assessment page
+    Given I complete the journey as far as check your answers
+    Then I click 'Save and continue'
+    Then I am on the benefit check results page
+    Then I click 'Continue'
+    Then I should be on a page showing 'Does your client use online banking?'
+    Then I choose "Yes"
+    Then I click 'Save and continue'
+    Then I am on the About the Financial Assessment page
+    And I click Check Your Answers Change link for 'Email'
```
Kalagan: You could use `Then I click link 'Change'` because there is only one `Change` link on the page.
I that case you wouldn't need the HTML to use `id="app-check-your-answers__<%= name %>">`

app/controllers/providers/update_email_addresses_controller.rb
```diff
@@ -0,0 +1,29 @@
+module Providers
+  class UpdateEmailAddressesController < ProviderBaseController
+    def show
+      authorize legal_aid_application
+      @form = Applicants::BasicDetailsForm.new(model: applicant)
```
Kalagan: Creating a new form would make sense I think. Because on this page we are only editing the email.

app/views/providers/about_the_financial_assessments/show.html.erb
willc-work: I meant to delete this

app/controllers/providers/update_email_addresses_controller.rb
willc-work: There actually is an email form already so I have used that.

Kalagan: `Applicants::BasicDetailsForm` is not an email form but more a form for applicant's details which happens to contains `email`. 
The risk in using this form just for the email is that it could potentially trigger some `presence` validation errors on the other attributes.

So it would make sense to create another form specific to this page (i.e. `Applicants::EmailForm`).

On the other hand, because this new email page is always after the applicant's details page, the other attributes (date of birth, NIN, first_name, last_name) should always be populated and valid so no other errors than emails errors should be raised.
So I guess it's fine to reuse this same form even though it's a bit strange to re-validate attributes that are not on the page.

Kalagan: Oops. I think I misunderstand your previous comment. I see now that you are re-using an already existing `Applicants::EmailForm`. I didn't realise there was already one

spec/requests/providers/email_addresses_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe 'update client email address before application confirmation', type: :request do
+  let(:application) { create(:legal_aid_application) }
+  let(:application_id) { application.id }
+  let(:provider) { application.provider }
+
+  describe 'GET /providers/applications/:legal_aid_application_id/email_address' do
+    subject { get "/providers/applications/#{application_id}/email_address" }
+
```
Kalagan: You could add this also in this section:
```ruby
    context 'when the provider is not authenticated' do
      before { subject }
      it_behaves_like 'a provider not authenticated'
    end
```

app/views/providers/update_email_addresses/show.html.erb
Kalagan: still applies

PR #545, opened 2019-05-30
AP-595 AP-597 non passported check vehicle answers

Kalagan:
## What

[AP-595](https://dsdmoj.atlassian.net/browse/AP-595)
[AP-597](https://dsdmoj.atlassian.net/browse/AP-597)

Adding links to the vehicle questions in the non-passported provider's check answers pages.

## Checklist

### DIFF
--- a/features/providers/complete_and_check_means_answers.feature
+++ b/features/providers/complete_and_check_means_answers.feature
@@ -48,5 +48,34 @@ Feature: Completing and checking means answers backwards and forwards
     Then I click 'Save and continue'
     Then I should be on a page showing "Check your answers"
 
+  @javascript
+  Scenario: I change the applicant answer about owning a vehicle
+    Given I am checking the applicant's means answers
+    Then I click Check Your Answers Change link for 'Vehicles'
+    Then I should be on a page showing 'Does your client own a vehicle?'
+    Then I choose 'No'
+    Then I click 'Save and continue'
+    Then I should be on a page showing 'Check your answers'
+
+  @javascript
+  Scenario: I change the applicant answers about details of their vehicle
+    Given I am checking the applicant's means answers
+    Then I click Check Your Answers Change link for 'Vehicles'
+    Then I choose 'Yes'
+    Then I click 'Save and continue'
+    Then I should be on a page showing "What is the estimated value of the vehicle?"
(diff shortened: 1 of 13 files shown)


spec/requests/providers/means_summaries_spec.rb
```diff
@@ -31,10 +33,36 @@
       expect(response.body).to include(ActiveSupport::NumberHelper.number_to_currency(total_amount))
     end
 
+    it 'displays the correct vehicles details' do
+      expect(response.body).to include(number_to_currency(vehicle.estimated_value, unit: '£'))
+      expect(response.body).to include(number_to_currency(vehicle.payment_remaining, unit: '£'))
+      expect(response.body).to include(vehicle.purchased_on.to_s)
+      expect(response.body).to include(I18n.translate('shared.check_answers_vehicles.providers.heading'))
+      expect(response.body).to include(I18n.translate('shared.check_answers_vehicles.providers.own'))
```
reggieb: Really picky - but as the `t` alias is universally used for I18n's `translate` method in views etc, I think you can use it here too `I18n.t('shared.check_answers_vehicles.providers.own')` and elsewhere in this PR.

features/providers/complete_and_check_means_answers.feature
```diff
@@ -48,5 +48,41 @@ Feature: Completing and checking means answers backwards and forwards
...
+  @javascript @vcr
+  Scenario: I view the client completed means answers and I change vehicle questions
+    Given I start the journey as far as the client completed means page
+    Then I should be on a page showing 'Your client has completed their financial assessment'
+    Then I click 'Continue'
+    Then I should be on a page showing "Your client's income"
+    Then I click 'Save and continue'
+    Then I should be on a page showing "Your client's regular payments"
+    Then I click 'Save and continue'
+    Then I should be on a page showing 'Check your answers'
```
reggieb: I wonder whether, if there was an "I start the journey as far as check your completed means page" step defined, this scenario could be broken down into smaller scenarios that each described one of the sub-journeys (Changing declared income, Change "Vehicles", etc.). That way we don't have such a large scenario where each sub-journey is so dependent on the success of the previous one.

PR #546, opened 2019-05-30
AP-599 Table header restyling

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-599)

Restyled headers for table.  Re-did select all link.  Added padding to cells.  Made table full-width.  

## Checklist


app/views/providers/transactions/show.html.erb
```diff
@@ -10,33 +10,30 @@
...
+          <th data-sort-type="date" class="govuk-table__header sort header-sort-desc table-combine_right_if_narrow" scope="col">
             <span class="govuk-visually-hidden">Sort by </span>
-            <span class="aria-sort-description"><%= t('.col_date') %></span>
-            <span class="table-hidden_right_title"> &amp; <%= t('.col_description') %></span>
+            <span class="aria-sort-description"><%= t('.col_date') %></span><span class="table-hidden_right_title"> &amp; <%= t('.col_description') %></span><span class="sort-direction-indicator"><span class="screen-reader-sort-indicator govuk-visually-hidden"> (currently sorted from newest to oldest)</span></span>
           </th>
           <th data-sort-type="alphabetic" class="govuk-table__header sort" scope="col">
             <span class="govuk-visually-hidden">Sort by </span>
-            <span class="aria-sort-description"><%= t('.col_description') %></span>
+            <span class="aria-sort-description"><%= t('.col_description') %></span><span class="sort-direction-indicator"><span class="screen-reader-sort-indicator govuk-visually-hidden"></span></span>
```
reggieb: Very minor: The direction indicator span is an additional item. It would be clearer if this was on it's own line, rather than being tacked onto the end of the description span.

app/assets/javascripts/table-select-all.es6
```diff
@@ -25,6 +26,12 @@ $(document).ready(function() {
       return false;
     });
 
+    $('.select-clear-all').keyup(function(ev) {
+      if (ev.which==13 || ev.which==32)  { //on space or return, the column is sorted
```
reggieb: It would be clearer and avoid magic numbers if 13 and 32 were assigned to variables:
```javascript
const spaceKey = 13
const returnKey = 32
if (ev.which == spaceKey || ev.which == returnKey) {
```
Also the default behavior is for a space key press to trigger `click` (see [this](https://stackoverflow.com/questions/27878940/spacebar-triggering-click-event-on-checkbox) for example). Will this cause click to be triggered twice? Is that a problem?

Kalagan: FYI you can use `return false` to prevent the default behaviour of the event.

app/views/providers/transactions/show.html.erb
MalcolmVonMoJ: I put it on the same line because I thought there might be some issues if there was whitespace between the two `<span>`s.  I can't remember why, but that might be a Monday morning thing.

app/assets/javascripts/table-select-all.es6
MalcolmVonMoJ: I added this in after trying the space bar and return key and seeing it didn't work.  I think this is because I have changed the link from an anchor to a normal header - this was done to increase the click area as inplied in the ticket and so styles could be reused.  

It should not affect the tick boxes themselves, they should still exhibit default behaviour.  

I'll push this to UAT and do my own checks cross-browser as well, just to be sure.

MalcolmVonMoJ: **Cross-browser testing**: Edge submitted page on sort.  I'll re-do to inculde `return false` as recommended.

reggieb: Nicely done.

MalcolmVonMoJ:
Manually reverting unexplained changes, mostly additions, which are not part of this change.

MalcolmVonMoJ:
Too many random changes that I can't explain are being introduced.  Closing PR and starting again from scratch.

PR #548, opened 2019-05-30
AP-579 - Add service to obtain CCMS document ids

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-579)

Create a new service to register all documents that need to be uploaded
to CCMS for an application and receive a document id in return.

If no documents exist for the application, the status is changed to
completed as at that point all relevant data has been sent to CCMS.

## Checklist

### DIFF
--- a/app/models/concerns/ccms_submission_state_machine.rb
+++ b/app/models/concerns/ccms_submission_state_machine.rb
@@ -11,6 +11,8 @@ module CCMSSubmissionStateMachine
       state :applicant_ref_obtained
       state :case_submitted
       state :case_created
+      state :document_ids_obtained
+      state :completed
       state :failed
 
       event :obtain_case_ref do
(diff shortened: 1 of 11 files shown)


app/models/concerns/ccms_submission_state_machine.rb
```diff
@@ -34,12 +36,21 @@ module CCMSSubmissionStateMachine
         transitions from: :case_submitted, to: :case_created
       end
 
+      event :obtain_document_ids do
+        transitions from: :case_created, to: :document_ids_obtained
+      end
+
+      event :complete do
+        transitions from: :case_created, to: :completed
+      end
```
stephenrichards: Shouldn't this be `transitions from: :document_ids_obtained, to: :completed` ?

mpw5: You're right that there will be another transition from :document_ids_obtained to: :completed, but I'll do that in the next story. This one covers the situation where there are no documents, so the status moves straight from :case_created to :completed

PR #558, opened 2019-06-04
Ap 543 financial content

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-543)

Adds the users email to the 'about financial assessments' page and includes a change link.
The change link goes to a new page that allows thew user to update the email address.

## Checklist


app/views/providers/email_addresses/show.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        method: :patch,
+        local: true
+      ) do |form| %>
+
+    <div class="govuk-!-padding-bottom-6"></div>
+
+    <%= form.govuk_text_field(
+          :email,
+          label: { size: :m },
+          class: 'govuk-input govuk-!-width-three-quarters'
```
Kalagan: you shouldn't need to add `govuk-input` here as `govuk_text_field` already adds it by default

app/views/providers/about_the_financial_assessments/show.html.erb
```diff
@@ -1,22 +1,37 @@
...
+    <%= t '.section_2.text' %>
   </p>
 
+    <dl class="govuk-summary-list">
+      <div class="govuk-summary-list__row">
+        <dt class="govuk-summary-list__value">
+          <%= @applicant.email %>
+        </dt>
+        <dd class="govuk-summary-list__actions">
+          <%= link_to(providers_legal_aid_application_email_address_path(anchor: :email), class: 'govuk-link change-link') do %>
```
Kalagan: Would this work?
```ruby
<%= link_to(t('generic.change'), providers_legal_aid_application_email_address_path(anchor: :email), class: 'govuk-link change-link') %>
```

With `link_to` you can pass the `text` as the first parameter https://apidock.com/rails/ActionView/Helpers/UrlHelper/link_to
You don't need a block

features/providers/civil_application_journey.feature
```diff
@@ -267,6 +267,25 @@ Feature: Civil application journeys
...
+    Then I am on the About the Financial Assessment page
+    And I click link 'Change'
+    Then I should be on a page showing 'Email address'
+    Then I fill 'email' with 'test@test.com'
+    Then I click 'Save and continue'
+    Then I am on the About the Financial Assessment page
+    Then I should be on a page showing 'test@test.com'
+
+
+
```
Kalagan: For consistency, I would just leave 1 space between this scenario and the next instead of 3 because.
That's what we do everywhere

PR #561, opened 2019-06-06
AP-601 New SCSS file with style for truncating email address

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-601)

New SCSS file with CSS to do the following:
- max-width that (for narrower displays) prevents email containing field from overlapping with the service title.  
- overflow, vertical alignment added to stop other undesirable effects.  

## Checklist

### DIFF
--- a/app/assets/stylesheets/long-email-truncation.scss
+++ b/app/assets/stylesheets/long-email-truncation.scss
@@ -1,15 +1,34 @@
 /********
-Note: this is only designed to account for narrow screens where the Ruby
-truncation to 20 characters is insufficient to prevent overlap.
+All screens
 /********/
 
 #navigation>li {
-  max-width:calc(50vw - 101px - 30px - 76px - 30px);
-  vertical-align:top;
+  max-width:100%;
   overflow:hidden;
   text-overflow: ellipsis;
-
   &:focus-within {
     color:#0b0c0c;
+  }
+}
+
+/********
+Approx 770px+
+/********/
+
+@media (min-width: 48.0625em) {
+  #navigation>li {
+    max-width:calc(50vw - 101px - 30px - 76px - 30px);
+    vertical-align:top;
+  }
+}
+
+/********
+1020px+
+(as per the govuk-width-container style)
+/********/
+
+@media (min-width: 1020px) {
+  #navigation>li {
+    max-width: 312px;
   }
 }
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/long-email-truncation.scss
```diff
@@ -0,0 +1,15 @@
...
+#navigation>li {
+  max-width:calc(50vw - 101px - 30px - 76px - 30px);
+  vertical-align:top;
+  overflow:hidden;
+  text-overflow: ellipsis;
+
+  &:focus-within {
+    color:#0b0c0c;
+  }
+}
```
reggieb: As this is designed to work on narrow screens, would it be worth wrapping it in a `@media (max-width: 350px)` [type directive](https://www.w3schools.com/cssref/css3_pr_mediaquery.asp), so that it is only applied when the screen width drops below a certain width (350px in my example).

MalcolmVonMoJ: Here, by narrow screens, we don't mean mobile, but screens of > about 770px but < about 920px.  
The media query could be added here to account for > 920px but because the Ruby truncation kicks in at 20 characters it isn't needed.  A string of 20 upper-case W's does not cause issues.  
<770px is dealt with already by the existing mobile styles

MalcolmVonMoJ: I have added the media queries to account for all situations.  This makes the CSS quite robust, dealing with mobile view, narrow desktop/tablet and normal/wide displays.  

Do you think the Ruby truncation to 20 characters is still needed?

MalcolmVonMoJ: Talked with @Kalagan, who agrees that the Ruby truncation is now redundant.  So I have removed that, and added a new test email address.

app/views/shared/_provider_header_link.html.erb
```diff
@@ -1,6 +1,6 @@
 <ul id="navigation" class="govuk-header__navigation" aria-label="Top Level Navigation">
   <li class="govuk-header__navigation-item">
-    <%= link_to(current_provider.username.truncate(20), providers_provider_path, class: 'govuk-header__link') %>
+    <%= link_to(current_provider.username, providers_provider_path, class: 'govuk-header__link') %>
```
reggieb: I presume this is the "Ruby truncation" that you mentioned in your comments. So yes - removing the truncation here makes sense :+1:

MalcolmVonMoJ: Yes.  This done in liaison with Julien.

MalcolmVonMoJ:
Shall I close this and re-do the PR then?

MalcolmVonMoJ:
Replaced with PR 564.

PR #562, opened 2019-06-06
AP556 Admin Confirm Delete

AbdirizakObsiye2018:
Add Admin Confirmation button upon deleting an application or all applications

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-556)

Have a confirmation button on the admin page upon clicking either:
-  the delete button on a legal aid application record
- the delete all applications button

Include the confirmation button into an alert/error summary message, which appears when 'Delete all applications' or 'Delete' buttons are pressed.

Have JavaScript toggle the view of the error summary/confirmation alert message when a button is pressed. Also, have JavaScript to adapt the 'Yes, delete' confirmation button's action route to the data record delete was clicked on or to all applications if 'Delete all applications' was clicked.

## Checklist

### DIFF
--- a/app/views/admin/legal_aid_applications/index.html.erb
+++ b/app/views/admin/legal_aid_applications/index.html.erb
@@ -6,9 +6,12 @@
 
   <div id="confirm-delete" class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="error-summary" data-cancel-text="<%= t('.cancel_delete') %>" hidden>
     <h2 class="govuk-error-summary__title" id="error-summary-title">
-      <span id="delete-message"></span>
-      <span id="delete-case-details"></span>
+      <span id="delete-message"><% t('.warning.delete') %></span>
     </h2>
+    <div class="govuk-inset-text" id="delete-case-details" hidden>
+      <p id="delete-case-name" class="govuk-body-m"></p>
+      <p id="delete-case-ref" class="govuk-body-m"></p>
+    </div>
     <div class="govuk-error-summary__body">
       <%= t('.warning.delete_consequence') %>
     </div>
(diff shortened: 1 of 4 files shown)


app/views/admin/legal_aid_applications/index.html.erb
```diff
@@ -4,21 +4,40 @@
...
           class: 'govuk-button'
         ) if create_test_applications_enabled? %>
 
-  <%= button_to(
-        t('.delete_all'),
-        destroy_all_admin_legal_aid_applications_path,
-        method: :delete,
-        class: 'govuk-button destructive'
-      ) if @applications.present? && destroy_enabled? %>
+    <%= submit_tag(
```
reggieb: I think this should be button rather than a submit input - as there is no enclosing form.

reggieb:
Further to point one - putting the url construction in the JS couples the JS to this implementation. It means we won't be easily able to reuse the same code on another page.

app/views/admin/legal_aid_applications/index.html.erb
```diff
@@ -4,21 +4,44 @@
       back_link: :none
     ) do %>
 
-  <p><%= link_to t('.edit_settings'), admin_settings_path, class: 'govuk-link' %></p>
+  <div id="confirm-delete" class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="error-summary" data-cancel-text="<%= t('.cancel_delete') %>" hidden>
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      <span id="delete-message"><% t('.warning.delete') %></span>
+    </h2>
```
reggieb: I think you have too many ids in the page. Are they really all necessary? Remember, with with a lot of ids there is a lot of uniqueness to be maintained.

PR #565, opened 2019-06-07
Tidy up ccms services

mpw5:
## What

Tidy up CCMS services: 

  - remove unused methods
  - remove nocov from `call` methods
  - add tests for `call` methods, mocking soap calls
  - minor naming corrections 

Also improve functionality of `pretty_postcode` method to prevent multiple spaces being added.

## Checklist

### DIFF
--- a/app/models/address.rb
+++ b/app/models/address.rb
@@ -26,7 +26,11 @@ class Address < ApplicationRecord
   end
 
   def pretty_postcode
-    postcode.insert(-4, ' ')
+    pretty_postcode? ? postcode : postcode.insert(-4, ' ')
+  end
+
+  def pretty_postcode?
+    postcode[-4] == ' '
   end
 
   def first_lines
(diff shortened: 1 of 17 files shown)


app/models/address.rb
```diff
@@ -26,7 +26,11 @@ def full_address
   end
 
   def pretty_postcode
-    postcode.insert(-4, ' ')
+    pretty_postcode? ? postcode : postcode.insert(-4, ' ')
```
Kalagan: I wonder if we could add this code to the `normalize_postcode` method which is run before the address is saved:
https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/models/address.rb#L59
```ruby
  def normalize_postcode
    return if postcode.blank?

    postcode.delete!(' ')
    postcode.upcase!
+   postcode.insert(-4, ' ') unless postcode[-4] == ' '
  end
```
Then we don't need the `pretty_postcode` method at all and we can use directly `postcode`.

mpw5: Thanks. That makes sense but breaks a lot of existing validations and tests. I think it might also cause problems with calls to the address lookup API. I think I'll merge this one now and then look into making those changes in a separate PR later.

PR #566, opened 2019-06-10
AP-563 Category column in transactions table

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-563)

Added new column.
CSS changes to account for narrow screens
CSS addition to account for new "tag" in category column
HTML changes for new column, and an annoying pseudo column needed for narrow screens.
JS changes for new functionality and to account for pseudo column.
New entries in YML files.

## Checklist

### DIFF
--- a/app/assets/javascripts/table-select-all.es6
+++ b/app/assets/javascripts/table-select-all.es6
@@ -13,6 +13,14 @@ $(document).ready(function() {
       }
     }
 
+    function setCategory(row) {
+      if (row.find('input:checked').length) {
+        row.find(".table-category-vacant").addClass("table-category-preview");
+      } else {
+        row.find(".table-category-vacant").removeClass("table-category-preview");
+      }
+    }
+
     $('.select-clear-all').click(function() {
       const table = $(this).parents('table');
       if($(this).hasClass('clear-all')) {
@@ -36,7 +44,8 @@ $(document).ready(function() {
     });
 
     $('table input').click(function() {
-      resetSelectClearLink($(this).parents('table'))
+      resetSelectClearLink($(this).parents('table'));
+      setCategory($(this).parents('tr'))
     });
 
     resetSelectClearLink($('.select-clear-all').parents('table'))
--- a/app/assets/javascripts/table-select-all.es6
+++ b/app/assets/javascripts/table-select-all.es6
@@ -14,10 +14,11 @@ $(document).ready(function() {
     }
 
     function setCategory(row) {
+      let vacantItem = row.find(".table-category-vacant");
       if (row.find('input:checked').length) {
-        row.find(".table-category-vacant").addClass("table-category-preview");
+        vacantItem.addClass("table-category-preview");
       } else {
-        row.find(".table-category-vacant").removeClass("table-category-preview");
+        vacantItem.removeClass("table-category-preview");
       }
     }
 
@@ -43,7 +44,7 @@ $(document).ready(function() {
       }
     });
 
-    $('table input').click(function() {
+    $('table').find("input").click(function() {
       resetSelectClearLink($(this).parents('table'));
       setCategory($(this).parents('tr'))
     });
--- a/app/assets/stylesheets/narrow-table-display.scss
+++ b/app/assets/stylesheets/narrow-table-display.scss
@@ -1,33 +1,68 @@
-th.table-combine_right_if_narrow .table-hidden_right_title {
+th.table-combine_right_if_narrow .table-hidden_right_title,
+th.table-combine_right_if_narrow_2 .table-hidden_right_title {
     display:none;
+}
+
+.table-merge_columns tr {
+  border-bottom: 1px solid #bfc1c3;
 }
 
 .govuk-checkboxes__label.table-checkbox_label {
   /*removing padding for labels in tables*/
   padding: 0;
 }
-@media (min-width: 350px) {
-  .select-clear-all {
-    white-space: nowrap;
-  }
+
+.nullcell {
+  margin:0;
+  padding:0;
+  font-size:0;
+  width:0;
 }
-@media (max-width: 435px) {
+@media (max-width: 555px) {
   td.table-combine_right_if_narrow,
   td.table-combine_right_if_narrow + td {
     display: block;
     border-bottom: 0px;
+    &.govuk-table__cell--numeric {
+      text-align:center
+    }
   }
   th.table-combine_right_if_narrow{
     .table-hidden_right_title {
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/table-select-all.es6
```diff
@@ -13,6 +13,14 @@ $(document).ready(function() {
       }
     }
 
+    function setCategory(row) {
+      if (row.find('input:checked').length) {
+        row.find(".table-category-vacant").addClass("table-category-preview");
+      } else {
+        row.find(".table-category-vacant").removeClass("table-category-preview");
+      }
```
reggieb: Try not to repeat the same code with each condition of a conditional. It means that there are two things that need to be changed and kept in step - when there could just be one:
```javascript
let vacantItem = row.find(".table-category-vacant")
if (row.find('input:checked').length) {
  vacantItem.addClass("table-category-preview");
} else {
  vacantItem.removeClass("table-category-preview");
}
```
Now if the class that identifies the vacant item needs to change - it only needs to be changed in one place, and it is much more obvious that the different behaviours are being applied to the same object.

app/assets/javascripts/table-select-all.es6
```diff
@@ -36,7 +44,8 @@ $(document).ready(function() {
     });
 
     $('table input').click(function() {
```
reggieb: This code adds a listener to every input in _every_ table on the page. As the table where this listener is operating is already identified with the variable `table` this would be better:
```javascript
table.find('input').click(function() {
```

app/views/providers/transactions/show.html.erb
```diff
@@ -20,15 +20,21 @@
           </th>
           <th data-sort-type="date" class="govuk-table__header sort header-sort-desc table-combine_right_if_narrow" scope="col">
             <span class="govuk-visually-hidden">Sort by </span>
-            <span class="aria-sort-description"><%= t('.col_date') %></span><span class="table-hidden_right_title"> &amp; <%= t('.col_description') %></span><span class="sort-direction-indicator"><span class="screen-reader-sort-indicator govuk-visually-hidden"> (currently sorted from newest to oldest)</span></span>
+            <span class="aria-sort-description"><%= t('.col_date') %></span><span class="table-hidden_right_title"> <%= t('.col_and_description') %></span><span class="sort-direction-indicator"><span class="screen-reader-sort-indicator govuk-visually-hidden"> (currently sorted from newest to oldest)</span></span>
```
reggieb: The text "currently sorted from newest to oldest" should be defined in a translation file - not here in the markup.

This code would be easier to read if the spans were split onto separate lines:
```html
<span class="aria-sort-description"><%= t('.col_date') %></span>
<span class="table-hidden_right_title"><%= t('.col_and_description') %></span>
<span class="sort-direction-indicator">
  <span class="screen-reader-sort-indicator govuk-visually-hidden"> (currently sorted from newest to oldest)</span>
</span>
```
Then sort out any spacing issues with CSS and not by compressing the markup.

app/assets/stylesheets/narrow-table-display.scss
```diff
@@ -1,33 +1,68 @@
-th.table-combine_right_if_narrow .table-hidden_right_title {
+th.table-combine_right_if_narrow .table-hidden_right_title,
+th.table-combine_right_if_narrow_2 .table-hidden_right_title {
```
reggieb: What is the significance of the `2`? That's a magic number! What is it about these `td` elements that means they need to be classified differently? Could an additional flag be used to identify the elements that need the special styling. For example, say the special condition is they are made of chocolate. `th` elements that need to be combined if narrow could be classified as:

- **Made of chocolate** `table-combine_right_if_narrow made_of_chocolate`
- **Made of something else** `table-combine_right_if_narrow`

With this the `table-combine_right_if_narrow` would still apply to the chocolate elements, and styling applied just to them could be applied to `.made_of_chocolate` or even `.table-combine_right_if_narrow.made_of_chocolate`

app/assets/stylesheets/table-category-column.scss
```diff
@@ -0,0 +1,37 @@
...
+@media (min-width: 655px) {
+  p.table-category, p.table-category-vacant {
+    padding: 4px 8px 1px;
+  }
+}
+@media (min-width: 705px) {
+  p.table-category, p.table-category-vacant {
+      white-space: nowrap;
+  }
+}
```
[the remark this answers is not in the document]
MalcolmVonMoJ: The styling is specified to the `p` tag to push it heigher up the heirarchy than the GDS out of the box style, which otherwise gets applied.  I can (and have) removed it from where it isn't needed.  

Other changes made as per your comments.

app/views/providers/transactions/show.html.erb
MalcolmVonMoJ: The arrangement of the `<span>` onto one line is to avoid unwanted whitespace.

reggieb: Why?

MalcolmVonMoJ: The sorting arrows get pushed onto a new line on some narrow displays.

app/assets/stylesheets/narrow-table-display.scss
MalcolmVonMoJ: I agree that the `2` might not be the best label.  Perhaps something else.  

The issue here is that pretty much the entire styling has to be separate because it all has to be in their respective @media queries.  So there is very little that can be combined - even though the styles are pretty much identical.  

So having `table-combine_right_if_narrow made_of_chocolate` will mean that the `table-combine_right_if_narrow` class virtually redundant.  (The style at line 3 is the only one that won't be reliant on `made_of_chocolate`.)

I am thinking a class `table-combine_right_if_narrow_at_555px` might be a good way of doing it, but that means changing the class name every time the width changes - as indeed it did as part of this change.  

Unless SASS has a clever way of dealing with identical styles coming into play at different widths, then I don't think these can be improved (minus, perhaps, my choice of class name).

MalcolmVonMoJ: I have changed the CSS class names:

- Removing the `2` class
- Adding new class `narrow_555` for the 555px class
- Adding new class `narrow_470` for the 470px class

reggieb: Those class names are very specific but I can't think of better alternatives (`narrow` and `slightly_narrower` :smile: ). So lets go for `narrow_555` and `narrow_470`

PR #580, opened 2019-06-13
Ap 533 and ap 565 savings page changes

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-533)

Add 'none of these' checkbox and change copy on the savings and investment pages on both 

## Checklist

### DIFF
--- a/app/views/providers/savings_and_investments/show.html.erb
+++ b/app/views/providers/savings_and_investments/show.html.erb
@@ -1,11 +1,9 @@
 <%= page_template page_title: t('.h1-heading'), template: :basic, show_errors_for: @form do %>
-    <%= render partial: 'shared/forms/savings_and_investments_form',
-               locals: {
-                 model: @form,
-                 show_draft: true,
-                 bank_accounts: [],
-                 or_break: t('generic.or'),
-                 none_selected: t('.none_selected'),
-                 form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application)
-               } %>
+    <%= render  'shared/forms/savings_and_investments_form',
+                model: @form,
+                show_draft: true,
+                bank_accounts: [],
+                or_break: t('generic.or'),
+                none_selected: t('.none_selected'),
+                form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application) %>
 <% end %>
--- a/app/views/providers/savings_and_investments/show.html.erb
+++ b/app/views/providers/savings_and_investments/show.html.erb
@@ -1,8 +1,11 @@
-<%= page_template page_title: t('.h1-heading') do %>
-<%= render(
-      'shared/forms/savings_and_investments_form',
-      bank_accounts: [],
-      form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application),
-      show_draft: true
-    ) %>
+<%= page_template page_title: t('.h1-heading'), template: :basic, show_errors_for: @form do %>
+    <%= render partial: 'shared/forms/savings_and_investments_form',
+               locals: {
+                 model: @form,
+                 show_draft: true,
+                 bank_accounts: [],
+                 or_break: t('.or_break'),
+                 none_selected: t('.none_selected'),
+                 form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application)
+               } %>
 <% end %>
--- a/app/views/providers/savings_and_investments/show.html.erb
+++ b/app/views/providers/savings_and_investments/show.html.erb
@@ -4,7 +4,7 @@
                  model: @form,
                  show_draft: true,
                  bank_accounts: [],
-                 or_break: t('.or_break'),
+                 or_break: t('generic.or'),
                  none_selected: t('.none_selected'),
                  form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application)
                } %>
--- a/app/views/shared/forms/_savings_and_investments_form.html.erb
+++ b/app/views/shared/forms/_savings_and_investments_form.html.erb
@@ -2,7 +2,7 @@
 
 <%= form_with(model: @form, url: form_path, method: :patch, local: true) do |form| %>
 
-  <%= govuk_form_group show_error_if: (@form.errors.present? && !@form.any_checkbox_checked?) do %>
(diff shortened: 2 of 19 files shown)


spec/forms/savings_amounts/savings_amounts_form_spec.rb
```diff
@@ -106,17 +106,26 @@
...
           savings_amount.reload
-
           attributes.each do |attr|
             val = savings_amount.send(attr)
-            expect(val).to eq(nil), "Attr #{attr}: expected nil, got #{val}"
+            if attr == :isa
+              expect(val).not_to eq(nil), "Attr #{attr}: expected not to be nil, got #{val}"
+            else
+              expect(val).to eq(nil), "Attr #{attr}: expected nil, got #{val}"
+            end
```
reggieb: Rather than checking the condition in every iteration of the loop, remove `:isa` from the array and test it separately:
```ruby
attributes_except_isa = attributes - [:isa]

attributes_except_isa.each do |attr|
  val = savings_amount.send(attr)
  expect(val).to eq(nil), "Attr #{attr}: expected nil, got #{val}"
end

expect(savings_amount.isa).not_to eq(nil)
```

app/views/shared/forms/_savings_and_investments_form.html.erb
```diff
@@ -1,26 +1,34 @@
...
-      <%= render(
-            partial: 'shared/forms/revealing_checkbox/attribute',
-            collection: attributes,
-            locals: {
-              model: @form,
-              form: form
-            }
-          ) %>
-    </div>
+  <%= govuk_form_group show_error_if: (@form.errors.present? && !@form.any_checkbox_checked?) do %>
```
reggieb: As the form method `any_checkbox_checked_or_draft` adds a base error, the code shouldn't have to check for that condition again here. 

That is - if `@form.any_checkbox_checked?` is false, there should already be an error in `@form.errors`

This should work:
```ruby
govuk_form_group show_error_if: (@form.errors.present?)
```

app/views/providers/savings_and_investments/show.html.erb
```diff
@@ -1,8 +1,11 @@
...
+<%= page_template page_title: t('.h1-heading'), template: :basic, show_errors_for: @form do %>
+    <%= render partial: 'shared/forms/savings_and_investments_form',
+               locals: {
+                 model: @form,
+                 show_draft: true,
+                 bank_accounts: [],
+                 or_break: t('.or_break'),
+                 none_selected: t('.none_selected'),
+                 form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application)
+               } %>
```
reggieb: If you are not passing a collection in, the preferred (as it's more succinct) syntax is:
```ruby
render 'path/to/partial', a: 1, b: 2
```
Rather than:
```ruby
render partial: 'path/to/partial', locals: {a: 1, b: 2}
```

So here:
```ruby
render  'shared/forms/savings_and_investments_form',
            model: @form,
            show_draft: true,
            bank_accounts: [],
            or_break: t('.or_break'),
            none_selected: t('.none_selected'),
            form_path: providers_legal_aid_application_savings_and_investment_path(@legal_aid_application)

File: app/views/shared/forms/_savings_and_investments_form.html.erb [hunk 4]
```diff
@@ -1,26 +1,34 @@
...
+    <%= govuk_fieldset_header do %>
+      <h1 class="govuk-fieldset__heading govuk-!-padding-bottom-7"><%= page_title %></h1>
+    <% end %>
+    <fieldset class="govuk-fieldset" aria-describedby="citizenship-conditional-hint">
```
reggieb: I can't find an id of "citizenship-conditional-hint" anywhere in this PR, and cannot find "citizenship" anywhere in the code base. So I cannot see how this `aria-describedby` can be correct.

config/locales/en/shared.yml
```diff
@@ -50,6 +50,10 @@ en:
         sub_heading: Select all that apply
         or_break: or
         none_selected: None of these
+      savings_and_investments_form:
+        sub_heading: Select all that apply
+        or_break: or
```
reggieb: Rather than duplicating `or_break: or` here and in two other translations, I think it would better to remove the local `or_break`, and use `generic.or` which has the same value, and can replace the usage of all these entries.

spec/forms/savings_amounts/savings_amounts_form_spec.rb
```diff
@@ -106,20 +106,33 @@
...
+          attributes_except_isa.each do |attr|
             val = savings_amount.send(attr)
             expect(val).to eq(nil), "Attr #{attr}: expected nil, got #{val}"
           end
         end
 
+        it 'does not empty amount if a checkbox is checked' do
+          subject.save
+          savings_amount.reload
+          val = savings_amount.send(:isa)
```
reggieb: Don't use `send` with static content. This could simply be:
```ruby
subject.save
expect(savings_amount.reload.isa).not_to eq(nil)
```

spec/forms/savings_amounts/savings_amounts_form_spec.rb
```diff
@@ -106,20 +106,33 @@
...
+        let(:amount_params) { attributes.each_with_object({}) { |attr, hsh| hsh[attr] = Faker::Number.decimal.to_s } }
+
+        it 'empties amounts if checkbox is unchecked' do
+          attributes_except_isa = attributes - [:isa]
           subject.save
           savings_amount.reload
-
-          attributes.each do |attr|
+          attributes_except_isa.each do |attr|
             val = savings_amount.send(attr)
```
reggieb: Not you're change, but as you're here .... `__send__` is preferred to `send`. So:
```ruby
val = savings_amount.__send__(attr)
```

app/views/providers/savings_and_investments/show.html.erb
willc-work: isn't bank_accounts a collection (when it exists?)

reggieb: The key thing is that you are not needing to pass in a `collection:` at the same level as `locals:`. All the key/value pairs are locals. So the more succinct form is fine.

PR #581, opened 2019-06-14
AP-661 blank applications

Kalagan:
## What

[AP-661](https://dsdmoj.atlassian.net/browse/AP-661)

- Rename path `/providers/applications/:id/applicant` to `/providers/applications/:id/applicant_details`
- Add new path `/providers/applicants`

This new controller is the new start of the provider flow. It creates the application and the applicant on submit. It then removes itself from the history and adds `/providers/applications/:id/applicant_details` instead.
That way, when navigating back to this step, the provider can edit the applicant's details rather than being presented with an empty page that would create an additional applicant and application when submitted

## Checklist

### DIFF
--- a/spec/requests/providers/applicant_details_spec.rb
+++ b/spec/requests/providers/applicant_details_spec.rb
@@ -1,0 +1,150 @@
+require 'rails_helper'
+
+RSpec.describe Providers::ApplicantDetailsController, type: :request do
+  let(:application) { create(:legal_aid_application) }
+  let(:application_id) { application.id }
+  let(:provider) { application.provider }
+
+  describe 'GET /providers/applications/:legal_aid_application_id/applicant_details' do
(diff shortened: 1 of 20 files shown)


spec/requests/providers/applicant_details_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+    let(:params) do
+      {
+        applicant: {
+          first_name: 'John',
+          last_name: 'Doe',
+          national_insurance_number: 'AA 12 34 56 C',
+          dob_year: '1981',
+          dob_month: '07',
+          dob_day: '11',
+          email: Faker::Internet.safe_email
```
reggieb: Could `attributes_for(:applicant)` be used here?

Kalagan: Indeed. I'll change it. 
(This code is not new but copied from the previous `applicants_spec.rb` file)

PR #582, opened 2019-06-14
AP-662 has Provider used Delegated Functions?

reggieb:
[Jira AP-662](https://dsdmoj.atlassian.net/browse/AP-662)

This adds a page for the provider to input whether they used Delegated Functions, and if so when.

On building this I found that I was copying into the form a lot of complication from other forms that use date. So I created `DateFieldBuilder` to put all that complication into one object that could be reused across such forms. To test it's reuse, I refactored `Incidents::LatestIncidentForm` to check that the functionality worked in another form. I've add a tech debt ticket for other Forms to be likewise refactored ([AP-740](https://dsdmoj.atlassian.net/browse/AP-740)).

I have added a holding action controller `Providers::SubstantiveApplicationsController`. The functionality for this should be added in PR-663

### DIFF
--- a/app/forms/concerns/date_field_builder.rb
+++ b/app/forms/concerns/date_field_builder.rb
@@ -1,0 +1,61 @@
+class DateFieldBuilder
+  DATE_PARTS = %i[year month day].freeze
+
+  attr_reader :form, :model, :method, :prefix
+
+  # :method is the target method on the model
+  # :prefix is used to construct date part fields
+  def initialize(form:, model:, method:, prefix:)
+    @form = form
+    @model = model
+    @method = method
+    @prefix = prefix
+  end
+
+  def blank?
+    from_form.all?(&:blank?)
+  end
+
+  def complete?
+    @complete ||= from_form.all?(&:present?)
+  end
+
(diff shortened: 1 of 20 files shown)


Kalagan:
Isn't this ticket number supposed to be 662 instead of 622?
https://dsdmoj.atlassian.net/browse/AP-662

app/forms/concerns/date_field_builder.rb
```diff
@@ -0,0 +1,61 @@
...
+
+    DATE_PARTS.each_with_object({}) { |part, hash| hash[field_for(part)] = model_date.__send__(part) }
+  end
+
+  def model_date
+    @model_date ||= model.__send__(method)
+  end
+
+  def form_date_invalid?
+    !Date.valid_date?(*from_form.map(&:to_i))
```
Kalagan: nice one 👍

reggieb: Yes - I've only recently found `Date.valid_date?` - so much better than catching the error.

reggieb:
662  - palm -> face

PR #584, opened 2019-06-14
AP-718 - injection of application from Apply into CCMS

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-718)

Introduce functionality to create a case in CCMS using calls from Apply. 

The code to make these calls was developed in earlier PRs. This change introduces a spec (`ccms_integration/ccms_spec.rb`) which ties together the process.

Much of the data that needs to be included in the case creation payload is not yet available and so has been replaced with dummy values or mocked out.

An Opponent model has been created.  There will be just one dummy opponent used in all cases, but the final values for this are still to be decided.

This PR also fixes a number of problems found with code developed in those earlier PRs which were preventing successful case creation.

## Checklist


app/models/opponent.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def other_party_relationship_to_client
+    relationship_to_client.capitalize
+  end
+
+  def other_party_relationship_to_case
+    relationship_to_case.capitalize
+  end
+
+  def method_missing(meth, *args)
```
Kalagan: `meth` always makes me think of Breaking Bad https://youtu.be/r5wpigdBfK8

app/models/legal_aid_application.rb
```diff
@@ -138,6 +140,26 @@ def submission_date
     transaction_period_finish_at.to_date
   end
 
+  # This is here so that this method can be stubbed in the ccms payload integration tests
+  # it will be later replaced by an opponents association to load a dummy opponent for CCMS
+  # injection
```
reggieb: In some classes there are these long explanations about why static content is being returned. In others (for example the Opponent model) there are similar methods with no comments. Can we mark them all with a simple easily identifiable comment. I'd suggest using a TODO that many IDE will find easily:
```ruby
def opponnents
  nil  # TODO: Replace with CCMS data
end
```
or
```ruby
def opponnents
  nil  # TODO: CCMS placeholder
end
```
And use it everywhere it's relevant.

app/models/opponent.rb
```diff
@@ -0,0 +1,108 @@
...
+  def assessed_assets
+    0
+  end
+
+  def full_name
+    "#{title.capitalize} #{first_name} #{surname}"
+  end
+
+  def person?
+    opponent_type.capitalize == 'Person'
```
reggieb: `'Person'` needs freezing. Same with `'Organisation'` in `organisation?`

app/models/opponent.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def method_missing(meth, *args)
+    if valid_relationship_to_case_method?(meth)
+      case_relationship?(meth)
+    elsif valid_relationship_to_client_method?(meth)
+      client_relationship?(meth)
+    else
+      super
+    end
+  end
```
reggieb: Using meta-programming when there is a simpler alternative is ALWAYS a mistake. The code is over complicated, difficult to debug and extend. I strongly recommend replacing this `method_missing` functionality is replaced with:

```ruby
def case_relationship?(relationship)
  relationship_to_case.capitalize == CASE_RELATIONSHIP_VALUES[relationship.to_sym]
end

def client_relationship?(relationship)
  relationship_to_client.capitalize == CLIENT_RELATIONSHIP_VALUES[relationship.to_sym]
end
```
And thus:
```
# rather than:
opponent.relationship_case_child?

# We do this:
opponent.relationship_case? :child
```

app/services/ccms/case_add_requestor.rb
```diff
@@ -24,13 +25,27 @@ def initialize(submission)
...
 
     private
 
+    # :nocov:
+    def save_request
+      # puts "CCMS payload saved to #{Rails.root.join('spec/integration/generated/add_case_request.xml')}"
+      File.open(Rails.root.join('ccms_integration/generated/add_case_request.xml'), 'w') do |fp|
+        fp.puts request_xml
+      end
+    rescue StandardError => e
```
reggieb: Why is this being rescued?

app/models/opponent.rb
reggieb: OK - I understand the requirement for this at the moment.

PR #590, opened 2019-06-18
AP-692 split prospects of success

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-692)

- Add new **Is the chance of a successful outcome 50% or better?** page
- Remove **more than 50%** from `success_prospect` page and change copy
- Modify FormBuilder to allow passing hint option to radio buttons

## Checklist

### DIFF
--- a/app/lib/govuk_elements_form_builder/form_builder.rb
+++ b/app/lib/govuk_elements_form_builder/form_builder.rb
@@ -67,9 +67,10 @@ module GovukElementsFormBuilder
 
       label_options = { value: value.to_s, class: 'govuk-label govuk-radios__label' }
       label_html = label(attribute, options[:label], label_options)
+      hint_html = hint_tag(attribute, options.merge(radio_button_value: value))
 
       content_tag(:div, class: 'govuk-radios__item') do
-        concat_tags(radio_html, label_html)
+        concat_tags(radio_html, label_html, hint_html)
       end
     end
 
@@ -227,13 +228,19 @@ module GovukElementsFormBuilder
     end
 
     def hint_message(attribute, options)
+      return options[:hint] if options.key?(:radio_button_value)
+
       options[:hint].presence || I18n.translate("helpers.hint.#{@object_name}.#{attribute}", default: nil)
     end
 
     def hint_tag(attribute, options)
       return unless hint?(attribute, options)
 
-      content_tag(:span, hint_message(attribute, options), class: 'govuk-hint', id: "#{attribute}-hint")
+      classes = ['govuk-hint']
+      classes << 'govuk-radios__hint' if options.key?(:radio_button_value)
+
+      id = [attribute, options[:radio_button_value], 'hint'].compact.join('-')
+      content_tag(:span, hint_message(attribute, options), class: classes.join(' '), id: id)
     end
 
     def error_tag(attribute, options)
--- a/app/models/merits_assessment.rb
+++ b/app/models/merits_assessment.rb
@@ -1,23 +1,15 @@
 class MeritsAssessment < ApplicationRecord
   belongs_to :legal_aid_application
 
-  # defining before enum so it can be used in the enum
-  def self.prospect_likely_to_succeed
-    :likely
-  end
-
-  enum(
-    success_prospect: {
-      prospect_likely_to_succeed => '50% or better'.freeze,
-      marginal: 'Marginal 45 - 50%'.freeze,
-      borderline: 'borderline'.freeze,
-      uncertain: 'uncertain'.freeze,
-      poor: 'poor'.freeze
-    },
-    _prefix: true
-  )
+  enum success_prospect: {
+    likely: 'likely'.freeze,
+    marginal: 'marginal'.freeze,
+    poor: 'poor'.freeze,
+    borderline: 'borderline'.freeze,
+    not_known: 'not_known'.freeze
+  }, _prefix: true
 
   def self.prospects_unlikely_to_succeed
-    @prospects_unlikely_to_succeed ||= success_prospects.except(prospect_likely_to_succeed.to_s)
+    MeritsAssessment.success_prospects.except(:likely).keys
   end
 end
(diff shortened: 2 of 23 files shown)


app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -227,13 +228,19 @@ def hint?(attribute, options)
     end
 
     def hint_message(attribute, options)
+      return options[:hint] if options[:radio_button]
+
```
reggieb: The effect of this change is to remove the option of putting translations into `helpers.hint` if they are radio buttons. I wonder if this is a better alternative:
```ruby
def hint_message(attribute, options)
  return options[:hint] if option[:hint].present?
  
  translation_path = ['helpers.hint', @object_name, attribute]
  translation_path << options[:value] if options[:radio_button]
  I18n.translate(translation_path.join('.'), default: nil)
end
```
That would then allow the translations for a `bar` attribute of `foo` with radios `yes` and `no` to have it hints at:
```yaml
en:
  helpers:
    hint:
      foo:
        bar:
          yes: Oh yes
          no: Hell no
```

app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -67,9 +67,10 @@ def govuk_radio_button(attribute, value, *args)
 
       label_options = { value: value.to_s, class: 'govuk-label govuk-radios__label' }
       label_html = label(attribute, options[:label], label_options)
+      hint_html = hint_tag(attribute, options.merge(radio_button: true, value: value))
```
reggieb: Rather than passing in two new options, would one work?
```ruby
hint_html = hint_tag(attribute, options.merge(radio_button: value))
```

app/lib/govuk_elements_form_builder/form_builder.rb
reggieb: If `options[:radio_button]` returned the current `value`, the code could change to:
```ruby
translation_path << options[:radio_button] if options[:radio_button]
```
Or you could do:
```ruby
I18n.translate(
  ['helpers.hint', @object_name, attribute, options[:radio_button]].compact.join('.'),
  default: true
)
```

app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -227,13 +228,19 @@ def hint?(attribute, options)
...
     end
 
     def hint_tag(attribute, options)
       return unless hint?(attribute, options)
 
-      content_tag(:span, hint_message(attribute, options), class: 'govuk-hint', id: "#{attribute}-hint")
+      classes = ['govuk-hint']
+      classes << 'govuk-radios__hint' if options[:radio_button]
+
+      id = [attribute, options[:value], 'hint'].compact.join('-')
```
reggieb: If `options[:radio_button]` returns the value, then this becomes:

```ruby
id = [attribute, options[:radio_button], 'hint'].compact.join('-')
```

app/models/merits_assessment.rb
```diff
@@ -1,23 +1,15 @@
...
-    },
-    _prefix: true
-  )
+  enum success_prospect: {
+    likely: 'likely',
+    marginal: 'marginal',
+    poor: 'poor',
+    borderline: 'borderline',
+    not_known: 'not_known'
+  }, _prefix: :success_prospect
```
reggieb: I think these enum options should still be frozen. And `_prefix: true` does already set the prefix to match the enum name.

db/migrate/20190617103359_add_success_more_than_fifty_percent_to_merits_assessments.rb
```diff
@@ -0,0 +1,5 @@
+class AddSuccessMoreThanFiftyPercentToMeritsAssessments < ActiveRecord::Migration[5.2]
+  def change
+    add_column :merits_assessments, :success_more_than_fifty_percent, :boolean
```
reggieb: What happens if the `likely` threshold is changed? If it was changed to 60% would this column need to be renamed? Would `success_likely` be a better name?

app/lib/govuk_elements_form_builder/form_builder.rb
Kalagan: The problem with
```yaml
en:
  helpers:
    hint:
      foo:
        bar:
          yes: Oh yes
          no: Hell no
```
Is that the hint message of the radio button collection (the one above the options) will now be the hash:
```ruby
{ yes: 'Oh yes', no: 'Hell no' }
```

Kalagan: Because that's how we use it for now:
i.e. : 
```yaml
en:
  helpers:
    hint:
      user:
        name: Enter your name
```

reggieb: OK

app/lib/govuk_elements_form_builder/form_builder.rb
Kalagan: Yes it works. I called it `radio_button_value`
And then, in the rest of the code I need to do 
`if options.key?(:radio_button_value)`
instead of 
`if options[:radio_button]`
because the value could `false`

db/migrate/20190617103359_add_success_more_than_fifty_percent_to_merits_assessments.rb
Kalagan: yes much better!

Kalagan: I changed everything to `success_likely`
attribute, route, controller, form.

The route is now `resources :success_likely, only: %i[index create]`

I'm using `resources` with `%i[index create]` so the controller doesn't have to be plural.

`likelies`!? :scream:

reggieb: There's always `likely_success` which would have the `likely_successes` plural if needed perhaps.

Kalagan: I think it's fine like that. We also use `resources` and `%i[index create]` for other pages.

PR #593, opened 2019-06-19
AP 663 Proceed to substantive application

reggieb:
[AP-663](https://dsdmoj.atlassian.net/browse/AP-663)

## Main change

Adds a page to the providers journey (after `used_delegated_functions` if the provider selects "Yes") that allows the provider to choose between continuing with a substantive application, or returning later.

## Knock on change

If the provider chooses to return later they are prompted with a date in the future by which they should complete the application. That date is _"the current day plus 10 working days"_. To work that out I needed to:

- Add a system to calculate working days (`WorkingDayCalculator`). This uses the `business` gem
- To accommodate changes in bank holiday and to avoid having to rely on the gems hardcoded list of bank holidays, they are retrieved from a [Gov-UK api](https://www.gov.uk/bank-holidays.json) by `BankHolidayRetriever` and stored in a model `BankHoliday`
- A worker (`BankHolidayUpdateWorker`) is triggered on calling `BankHoliday.dates` that:
  - Checks to see if there is a current store of bank holiday data (current being less than 2 days old) and closes without doing anything if there is
  - Gets the latest data from the API and if the data is the same as the currently stored version, just updates the `updated_at` of the `BankHoliday` instance holding that data.
  - If the data is different to the current data, it creates a new `BankHoliday` instance to store the data


app/models/bank_holiday.rb
```diff
@@ -0,0 +1,18 @@
...
+  serialize :dates, Array
+
+  scope :by_updated_at, -> { order(updated_at: :asc) }
+  before_validation :populate_dates
+
+  validates :dates, presence: true
+
+  def self.dates
+    BankHolidayUpdateWorker.perform_in 10.seconds
+    instance = by_updated_at.last || create
```
Kalagan: `create!` with an exclamation mark?
😃

app/forms/legal_aid_applications/substantive_application_form.rb
```diff
@@ -0,0 +1,38 @@
...
+    validates :substantive_application, presence: { unless: :draft? }
+
+    def substantive_application_deadline_on
+      return model.substantive_application_deadline_on if model.substantive_application_deadline_on?
+
+      attributes['substantive_application_deadline_on'] ||= substantive_application_deadline
+    end
+
+    def save
+      super && model.provider_used_delegated_functions!
```
Kalagan: I'm a bit uneasy by putting the change of state here because I think that for all the other states it's done in the controller.
So for consistency sake, it would be better to follow the same pattern.

app/services/bank_holiday_retriever.rb
```diff
@@ -0,0 +1,37 @@
...
+  def data
+    return raise_error unless response.is_a?(Net::HTTPOK)
+
+    @data ||= JSON.parse(response.body)
+  end
+
+  def dates(group)
+    return if data.empty?
+
+    data.dig(group, 'events')&.map { |details| details['date'] }
```
Kalagan: would `.pluck('date')` work instead of `.map { |details| details['date'] }` ?

app/models/bank_holiday.rb
reggieb: Damn! I was so pleased with myself for using `save!` in the worker. I even did a quick "Check for non-bang save methods before Julien's spots them" check.

app/forms/legal_aid_applications/substantive_application_form.rb
reggieb: And it would mean the overloading of `save` would no longer be necessary. Nice one.

app/services/bank_holiday_retriever.rb
reggieb: Yes - it might well do.

PR #604, opened 2019-06-25
Ap 675 latest incident

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-675)

Add date the incident_occurred to the date_told page, add validation for the timeline of the dates
removed old pages

## Checklist


app/forms/incidents/told_on_form.rb
```diff
@@ -1,69 +1,84 @@
...
-    def incomplete_date
-      return nil if told_on_methods.all?(&:blank?)
+    def exclude_from_model
+      (told_on_date_fields.fields + occurred_on_date_fields.fields)
+    end
 
-      @incomplete_date = true
-      :invalid
+    def draft_and_not_partially_complete_date?
+      draft? && !told_on_date_fields.partially_complete?
```
reggieb: This only checks whether `told_on_date_fields` are partially complete. It will not detect if `occurred_on_date_fields` are partially complete. A simple solution would be:
```ruby
draft? && !told_on_date_fields.partially_complete? && occurred_on_date_fields.partially_complete?
```
However, that could cause an incorrect error if one date was empty, the other partially complete, and draft selected. In this case both sets of date fields would get an error - were the correct behaviour would be just for the incomplete date fields to get the error.

So, I think this is a better solution:

```
def draft_and_not_partially_complete_told_on_date?
  draft? && !told_on_date_fields.partially_complete?
end

def draft_and_not_partially_complete_occurred_on_date?
  draft? && !occurred_on_date_fields.partially_complete?
end
```
Then use `draft_and_not_partially_complete_told_on_date?` in the `unless` clause of the `told_on` validation, and the other method with the `occurred_on` validation.

app/views/providers/check_merits_answers/show.html.erb
```diff
@@ -19,20 +19,20 @@
...
-  <dl class="govuk-summary-list govuk-!-margin-bottom-2">
+  <%# <!-- <dl class="govuk-summary-list govuk-!-margin-bottom-2">
     <%= check_answer_link(
           name: :details_of_latest_incident,
           url: providers_legal_aid_application_details_latest_incident_path,
           question: t('.items.details_of_latest_incident'),
           answer: nil
         ) %>
-  </dl>
+  </dl> --> %>
```
reggieb: Can this be removed rather than commented out.

app/views/providers/date_client_told_incidents/show.html.erb
```diff
@@ -10,6 +10,8 @@
     <%= govuk_form_group show_error_if: @form.errors.present? do %>
       <%= govuk_fieldset_header page_title %>
 
+      <div class="govuk-!-padding-bottom-4"></div>
```
reggieb: The method `govuk_fieldset_header` includes an option to add padding to the bottom of this field. So rather than added in a spacing div and therefore a `div` without context, remove this and modify the `govuk_fieldset_header` call:
```
<%= govuk_fieldset_header page_title, padding_below: 4 %>
```

PR #606, opened 2019-06-27
AP-737 Write first Javascript unit test

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-737)

- I wrote a test for the javascript that handles the loading page when fetching banking data.
- I replaced `$.getJSON` with using [axios](https://www.npmjs.com/package/axios) so it's easier to mock the whole module like on the documentation (https://jestjs.io/docs/en/mock-functions#mocking-modules)
- Added javascript tests to be run with the `bundle exec rake` task so CircleCI will run them.

## Checklist


spec/javascript/worker_waiter.spec.js
```diff
@@ -0,0 +1,70 @@
...
+  it('does not reloads the page', done => {
+    require('worker_waiter')({
+      poll_interval: 0,
+      callback: () => {
+        expect(window.location.reload).not.toBeCalled()
+        done()
+      }
+    })
+  })
+})
```
reggieb: The two main test conditions here are I think:

1. There is a `.worker-waiter` on the page and the endpoint initially returns a "working" status, and on a second attempt, a non-"working" status.
2. There is not a `.working-waiting` on the page.

I'd suggest adding at least one other test - a successful test without a working waiting. That is, the first response returns "complete", just to make sure the normal behaviour isn't always just doing two calls. You could even do two "working" before the complete - but that's probably over the top.

PR #608, opened 2019-06-28
AP-720 Do you have dependants

AbdirizakObsiye2018:
A page with a yes or no checkbox form, which says if a legal aid application has or doesn't have a dependant or dependants

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-720)

- Added column has_dependants to legal aid applications table
- Created controller for has_dependants as well as a form and view
- Created request tests and a feature test for 'no'
- Updated routes and translation files

### DIFF
--- a/app/views/shared/forms/_has_dependants_form.html.erb
+++ b/app/views/shared/forms/_has_dependants_form.html.erb
@@ -1,0 +1,42 @@
+<%= form_with(
+      model: @form,
+      url: form_path,
+      method: :patch,
+      local: true
+    ) do |form| %>
+
+    <%= govuk_form_group show_error_if: @form.errors.present? do %>
+      <%= govuk_fieldset_header page_title %>
+
+      <div class="govuk-!-padding-bottom-4"></div>
+
+      <p class="govuk-body">
+        <%= dependant_info %>
+      </p>
+
+      <div class="govuk-inset-text">
+        <p class="govuk-body">
+          <%= dependant_extra_info %>
+        </p>
+      </div>
+
+      <%= form.govuk_radio_button(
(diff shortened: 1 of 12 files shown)


app/views/shared/forms/_has_dependants_form.html.erb
```diff
@@ -0,0 +1,42 @@
...
+        <p class="govuk-body">
+          <%= dependant_extra_info %>
+        </p>
+      </div>
+
+      <%= form.govuk_radio_button(
+            :has_dependants,
+            true,
+            label: t('generic.yes'),
+            'data-aria-controls' => 'conditional-true'
```
Kalagan: is this useful?

spec/requests/citizens/has_dependants_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'rails_helper'
+
+RSpec.describe 'citizen own home requests', type: :request do
+  let(:application) { create :application, :with_applicant }
+  let(:application_id) { application.id }
```
Kalagan: this is not used anywhere

app/views/shared/forms/_has_dependants_form.html.erb
AbdirizakObsiye2018: No, it shouldn't be there, thanks.

spec/requests/citizens/has_dependants_spec.rb
AbdirizakObsiye2018: Removed it

PR #615, opened 2019-07-01
AP-721 Dependant's details page

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-721)

- create two new controllers using the same form. One to create a new dependant and one to edit a dependant
- Add a way to pass a parameter from the controller to the flow service

## Checklist


app/controllers/concerns/flowable.rb
```diff
@@ -22,6 +22,10 @@ def step_prefix
     delegate :forward_path, to: :flow_service
     helper_method :forward_path, :journey_type
 
+    def flow_param(flow_param)
+      @flow_param = flow_param
+    end
```
reggieb: I wonder if rather than relying on two calls to the flow service, the flow params could be passed into `go_forward`. I think it requires a little refactoring to ensure that memoized objects aren't reused out of context by other processes. So ...

First step remove the line:
```ruby
delegate :forward_path, to: :flow_service
```
Then refactor the code:
```ruby

    def go_forward(flow_param=nil)
      forward_path = flow_service(flow_params).forward_path
      if path?(forward_path)
        redirect_to forward_path
      else
        render plain: forward_path
      end
    end

    def flow_service(flow_params=nil)
      Flow::BaseFlowService.flow_service_for(
        journey_type,
        legal_aid_application: legal_aid_application,
        current_step: current_step,
        params: flow_param
      )
    end
```
Then in the controller rather than:
```ruby
flow_param(dependant)
go_forward
```
We use:
```ruby
go_forward(dependant)
```

app/controllers/citizens/dependant_details_controller.rb
```diff
@@ -1,35 +1,37 @@
 module Citizens
-  class DependantDetailsController < BaseController
-    include ApplicationFromSession
-    helper_method :other_dependants
+  module Dependants
+    class DetailsController < BaseController
```
reggieb: I think this controller needs to be removed - it's been superseded by the file at `app/controllers/citizens/dependants/details_controller.rb`

app/controllers/citizens/dependants/new_controller.rb
```diff
@@ -0,0 +1,47 @@
+module Citizens
+  module Dependants
+    class NewController < BaseController
```
reggieb: The pattern I used in vehicles was to have a `VehiclesController` that created the main object, and then the controllers that modified vehicles would be in the `Vehicles` namespace. So `Vehicles::FooController`.

`NewController` feels a little strange. Would making this `Citizens::DependantsController` be simpler?

Also - the `dependents` step prefix wouldn't be necessary in this controller. It would just have the step name `:dependents`, and `Dependents::DetailsController` would stay as `dependents_details` (it would need the prefix).

config/routes.rb
```diff
@@ -78,8 +78,10 @@
     end
     resource :means_test_result, only: [:show]
     resource :declaration, only: %i[show update]
-    resources :dependants, only: %i[new create]
-    resources :dependant_details, only: %i[show update]
+
+    resources :dependants, only: %i[index create], controller: 'dependants/new' do
+      resource :details, only: %i[show update], controller: 'dependants/details'
+    end
```
reggieb: Comparing this with vehicles, I notice vehicles uses `resource :vehicle` rather than `resources :vehicles` and then a namespace, so perhaps the same pattern won't work. But try this:
```ruby
resources :dependants, only: %i[index create] do
  scope module: :dependants do
    resource :details, only: %i[show update]
  end
end
```
On my system that creates:
```bash
citizens_dependant_details GET   /citizens/dependants/:dependant_id/details(.:format) citizens/dependants/details#show
                           PATCH /citizens/dependants/:dependant_id/details(.:format) citizens/dependants/details#update
                           PUT   /citizens/dependants/:dependant_id/details(.:format) citizens/dependants/details#update
       citizens_dependants GET   /citizens/dependants(.:format)                       citizens/dependants#index
                           POST  /citizens/dependants(.:format)                       citizens/dependants#create
```

PR #619, opened 2019-07-03
AP-722 dependant relationship

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-722)

Dependant relationship page.

## Checklist

### DIFF
--- a/spec/requests/citizens/dependants_spec.rb
+++ b/spec/requests/citizens/dependants_spec.rb
@@ -31,6 +31,7 @@ RSpec.describe Citizens::DependantsController, type: :request do
         }
       }
     end
+    let(:latest_dependant) { Dependant.order(:created_at).last }
 
     subject { post citizens_dependants_path, params: params }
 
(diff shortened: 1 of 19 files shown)


spec/requests/citizens/dependants_spec.rb
```diff
@@ -41,8 +41,10 @@
     end
 
     it 'redirects to the dependant relationship page if the dependant is more than 15 years old' do
+      dependant_id = SecureRandom.uuid
+      allow_any_instance_of(Dependant).to receive(:id).and_return(dependant_id)
```
reggieb: I don't think you need to stub here and can avoid using `allow_any_instance_of`. Instead identify the dependent that was just created and use its id. So:
```ruby
let(:latest_dependant) { Dependant.order(:created_at).last }

it 'redirects to the dependant relationship page if the dependant is more than 15 years old' do
  subject
  expect(response).to redirect_to(citizens_dependant_relationship_path(latest_dependant.id))
end
```

Kalagan: Good idea 👍

PR #623, opened 2019-07-05
AP-755 - remove proceeding types not needed for MVP

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-755)

Most of the proceedings in the current seed file and proceeding_types database table aren't needed for MVP. This PR removes them by

* removing the data from the seed file
* deleting the data from the database (after amending any existing records that use those proceedings to ensure referential integrity)
* amending feature tests to use valid proceedings

It also adds one missing proceeding that is needed - DA020.

## Checklist


db/migrate/20190705091659_delete_unnecessary_proceeding_types.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def proceeding_ids_to_be_deleted
+    ProceedingType.where.not(ccms_code: %w[DA001 DA002 DA003 DA004 DA005 DA006 DA007]).pluck(:id)
+  end
+
+  def up
+    ApplicationProceedingType.where(proceeding_type_id: proceeding_ids_to_be_deleted).update_all(proceeding_type_id: safe_proceeding_id)
+    ProceedingType.delete(proceeding_ids_to_be_deleted)
+    ProceedingType.create(code: 'PR0220',
```
Kalagan: `.create!` instead of `.create`

Kalagan: Anyway, I don't think we need to add this proceeding type here. I think that it will be added by `db:seed` which is run when the container starts: https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/docker/setup_db#L8

db/migrate/20190705091659_delete_unnecessary_proceeding_types.rb
```diff
@@ -0,0 +1,26 @@
...
+                          ccms_category_law: 'Family',
+                          ccms_matter_code: 'MINJN',
+                          ccms_matter: 'Domestic Abuse',
+                          meaning: 'FGM Protection Order-Domestic Abuse ',
+                          description: 'To be represented on an application for a Female Genital Mutilation Protection Order under the Female Genital Mutilation Act.')
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
```
Kalagan: In my opinion, we should leave this method empty. Which means that you have the option to rollback passed that migration.
Otherwise, with this `ActiveRecord::IrreversibleMigration` you can never rollback more than this migration

db/schema.rb
```diff
@@ -338,6 +338,17 @@
     t.index ["original_file_id"], name: "index_pdf_files_on_original_file_id", unique: true
   end
 
+  create_table "proceeding_type_scope_limitations", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
Kalagan: The changes to this file other than line 13 don't seem to belong to this PR.
Use `db:schema:load` to align your DB with the `schema.rb`

db/migrate/20190705091659_delete_unnecessary_proceeding_types.rb
mpw5: Thanks. I tried it using just the additional seed entry but that didn't work for me, so I added it to the migration too. I've removed it from the migration and it works now though, I must have been doing something wrong 🤷‍♂

PR #624, opened 2019-07-05
AP-757 - Create and populate scope limitations table and model

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-757)

Create a database table, model and seed data for scope limitations.

## Checklist

### DIFF
--- a/app/services/scope_limitations_populator.rb
+++ b/app/services/scope_limitations_populator.rb
@@ -1,0 +1,14 @@
+class ScopeLimitationsPopulator
+  def self.call
+    new.call
+  end
+
+  def call
+    file_path = Rails.root.join('db', 'seeds', 'scope_limitations.csv')
+
+    CSV.read(file_path, headers: true, header_converters: :symbol).each do |row|
+      scope_limitation = ScopeLimitation.where(code: row[:code]).first_or_initialize
+      scope_limitation.update! row.to_h
+    end
+  end
+end
--- a/db/seeds/scope_limitations.rb
+++ b/db/seeds/scope_limitations.rb
@@ -1,0 +1,2 @@
+Rails.logger.info 'Populating scope limitations'
+ScopeLimitation.populate
--- a/spec/factories/scope_limitations.rb
+++ b/spec/factories/scope_limitations.rb
@@ -5,13 +5,13 @@ FactoryBot.define do
     description { 'Description' }
     substantive { true }
     delegated_functions { false }
-  end
 
-  trait :with_real_data do
-    code { 'AA009' }
-    meaning { 'Warrant of arrest FLA' }
-    description { 'As to an order under Part IV Family Law Act 1996 limited to an application for the issue of a warrant of arrest.' }
-    substantive { true }
-    delegated_functions { true }
+    trait :with_real_data do
+      code { 'AA009' }
+      meaning { 'Warrant of arrest FLA' }
+      description { 'As to an order under Part IV Family Law Act 1996 limited to an application for the issue of a warrant of arrest.' }
+      substantive { true }
+      delegated_functions { true }
+    end
   end
 end
--- a/spec/factories/scope_limitations.rb
+++ b/spec/factories/scope_limitations.rb
@@ -1,0 +1,17 @@
+FactoryBot.define do
+  factory :scope_limitation do
+    sequence(:code) { |n| format('AA%03d', n) }
+    meaning { 'Meaning' }
+    description { 'Description' }
+    substantive { true }
+    delegated_functions { false }
+  end
+
+  trait :with_real_data do
+    code { 'AA009' }
+    meaning { 'Warrant of arrest FLA' }
+    description { 'As to an order under Part IV Family Law Act 1996 limited to an application for the issue of a warrant of arrest.' }
+    substantive { true }
+    delegated_functions { true }
+  end
+end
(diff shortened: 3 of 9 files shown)


db/seeds/scope_limitations.rb
```diff
@@ -0,0 +1,18 @@
...
+data.each do |row|
+  code, meaning, description, substantive, delegated_functions = row
+  scope_limitation = ScopeLimitation.where(code: code).first_or_initialize
+  attrs = {
+    meaning: meaning,
+    description: description,
+    substantive: substantive,
+    delegated_functions: delegated_functions
+  }
+  scope_limitation.update_attributes(attrs)
```
Kalagan: Would `.update` work here? `.update_attributes` seems to be an alias of  `.update` (https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attributes)

Also, I recommend using the version with the exclamation mark (`.update!`). Otherwise, if it fails, it won't raise any error.

spec/models/scope_limitation_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  describe 'validation' do
+    subject { described_class.new }
+
+    before do
+      subject.code = 'AA001'
+      subject.meaning = 'test_limitation'
+      subject.description = 'this is a test limitation'
+      subject.substantive = true
+      subject.delegated_functions = false
+    end
```
reggieb: Why not use a factory here?

db/seeds/scope_limitations.rb
```diff
@@ -0,0 +1,18 @@
+require 'csv'
+
+file_path = Rails.root.join('db', 'seeds', 'scope_limitations.csv')
+
+data = CSV.read(file_path)
+data.shift
```
reggieb: If you use [the `headers: true` option of `CSV.read`](http://ruby-doc.org/stdlib-2.6.3/libdoc/csv/rdoc/CSV.html#class-CSV-label-CSV+with+headers), the code here can be simplified a lot. I think this will work:
```ruby
CSV.read(file_path, header:true, header_converters: :symbol).do |row|
  scope_limitation = ScopeLimitation.where(code: row[:code]).first_or_initialize
  scope_limitation.update! row.to_h
end
```

reggieb:
I think the data load is a little more complicated than it was for restrictions: hence my suggestion of using a service. It would be neat if the service could be triggered by `ScopeLimitation.populate`

mpw5:
@Kalagan @reggieb I've made changes that I think address your comments. Can you re-review when you have time please?

spec/models/scope_limitation_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+RSpec.describe ScopeLimitation, type: :model do
+  let(:scope_limitations_populator) { ScopeLimitationsPopulator.new }
+
+  before do
+    allow(ScopeLimitationsPopulator).to receive(:new).and_return(scope_limitations_populator)
+  end
+
+  describe '.populate' do
+    it 'calls the scope_limitations_populator service' do
+      expect(scope_limitations_populator).to receive(:populate).with(no_args)
```
Kalagan: If you add the class method like on my suggestion, you could just write `expect(ScopeLimitationsPopulator).to receive(:call).with(no_args)` and you wouldn't need the extra `let(:scope_limitations_populator) { ScopeLimitationsPopulator.new }`

spec/factories/scope_limitations.rb
```diff
@@ -0,0 +1,17 @@
+FactoryBot.define do
+  factory :scope_limitation do
+    sequence(:code) { |n| format('AA%03d', n) }
+    meaning { 'Meaning' }
+    description { 'Description' }
+    substantive { true }
+    delegated_functions { false }
+  end
+
+  trait :with_real_data do
```
Kalagan: I think the `trait` should be defined inside the `factory :scope_limitation` block. Like for the other factories.

reggieb:
Agree with Julien's points

mpw5:
Thanks. Fixed.

app/services/scope_limitations_populator.rb
```diff
@@ -0,0 +1,14 @@
...
+  def call
+    file_path = Rails.root.join('db', 'seeds', 'scope_limitations.csv')
+
+    CSV.read(file_path, headers: true, header_converters: :symbol).each do |row|
+      scope_limitation = ScopeLimitation.where(code: row[:code]).first_or_initialize
+      scope_limitation.update! row.to_h
+    end
+  end
+
+  def self.call
```
Kalagan: usually the class methods are placed at the top before the other ones

db/seeds/scope_limitations.rb
```diff
@@ -0,0 +1,2 @@
+Rails.logger.info 'Populating scope limitations'
+ScopeLimitation.populate
```
Kalagan: `ScopeLimitation.call`

spec/services/scope_limitations_populator_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'rails_helper'
+require 'csv'
+
+RSpec.describe ScopeLimitationsPopulator do
+  describe '#populate' do
```
Kalagan: `'.call'`

db/seeds/scope_limitations.rb
Kalagan: oops no sorry forget that. That's the one in the model

PR #633, opened 2019-07-08
Stop Faker warnings

Kalagan:
Remove annoying Faker's deprecation warnings by removing the use of `Faker::Number.decimal`

### DIFF
--- a/spec/factories/other_assets_declarations.rb
+++ b/spec/factories/other_assets_declarations.rb
@@ -3,21 +3,21 @@ FactoryBot.define do
     legal_aid_application
 
     trait :with_second_home do
-      second_home_value { Faker::Number.decimal(6, 2) }
-      second_home_mortgage { Faker::Number.decimal(6, 2) }
-      second_home_percentage { Faker::Number.decimal(2, 2) }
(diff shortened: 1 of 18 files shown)


spec/factories/other_assets_declarations.rb
```diff
@@ -45,10 +45,10 @@
     end
 
     trait :mix_of_values do
-      second_home_value { Faker::Number.decimal(6, 2) }
-      second_home_mortgage { Faker::Number.decimal(6, 2) }
-      second_home_percentage { Faker::Number.decimal(6, 2) }
-      timeshare_property_value { Faker::Number.decimal(6, 2) }
+      second_home_value { rand(1...1_000_000.0).round(2) }
+      second_home_mortgage { rand(1...1_000_000.0).round(2) }
+      second_home_percentage { rand(1...1_000_000.0).round(2) }
```
mpw5: As this is a percentage field I think it needs to be `second_home_percentage { rand(1...99.0).round(2) }`.

Kalagan: Good spot. I missed this one

reggieb:
:+1: on this being a welcome change

PR #634, opened 2019-07-08
AP723 Dependants Income

Obsiye:
Dependants income page, which records whether the dependant has an income or not. If it does then record this value

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-723)

- Add attributes/columns to the Dependants table: has_income and monthly_income
- Create controller, form, view and route for Dependants income get and patch paths
- Update locales
- Update Citizen dependant flow path
- Create request and feature tests

## Checklist

### DIFF
--- a/app/forms/dependant_form/monthly_income_form.rb
+++ b/app/forms/dependant_form/monthly_income_form.rb
@@ -1,0 +1,33 @@
+module DependantForm
+  class MonthlyIncomeForm
+    include BaseForm
+    form_for Dependant
+
+    before_validation :clear_monthly_income
+
+    attr_accessor :has_income, :monthly_income
+
+    validate :income_presence
+    validates :monthly_income, presence: true, if: proc { |form| form.has_income.to_s == 'true' }
+    validates :monthly_income, allow_blank: true, currency: { greater_than: 0.0 }
+
+    def attributes_to_clean
+      [:monthly_income]
+    end
+
+    private
+
+    def clear_monthly_income
+      monthly_income&.clear unless ActiveModel::Type::Boolean.new.cast(has_income.to_s)
+    end
+
+    def income_presence
+      return if has_income.present?
+
+      errors.add(
+        :has_income,
+        I18n.t('activemodel.errors.models.dependant.attributes.has_income.blank', name: model.name)
+      )
+    end
+  end
+end
--- a/app/forms/dependant_form/monthly_income_form.rb
+++ b/app/forms/dependant_form/monthly_income_form.rb
@@ -7,7 +7,7 @@ module DependantForm
 
     attr_accessor :has_income, :monthly_income
 
-    validate :income_presence
+    validates :has_income, presence: { message: ->(form, _) { form.error_message } }
     validates :monthly_income, presence: true, if: proc { |form| form.has_income.to_s == 'true' }
     validates :monthly_income, allow_blank: true, currency: { greater_than: 0.0 }
 
@@ -15,19 +15,14 @@ module DependantForm
       [:monthly_income]
     end
 
+    def error_message
+      I18n.t('activemodel.errors.models.dependant.attributes.has_income.blank_message', name: model.name)
+    end
+
     private
 
     def clear_monthly_income
       monthly_income&.clear unless ActiveModel::Type::Boolean.new.cast(has_income.to_s)
-    end
-
-    def income_presence
-      return if has_income.present?
-
-      errors.add(
-        :has_income,
-        I18n.t('activemodel.errors.models.dependant.attributes.has_income.blank', name: model.name)
-      )
     end
   end
 end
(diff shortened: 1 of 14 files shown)


app/services/flow/flows/citizen_dependants.rb
```diff
@@ -15,13 +15,17 @@ class CitizenDependants < FlowSteps
...
+          forward: ->(_, dependant) { dependant.adult_relative? || dependant.eighteen_or_less? ? :dependants_monthly_incomes : :dependants_education }
         },
         dependants_education: {
           path: ->(_, dependant) { "[PLACEHOLDER] #{dependant.name} dependants education" }
         },
-        dependants_income: {
-          path: ->(_, dependant) { "[PLACEHOLDER] #{dependant.name} dependants income" }
+        dependants_monthly_incomes: {
+          path: ->(_, dependant) { urls.citizens_dependant_monthly_income_path(dependant) },
+          forward: ->(_) { :dependants_assets }
```
Kalagan: you can simply write `forward: :dependants_assets` here.

db/schema.rb
```diff
@@ -308,25 +310,25 @@
...
-    t.decimal "trust_value", precision: 14, scale: 2
+    t.decimal "second_home_value"
+    t.decimal "second_home_mortgage"
+    t.decimal "second_home_percentage"
+    t.decimal "timeshare_property_value"
+    t.decimal "land_value"
+    t.decimal "valuable_items_value"
+    t.decimal "money_assets_value"
+    t.decimal "money_owed_value"
+    t.decimal "trust_value"
```
Kalagan: This file contains lots of change irrelevant to this PR.
Doing this should fix it:
1. revert the changes by copying the value of this file from master (https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/db/schema.rb)
2. `bin/rails db:schema:load db:migrate db:seed`

features/citizens/citizens.feature
```diff
@@ -298,4 +298,24 @@ Feature: Citizen journey
     Then I should be on a page showing "What is your relationship to"
     Then I choose "They're a child relative"
     Then I click "Save and continue"
-    Then I should be on a page showing "dependants income"
+    Then I should be on a page showing "receive any income?"
+
+  @javascript
+  Scenario: I have an adult dependant more than 18 years old
```
Kalagan: Rather than creating a new scenario, I suggest carrying on with the previous one.

app/forms/dependant_form/monthly_income_form.rb
```diff
@@ -0,0 +1,33 @@
+module DependantForm
+  class MonthlyIncomeForm
+    include BaseForm
+    form_for Dependant
+
+    before_validation :clear_monthly_income
+
+    attr_accessor :has_income, :monthly_income
+
+    validate :income_presence
```
reggieb: Why doesn't this work:
```ruby
validates :has_income, presence: true
```
That would avoid using a custom method.

app/services/flow/flows/citizen_dependants.rb
reggieb: I think this will work:
```ruby
forward: :depandants_assets
```

app/views/shared/forms/dependants/_monthly_incomes.html.erb
```diff
@@ -0,0 +1,45 @@
+<%= page_template(
+      page_title: t('.page_title', name: @form.model.name),
+      back_link: { text: t('generic.back') },
```
reggieb:
```ruby
back_link: { text: t('generic.back') },
```
You don't need that line. `page_template` defaults to using `t('generic.back')`

db/schema.rb
```diff
@@ -264,23 +266,23 @@
     t.boolean "open_banking_consent"
     t.datetime "open_banking_consent_choice_at"
     t.string "own_home"
-    t.decimal "percentage_home"
     t.decimal "property_value", precision: 10, scale: 2
     t.string "shared_ownership"
     t.decimal "outstanding_mortgage_amount"
+    t.decimal "percentage_home"
```
reggieb: These fields moving suggest you didn't `rake db:schema:load` on starting this branch. The schema needs to be fixed, so that only the entries made in the migration modify the schema

app/forms/dependant_form/monthly_income_form.rb
AbdirizakObsiye2018: I used the custom method because the error message required the dependants name within it e.g. Select yes if John receives any income

features/citizens/citizens.feature
AbdirizakObsiye2018: Carrying onto creating a new dependant that's more than 18 years old, will require the add assets page / feature which hasn't been implemented yet ( add income page - > add assets page - > any other dependants page )

PR #645, opened 2019-07-11
AP-758 Create and populate proceeding_type_scope_limitations join table

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-758)

A join table is needed to provide a many-to-many relationship between `proceeding_types` and `scope_limitations`:

* add model and migration to create the table - `proceeding_types_scope_limitations` with foreign keys to the existing `proceeding_types` and `scope_limitations` tables and indexes to ensure unique records and only one `substantive_default` and `delegated_functions` default exists per `proceeding_type`
* amend existing `proceeding_types` and `scope_limitations` models to add `has_many` relationships
* add seed data and create a service class to populate the seed data via the model
* amend seeding process to ensure that `proceeding_types`, `scope_limitations` and `proceeding_types_scope_limitations` are always seeded in the correct order

## Checklist


app/models/proceeding_type_scope_limitation.rb
```diff
@@ -0,0 +1,11 @@
+class ProceedingTypeScopeLimitation < ApplicationRecord
+  belongs_to :proceeding_type
+  belongs_to :scope_limitation
+  validates_uniqueness_of :proceeding_type_id, scope: :scope_limitation_id
+  validates_uniqueness_of :proceeding_type_id, conditions: -> { where(substantive_default: true) }, scope: :substantive_default
+  validates_uniqueness_of :proceeding_type_id, conditions: -> { where(delegated_functions_default: true) }, scope: :delegated_functions_default
+
```
stephenrichards: It looks to me as though you have a unique index on the database for ensuring there is only one true substantive default and one true delegated_functions_default per proceeding_type, so do you need this `validate_uniqueness_of` lines?  `validates_uniqueness_of` is dog slow on inserts, not that that is really a worry for this table, but if it is taken care of by the database, then we should remove them from here.

mpw5: I was working on the basis that it's best to validate in the model and the database (https://stackoverflow.com/a/13123347 for example). 

I actually do need the `validates_uniqueness` at line 4 - if `db:seed` is run more than once the database throws a duplicate error. Having that line allows me to handle the `ActiveRecord::RecordInvalid` exception that is thrown by the model before the data reaches the database, and means the seed can be run multiple times without error.

The other two are just belt-and-braces, and I hadn't considered the performance impacts, so I'm happy to remove line 5 and 6 if you think it's better.

mpw5: Ignore my second paragraph above. I've improved the way the seeding works so that's no longer an issue. I'll  remove the three `validates_uniqueness_of` lines if you think they're not necessary.

Kalagan: I agree with Stephen and I don't think these validations are useful.
Like you said, these validations throw `ActiveRecord::RecordInvalid`. However, even without them the DB would prevent the insertion and `ActiveRecord::RecordNotUnique` would be thrown.

mpw5: Thanks. Removed now.

db/migrate/20190705150936_create_proceeding_type_scope_limitations.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    add_index :proceeding_type_scope_limitations,
+              %i[proceeding_type_id scope_limitation_id],
+              unique: true,
+              name: 'index_proceedings_scopes_unique_on_ids'
+
+    add_index :proceeding_type_scope_limitations,
+              %i[proceeding_type_id substantive_default],
+              unique: true,
+              where: 'substantive_default = true',
```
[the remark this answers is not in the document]
mpw5: Neither did I until yesterday!

PR #648, opened 2019-07-12
AP-756 Add additional proceeding fields

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-756)

New fields need to be added to the `proceeding_type` database table, with appropriate data added to existing records in the seed data.

* create a migration to add the required fields
* add additional data to the `proceeding_types` seed file
* tidy up data in that seed file - remove duplicate `meaning` field; rename headings to match table column names to allow easier seeding; set all substantive cost limitations to £25000
* refactor proceeding seed code to bring it in line with the way `scope_limitations` and `proceeding_type_scope_limitations` are seeded
* create table, model, and seed data for `service_levels` and add foreign key relationship from that to `proceeding_types`

## Checklist


app/services/proceeding_type_populator.rb
```diff
@@ -0,0 +1,18 @@
+require 'csv'
+
+class ProceedingTypePopulator
+  def self.call
+    new.call
+  end
+
+  def file_path
+    Rails.root.join('db', 'seeds', 'proceeding_types.csv').freeze
```
reggieb: Pedantic, and certainly a matter of taste ... but I think this is easier to read:
```ruby
Rails.root.join('db/seeds/proceeding_types.csv').freeze
```

db/migrate/20190711135901_add_fields_to_proceeding_types.rb
```diff
@@ -0,0 +1,9 @@
+class AddFieldsToProceedingTypes < ActiveRecord::Migration[5.2]
+  def change
+    add_column :proceeding_types, :default_level_of_service_id, :integer
+    add_column :proceeding_types, :default_level_of_service_name, :string
```
reggieb: Having a series of methods with a shared prefix or suffix is often a sign that another object is needed, and I think that maybe the case here. I suspect that the id/name relationship is constant (that is, a level of service with id n will always have the same name), and that there are a smaller number of service levels than there are Proceeding types.

I think what is needed here is another object `ServiceLevel`, which has the attributes `service_id` and `name`. Then in `ProceedingType`:

```ruby
belongs_to :default_level_of_service, class_name: 'ServiceLevel', foreign_key: "default_service_level_id"
```
or perhaps
```ruby
belongs_to :default_service_level, class_name: 'ServiceLevel', foreign_key: "default_service_level_id"
```
Otherwise it is likely that inconsistencies between service id and name will creep into the data.

db/migrate/20190711135901_add_fields_to_proceeding_types.rb
```diff
@@ -0,0 +1,9 @@
...
+    add_column :proceeding_types, :default_cost_limitation_delegated_functions, :integer
+    add_column :proceeding_types, :default_cost_limitation_substantive, :integer
```
reggieb: Another two attributes sharing a common prefix, `default_cost_limitation_` in this case. However, I expect that the two are not closely coupled ( that is, that two proceeding types could share the same `default_cost_limitation_delegated_functions` and still have different `default_cost_limitation_substantive` values.

However, if that is not the case (that is, if a proceeding type has a `default_cost_limitation_delegated_functions` value of X, it will always have a `default_cost_limitation_substantive` value of Y), then a separate object should be considered.

reggieb: Of course, if default cost limitations are tightly coupled to the service level, they would belong in `ServiceLevel`

db/migrate/20190711135901_add_fields_to_proceeding_types.rb
stephenrichards: yep, on reflection I agree with Rob

db/migrate/20190711135901_add_fields_to_proceeding_types.rb
stephenrichards: They are not related to service level, and they are not tightly coupled. - they simply represent the default cost limitation for each type of applicaiton: a substantive application and a delegated function application.  So I think these are best left as-is.

db/migrate/20190711135901_add_fields_to_proceeding_types.rb
mpw5: I agree with Rob too. It crossed my mind to make `level_of_service` a separate object when I was making the original change, but as we currently only have one level of service it didn't seem like a priority at this point.

But you've convinced me it's worth doing, so I've pushed a commit to change it.

db/migrate/20190711135901_add_fields_to_proceeding_types.rb
mpw5: Agreed, I'll leave this one alone.

PR #650, opened 2019-07-12
Add a Mock Benefit Checker that can be used in dev and test environments

reggieb:
Adds a simple lookup Mock of the benefit checker, for when we have problems accessing the Benefits Checker API

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,9 @@ helm list --tiller-namespace=laa-apply-for-legalaid-uat --namespace=laa-apply-fo
 helm delete <name-of-the-release> --tiller-namespace=laa-apply-for-legalaid-uat --purge
 ```
 
+## Dev: running locally
 
-## LAA Portal Authentication dev setup
+### LAA Portal Authentication dev setup
 
 Add the following to .env.development
 ```
@@ -116,12 +117,26 @@ LAA_PORTAL_MOCK_SAML=false
 ```
 ** Note the above keys can be found in rattic
 
+### Authentication
+
 Mock Saml request on dev add the the following settings
 
 ```
 LAA_PORTAL_IDP_SSO_TARGET_URL=http://localhost:3002/saml/auth
 LAA_PORTAL_MOCK_SAML=true
 ```
+
+### Benefits checker
+
+To mock the benefits check in development and testing add the following environment
+variable:
+
+```
+USE_BC_DEV_MOCK=true
+```
+
+This will enable `MockBenefitCheckService`. See `MockBenefitCheckService::KNOWN for
+credentials that will return 'Yes' for has qualifying benefits.
 
 ## Admin Portal
 
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ To mock the benefits check in development and testing add the following environm
 variable:
 
 ```
-USE_BC_DEV_MOCK=true
+BC_USE_DEV_MOCK=true
 ```
 
 This will enable `MockBenefitCheckService`. See `MockBenefitCheckService::KNOWN for
--- a/app/services/benefit_check_service.rb
+++ b/app/services/benefit_check_service.rb
@@ -1,6 +1,6 @@
 class BenefitCheckService
   BENEFIT_CHECKER_NAMESPACE = 'https://lsc.gov.uk/benefitchecker/service/1.0/API_1.0_Check'.freeze
-  USE_MOCK = ActiveModel::Type::Boolean.new.cast(ENV['USE_BC_DEV_MOCK'])
+  USE_MOCK = ActiveModel::Type::Boolean.new.cast(ENV['BC_USE_DEV_MOCK'])
 
   class ApiError < StandardError
     include Nesty::NestedError
@@ -9,10 +9,7 @@ class BenefitCheckService
   def self.call(application)
     return MockBenefitCheckService.call(application) if USE_MOCK && !Rails.env.production?
 
-    # :nocov:
-    # Not covered as this part of method will be skipped if USE_MOCK enabled
     new(application).call
-    # :nocov:
   end
 
   def initialize(application)
--- a/app/services/benefit_check_service.rb
+++ b/app/services/benefit_check_service.rb
@@ -1,6 +1,19 @@
 class BenefitCheckService
   BENEFIT_CHECKER_NAMESPACE = 'https://lsc.gov.uk/benefitchecker/service/1.0/API_1.0_Check'.freeze
-  ApiError = Class.new(StandardError)
+  USE_MOCK = ActiveModel::Type::Boolean.new.cast(ENV['USE_BC_DEV_MOCK'])
+
+  class ApiError < StandardError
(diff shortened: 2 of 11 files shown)


app/services/benefit_check_service.rb
```diff
@@ -1,6 +1,19 @@
 class BenefitCheckService
   BENEFIT_CHECKER_NAMESPACE = 'https://lsc.gov.uk/benefitchecker/service/1.0/API_1.0_Check'.freeze
-  ApiError = Class.new(StandardError)
+  USE_MOCK = ActiveModel::Type::Boolean.new.cast(ENV['USE_BC_DEV_MOCK'])
```
Kalagan: Shouldn't this `ENV` be read in `config/application.rb` like the other BenefitChecker related settings?
https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/config/application.rb#L20

app/services/benefit_check_service.rb
```diff
@@ -11,8 +24,6 @@ def call
     soap_client.call(:check, message: benefit_checker_params).body.dig(:benefit_checker_response)
   rescue Savon::SOAPFault => e
     raise ApiError, "HTTP #{e.http.code}, #{e.to_hash}"
-  rescue Net::ReadTimeout, Net::OpenTimeout => e
-    raise ApiError, e
```
Kalagan: Why do you remove these 2?
On this PR you chose to add `Net::OpenTimeout`: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/638/files

README.md
```diff
@@ -116,13 +117,27 @@ LAA_PORTAL_MOCK_SAML=false
...
 LAA_PORTAL_MOCK_SAML=true
 ```
 
+### Benefits checker
+
+To mock the benefits check in development and testing add the following environment
+variable:
+
+```
+USE_BC_DEV_MOCK=true
```
Kalagan: I think the name `BC_USE_DEV_MOCK` would fit better with the other benefit checker settings : 
```
BC_LSC_SERVICE_NAME
BC_CLIENT_ORG_ID
BC_CLIENT_USER_ID
BC_WSDL_URL
```

spec/requests/providers/check_benefits_spec.rb
```diff
@@ -28,7 +28,7 @@
       let!(:benefit_check_result) { create :benefit_check_result, legal_aid_application: application }
 
       it 'does not generate a new one' do
-        expect_any_instance_of(BenefitCheckService).not_to receive(:call).and_call_original
+        expect(BenefitCheckService).not_to receive(:call).and_call_original
```
Kalagan: I know it's not your code but I don't think `.and_call_original` is needed if `BenefitCheckService` is not supposed to receive `call` anyway

app/services/benefit_check_service.rb
```diff
@@ -1,6 +1,19 @@
...
+    include Nesty::NestedError
+  end
+
+  def self.call(application)
+    return MockBenefitCheckService.call(application) if USE_MOCK && !Rails.env.production?
+
+    # :nocov:
+    # Not covered as this part of method will be skipped if USE_MOCK enabled
+    new(application).call
+    # :nocov:
```
Kalagan: I don't understand why we can't cover this. Don't we have tests calling the real benefit checker but using VCR?

app/services/benefit_check_service.rb
reggieb: Because on building this PR and suffering from having the endpoint down, I have had more time to look at the errors and how they behave. From this I came to the following conclusions:

- `BenefitCheckService::ApiError` isn't caught anywhere so it's not being used for process flow. 
- Catching the `Savon::SOAPFault` is worthwhile because the default error message is ambiguous and contains limited information for debugging. I added `nesty` so we get the best of both worlds (the Savon error and the reformatted version).
- The `Net` errors are fine and provide the information needed to debug. Catching them and then raising a new error just obscures the data returned within the `Net` error. It's far better to just let the original exception bubble to the surface.

app/services/benefit_check_service.rb
reggieb: If you look at the specs for `BenefitCheckService` they use VCR. The key thing to notice is that the instance tested is initiated directly via `BenefitCheckService.new`. So all the instance methods (including `#call`) are thoroughly tested using VCR.

All the other specs (for example from the controller) mock the call to the `call` method. To test a non-mocked version would require a new spec with a working VCR cassette. As I still don't have a working API connection from dev, there isn't a simple way of building one.

As the code `new(application).call` is already thoroughly tested, putting this call within a `:nocov:` seemed the simplest solution, and had minimal risks.

reggieb: You're right to query this. There is a simple way this can be avoided.

app/services/benefit_check_service.rb
reggieb: I wanted to limit the scope of this environment variable. It should never be used in production, so shouldn't suffer from the issues that make going through `Rails.configuration.x` worthwhile, and therefore passing it though `Rails.configuration.x` seems like an unnecessary complication.

app/services/benefit_check_service.rb
Kalagan: Changing the test to use `BenefitCheckService.call(application)` instead of `BenefitCheckService.new` should do it.
I had a go at it: https://github.com/ministryofjustice/laa-apply-for-legal-aid/compare/add-dev-mock-for-benefit-checker...add-dev-mock-for-benefit-checker-julien?expand=1

app/services/benefit_check_service.rb
reggieb: I'd also add that having to already add `Net::OpenTimeout` to the list of caught exceptions suggested that all the likely error were in the list - which then begs the question "Why are some `Net` errors caught and others not?". Better not to catch any unless there is value added by doing so.

app/services/benefit_check_service.rb
Kalagan: good point well made

app/services/benefit_check_service.rb
Kalagan: good point well made

app/services/benefit_check_service.rb
reggieb: I found another way to do it - add a specific test of `BenefitCheckService.call` and mock the called to the instance method `call`. It ensures the behaviour is tested, with minimal change to the existing code.

Kalagan: Up to you but It feels a bit like a shortcut, to be honest 😄 
If you were to re-write the `class BenefitCheckService` from scratch which approach would you choose?

reggieb: I'd do it the same way (that is build most of the specs using a instance initiated via `.new`) - but not mock the `BenefitCheckService.call` call in this latest spec. However, as I don't have a working API callout, I can't build the VCR cassette simply. So .... shortcut.

PR #656, opened 2019-07-15
AP-773 Add new feature to test journey immediately after check answers

reggieb:
[Jira AP-773](https://dsdmoj.atlassian.net/browse/AP-773)

Ensures that all the journeys detailed in the workflow on the AP-773 ticket are represented within the cucumber tests. To make this clearer and to keep the related scenarios together, I moved the existing scenarios to a new feature, and added there the variants that ensured all the pathways were tested.

Also fixes a bug in mock benefit checker - there needs to be a match for the Walker entity that is already defined in the specs and features.


features/providers/civil_application_journey.feature
```diff
@@ -302,32 +302,6 @@ Feature: Civil application journeys
...
-    Then I click 'Save and continue'
-    Then I should be on a page showing 'must complete a financial assessment'
-    Then I click 'Continue'
-    Then I should be on a page showing 'Do you want to make a substantive application now?'
-    Then I choose 'Yes'
-    Then I click 'Save and continue'
-    Then I should be on a page showing 'Does your client use online banking?'
-
-  @javascript @vcr
-  Scenario: I use delegated functions and will return later
```
Kalagan: Can we delete the VCR cassettes matching this scenario and the one above?

reggieb: Good point!

PR #660, opened 2019-07-16
AP-796 Generate delegated function blocks from application

stephenrichards:
## Generate delegated functions blocks from data 

[Link to story](https://dsdmoj.atlassian.net/browse/AP-796)

This story is to replace hard-coded dummy data for the delegated functions blocks (in the means section and the merits sections of the payload) with data from the applications

- Created an `Opponent.dummy_opponent` method to return the standard dummy opponent, or create it if it doesn't already exist
- Modified `LegalAidApplication` to return a collection comprising the dummy opponent for the `opponent` and `other_parties` methods.
- Added methods `supervisor_contact_id` and `fee_earner_contact_id` to `Provider`.  These return dummy data until such time as we are provided with real data from the `ProviderDetails` API
- Created an Rspec support class `XmlExtractor` to retrieve attribute blocks from deep within the payload
- Added two custom matchers `have_response_type` and `have_response_value` for verifying the contents of an attribute block
- updated the entries for the two `DELEGATED_FUNCTIONS_DATE` attributes in `config/ccms_keys.yml` to replace hard-coded data with data from the `LegalAidApplication` and to suppress generation of the blocks if delegated functions not used

## Checklist


app/models/opponent.rb
```diff
@@ -36,6 +36,26 @@ class Opponent < ApplicationRecord
     interested_party: 'Interested party'
   }.freeze
 
+  def self.dummy_opponent
+    Opponent.find_by(other_party_id: 'OPPONENT_00000001') || create_dummy_opponent
```
Kalagan: I suggest one of these:
```ruby
    Opponent.create_with(
      title: 'MR.',
      first_name: 'Dummy',
      surname: 'Opponent',
      date_of_birth: Date.new(2015, 1, 1),
      relationship_to_case: 'OPP',
      relationship_to_client: 'EX_SPOUSE',
      opponent_type: 'PERSON',
      opp_relationship_to_case: 'Opponent',
      opp_relationship_to_client: 'Ex Husband/ Wife',
      child: false
    ).find_or_create_by!(other_party_id: 'OPPONENT_00000001')
```
```ruby
    Opponent.find_or_create_by!(other_party_id: 'OPPONENT_00000001') do |opponent|
      opponent.title = 'MR.'
      opponent.first_name = 'Dummy'
      opponent.surname = 'Opponent'
      opponent.date_of_birth = Date.new(2015, 1, 1)
      opponent.relationship_to_case = 'OPP'
      opponent.relationship_to_client = 'EX_SPOUSE'
      opponent.opponent_type = 'PERSON'
      opponent.opp_relationship_to_case = 'Opponent'
      opponent.opp_relationship_to_client = 'Ex Husband/ Wife'
      opponent.child = false
    end
```
Both are equivalent (see https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by)

app/models/opponent.rb
```diff
@@ -36,6 +36,26 @@ class Opponent < ApplicationRecord
...
+  end
+
+  def self.create_dummy_opponent
+    Opponent.create(
```
Kalagan: I suggest always using `create!` instead of `create` unless you are processing the result of `create` and showing the errors

app/models/opponent.rb
```diff
@@ -36,6 +36,26 @@ class Opponent < ApplicationRecord
...
+      other_party_id: 'OPPONENT_00000001',
```
Kalagan: Unless you decide to use `find_or_create_by!`, I think this string should be a constant as it is repeated twice.
Does it have to be `OPPONENT_00000001` ? Would `OPPONENT_1` be a bit simpler?

app/models/opponent.rb
stephenrichards: Yep, good point, thanks.

app/models/opponent.rb
stephenrichards: Agreed

app/models/opponent.rb
stephenrichards: It does have to be OPPONENT_nnnnnnnn I'm afraid.  In fact, this is a value that is going to be dictated by CCMS - a different value for each environment, so there will be more work to do in this area, but this will suffice for now.

PR #662, opened 2019-07-17
Fix intermittent failing tests

Kalagan:
## What

I noticed that the `it 'redirects to the assets page'` test of `monthly_incomes_spec.rb` is intermittently failing if the dependant happens to be less than 18 years old.
So I fixed this test by fixing the age of the dependant using the `:over_18` factory trait.
- I also added another test for the scenario if the dependant is less than 18.
- Added all attributes in the dependant's factory.
- Removed a duplicate in `schema.rb`

---

Additionally, this test is also failing intermittently: https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/spec/forms/incidents/told_on_form_spec.rb#L99
image
Because the test only passed `told_on` as a parameter to the form but not `occurred_on`, it can happen that `occurred_on` which is generated as `Faker::Date.backward(90)` by the factory could be after the `told_on` in the param (`3.days.ago.to_date`). In that case, the validation error `"Occurred on Date the incident occurred must be before the date your client told you about it"` prevents the `.save` and the test fails.
Modifying the factory to make `occurred_on` to be much earlier prevents this from happening.

## Checklist

Before you ask people to review this PR:a

### DIFF
--- a/spec/factories/dependants.rb
+++ b/spec/factories/dependants.rb
@@ -4,7 +4,12 @@ FactoryBot.define do
     number { Faker::Number.number(2) }
     name { Faker::Name.name }
     date_of_birth { Faker::Date.birthday(7, 77) }
-    relationship {  Dependant.relationships.keys.sample }
+    relationship { Dependant.relationships.keys.sample }
+    monthly_income { rand(1...10_000.0).round(2) }
+    has_income { Faker::Boolean.boolean }
+    in_full_time_education { Faker::Boolean.boolean }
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -213,7 +213,6 @@
     t.string "relationship"
     t.decimal "monthly_income"
     t.boolean "has_income"
-    t.string "relationship"
```
mpw5: Should this have been deleted?

mpw5: Oh, it's a duplicate. Ignore me.

PR #667, opened 2019-07-18
AP-802 Generate CCMS payload blocks for respondent

stephenrichards:
There are a number of XML attribute blocks that need to be generated depending
on values on the Respondent record.  This PR replaces the code that produced
dummy values for proof of concept with values from the record.

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-802)

- Added methods `receives_tax_credits?`, `receives_wages?` and `receives_benefits?` to the `BankAccount` model.  These are required to generate the payload without using test doubles, and are all hard coded to true for the time being and will be replaced with live code when we have the HMRC API in place.
- Added `ccms_case_reference` method on `LegalAidApplication`
- Changed the `config/ccms_keys.yml` to generate values from the `Respondent` record for the following attributes:
    - `WARNING_LETTER_SENT`
    - `POLICE_NOTIFIED`
    - `INJ_RESPONDENT_CAPACITY`
    - `BAIL_CONDITIONS_SET`
    - `INJ_REASON_NO_WARNING_LETTER`
    - `INJ_RECENT_INCIDENT_DETAIL`
- Set the `FAM_PROSP_GOOD` and related attributes to hard coded values (they will be dealt with in a subsequent ticket once the  legal framework data has been imported).
- Split the spec for individual XML attribute blocks into a separate file

## Checklist


app/models/bank_account.rb
```diff
@@ -19,4 +19,16 @@ def holders
...
+    true # TODO: CCMS placeholder
+  end
+
+  def receives_wages?
+    true # TODO: CCMS placeholder
+  end
+
+  def receives_benefits?
+    true # TODO: CCMS placeholder
+  end
```
reggieb: Is `BankAccount` really the right place for these fields. I would expect to see them in the `Applicant` model. `has_one :applicant, through: :legal_aid_application` could be added to `BankAccount` and the methods delegated to make then easily accessible from instances of this model.

If these methods are flags pertaining to the nature of the data stored within a particular set of bank transactions (for which `BankAccount` is the container) would `includes_` or `contains_` be a better prefix. For example `includes_tax_credits?`

app/services/ccms/case_add_requestor.rb
```diff
@@ -458,9 +458,11 @@ def provider
       @provider ||= @legal_aid_application.provider
     end
 
+    # :nocov:
     def lead_proceeding
       @lead_proceeding ||= @legal_aid_application.lead_proceeding
     end
+    # :nocov:
```
reggieb: If you add a `legal_aid_application` getter method, the methods `applicant`, `provider`, `lead_proceeding`, `proceedings`, `vehicle`, and `wage_slips` can be delegated to `legal_aid_application` and some complication will be removed and you won't have to `nocov` this method.

```ruby
attr_reader :legal_aid_application
delegate :applicant, :provider, :lead_proceeding, :proceedings, :vehicle, :wage_slips, to :legal_aidd_application
```

spec/factories/applicants.rb
```diff
@@ -23,5 +27,14 @@
...
     end
+
+    after(:create) do |applicant, evaluator|
+      if evaluator.num_bank_accounts > 0
+        provider = create :bank_provider, applicant: applicant
+        evaluator.num_bank_accounts.times do
+          create :bank_account, bank_provider: provider
+        end
+      end
+    end
```
reggieb: I'd suggest that `with_bank_accounts` would better match to the exiting traits. Also I don't think it is obvious what this code does or how to use it, so a comment would be useful. Something like:

 _"add `with_bank_accounts: 2` to add two bank accounts to the applicant"_.

Also as the `transient` is only used within this `after` block I would move the code so that the `transient` declaration was immediately before the `after` block.

app/services/ccms/case_add_requestor.rb
stephenrichards: Actually, it's no longer needed.  It's always true for MVP, because we only ever have one proceeding per applciation.

app/models/bank_account.rb
reggieb: OK - but "receiving benefit" has a specific meaning and is something that applies to a person and not a bank account. Surely `includes_benefits?` or `contains_benefits?` is clearer as to it's meaning.

spec/factories/applicants.rb
stephenrichards: `with_bank_accounts` is not really a trait, but agree it makes sense to keep the naming convention.

PR #670, opened 2019-07-19
AP-707 Fix issues raised by webhint

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-707)

- Fixed the CYA template - replaced `div` tags with `dt` and `dd` tags as needed, and as per the Design System. 
- Changed aria-describedby to return _nil_ instead of a zero-length string for not applicable instances.  Also changed `form_builder_spec.rb` file accordingly.
- Changed `statement_of_case_file_upload.js` file to include alt text with image
- Change date entry template to remove `default` (never used aside from for subtitles) and to align the `for` values with the `id`s of their corresponding inputs.
- Made the panel title an `h1` as per GDS design system.
- Made `h3` an `h2` in `providers\legal_aid_applications\index.html.erb` - as there was no `h2` above it.
- Made the `h3` an `h2` in the default page template, as this was throwing an accessibility error in some instances (`h2` needed before any `h3`).
- Removed spaces from href for telephone numbers.
- Removed "label: nil" from estimated vehicle value page
- Removed unneeded methods from `form_builder.rb`
- Date template change for delegated functions date
- Moved <noscript> with <style> tag to <head> - this now appears for all pages - might need to address this
- Style type tag removed
- Added alt text for plus sign images
- Removed the superfluous bold method - this is no longer needed now the main method has been fixed.  
- Removed duplicate H1s - some pages had  <%= govuk_fieldset_header page_heading %> instead of <%= govuk_fieldset_header page_title %>
- Changed rogue label to div on address selection screen, removed the hidden labels that came with this.
- Changed ID to match label on address selection screen
- Added unique IDs to list using each_with_index method
- Removed the DT and DD tags from the only_link_section.html.erb
- Removed the header from the vehicle date screen and moved it to be part of the date question template. 
- Added H1 content for the read only CYA screen - in liaison with Jim.
- Reformatted the structure of the restrictions on assets CYA entry.
- Corrected HTML on page not found page

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,11 +307,10 @@ workflows:
         - build_and_test
     - hold_uat:
         type: approval
-        requires:
-        - build_and_test
     - deploy_uat:
         requires:
         - hold_uat
+        - build_and_test
     - hold_staging:
         type: approval
         requires:
--- a/.hintrc
+++ b/.hintrc
@@ -6,7 +6,15 @@
   "parsers": [ "html", "css" ],
   "hintsTimeout": 120000,
   "hints": {
-    "axe": "error",
+    "axe": ["error", {
+      "rules": {
+        "definition-list": { "enabled": false },
+        "aria-allowed-attr": { "enabled": false },
+        "region": { "enabled": false },
+        "checkboxgroup": { "enabled": false },
+        "radiogroup": { "enabled": false }
+      }
+    }],
     "button-type": "error",
     "content-type": "error",
     "create-element-svg": "error",
@@ -14,15 +22,21 @@
     "disown-opener": "error",
     "highest-available-document-mode": "error",
     "html-checker": ["error", {
-        "ignore": [
-          "Start tag seen without seeing a doctype first. Expected “<!DOCTYPE html>”.",
-          "Attribute “src” not allowed on element “image” at this point.",
-          "The “banner” role is unnecessary for element “header”.",
-          "The “button” role is unnecessary for element “button”.",
-          "The “main” role is unnecessary for element “main”.",
-          "The “contentinfo” role is unnecessary for element “footer”.",
-          "Element “h1” not allowed as child of element “legend” in this context. (Suppressing further errors from this subtree.)"
-        ]
+      "ignore": [
+        "Start tag seen without seeing a doctype first. Expected “<!DOCTYPE html>”.",
+        "Element “input” is missing required attribute “role”.",
+        "Element “image” is missing required attribute “width”.",
+        "Element “image” is missing required attribute “height”.",
+        "Possible misuse of “aria-label”. (If you disagree with this warning, file an issue report or send e-mail to www-validator@w3.org.)",
+        "Attribute “aria-expanded” not allowed on element “input” at this point.",
+        "Attribute “src” not allowed on element “image” at this point.",
+        "The “banner” role is unnecessary for element “header”.",
+        "The “button” role is unnecessary for element “button”.",
+        "The “main” role is unnecessary for element “main”.",
+        "The “contentinfo” role is unnecessary for element “footer”.",
+        "Element “legend” not allowed as child of element “div” in this context. (Suppressing further errors from this subtree.)",
+        "Element “h1” not allowed as child of element “legend” in this context. (Suppressing further errors from this subtree.)"
+      ]
     }],
     "http-cache": "error",
     "http-compression": "error",
@@ -35,7 +49,7 @@
     "no-http-redirects": "error",
     "no-protocol-relative-urls": "error",
     "no-vulnerable-javascript-libraries": "error",
-    "sri": "error",
+    "sri": "off",
     "ssllabs": "error",
     "strict-transport-security": "error",
     "stylesheet-limits": "error",
--- a/app/javascript/src/statement_of_case_file_upload.js
+++ b/app/javascript/src/statement_of_case_file_upload.js
@@ -32,7 +32,7 @@ $(document).ready(() => {
             <td class="govuk-table__cell">${file.name}</td>
             <td class="govuk-table__cell"></td>
             <td class="govuk-table__cell no-wrap">
-              <img src="${image_loading_small}" class="small-loading-image"/>
+              <img src="${image_loading_small}" class="small-loading-image" alt="small loading image"/>
               ${copy_uploading}
             </td>
             <td class="govuk-table__cell"></td>
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -8,6 +8,12 @@
     <%= csp_meta_tag %>
 
     <%= stylesheet_link_tag 'application', media: 'all' %>
+    <noscript>
+      <style>
+        .proceeding-item { display: block; }
+        #search-field { display: none; }
+      </style>
+    </noscript>
     <%= javascript_pack_tag 'application' %>
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="theme-color" content="#0b0c0c">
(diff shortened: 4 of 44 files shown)


.circleci/config.yml
```diff
@@ -307,11 +307,11 @@ workflows:
         - build_and_test
     - hold_uat:
         type: approval
-        requires:
-        - build_and_test
     - deploy_uat:
         requires:
         - hold_uat
+        - build_and_test
+        - run_webhint
```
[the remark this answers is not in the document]
MalcolmVonMoJ: Essentially yes, but also, it isn't useful to deploy to UAT unless Webhint passes.  

I appreciate that people might not want to wait for Webhint to pass before authorising UAT (which is why I brought the hold UAT forward), but if Webhint doesn't pass, it shouldn't be sent to UAT.  

That was my thinking anyway.

app/views/layouts/application.html.erb
```diff
@@ -8,6 +8,12 @@
     <%= csp_meta_tag %>
 
     <%= stylesheet_link_tag 'application', media: 'all' %>
+    <noscript>
+      <style>
+        .proceeding-item { display: block; }
+        #search-field { display: none; }
+      </style>
+    </noscript>
```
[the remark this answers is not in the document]
MalcolmVonMoJ: It needs to be in the `<head>` tag, I couldn't see how to do that and make it only appear in the proceeding types page.  Adding in extra code for this seemed overkill - since we pull in all the other CSS on all pages, whether it is used or not.

app/javascript/src/statement_of_case_file_upload.js
```diff
@@ -32,7 +32,7 @@ $(document).ready(() => {
             <td class="govuk-table__cell">${file.name}</td>
             <td class="govuk-table__cell"></td>
             <td class="govuk-table__cell no-wrap">
-              <img src="${image_loading_small}" class="small-loading-image"/>
+              <img src="${image_loading_small}" class="small-loading-image" alt="small loading image"/>
```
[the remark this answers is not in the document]
MalcolmVonMoJ: Yes, probably should be.  
Another ticket (AP-605) will replace this image, so this alt text is just to pass webhints.

.hintrc
```diff
@@ -6,27 +6,41 @@
...
+        "The “main” role is unnecessary for element “main”.",
+        "The “contentinfo” role is unnecessary for element “footer”.",
+        "Element “legend” not allowed as child of element “div” in this context. (Suppressing further errors from this subtree.)",
+        "Element “h1” not allowed as child of element “legend” in this context. (Suppressing further errors from this subtree.)"
+      ]
     }],
     "http-cache": "error",
     "http-compression": "error",
-    "meta-charset-utf-8": "error",
+    "meta-charset-utf-8": "off",
```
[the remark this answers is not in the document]
MalcolmVonMoJ: Will change

.circleci/config.yml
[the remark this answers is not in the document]
Kalagan: _"it isn't useful to deploy to UAT unless Webhint passes"_
I don't agree with this. The webhint job could return errors that are invisible from the UAT tester.
At the moment, `deploy_uat` is after `build_and_test` only because  `deploy_uat` requires the docker image built by `build_and_test`

app/javascript/src/statement_of_case_file_upload.js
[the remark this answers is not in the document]
Kalagan: Ok, fair enough. There's a couple of other `alt` attributes in this PR

app/views/shared/check_answers/_only_link_section.html.erb
```diff
@@ -2,10 +2,11 @@
   <div class="govuk-grid-column-two-thirds">
     <h2 class="govuk-heading-m"><%= question %></h2>
   </div>
+  <%= link_to(url, class: 'govuk-link change-link') do %>
     <div class="govuk-grid-column-one-third govuk-summary-list--no-border align-text-right">
```
Kalagan: I'm surprised webhint is not complaining about a `div` being inside a

app/views/providers/address_selections/show.html.erb
```diff
@@ -24,11 +24,10 @@
                 class: 'govuk-body change-link change-postcode-link'
               ) %>
         </p>
-      <% end %>
-      <%= f.hidden_field :postcode %>
```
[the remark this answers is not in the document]
MalcolmVonMoJ: Okay, I have put this back.

app/views/shared/check_answers/_only_link_section.html.erb
MalcolmVonMoJ: I just looked this up:
> HTML 5 states that the <a> element "may be wrapped around entire paragraphs, lists, tables, and so forth, even entire sections, so long as there is no interactive content within (e.g. buttons or other links)".

Webhint is complaining about `legend`s without `fieldset`s, but I shall fix that separately - this ticket is already big enough.

app/views/shared/forms/vehicles/_estimated_value.html.erb
```diff
@@ -9,7 +9,6 @@
     <%= form.govuk_text_field(
           :estimated_value,
           hint: t('.use_car_valuation_sites'),
-          label: nil,
```
Kalagan: `label: nil` means that we don't want to have a label for that input field. In that case it's because the title acts as a label : 
image

If you remove `label: nil`, The form builder inserts a label and Rails makes up the text based on the attribute (`estimated_value` becomes "Estimated value") :
image

app/views/shared/check_answers/_only_link_section.html.erb
Kalagan: Thanks for looking up the info 👍

.circleci/config.yml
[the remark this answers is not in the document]
MalcolmVonMoJ: Okay, I'll change this.

app/views/shared/forms/vehicles/_estimated_value.html.erb
MalcolmVonMoJ: Oh damn!  

This was removed because it was failing Webhint aXe test:
> Form elements should have a visible label
> Form elements must have labels

Kalagan: I think this does it:
```ruby
  <%= govuk_form_group show_error_if: @form.errors.present? do %>
    <%= govuk_fieldset_header do %>
      <%= label_tag(:estimated_value, page_title) %>
    <% end %>

    <%= form.govuk_text_field(
          :estimated_value,
          hint: t('.use_car_valuation_sites'),
          label: nil,
          value: number_to_currency_or_original_string(@form.model.estimated_value),
          input_prefix: t('currency.gbp'),
          class: 'govuk-!-width-one-third'
        ) %>
  <% end %>
```

MalcolmVonMoJ: This removes the `<h1>` and so fails for a different reason.

.circleci/config.yml
```diff
@@ -307,11 +307,10 @@ workflows:
         - build_and_test
     - hold_uat:
         type: approval
-        requires:
-        - build_and_test
     - deploy_uat:
         requires:
         - hold_uat
+        - build_and_test
```
Kalagan: why moving this here?
`deploy_uat` already requires `hold_uat` which requires `build_and_test`

app/views/providers/income_summary/_add_other_income.html.erb
```diff
@@ -1,7 +1,7 @@
 <li>
   <h2 class="app-task-list__section">
     <span class="app-task-list__section-number">
-      <%= image_tag('plus_icon.svg') %>
+      <%= image_tag('plus_icon.svg', alt: 'Plus sign') %>
```
Kalagan: ideally, `'Plus sign'` should be from locale but maybe it doesn't matter that much...

PR #680, opened 2019-07-24
AP568 and AP535 Restrictions page for passported and non-passported flow

Obsiye:
Restrictions page for passported and non-passported

[AP568 passported restrictions story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-568)

[AP535 non-passported restrictions story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-535)

I removed the old way of doing restrictions to a yes-no with a description box style.

- Remove the restrictions and legal_aid_application_restrictions tables
- Add "has_restrictions" and "restrictions_details" columns to the legal_aid_applications table
- Create Restrictions form
- Create Restrictions controller for Provider and Citizen (passported and non-passported)
- Update views for restrictions page
- Update views for check_answer passported and non-passported
- Update routes for Provider and Citizen
- Update unit and feature tests
- Update factories for restrictions
- Remove inclusion of restriction in  app/services/save_applicant_means_answers.rb as it's now in the application

Add logic to remove the border/section break underneath the check answers item headers and add one below the details/text-area when required.

Add citizen or provider prefix for list_from_translation_path method so the translation passed into this method is dynamically either citizen or provider, which  removes hard coding citizen or provider in views using this list method

Add functionality to customise fieldset legend heading size because the page title is too long and thus a smaller font is required from --xl to --l (extra large to just large) to fit better on the page. 

Remove check_answer_no_link_list method and related view partial
Remove unnecessary file app/views/shared/forms/conditional_checkbox_with_numeric_input
Remove spec/models/admin_user_spec.rb which has nothing in it.

## Checklist


db/schema.rb
```diff
@@ -211,8 +211,6 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.string "relationship"
-    t.decimal "monthly_income"
-    t.boolean "has_income"
```
Kalagan: these 2 lines seem to have sneaked in from another PR

spec/forms/legal_aid_applications/percentage_home_form_spec.rb
```diff
@@ -16,7 +16,7 @@
 
   describe '#save' do
     it 'updates application.percentage_home' do
-      expect { subject.save }.to change { application.reload.percentage_home.to_s.to_d }.to(percentage_home.to_d)
+      expect { subject.save }.to change { application.reload.percentage_home }.to(percentage_home)
```
Kalagan: I think these `.to_s` and `.to_d` were introduced because this test was intermittently failing so I would leave them if I were you.

spec/forms/legal_aid_applications/restrictions_form_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+RSpec.describe LegalAidApplications::RestrictionsForm, type: :form do
+  let(:application) { create :legal_aid_application, :with_applicant }
+  let(:params) do
+    {
+      has_restrictions: has_restrictions,
+      restrictions_details: restrictions_details
+    }
+  end
+  let(:restrictions_details) { 'xxxxx' }
```
Kalagan: I suggest using `Faker::Lorem.paragraph` like you did on some other tests

spec/forms/legal_aid_applications/restrictions_form_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:has_restrictions) { 'true' }
+  let(:form_params) { params.merge(model: application) }
```
Kalagan: You could also pass a `mode`. To make sure it is used like in the controller.
`let(:mode) { %i[provider citizen].sample }`

spec/requests/citizens/check_answers_spec.rb
```diff
@@ -127,9 +130,10 @@
 
     context 'applicant does not have any capital restrictions' do
       let(:legal_aid_application) { create :legal_aid_application, :with_everything }
-      let!(:restriction) { nil }
+      let(:has_restrictions) { false }
+      let!(:restrictions_details) { '' }
```
Kalagan: because `restrictions_details` is an attribute of `legal_aid_application` and not another table, you don't need the `!`.

https://relishapp.com/rspec/rspec-core/v/2-5/docs/helper-methods/let-and-let
*let is lazy-evaluated: it is not evaluated until the first time the method it defines is invoked. You can use let! to force the method's invocation before each example.*

And this line is not needed either:
`let(:legal_aid_application) { create :legal_aid_application, :with_everything }`.
We can use the same `legal_aid_application` from line 7

spec/requests/citizens/restrictions_spec.rb
```diff
@@ -14,54 +14,64 @@
...
         legal_aid_application: {
-          restriction_ids: restriction_ids
+          has_restrictions: has_restrictions,
+          restrictions_details: restrictions_details
         }
       }
     end
-    subject { post citizens_restrictions_path, params: params }
+    let(:restrictions_details) { Faker::Lorem.paragraph }
+    let(:has_restrictions) { true }
```
Kalagan: `'true'` better matches what the real param to the controller is
```suggestion
    let(:has_restrictions) { 'true' }
```

spec/requests/providers/restrictions_spec.rb
```diff
@@ -14,38 +14,33 @@
...
-          restriction_ids: restriction_ids
+          has_restrictions: has_restrictions,
+          restrictions_details: restrictions_details
         }
       }
     end
-
-    subject { post providers_legal_aid_application_restrictions_path(legal_aid_application), params: params.merge(submit_button) }
+    let(:restrictions_details) { Faker::Lorem.paragraph }
+    let(:has_restrictions) { true }
```
Kalagan:
```suggestion
    let(:has_restrictions) { 'true' }
```

db/schema.rb
Kalagan: Actually, these 2 were duplicated by this PR: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/667/files#diff-1acd2e7e27a227829d5d14a91c863bb6R219
So it's good to remove them.

app/views/providers/about_the_financial_assessments/show.html.erb
```diff
@@ -2,7 +2,7 @@
   <p class="govuk-body"><%= t '.sign_in_para' %></p>
 
   <p class="govuk-body">
-    <%= list_from_translation_path 'providers.about_the_financial_assessments.show.list' %>
+    <%= list_from_translation_path 'about_the_financial_assessments.show.list' %>
```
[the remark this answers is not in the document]
Obsiye: Good spot thanks

app/views/shared/check_answers/_item.html.erb
```diff
@@ -1,6 +1,9 @@
-<% read_only = local_assigns[:read_only] ? read_only : false %>
+<%
+read_only = local_assigns[:read_only] ? read_only : false
+no_border = local_assigns[:no_border] ? '--no-border' : ''
+%>
 
-<div class="govuk-summary-list__row normal-word-break" id="app-check-your-answers__<%= name %>">
+<div class="govuk-summary-list__row<%= no_border %> normal-word-break govuk-!-margin-bottom-2" id="app-check-your-answers__<%= name %>">
```
Kalagan: This will add the `govuk-!-margin-bottom-2` class to all check answers items. Regardless of the `no_border` parameter.
Are you sure this is what we want?

Obsiye: It shouldn't be there thanks.

PR #681, opened 2019-07-25
AP-816 Fix providers-offices relation

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-816)

Changing the relations between providers and offices.

**Before:**
- A provider has one firm.
- A firm has several providers
- A firm has several offices

To get the list of offices a provider can access, we would do `provider.firm.offices` because we assumed that a provider could access all offices withing their firm.

**After**
We were wrong.
Providers withing the same firm don't always have access to all offices within their firm.
Therefore, this PR adds the `provider.offices` relation and prevent providers from selecting offices they don't have access to even if they belong to the same firm.
And changed `provider.office` to `provider.selected_office` to make it more obvious that this is the office they selected when they login but they can access all of `provider.offices` if they want.

## Checklist

### DIFF
--- a/app/models/office.rb
+++ b/app/models/office.rb
@@ -1,6 +1,5 @@
 class Office < ApplicationRecord
-  has_many :providers
-  has_many :legal_aid_applications
-
   belongs_to :firm
+  has_many :legal_aid_applications
+  has_and_belongs_to_many :providers
 end
--- a/spec/requests/providers/select_offices_spec.rb
+++ b/spec/requests/providers/select_offices_spec.rb
@@ -1,8 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe 'provider selects office', type: :request do
-  let(:provider) { create :provider, firm: firm }
   let(:firm) { create :firm }
+  let!(:office_1) { create :office, firm: firm }
+  let!(:office_2) { create :office, firm: firm }
+  let!(:office_3) { create :office, firm: firm }
+
+  let(:provider) { create :provider, firm: firm, offices: [office_1, office_2] }
 
   describe 'GET providers/select_office' do
     subject { get providers_select_office_path }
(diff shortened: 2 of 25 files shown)


app/models/office.rb
```diff
@@ -1,6 +1,5 @@
 class Office < ApplicationRecord
-  has_many :providers
-  has_many :legal_aid_applications
-
   belongs_to :firm
+  has_many :legal_aid_applications
+  has_and_belongs_to_many :providers
```
reggieb: It's a long time since I've used `has_and_belongs_to_many`. Using a through relationship is so much more flexible, that I've always found that HABTM eventually get changed to using a through model. However, if it works for now fine - it's easy enough to change it later.

reggieb: Thinking about this a little further, one thing a through model will allow is the addition of a `firm` filter to a provider's offices. The HABTM would allow a providers to be associated with offices that do not belong to the provider's firm. Obviously there are other ways of restricting that, but it would be nice if the relationship handled that.

I'm not sure exactly what the provider <-> firm relationship is. I think that we have discussed providers moving firms, so that historically applications can exist for a particular provider that span multiple firms. There may even be instances where providers work for more that one firm at the same time.

Having a through model that also had an association with firm would allow for queries like `provider.offices.where(:provider_offices: { firm: firm })` which would be more complicated with a HABTM relationship.

spec/requests/providers/select_offices_spec.rb
```diff
@@ -1,8 +1,12 @@
 require 'rails_helper'
 
 RSpec.describe 'provider selects office', type: :request do
-  let(:provider) { create :provider, firm: firm }
   let(:firm) { create :firm }
+  let!(:office_1) { create :office, code: 'office1', firm: firm }
+  let!(:office_2) { create :office, code: 'office2', firm: firm }
+  let!(:office_3) { create :office, code: 'office3', firm: firm }
```
reggieb: I think you don't need to specify the code in these `let` statements as you are not using the code string in any matching (good!) but rather using matches on `office_<n>.code`. So:
```ruby
  let!(:office_1) { create :office, firm: firm }
  let!(:office_2) { create :office, firm: firm }
  let!(:office_3) { create :office, firm: firm }
```

app/models/office.rb
stephenrichards: A firm has many offices.  A provider belongs to a firm and has a subset of that firm's offices, so I think HABTm works ok.

reggieb: Have discussed this with Julien and am happy for the association to be HABTM. We can change it when we need to.

spec/requests/providers/select_offices_spec.rb
Kalagan: 👍 
This was to make sure these code were unique.
But I'm removing them and replacing with with `Faker::Number.unique.number(10)` in the factory

PR #683, opened 2019-07-26
AP-759 & AP-779 Create Scope Limitation(s) for an application 

stephenrichards:
## Create scope limitation records for an application

[Link to story AP-759](https://dsdmoj.atlassian.net/browse/AP-759) 
[Link to story AP-779](https://dsdmoj.atlassian.net/browse/AP-779)
 
This PR creates the join records that link a substantive scope limitation to an application, and if the application used delegated functions, additionally adds the delegated functions scope limitation to the application as well.

Each proceeding type has a number of scope limitations, one of which will be marked as the default for substantive applications, and another will be marked as the default for delegated function applications.

When the proceeding type is added to the application, new code in this PR adds the  default substantive scope limitation to the application by writing a record in the application_scope_limitations table.

If delegated functions are subsequently added, the default delegated function scope limitation is added to the application as well.

Currently, only the default scope limitations are added, although there are plans to enable providers to choose other valid scope limitations in the future.

Similarly, currently there is only ever one proceeding type per application, although this will change in the future.  In this PR, we only look at that first proceeding type to work out what scope limitations to add into the application

The scope limitation text, level of service and default costs on the limitation page are now all taken from the proceeding type and scope limitation records instead of from hard coded constants.

### Done

- create application_scope_limitations join table
- add default substantive scope limitation when adding proceeding type
- add default delegated functions scope limitation when adding proceeding type
- add methods on `LegalAidApplication` to retrieve default scope limitations for the first proceeding type
- add methods on `LegalAidApplication`to add default scope limitations
- add unique indexes to `application_proceeding_types` and `application_scope_limitations` tables
- add `:with_scope_limitations` trait to `LegalAidApplication` factory
- add specs for the `:with_scope_limitations` trait

## Checklist


spec/factories/legal_aid_applications.rb
```diff
@@ -69,6 +71,43 @@
...
+    #
+    #    create :legal_aid_application, :with_scope_limitations, :with_delegated_functions
+    # will NOT work
+    #
+    # See factory specs at spec/factory_specs/legal_aid_application_factory_spec.rb
+    #
+    trait :with_scope_limitations do
+      after(:create) do |application, evaluator|
+        proceeding_types = if evaluator.proceeding_types.empty?
+                             create_list(:proceeding_type, 1)
```
reggieb: If this is changed to:
```ruby
create_list(:proceeding_type, 1, legal_aid_application: application)
```
line 101 becomes redundant, and line 103 can be simplified to:
```ruby
proceeding_type = proceeding_types.first
```

spec/factories/legal_aid_applications.rb
```diff
@@ -69,6 +71,43 @@
...
+                           end
+        application.proceeding_types = proceeding_types
+
+        pt = application.proceeding_types.first
+        create :scope_limitation,  :substantive_default, joined_proceeding_type: pt
+        create :scope_limitation,  :delegated_functions_default, joined_proceeding_type: pt
+        application.add_default_substantive_scope_limitation!
+        application.add_default_delegated_functions_scope_limitation! if application.used_delegated_functions?
+      end
+    end
```
reggieb: Rather than than having one trait that created both scope limitations, would it be worth having two traits. One for just creating the substantive scope limitation, and one the delegated function scope limitation. This may simplify the code a little and remove the dependency on the order of `:with_scope_limitations` and  `:with_delegated_functions`

spec/factories/scope_limitations.rb
```diff
@@ -1,11 +1,45 @@
...
+      delegated_functions { true }
+    end
+
+    transient do
+      joined_proceeding_type { create :proceeding_type }
+    end
+
+    trait :substantive_default do
+      substantive { true }
+      after(:create) do |sl, evaluator|
```
reggieb: Use `scope_limitation` here as is done in `delegated_functions_default` rather than using the unnecessary abbreviation `sl`

app/models/proceeding_type.rb
```diff
@@ -7,11 +7,15 @@ class ProceedingType < ApplicationRecord
...
   def self.populate
     ProceedingTypePopulator.call
   end
+
+  def default_substantive_scope_limitation
+    proceeding_type_scope_limitations.where(substantive_default: true).first.scope_limitation
+  end
+
+  def default_delegated_functions_scope_limitation
+    proceeding_type_scope_limitations.where(delegated_functions_default: true).first.scope_limitation
```
Kalagan: would this work?
```suggestion
    proceeding_type_scope_limitations.find_by(delegated_functions_default: true).scope_limitation
```

app/models/proceeding_type.rb
```diff
@@ -7,11 +7,15 @@ class ProceedingType < ApplicationRecord
...
-  scope :with_ccms_code_starting, ->(_text) do
-    where(arel_table[:ccms_code].matches_regexp('^DA\d+$'))
-  end
-
   def self.populate
     ProceedingTypePopulator.call
   end
+
+  def default_substantive_scope_limitation
+    proceeding_type_scope_limitations.where(substantive_default: true).first.scope_limitation
```
Kalagan: would this work?
```suggestion
    proceeding_type_scope_limitations.find_by(substantive_default: true).scope_limitation
```

app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -23,8 +23,11 @@ def create
     # PATCH /provider/applications/:legal_aid_application_id/proceedings_types/:id
     def update
       authorize legal_aid_application
-      legal_aid_application.proceeding_types.clear # Remove this when multiple proceeding types required!
-      legal_aid_application.proceeding_types << proceeding_type unless legal_aid_application.proceeding_types.include?(proceeding_type)
+      ActiveRecord::Base.transaction do
+        legal_aid_application.reset_proceeding_types! # This will probably change when multiple proceeding types implemented!
+        legal_aid_application.proceeding_types << proceeding_type unless legal_aid_application.proceeding_types.include?(proceeding_type)
+        legal_aid_application.add_default_scope_limitation!
```
Kalagan: shouldn't `.add_default_scope_limitation!` only be done on the next page (`used_delegated_functions_controller.rb`)?

app/models/legal_aid_application.rb
```diff
@@ -157,6 +159,57 @@ def ccms_case_reference
...
+      substantive: false
+    )
+  end
+
+  def default_substantive_scope_limitation_for_first_proceeding_type
+    proceeding_types.first.default_substantive_scope_limitation
+  end
+
+  def default_delegated_functions_scope_limitation_for_first_proceeding_type
+    proceeding_types.first.default_delegated_functions_scope_limitation
```
Kalagan: `proceeding_types.first` is used a lot in this PR.
I suggest creating and using this method:
```ruby
def proceeding_type
 proceeding_types.first
end
```

bin/rails
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
-  load File.expand_path('../spring', __FILE__)
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
```
Kalagan: why removing spring?

spec/factory_specs/legal_aid_application_factory_spec.rb
```diff
@@ -47,4 +47,73 @@
       end
     end
   end
+
+  describe ':with scope limitations' do
+    context "don't specify proceeding types" do
+      context "don't specify the number of proceeding types" do
+        it 'creates one proceeding type' do
+          application = create :application, :with_scope_limitations
+          expect(ProceedingType.count).to eq 1
```
reggieb: With FactoryBot, I don't think you can be absolutely sure that `ProceedingType.count` returns zero at the start of the spec. So this would be safer:
```ruby
let(:application) {  create :application, :with_scope_limitations }
it 'creates one proceeding type' do
  expect { application }.to change { ProceedingType.count }.by(1)
  expect(application.proceeding_types.count).to eq(1)
end
```

config/puma.rb
```diff
@@ -4,7 +4,7 @@
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
 # and maximum; this matches the default thread size of Active Record.
 #
-threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 1 }
```
Kalagan: why changing this?

db/migrate/20190724135655_add_enum_column_to_application_scope_limitations.rb
```diff
@@ -0,0 +1,5 @@
+class AddEnumColumnToApplicationScopeLimitations < ActiveRecord::Migration[5.2]
+  def change
+    add_column :application_scope_limitations, :substantive, :boolean, default: true
```
Kalagan: why creating a new migration rather than sticking this in the previous migration when this table gets created?

spec/models/legal_aid_application_spec.rb
```diff
@@ -361,4 +361,86 @@
...
+      let!(:scope_limitation_3) { create :scope_limitation }
+      let(:application) { create :legal_aid_application }
+
+      it 'adds multiple scope limitations' do
+        application.scope_limitations << scope_limitation_1
+        application.scope_limitations << scope_limitation_2
+        application.save!
+        expect(application.scope_limitations).to match_array [scope_limitation_1, scope_limitation_2]
+      end
+    end
```
reggieb: This is just testing that ActiveRecord's `has_many` works. I don't think it's needed.

db/schema.rb
```diff
@@ -357,15 +367,15 @@
...
-    t.decimal "trust_value", precision: 14, scale: 2
+    t.decimal "second_home_value"
+    t.decimal "second_home_mortgage"
+    t.decimal "second_home_percentage"
+    t.decimal "timeshare_property_value"
+    t.decimal "land_value"
+    t.decimal "valuable_items_value"
+    t.decimal "money_assets_value"
+    t.decimal "money_owed_value"
+    t.decimal "trust_value"
```
Kalagan: can we not modify the schema.rb for things unrelated to the PR?
when on the master branch use the `bin/rails db:schema:load` command to align your local DB with the `schema.rb` file in master

spec/models/legal_aid_application_spec.rb
```diff
@@ -361,4 +361,86 @@
...
+    let!(:proceeding_type) { create :proceeding_type }
+    let!(:sl_substantive_default) { create :scope_limitation, :substantive_default, joined_proceeding_type: proceeding_type, meaning: 'Default substantive SL' }
+    let!(:sl_delegated_default) { create :scope_limitation, :delegated_functions_default, joined_proceeding_type: proceeding_type, meaning: 'Default delegated functions SL' }
+    let!(:sl_non_default) { create :scope_limitation }
+
+    context 'substantive application' do
+      let(:application) { create :legal_aid_application, proceeding_types: [proceeding_type] }
+
+      describe '#add_default_scope_limitation!' do
+        it 'adds the default substatntive scope limitation to the application' do
```
reggieb: typo: `substatntive`

features/step_definitions/civil_journey_steps.rb
```diff
@@ -167,7 +170,8 @@
   )
   @legal_aid_application = create(
     :legal_aid_application,
-    :with_proceeding_types,
+    # :with_proceeding_types,
```
Kalagan: remove comment

spec/factories/legal_aid_applications.rb
```diff
@@ -69,6 +71,43 @@
...
+    # See factory specs at spec/factory_specs/legal_aid_application_factory_spec.rb
+    #
+    trait :with_scope_limitations do
+      after(:create) do |application, evaluator|
+        proceeding_types = if evaluator.proceeding_types.empty?
+                             create_list(:proceeding_type, 1)
+                           else
+                             evaluator.proceeding_types
+                           end
+        application.proceeding_types = proceeding_types
```
Kalagan: Your code is fine but this should also work:
```ruby
application.proceeding_types = evaluator.proceeding_types.presence || create_list(:proceeding_type, 1)
```

spec/factories/legal_aid_applications.rb
```diff
@@ -69,6 +71,43 @@
...
+    # when created, e.g.
+    #
+    #     create :legal_aid_application, :with_delegated_functions, :with_scope_limitations
+    # will work.
+    #
+    #    create :legal_aid_application, :with_scope_limitations, :with_delegated_functions
+    # will NOT work
+    #
+    # See factory specs at spec/factory_specs/legal_aid_application_factory_spec.rb
+    #
```
Kalagan: Nice comment 👍

app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -23,8 +23,11 @@ def create
     # PATCH /provider/applications/:legal_aid_application_id/proceedings_types/:id
     def update
       authorize legal_aid_application
-      legal_aid_application.proceeding_types.clear # Remove this when multiple proceeding types required!
-      legal_aid_application.proceeding_types << proceeding_type unless legal_aid_application.proceeding_types.include?(proceeding_type)
+      ActiveRecord::Base.transaction do
+        legal_aid_application.reset_proceeding_types! # This will probably change when multiple proceeding types implemented!
+        legal_aid_application.proceeding_types << proceeding_type unless legal_aid_application.proceeding_types.include?(proceeding_type)
```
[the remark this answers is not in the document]
stephenrichards: I left it in because we are seeing very weird behaviour whereby two or more PATCH requests are sometimes being sent (if the number of threads on puma is greater than 1 - but we've also seen it on UAT.) I presumed that you left it there on purpose.  I've created a ticket for us to investigate this issue a bit further - AP-826.  Any insight you can provide is very welcome!

app/models/legal_aid_application.rb
```diff
@@ -157,6 +159,57 @@ def ccms_case_reference
...
+  end
+
+  def substantive_scope_limitation
+    application_scope_limitations.find_by(substantive: true).scope_limitation
+  end
+
+  def delegated_functions_scope_limitation
+    application_scope_limitations.find_by(substantive: false)&.scope_limitation
+  end
```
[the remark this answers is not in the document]
stephenrichards: Well, there should ALWAYS be a `susbstantive_scope_limitation` for a legal aid application once a proceeding type has been added, but there isn't always a `delegated_functions_scope_limtation`.  However, I guess it's possible for this method to be called before the proceeding type and `substantive_scope_limitation` have been added, so I'll put in the safe navigation operator.

I don't quite see what you mean regarding scopes on the `ApplicationScopeLimitation` model.  We want to call `#substantive_scope_limitation` on `LegalAidApplication`, so do you  mean something_like this
```def substantive_scope_limitation
  application_scope_limitations.substantive
end```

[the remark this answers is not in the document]
Comment (stephenrichards) [reply app/models/legal_aid_application.rb]: Having read the comment below, I now see what you mean.

File: app/models/proceeding_type.rb [hunk 20]
```diff
@@ -7,11 +7,15 @@ class ProceedingType < ApplicationRecord
...
+  end
```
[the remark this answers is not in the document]
stephenrichards: 👍

app/controllers/providers/proceedings_types_controller.rb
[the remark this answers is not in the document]
stephenrichards: Julien cleverly found out what was causing this (JS nonsense as I suspected), so I'll take this out now.

app/models/legal_aid_application.rb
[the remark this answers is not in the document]
stephenrichards: 👍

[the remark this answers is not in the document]
stephenrichards: have created a method `#lead_proceeding_type`.  There is a concept of lead proceeding type which we need to identify for CCMS, so this fits quite well.

spec/factories/legal_aid_applications.rb
stephenrichards: 👍

app/models/proceeding_type.rb
stephenrichards: I've taken Rob's suggestion and delegated this to the join table

app/models/proceeding_type.rb
stephenrichards: have delegated this out to the proceeding_types_scope_limtations table..

But is `where...first` preferable to  `find_by`

app/controllers/providers/proceedings_types_controller.rb
stephenrichards: No.  We always add the default substatntive scope limitation when adding a proceeding type.  Then, if delegated functions are used, then we add and *additional* scope limitation for delegated functions.  I agree that it is not clear what is going on here, so I've changed it to explicitly add the *substantive* scope limitation here.

app/models/proceeding_type.rb
reggieb: I think nowadays both result in the same SQL call. Personally, I think `find_by` is more succinct; why use two methods when one will do.

app/models/legal_aid_application.rb
stephenrichards: I've created a method called lead_proceeding_type which is a concept we  need for injecting the application into CCMS, and will be something we need when we have multiple proceeding types per application

db/schema.rb
stephenrichards: 👍

spec/factories/legal_aid_applications.rb
stephenrichards: No need for the comment now - have refactored along the lines suggested by Rob which means that theere isn't the need to make sure the traits are specified in the right order.

spec/factories/legal_aid_applications.rb
stephenrichards: Thanks for this - I'd never seen the `#presence` method before - very useful.

app/models/legal_aid_application.rb
```diff
@@ -29,6 +29,8 @@ class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLen
   has_many :ccms_submissions, class_name: 'CCMS::Submission', dependent: :destroy
   has_one :most_recent_ccms_submission, -> { order(:created_at) }, class_name: 'CCMS::Submission'
   has_one :vehicle, dependent: :destroy
+  has_many :application_scope_limitations, dependent: :destroy
+  has_many :scope_limitations, through: :application_scope_limitations, dependent: :destroy
```
Kalagan: I'm a bit worried about the `dependent: :destroy` here.
I wonder if it means that when we delete the `legal_aid_application` it will also delete the `scope_limitations`.
And we don't want that do we?

app/controllers/providers/check_provider_answers_controller.rb
```diff
@@ -20,7 +20,7 @@ def continue
     private
 
     def set_variables
-      @proceeding_types = legal_aid_application.proceeding_types
+      @proceeding_type = legal_aid_application.lead_proceeding_type
```
reggieb: Should the instance variable also include the `lead_` prefix, as when this functionality extends it will have more significance in the view? So `@lead_proceeding_type`

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -97,20 +97,29 @@
...
       <dt class="govuk-summary-list__key govuk-!-width-one-half">
         <%= t '.section_proceeding.work' %>
       </dt>
       <dd class="govuk-summary-list__value">
-        <%= t '.section_proceeding.work_answer' %>
+        <% if @legal_aid_application.used_delegated_functions? %>
+          <%= @legal_aid_application.delegated_functions_scope_limitation.description %>
+        <% else %>
+          <%= @legal_aid_application.substantive_scope_limitation.description %>
+        <% end %>
```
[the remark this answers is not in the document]
Kalagan: Could we move this logic to a method on `legal_aid_application ` ?

e.g.  :
```ruby
def scope_limitation
  used_delegated_functions? ? delegated_functions_scope_limitation : substantive_scope_limitation
end
```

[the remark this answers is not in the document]
reggieb: This still feels like there should be default scope limitation on the application, that matches the default cost limitation. So if the `LegalAidApplication` model contained:
```ruby
def default_scope_limitation
  used_delegated_functions? ? delegated_functions_scope_limitation : substantive_scope_limitation
end
```
Then lines 114-118 reduce to:
```
<%= @legal_aid_application.default_scope_limitation.description %>
```

config/puma.rb
reggieb: I agree - if you've had to change this, that feels like a bug.

app/models/legal_aid_application.rb
```diff
@@ -157,6 +168,53 @@ def ccms_case_reference
...
+    used_delegated_functions? ? default_delegated_functions_cost_limitation : default_substantive_cost_limitation
+  end
+
+  def default_substantive_cost_limitation
+    lead_proceeding_type.default_cost_limitation_substantive
+  end
+
+  def default_delegated_functions_cost_limitation
+    lead_proceeding_type.default_cost_limitation_delegated_functions
+  end
```
reggieb: I think `default_substantive_scope_limitation`, `default_delegated_functions_scope_limitation`, `default_cost_limitation_substantive`, and `default_cost_limitation_delegated_functions` should all be delegated to `lead_proceeding_type`. In that way, it is made clear that construction of those values belongs to the `lead_proceeding_type`. It also makes it simpler to ensure that they are all handled the same way, rather than having two of them suffixed with `_for_first_proceeding_type` and two of them not.

app/views/providers/check_provider_answers/index.html.erb
```diff
@@ -76,7 +76,7 @@
         name: :proceeding_type,
         url: providers_legal_aid_application_proceedings_types_path(anchor: 'proceeding-search-input'.freeze),
         question: t('.section_proceeding.proceeding'),
-        answer: @proceeding_types.first.meaning,
+        answer: @proceeding_type.meaning,
```
reggieb: see comment in controller about using `@lead_proceeding_type` as the variable name

app/models/legal_aid_application.rb
stephenrichards: No we don't!

app/views/providers/check_provider_answers/index.html.erb
stephenrichards: Good spot

bin/rails
stephenrichards: It was when we were having issues with double clicks.  I'll put it back.

app/models/legal_aid_application.rb
stephenrichards: I've removed the methods `default_substantive_scope_limitation_for_first_proceeding_type` and `default_delegated_functions_scope_limitation_for_first_proceeding_type` and replaced the lines that referenced them with `lead_proceeding_type.default_substantive_scope_limitation` and `lead_proceeding_type.default_delegated_functions_scope_limitation`.  This makes it absolutely clear where it's coming from.  Just calling it `default ....` and delegating it to the lead_proceeding_type which cause more pain when we come to add in multiple proceeding types, I feel.

bin/rails
Kalagan: the double click issue is fixed now 🙂

PR #684, opened 2019-07-26
AP-747 Add end of application

reggieb:
[Jira AP-747](https://dsdmoj.atlassian.net/browse/AP-747)

Adds end of application controller and show view. Changes ending state to `application_submitted`

### DIFF
--- a/app/models/concerns/legal_aid_application_state_machine.rb
+++ b/app/models/concerns/legal_aid_application_state_machine.rb
@@ -17,7 +17,7 @@ module LegalAidApplicationStateMachine
       state :provider_checked_citizens_means_answers
       state :checking_merits_answers
       state :checked_merits_answers
-      state :merits_completed
+      state :assessment_submitted
 
       event :check_your_answers do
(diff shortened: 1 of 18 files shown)


app/models/concerns/legal_aid_application_state_machine.rb
```diff
@@ -17,7 +17,7 @@ module LegalAidApplicationStateMachine
       state :provider_checked_citizens_means_answers
       state :checking_merits_answers
       state :checked_merits_answers
-      state :merits_completed
+      state :assessment_submitted
```
Kalagan: Do we need a run an SQL query like this?
`update legal_aid_applications set state = 'assessment_submitted' where state = 'merits_completed'`

It might not be necessary if we are allowed to simply delete all applications in staging/prod

reggieb: `LegalAidApplication.where(state: 'merits_completed').update_all(state: 'assessment_submitted')`

Good point!

PR #690, opened 2019-07-30
AP-792 Migrating to live-1

Kalagan:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-792)

- changing region to `eu-west-2`
- use non-run-as-root images for PostgreSQL and ClamAV
- reference ClamAV from ECR repo in live-1
- remove `imagePullPolicy` so it always uses the latest image

## Checklist


.circleci/config.yml
```diff
@@ -21,7 +21,7 @@ references:
         GITHUB_TEAM_NAME_SLUG: laa-apply-for-legal-aid
     - image: postgres:10.5
     - image: redis:5.0
-    - image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/laa-apply-for-legal-aid/clamav:1.0.0
+    - image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-apply-for-legal-aid/clamav:1.0.2
```
Kalagan: Using version `1.0.2` of ClamAV (https://github.com/ministryofjustice/clamav/releases)
The only difference with `1.0.0` is that it does not run as root.
(`1.0.1` was a failed attempt)

Kalagan: `clamav`'s repo has also been migrated to `live-1` and we are now pulling the image from the ECR repo on `live-1`
I tried pulling a `live-0` image from `live-1` but it didn't work.

helm_deploy/apply-for-legal-aid/requirements.lock
```diff
@@ -1,7 +1,6 @@
...
 - name: postgresql
-  repository: https://kubernetes-charts.storage.googleapis.com/
-  condition: postgresql.enabled
-  version: 0.15.0
-digest: sha256:baab52c09113374d85d22a45f6606c4f86cb20cfbb77be7a8f2f065fc08c0ff9
-generated: 2018-08-21T13:06:22.312036363+01:00
+  repository: https://charts.bitnami.com
+  version: 6.2.0
+digest: sha256:c1d53c0e210361522bcd20e01c0c6a5ad680858145d328476b457daa3a1711dd
+generated: 2019-07-26T13:58:57.870599881+01:00
```
Kalagan: The original PostgreSQL image runs as root so we are now using bitnami's PostgreSQL
as recommended by cloud-platform.
It's also what is used by `check-financial-eligibility` at the moment

helm_deploy/apply-for-legal-aid/requirements.yaml
```diff
@@ -1,5 +1,5 @@
 dependencies:
 - name: postgresql
-  version: ^0.8.3
+  version: ^6.2.0
   condition: postgresql.enabled
-  repository: https://kubernetes-charts.storage.googleapis.com/
+  repository: https://charts.bitnami.com
```
Kalagan: The original PostgreSQL image runs as root so we are now using bitnami's PostgreSQL
as recommended by cloud-platform.
It's also what is used by `check-financial-eligibility` at the moment

helm_deploy/apply-for-legal-aid/templates/NOTES.txt
```diff
@@ -1,7 +1,7 @@
 1. Get the application URL by running these commands:
 {{- if .Values.ingress.enabled }}
 {{- range .Values.ingress.hosts }}
-  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ . }}{{ $.Values.ingress.path }}
+  https://{{ . }}{{ $.Values.ingress.path }}
```
Kalagan: This condition was not working properly so I'm removing it.
It now assumes that it's always https.

helm_deploy/apply-for-legal-aid/templates/secret.yaml
```diff
@@ -8,9 +8,9 @@ data:
   secretKey: {{ .Values.deploy.secretKey | b64enc | quote }}
   railsMasterKey: {{ .Values.deploy.rails_master_key | b64enc | quote }}
   {{if .Values.postgresql.enabled }}
-  postgresDatabase: {{ .Values.postgresql.postgresDatabase | b64enc | quote }}
-  postgresUser: {{ .Values.postgresql.postgresUser | b64enc | quote }}
-  postgresPassword: {{ .Values.postgresql.postgresPassword | b64enc | quote }}
+  postgresqlDatabase: {{ .Values.postgresql.postgresqlDatabase | b64enc | quote }}
+  postgresqlUsername: {{ .Values.postgresql.postgresqlUsername | b64enc | quote }}
+  postgresqlPassword: {{ .Values.postgresql.postgresqlPassword | b64enc | quote }}
```
Kalagan: bitnami requires these values to be named this way

helm_deploy/apply-for-legal-aid/templates/deployment_web.yaml
```diff
@@ -31,7 +31,6 @@ spec:
 {{ toYaml .Values.clamav.resources | indent 12 }}
         - name: web
           image: '{{ .Values.image.repository }}:{{ .Values.image.tag }}'
-          imagePullPolicy: IfNotPresent
```
Kalagan: `imagePullPolicy`'s default value is `IfNotPresent`.
The effect of omitting it is:
- if the image's tag is `latest`, force a pull
- otherwise, use an existing image if it's there.

See https://kubernetes.io/docs/concepts/containers/images/#updating-images :

Kalagan: Without this change, the container would not be an updated when re-deploying new code with the tag `latest`.
It's strange because it worked in `live-0` but it doesn't in `live-1`.
Maybe it's a different cluster setting or a different version

Kalagan: Actually, I'm removing this change. Somehow, I was under the impression that we used `latest` when deploying to staging/prod. But we are always using the git commit ID. So the previous code worked

PR #698, opened 2019-07-31
Connect 6dg ccms uat

scruti:
## What does this do
- Setup the connection between Apply Service and CCMS SOA service in UAT.
- Enable the logging of our connections to CCMS in debug mode.
- Add a button on the admin page to do a basic request test to CCMS.

## Points to highlight

The environment variables we setup to connect to CCMS SOA have been renamed, so probably you will need to update them in your `.env`, even for `Sit10` connections.
Check the `.env.sample` for the updated names.

We are still missing Staging and Production values, as the setup for those environments would need to be done in both 6dg and AWS Ops sides once UAT gets properly tested.

## Couldn't you just point Savon to the wsdl url in CCMS instead of using local WSDL files?

That is what we originally tried to do, but then realised that, even in the case of being able to get the files from the URL directly, the returned WSDL files contain the `ccmssoa.uat.legalservices..` addresses that are private to the `6dg` network.
And we are accessing to the `6dg` network through 
`[Apply CP] <---> [AWS API GW] <---> [6dg CCMS SOA UAT]`
So we need to point to the `AWS API GW` "proxy" in all our requests. 

This is the reason why we downloaded the wsdls from `6dg CCMS SOA UAT` and manually modified the contained urls one by one to point to the `AWS API GW`.
Seems it works!
 
### Missing bits:
- [x] Get a gateway to CCMS SOA URL from LAA AWS team.
- [x] Get the API Key for the Gateway and configure it in our Savon client to authenticate the requests.
- [x] Get the WSDL endpoints files from CCMS SOA UAT
- [x] Update the urls/files where is needed. 
- [x] Place a button in Admin page to trigger request of Application CCMS Reference Data: https://dsdmoj.atlassian.net/browse/AP-946 
- [x] Deploy into UAT
- [x] Test connection to CCMS in UAT using that button.


app/services/ccms/base_requestor.rb
```diff
@@ -30,13 +30,15 @@ def transaction_request_id
     # :nocov:
     def soap_client
       @soap_client ||= Savon.client(
+        headers: { 'x-api-key' => Rails.configuration.x.ccms_soa.aws_gateway_api_key },
```
scruti: ❓ is going to work on `Sit10` after merging this? (I guess will just ignore the empty header). Can anyone who currently can run the testing script correctly over `Sit10` check if it works on this branch?

spec/requests/admin/ccms_connectivity_tests_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    let(:requestor) do
+      instance_double(CCMS::ReferenceDataRequestor, call: response_stub)
+    end
+
+    subject { get admin_ccms_connectivity_test_path(legal_aid_applications.first.id) }
+
+    before do
+      allow(CCMS::ReferenceDataRequestor).to receive(:new).and_return(
+        instance_double(CCMS::ReferenceDataRequestor, call: response_stub)
```
JulienSansot: should you use `requestor` defined on line 20 here?

app/services/ccms/base_requestor.rb
stephenrichards: Mark is testing this.

app/services/ccms/base_requestor.rb
```diff
@@ -53,17 +55,19 @@ def soap_envelope(namespaces)
...
       end
     end
 
     def ns3_header_rq(xml)
       xml.__send__('ns3:TransactionRequestID', transaction_request_id)
       xml.__send__('ns3:Language', 'ENG')
-      xml.__send__('ns3:UserLoginID', ENV['USER_LOGIN'])
-      xml.__send__('ns3:UserRole', ENV['USER_ROLE'])
+      xml.__send__('ns3:UserLoginID', Rails.configuration.x.ccms_soa.user_login)
+      xml.__send__('ns3:UserRole', Rails.configuration.x.ccms_soa.user_role)
```
JulienSansot: if you add a `config` method in `BaseRequestor `
```ruby
def config
  Rails.configuration.x.ccms_soa
end
```

could you simplify all the uses of `Rails.configuration.x.ccms_soa....` to `config...` ?

spec/requests/admin/ccms_connectivity_tests_spec.rb
scruti: Wops, my bad. Originally defined the `requestor` object, but later decided to embed it in the `before` stub in order to make it more compact and easier to understand (as would not use it outside this single case). But seems I forgot to remove the `requestor` definition.

Going to remove the `let(:requestor)` instead.

app/services/ccms/base_requestor.rb
scruti: Nice suggestion, doing it!

PR #699, opened 2019-07-31
AP-807 SPIKE Generating PDF in the background

Kalagan:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-807)

**This PR doesn't need to be merged but is an example of how to create a PDF in the background (as opposed to with the browser)**

This PR explains how to generate a PDF in the background with an example called `pdf_spike` which is specific to a legal_aid_application.
When working on implementing the PDF, you can navigate to 
`/providers/applications/:uuid/pdf_spike` to see the pdf
add `?debug=1` at the end of the URL to see an HTML version. It makes the implementation easier.

You can use the `app/assets/stylesheets/pdf.scss` file to apply pdf specific CSS.
`@media print` also works (https://developer.mozilla.org/en-US/docs/Web/CSS/@media#Examples)

The PDF can be generated in the backround with this code:
```ruby

legal_aid_application = LegalAidApplication.last
pdf_html = ActionController::Base.new.render_to_string(template: 'providers/pdf_spike/index', formats: :pdf, layout: 'pdf', locals: { :@legal_aid_application => legal_aid_application } )
pdf = WickedPdf.new.pdf_from_string(pdf_html)
save_path = Rails.root.join('tmp','pdf.pdf')
File.open(save_path, 'wb') { |file| file << pdf }

```

The gem [`wicked_pdf`](https://github.com/mileszs/wicked_pdf) is used.
Its configuration is in the file `config/initializers/wicked_pdf.rb`.
Check the Github page of the gem for additional configuration if necessary

## Checklist

### DIFF
--- a/app/controllers/providers/pdf_spike_controller.rb
+++ b/app/controllers/providers/pdf_spike_controller.rb
@@ -1,0 +1,15 @@
+module Providers
+  class PdfSpikeController < ProviderBaseController
+    def index
+      authorize @legal_aid_application
+      respond_to do |format|
+        format.html
+        format.pdf do
+          render pdf: 'PDF Spike',
+                 layout: 'pdf',
+                 show_as_html: params.key?(:debug)
+        end
+      end
+    end
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -78,6 +78,9 @@ gem 'webpacker', '~> 4.x'
 
 gem 'wdm', '>= 0.1.0' if Gem.win_platform?
 
+gem 'wicked_pdf'
+gem 'wkhtmltopdf-binary'
+
 group :development, :test do
   gem 'awesome_print', '~> 1.8.0'
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -451,6 +451,9 @@ GEM
     websocket-driver (0.7.0)
       websocket-extensions (>= 0.1.0)
     websocket-extensions (0.1.3)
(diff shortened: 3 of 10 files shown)


app/controllers/providers/pdf_spike_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Providers
+  class PdfSpikeController < ProviderBaseController
+    layout 'pdf'
+
+    def index
+      authorize @legal_aid_application
+      render pdf: 'file_name',
+             show_as_html: params.key?('debug')
+    end
```
reggieb: I think it is worth trying to see if we can make this a little more restful. Specifically, so that:

- `providers/pdf_spikes` renders HTML
- `providers/pdf_spikes.pdf` renders PDF

I think something like this will do it:
```ruby
respond_to do |format|
  format.html # renders "normally" as html
  format.pdf do
    render pdf: 'file_name',
           show_as_html: params.key?(:debug),
           layout: :pdf
  end
end
```
Note that here the layout is specified in the `render` call rather than at the class level - and therefore is only used for the PDF view.

If that works, it will be much easier to add PDF actions to current views, and we won't need separate PDF controllers.

Kalagan: Good idea. I updated the code.
With this solution, we need two templates. `index.html.erb` and `index.pdf.erb` but they can both render the same partial

Kalagan:
I'm closing this spike PR as we now have the `merits_report` PR merged which can serve as an example when we want to generate other PDFs

PR #702, opened 2019-08-01
Ap835 hard coded ccms true results and 839 APP_AMEND_TYPE

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-835)
[Link to story](https://dsdmoj.atlassian.net/browse/AP-839)

renamed merits_assessment and means_assessment in the XmlExtractor to deal with a naming disparity with the xml file.

checked that hard coded true values were correct and added tests for these values.

reordered ccms case submission states to make it more readable

## Checklist

### DIFF
--- a/app/models/ccms/submission.rb
+++ b/app/models/ccms/submission.rb
@@ -20,14 +20,14 @@ module CCMS
         ObtainApplicantReferenceService.call(self)
       when 'applicant_submitted'
         CheckApplicantStatusService.call(self)
-      when 'applicant_ref_obtained'
-        AddCaseService.call(self, options)
-      when 'case_submitted'
-        CheckCaseStatusService.call(self)
       when 'case_created'
         ObtainDocumentIdService.call(self)
       when 'document_ids_obtained'
         UploadDocumentsService.call(self)
+      when 'applicant_ref_obtained'
+        AddCaseService.call(self, options)
+      when 'case_submitted'
+        CheckCaseStatusService.call(self)
       else
         raise CcmsError, "Unknown state: #{aasm_state}"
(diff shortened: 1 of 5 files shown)


app/models/ccms/submission.rb
```diff
@@ -20,14 +20,14 @@ def process!(options = {}) # rubocop:disable Metrics/MethodLength
         ObtainApplicantReferenceService.call(self)
       when 'applicant_submitted'
         CheckApplicantStatusService.call(self)
-      when 'applicant_ref_obtained'
```
mpw5: Any reason why these lines were moved? I don't think it's a problem, just curious.

app/services/ccms/attribute_value_generator.rb
```diff
@@ -86,6 +86,10 @@ def used_delegated_functions_on(_options)
       @legal_aid_application.used_delegated_functions_on.strftime('%d-%m-%Y')
     end
 
+    def app_ammendment_type(_options)
```
mpw5: Spelling: `app_ammendment_type` should be `app_amendment_type`. You'll have to make changes to `config/ccms/ccms_keys.yml` as a result.

app/models/ccms/submission.rb
willc-work: The change to the order more accurately covers the order in which the requests are made

app/services/ccms/attribute_value_generator.rb
willc-work: thanks, at least its not like ccms where we have to live with the spelling errors

PR #705, opened 2019-08-02
AP-748 provider view of full application

reggieb:
[Jira AP-748](https://dsdmoj.atlassian.net/browse/AP-748)

Add an application completed controller with a show action that shows the details of the the whole application.

As this new page reuses a lot of existing pages (in particular, check answers pages) some additional work was required. Also as this work involved bring together pages from different parts of the app, inconsistencies became obvious and where found were removed or reduced:

- Move check answers to `shared/check_answers` partials. There was a little inconsistency in that some check answers partials were in the `check_answers` folders, and others with prefixed with `check_answers`. Moved and renamed the later so that all are now in `check_answers` folder.
- Added `read_only` behaviour to many of the check answer partials so that they could be reused in the new display only view.
- Move `date_from` and `date_to` from controllers to a helper as not used in the controller, and this removed the duplication of code.
- Tidied up some of the `number_to_currency` calls as they were specifying the existing default currency.
- Removed an inconsistency where `user_type` was being used where `journey_type` is used elsewhere in the app. This removed the inconsistency, but also simplified the code as `journey_type` is available as a helper method and therefore doesn't need to be passed around in the way `user_type` was.
- There was some inconsistency as to whether `govuk-summary-list` elements were grouped within a dictionary list or within divs. As using divs seemed to be the most used in recently updated/added parts of the code, I changed all the `dl`/`dt`/`dd` instances to `div`.


spec/support/common_methods.rb
```diff
@@ -12,5 +12,5 @@ def save_xml_in_temp_file(_xml)
     fp.puts "<!-- generated by rspec #{Time.now} -->"
     fp.puts requestor.formatted_xml
   end
-  puts "XML saved in #{filename}"
+  Rails.logger.info "XML saved in #{filename}"
```
reggieb: I wasn't sure if this was necessary - or could be removed - so compromised by sending to logger.

app/views/shared/check_answers/_restrictions.html.erb
```diff
@@ -0,0 +1,16 @@
+<dl class="govuk-summary-list govuk-!-margin-bottom-9">
+  <%= check_answer_link(
+        name: :restrictions,
+        url: check_answer_url_for(journey_type, :restrictions, @legal_aid_application),
+        question: t(".#{journey_type}.question"),
+        answer: yes_no(@legal_aid_application.has_restrictions),
+        no_border: @legal_aid_application.has_restrictions?
+      ) %>
```
Kalagan: no `read_only` param here?

reggieb: Good spot - thank you

PR #707, opened 2019-08-02
Ap 837 omit from payload

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-837)

Check attributes are set to not generate a block if they are in the collection of attributes that can be omitted and created tests to check these attributes are set correctly
## Checklist


spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -130,14 +130,14 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
-        end
-      end
+      # context 'PROC_MATTER_TYPE_MEANING' do
+      #   it 'should be the meaning from the proceeding type record' do
+      #     block = XmlExtractor.call(xml, :global_merits, 'PROC_MATTER_TYPE_MEANING')
+      #     expect(block).to be_present
+      #     expect(block).to have_response_type 'text'
+      #     expect(block).to have_response_value legal_aid_application.proceeding_types.first.meaning
+      #   end
+      # end
```
Kalagan: can we delete this block if it's not needed?

PR #708, opened 2019-08-03
AP-858 Avoid webhint false positives

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-858)

Avoid these WebHint false positives:
 - Don't check pages from TrueLayer
 - Doctype not being present
 -  "'charset' meta element should be the first thing in '<head>'."

The last one happened because Capybara's `page.save_page` inserts `<base href="http://localhost:3004" />`  as the first element in the `<head>` tag.
So this PR implements it better to avoid the WebHint error.

For the Doctype error, the problem is that the page saved to file does not include the Doctype and I didn't manage to fix that so I added this error as an exception in Webhint's config.

## Checklist

### DIFF
--- a/.hintrc
+++ b/.hintrc
@@ -15,6 +15,7 @@
     "highest-available-document-mode": "error",
     "html-checker": ["error", {
         "ignore": [
+          "Start tag seen without seeing a doctype first. Expected “<!DOCTYPE html>”.",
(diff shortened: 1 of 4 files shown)


Kalagan:
Capybara's `page.html` which is delegated to Selenium and then to the chrome driver returns the html without the `DOCTYPE` unfortunately

PR #724, opened 2019-08-09
Ap 836 ccms various hard coded

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-836)

Added tests for the various CCMS attributes.

Currently have an issue matching the expected values for some tests. The expected value is a long string, the matcher fails for these cases. Currently these tests are commented out until i can work out the issue.

## Checklist

### DIFF
--- a/app/services/ccms/attribute_value_generator.rb
+++ b/app/services/ccms/attribute_value_generator.rb
@@ -63,9 +63,10 @@ module CCMS
       0 # TODO: CCMS placeholder
     end
 
-    def lead_proceeding_category(_options)
-      'MAT' # TODO: CCMS placeholder
-    end
+    #  commented out pending check with Stephen, AP-836
+    # def lead_proceeding_category(_options)
+    #   'MAT' # TODO: CCMS placeholder
+    # end
 
     def main_dwelling_third_party_name(_options)
       'Mrs Fabby Fabby' # TODO: CCMS placeholder
--- a/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
+++ b/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
@@ -687,6 +687,222 @@ module CCMS # rubocop:disable Metrics/ModuleLength
           end
         end
 
+        it 'should be type of boolean hard coded to false' do
+          attributes = [
+            [:global_means, 'GB_INFER_B_1WP1_1A'],
+            [:global_means, 'GB_INPUT_B_14WP2_7A'],
+            [:global_means, 'GB_INPUT_B_17WP2_7A'],
+            [:global_means, 'GB_INPUT_B_18WP2_2A'],
+            [:global_means, 'GB_INPUT_B_18WP2_4A'],
+            [:global_means, 'GB_INPUT_B_1WP1_2A'],
+            [:global_means, 'GB_INPUT_B_1WP3_175A'],
+            [:global_means, 'GB_INPUT_B_1WP4_1B'],
+            [:global_means, 'GB_INPUT_B_1WP4_2B'],
+            [:global_means, 'GB_INPUT_B_1WP4_3B'],
+            [:global_means, 'GB_INPUT_B_39WP3_70B'],
+            [:global_means, 'GB_INPUT_B_41WP3_40A'],
+            [:global_means, 'GB_INPUT_B_5WP1_22A'],
+            [:global_means, 'GB_INPUT_B_5WP1_3A'],
+            [:global_means, 'GB_INPUT_B_8WP2_1A'],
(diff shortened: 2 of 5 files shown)


spec/support/xml_attribute_block_matcher.rb
```diff
@@ -18,11 +18,11 @@
 
 RSpec::Matchers.define :have_response_value do |expected|
   match do |actual|
-    actual_response_type = actual.css('ResponseValue').text
-    actual_response_type == expected
+    actual_response_value = actual.css('ResponseValue').text
+    actual_response_value == expected
```
reggieb: My guess is the problem is trailing white space. What happens if you change this to:
```ruby
actual_response_value.strip == expected.strip
```

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -846,6 +849,189 @@ module CCMS # rubocop:disable Metrics/ModuleLength
             expect(block).to have_response_value 'false'
           end
         end
+
+        context 'attributes hard coded to false' do
+          it 'should be type of text hard coded to false' do
+            attributes = [
+              [:global_means, 'COST_LIMIT_CHANGED_FLAG']
+            ]
+            attributes.each do |entity_attribute_pair|
```
reggieb: Why are you using a block here when there is only one set of attributes being tested? Just do:
```ruby
block = XmlExtractor.call(xml, :global_means, 'COST_LIMIT_CHANGED_FLAG')
```

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -846,6 +849,189 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
+            [:global_merits, 'INJ_REASON_POLICE_NOT_NOTIFIED'],
+            [:proceeding_merits, 'INJ_RECENT_INCIDENT_DETAIL']
+          ]
+          attributes.each do |entity_attribute_pair|
+            entity, attribute = entity_attribute_pair
+            block = XmlExtractor.call(xml, entity, attribute)
+            expect(block).to be_present
+            expect(block).to have_response_type 'text'
+            # Issue with this as it is not matching the text correctly, seems to be due to length of the string
+            # expect(block).to have_response_value 'Apply Service application. See provider statement of case and report uploaded as evidence'
```
reggieb: One problem with defining spec expectations within a loop, is that in their standard form, it is very difficult to determine which iteration of the loop failed. Therefore, I would strongly recommend that you use a [custom message](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/customized-message) inside loops so that the current value of entity being tested can be identified.

Here, that would be something like:
```ruby
expect(block).to have_response_value 'Apply Service application. See provider statement of case and report uploaded as evidence', "expected #{entity} #{attribute} to have matching value"
```

app/services/ccms/attribute_value_generator.rb
```diff
@@ -62,9 +62,10 @@ def means_assessment_capital_contribution(_options)
       0 # TODO: CCMS placeholder
     end
 
-    def lead_proceeding_category(_options)
-      'MAT' # TODO: CCMS placeholder
-    end
+    #  commented out pending check with Stephen, AP-836
+    # def lead_proceeding_category(_options)
+    #   'MAT' # TODO: CCMS placeholder
+    # end
```
reggieb: Why does this need to be commented out? What is the purpose of removing this attribute?

spec/support/xml_attribute_block_matcher.rb
reggieb: Also can you post an example of the failure message.

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
willc-work: this should actually be a block of 2 tests, global_merits and gloabl_means, fixed

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
willc-work: ```have_response_value``` already does this

app/services/ccms/attribute_value_generator.rb
willc-work: This is not used, so it causes the coverage test to fail. There is an inconsistency with the tickets so I was leaving it here until I clarified that issue.

willc-work: This has now been removed in a different PR.

PR #734, opened 2019-08-12
AP-804 Print application

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-804)

Print application by running `window.print()`

## Checklist

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -74,4 +74,8 @@ module ApplicationHelper
   def yes_no(boolean)
     boolean ? t('generic.yes') : t('generic.no')
   end
+
+  def print_button(text)
+    content_tag :button, text, class: 'govuk-button no-print', type: 'button', onclick: 'window.print()'
+  end
 end
--- a/app/views/providers/submitted_applications/show.html.erb
+++ b/app/views/providers/submitted_applications/show.html.erb
@@ -11,9 +11,7 @@
     </dd>
   </dl>
 
-  <button class="govuk-button" type="button">
-    <%= t('.print_button') %>
-  </button>
+  <%= print_button t('.print_button') %>
 
   <section class="client_details">
     <h2 class="govuk-heading-l"><%= t('.client_details_heading') %></h2>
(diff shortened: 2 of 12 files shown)


app/helpers/application_helper.rb
```diff
@@ -21,7 +21,7 @@ def controller_t(lazy_t, *args)
 
   def menu_button
     button_tag(t('generic.menu'),
-               type: 'button', role: 'button', class: 'govuk-header__menu-button js-header-toggle',
+               type: 'button', role: 'button', class: 'govuk-header__menu-button js-header-toggle no-print',
```
reggieb: The only place I can find `menu_button` being used is within `app/views/layouts/_header.html.erb`:
```html
    <div class="govuk-header__content">
      <%= home_link %>
      <%= menu_button if user_header_link %>
      <nav>
        <%= user_header_link %>
      </nav>
    </div>
```
And to be honest, I'm surprised that whole block isn't `no-print` rather than just this element.

If the layout is problematic if there isn't a `.govuk-header__content`, perhaps an inner `.no-print` `div` around the links would do it.

app/views/shared/_provider_header_link.html.erb
```diff
@@ -1,4 +1,4 @@
-<ul id="navigation" class="govuk-header__navigation" aria-label="Top Level Navigation">
+<ul id="navigation" class="govuk-header__navigation no-print" aria-label="Top Level Navigation">
```
reggieb: Again, I think this element only appears within `app/views/layouts/_header.html.erb` and that this should be set as no print at that level.

app/views/providers/submitted_applications/show.html.erb
```diff
@@ -11,7 +11,7 @@
     </dd>
   </dl>
 
-  <button class="govuk-button">
+  <button class="govuk-button no-print" onclick="window.print()">
```
reggieb: I think this is crying out to be put into a helper method:
```ruby
def print_button
  content_tag :button, t('generic.print_button'), class: 'govuk-button no-print', onclick: 'window.print()'
end
```
That way the print functionality is separated from the markup.

Then here:
```
<%= print_button %>
```
Note the movement of the label translation to generic. Alternatively the label could be passed in:
```
<%= print_button t('.print_button') %>
```

app/views/providers/submitted_applications/show.html.erb
```diff
@@ -67,9 +67,33 @@
...
+          <%= t('.client_declaration.date') %>
+        </dt>
+      </div>
+    </dl>
+  </section>
+
+  <div class="no-print">
+    <button class="govuk-button" onclick="window.print()">
+      <%= t('.print_button') %>
+    </button>
```
reggieb: `<%= print_button %>` here too.

PR #736, opened 2019-08-13
AP-795 Generate unique Proceeding Case ID

stephenrichards:
## Autogenerate Proceeding Case ids for CCMS payload

[Link to story](https://dsdmoj.atlassian.net/browse/AP-795)

Each proceeding on each case needs a unique id to be assigned and included in the CCMS payload.  The id is in the form `P_nnnnnnnn` where `nnnnnnnn` is a number greater than 55,000,000.

This is achieved by adding a `proceeding_case_id` column to the `application_proceeding_types` table, and linking it to a Postgres sequence starting at 55,000,001.  There is no Rails Migration DSL keyword for adding a sequence, so it needs to be added with a hand-coded `ADD SEQUENCE` SQL statement, which unfortunately means it is not dumped to the `db/schema.rb` file.  It has therefore been added in the seeds, and will only create the sequence and alter the column to use the sequence if it hasn't already been done before.

## Checklist

### DIFF
--- a/app/services/ccms/case_add_requestor.rb
+++ b/app/services/ccms/case_add_requestor.rb
@@ -477,10 +477,6 @@ module CCMS
       @application_proceeding_types ||= @legal_aid_application.application_proceeding_types
     end
 
-    # def proceedings
-    #   @proceedings ||= @legal_aid_application.proceeding_types
-    # end
-
     def bank_accounts
       @bank_accounts ||= applicant.bank_accounts
     end
--- a/app/services/ccms/case_add_requestor.rb
+++ b/app/services/ccms/case_add_requestor.rb
@@ -147,24 +147,26 @@ module CCMS
     end
 
     def generate_proceedings(xml)
-      @legal_aid_application.proceeding_types.each { |p| generate_proceeding(xml, p) }
+      @legal_aid_application.application_proceeding_types.each { |apt| generate_proceeding(xml, apt) }
     end
 
-    def generate_proceeding(xml, proceeding) # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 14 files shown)


app/services/ccms/case_add_requestor.rb
```diff
@@ -458,10 +473,14 @@ def provider
...
 
-    def proceedings
-      @proceedings ||= @legal_aid_application.proceeding_types
+    def application_proceeding_types
+      @application_proceeding_types ||= @legal_aid_application.application_proceeding_types
     end
 
+    # def proceedings
+    #   @proceedings ||= @legal_aid_application.proceeding_types
+    # end
```
Kalagan: delete this comment?

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -11,19 +11,48 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
+      end
+
       context 'DELEGATED_FUNCTIONS_DATE blocks' do
         context 'delegated functions used' do
           before do
             legal_aid_application.update(used_delegated_functions_on: Date.today, used_delegated_functions: true)
           end
 
           it 'generates the delegated functions block in the means assessment section' do
+            # File.open(Rails.root.join('ccms_integration/generated/tmp.xml'), 'w') {|f| f.puts xml }
```
Kalagan: remove this comment?

stephenrichards: I'm going to leave it in because sometimes we want to generate a file of the payload for manual checking.  I'll put some comments around it.

PR #742, opened 2019-08-14
Ap 840 client assets

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-840)

Added attributes to check application and updates the XML accordingly.

## Checklist

### DIFF
--- a/app/services/ccms/attribute_value_generator.rb
+++ b/app/services/ccms/attribute_value_generator.rb
@@ -88,39 +88,35 @@ module CCMS
       @legal_aid_application.used_delegated_functions? ? 'SUBDP' : 'SUB'
     end
 
-    def client_owns_vehicle(_options)
-      @legal_aid_application.own_vehicle?
-    end
-
-    def client_owed_money(_options)
+    def applicant_owed_money?(_options)
       not_zero? @legal_aid_application.other_assets_declaration.money_owed_value
     end
 
-    def client_has_interest_in_a_trust(_options)
+    def applicant_has_interest_in_a_trust?(_options)
       not_zero? @legal_aid_application.other_assets_declaration.trust_value
     end
 
-    def client_has_valuable_posessions(_options)
+    def applicant_has_valuable_posessions?(_options)
       not_zero? @legal_aid_application.other_assets_declaration.valuable_items_value
     end
 
-    def client_owns_timeshare(_options)
+    def applicant_owns_timeshare?(_options)
       not_zero? @legal_aid_application.other_assets_declaration.timeshare_property_value
     end
 
-    def client_owns_land(_options)
+    def applicant_owns_land?(_options)
       not_zero? @legal_aid_application.other_assets_declaration.land_value
     end
 
-    def client_has_investments(_options)
+    def applicant_has_investments?(_options)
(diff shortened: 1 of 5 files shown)


app/services/ccms/attribute_value_generator.rb
```diff
@@ -92,6 +87,42 @@ def app_amendment_type(_options)
       @legal_aid_application.used_delegated_functions? ? 'SUBDP' : 'SUB'
     end
 
+    def client_owns_vehicle(_options)
+      @legal_aid_application.own_vehicle? ? 'true' : 'false'
+    end
```
reggieb: Do `true` and `false` have to be returned as strings? If they can be returned as Booleans, this whole section of code gets a lot easier. For example this just becomes:
```ruby
def client_owns_vehicle(_options)
  @legal_aid_application.own_vehicle?
end
```

app/services/ccms/attribute_value_generator.rb
```diff
@@ -92,6 +87,42 @@ def app_amendment_type(_options)
...
+
+    def client_owed_money(_options)
+      @legal_aid_application.other_assets_declaration.money_owed_value.nil? ? 'false' : 'true'
```
reggieb: There is a lot of repetition here of the pattern `x.nil? ? 'true' : 'false'`. Firstly, if we can return Booleans, could this simply be:
```ruby
@legal_aid_application.other_assets_declaration.money_owed_value.present?
```
If the pattern used is really the only way to do it, I'd suggest wrapping in a method:
```ruby
def not_nil?(value)
  value.nil? ? 'false' : 'true'
end

def client_owed_money(_options)
  not_nil? @legal_aid_application.other_assets_declaration.money_owed_value
end
```
etc.

config/ccms/ccms_keys.yml
```diff
@@ -409,7 +409,7 @@ global_means:
     :response_type: text
     :user_defined: false
   GB_INPUT_B_16WP2_7A:
-    :value: false
+    :value: '#client_has_interest_in_a_trust'
     :br100_meaning: 'Trust: The client has an interest in a trust?'
     :response_type: boolean
```
reggieb: Boolean!

app/services/ccms/attribute_value_generator.rb
reggieb: I've just had a bit of a play:
```ruby
x = Nokogiri::XML('<body><foo></foo><bar></bar></body>')

f = x.at_xpath '/body/foo'
f.content = 'true'
f.set_attribute 'bol', 'true'

b = x.at_xpath '/body/bar'
b.content = true
b.set_attribute 'bol', true

puts x.to_s
```
Outputs:
```xml
<?xml version="1.0"?>
<body>
  <foo bol="true">true</foo>
  <bar bol="true">true</bar>
</body>
```
As you can see setting `true` or `"true"` results in the same XML output.

reggieb: So remove all the text "true" and "false" and just use boolean operations `.nil?`, `.blank?` and `.present?` etc.

willc-work: yes all works, thank you much cleaner

PR #743, opened 2019-08-14
AP-831a Update Citizen pages so to prevent update after declaration.

reggieb:
## What

[Jira AP-831](https://dsdmoj.atlassian.net/browse/AP-831)

Updates most Citizen pages to use common parent `CitizenBaseController`.

Adds action to redirect to `error_path(:assessment_already_completed)` for pages on Citizen journey. Adds method to over-write this behaviour for final page, so page can be refreshed.

### DIFF
--- a/app/controllers/base_controller.rb
+++ b/app/controllers/base_controller.rb
@@ -1,6 +1,15 @@
 class BaseController < ApplicationController
   include Flowable
 
+  # This stops the browser caching these pages.
+  # This is done so that someone can't use the Back button to return to a users pages
+  # after they have logged out. There is a cost to page load speeds as a result
+  def set_cache_buster
+    response.headers['Cache-Control'] = 'no-cache, no-store, max-age=0, must-revalidate'
+    response.headers['Pragma'] = 'no-cache'
+    response.headers['Expires'] = 'Fri, 01 Jan 1990 00:00:00 GMT'
+  end
+
   def merge_with_model(model, options = {})
     yield.tap do |hash|
       hash.merge! options if options.present?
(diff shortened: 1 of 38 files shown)


app/controllers/base_controller.rb
```diff
@@ -1,6 +1,15 @@
 class BaseController < ApplicationController
   include Flowable
 
+  # This stops the browser caching these pages.
+  # This is done so that someone can't use the Back button to return to a users pages
+  # after they have logged out. There is a cost to page load speeds as a result
+  def set_cache_buster
```
Kalagan: This method still seems to be only called by the providers controllers but not the citizens controllers.
Is this an oversight or the intended use?

app/controllers/citizens/citizen_base_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Citizens
+  class CitizenBaseController < BaseController
+    include ApplicationFromSession
+    before_action :authenticate_applicant!
+    before_action :check_not_complete
+
```
Kalagan: Don't you need to add `before_action :set_cache_buster` here?

app/controllers/citizens/accounts_controller.rb
```diff
@@ -1,6 +1,5 @@
 module Citizens
-  class AccountsController < BaseController
-    include ApplicationFromSession
+  class AccountsController < CitizenBaseController
     before_action :authenticate_applicant!
```
Kalagan: we can remove all these now that `before_action :authenticate_applicant!` is in `CitizenBaseController`

Kalagan: this comment also applies to other controllers

app/controllers/citizens/citizen_base_controller.rb
reggieb: Thank you! I had a niggling feeling I'd missed something there ... and that was it!

PR #745, opened 2019-08-14
AP-812 caseworker merits report

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-812)

Generate a merits report at the end of the application.
The PDF is generated in the background, saved to S3 via ActiveStorage and attached to the application.
The PDF is based on an HTML page similar to a check answers page.
You can access it via `/providers/applications/:uuid/merits_report`

## Checklist


app/models/concerns/legal_aid_application_state_machine.rb
```diff
@@ -88,7 +88,8 @@ module LegalAidApplicationStateMachine
       end
 
       event :submit_assessment do
-        transitions from: :checked_merits_answers, to: :assessment_submitted
+        transitions from: :checked_merits_answers, to: :assessment_submitted,
```
stephenrichards: The only thing that worries me here is that the application is transitioned to `assessment_submitted`, before the pdfs are actually generated.  If there is a problem with the generation, or the background job system isn't available for example, we will have an application which is ostensibly complete, but no means/merits reports for them.

I t hink it would be better to have another state, `generating_merits_report` and transition to that when we fire off the background jobs, and have the `MeritsReportCreator`  transition to `assessment_submitted` once it's complete.

app/models/legal_aid_application.rb
```diff
@@ -148,8 +149,10 @@ def checking_answers?
       provider_checking_citizens_means_answers?
   end
 
+  # TODO: decide when is the submission_date set for passported and non-passported applications
+  # and save it to the DB
```
stephenrichards: In the CCMS payload, I've been assuming that the creation date of the CCMS::Submission record is the submission date.

Kalagan: where do you get the submission date?

PR #748, opened 2019-08-15
AP-829 - update GOVUK styles

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-829)

Upgrade GOVUK Design System to 3.0.

This contains breaking changes, so I have made a number of fixes in line with recommendations here: https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0

- fix stylesheet import paths, by adding `/govuk/` to all filepaths
- fix `data-module` elements, by adding `govuk-` as prefix to all
- add new fonts
- use preferred colours
- fix broken menu link, by adding `govuk-` to `js-header-toggle`
- fix appearance of start buttons, by adding a helper method to add the chevron which was previously present but now needs to be manually added
- css changes to fix button highlighting and broken focus state

## Checklist


Kalagan:
This adds 2 new fonts. Don't we need to remove some old ones?

mpw5:
> This adds 2 new fonts. Don't we need to remove some old ones?

Good point. I'll remove the old ones.

Kalagan:
Could you also modify https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/views/layouts/_govuk_base64_fonts.html.erb to use the new fonts?
This partial is used for PDFs and insert the fonts in the page

mpw5:
> Could you also modify https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/views/layouts/_govuk_base64_fonts.html.erb to use the new fonts?
> This partial is used for PDFs and insert the fonts in the page

Thanks, I didn't know about that file. I'll change it. 

The design system no longer provides `tabular` fonts. Any thoughts on what I should do with these entries? Remove them? Point them at the regular light/bold fonts?

image

Kalagan:
I think you can remove them

PR #766, opened 2019-08-20
AP-746 Generate means report PDF

Kalagan:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-746)

Generate a means report at the end of the application.
The PDF is generated in the background, saved to S3 via ActiveStorage and attached to the application.
The PDF is based on an HTML page similar to a check answers page.
You can access it via `/providers/applications/:uuid/means_report`

## Checklist


app/views/shared/_client_details_v2.html.erb
```diff
@@ -1,3 +1,6 @@
+
```
mpw5: Do we need the `_v2` in the name of this?

spec/requests/providers/merits_declarations_spec.rb
```diff
@@ -41,19 +41,19 @@
...
 
         it 'redirects to next page' do
           subject
           expect(response).to redirect_to(providers_legal_aid_application_end_of_application_path(legal_aid_application))
         end
 
-        it 'creates a merits report pdf' do
-          MeritsReportCreatorWorker.clear
-          expect(MeritsReportCreator).to receive(:call).with(legal_aid_application)
+        it 'creates a pdf reports' do
```
mpw5: Being really picky, this should probably be `creates pdf reports` or `creates a pdf report`.

spec/services/reports/reports_creator_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Reports::ReportsCreator do
+  let(:legal_aid_application) { create :legal_aid_application, :with_proceeding_types, :with_everything, state: :generating_reports }
+
+  subject { described_class.call(legal_aid_application) }
+
+  describe '.call' do
+    it 'attaches merits_report.pdf to the application' do
```
mpw5: As this tests for both reports, would `attaches merits_report.pdf and means_report.pdf to the application` be more accurate?

Kalagan: yes definitely 👍

app/views/shared/_client_details_v2.html.erb
Kalagan: it's because there's already a `_client_details.html.erb` but it shows different information :/

mpw5: 👌

PR #773, opened 2019-08-22
Simplify CircleCI process by using a separate job and workflow rather than parallelism. And some other improvements

Kalagan:
## What

- Remove use of parallelism but use different jobs instead. It simpler and make more sense
- move `webhint` to the end of the previous job running all the tests. Rather than a separate job. With 2 jobs, there was twice the overhead of setting up the image.
- modify workflows to move jobs that don't rely on each other into separate workflows. Split `build_and_deploy` into 2 workflows :
  - `test` runs all the tests. Has only one job. This should be the only mandatory job when a PR needs to be merged
  - `build_and_deploy` builds the docker image, push it to ECR, and can be deployed to UAT/Staging/Prod
- in addition to saving the webhint reports. We also save the html pages. That was a request from Malcolm
- use CircleCI versions of `postgres` and `redis`. They have some improvements that make them run faster than the regular images

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,11 +1,20 @@
 version: 2.0
 
 references:
-  deploy_container_config: &deploy_container_config
+  cloud_platform_container_config: &cloud_platform_container_config
     docker:
     - image: ${ECR_ENDPOINT}/cloud-platform/tools:circleci
       environment:
         GITHUB_TEAM_NAME_SLUG: laa-apply-for-legal-aid
+
+  build_container_config: &build_container_config
+    docker:
+    - image: circleci/ruby:2.6.3-node-browsers
+      environment:
+        GITHUB_TEAM_NAME_SLUG: laa-apply-for-legal-aid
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -93,103 +45,98 @@ references:
...
-        paths:
-          - webhint_inputs
 
   deploy_uat:
-    <<: *deploy_container_config
+    <<: *cloud_platform_container_config
     steps:
     - checkout
-    - setup_remote_docker:
-        docker_layer_caching: true
```
mpw5: What's the reason for removing `docker_layer_caching`?

Kalagan: Because it doesn't work at the moment. It's a CircleCI feature that needs to be activated on a project per-project basis and it costs £15.
https://mojdt.slack.com/archives/CCSD5F397/p1562319821001800

mpw5: Thanks 👍💰

PR #774, opened 2019-08-23
AP-904 - Create substantive cases with CCMS integration spec

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-904)

Amend `ccms_integration/ccms_spec.rb` so that it injects two cases into
CCMS when it is run - one delegated functions (as it does already) and
one substantive.

- Add an extra `context` to account for the substantive case. This
involves some refactoring of the existing spec.

- Amend `save_request` functionality so that it includes the case type in
the filename, to make it easier to see what each payload is.

- Use factories to create `proceeding_types` and `scope_limitations`,
instead of doubles.

- Changes to `case_add_requestor` and `ccms_keys` to ensure correct values
are being populated in the payload.

## Checklist

### DIFF
--- a/ccms_integration/ccms_spec.rb
+++ b/ccms_integration/ccms_spec.rb
@@ -2,99 +2,6 @@ require 'rails_helper'
 
 module CCMS
   RSpec.describe Submission do
-    let(:proceeding_type_1) do
-      double ProceedingType,
-             bail_conditions_set?: false,
-             case_id: 'P_11594790',
-             ccms_category_law: 'Family',
-             ccms_category_law_code: 'MAT',
-             ccms_code: 'DA004',
-             ccms_matter: 'Domestic Abuse',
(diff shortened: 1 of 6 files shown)


ccms_integration/ccms_spec.rb
```diff
@@ -2,99 +2,6 @@
...
-             delegated_functions_apply: false
-    end
-
-    let(:scope_limitation_3) do
-      double 'ScopeLimitation',
-             limitation: 'FM049',
-             wording: scope_limitation_wording_3,
-             delegated_functions_apply: false
-    end
-
```
stephenrichards: Great work in getting rid of these doubles.  we can also delete the `scope_limitation_wording_n` methods in this spec as well as they are only referenced from this deleted code.

mpw5: Done 👍

PR #776, opened 2019-08-23
AP869 global means and merits values generated from application

Obsiye:
There are a number CCMS attributes in the global means section that will now have their values generated from values in the application

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&modal=detail&selectedIssue=AP-869)

Update ccms keys attributes to call methods in the application and applicant instead of having hard coded values. The below attributes were updated:

1. DATE_ASSESSMENT_STARTED
2. DATE_OF_BIRTH
3. DEFAULT_COST_LIMITATION
4. FIRST_NAME
5. GB_DECL_B_38WP3_11A

- Hard code value for ccms_key attribute COUNTRY to 'GBR'

- Update tests

- Create a `passported?` method on the legal aid application model and update the app to use it where needed

The below two attributes I left alone as they were either already calling a method or has a hard coded value:

1. DELEGATED_FUNCTIONS_DATE
2. DEVOLVED_POWERS_CONTRACT_FLAG

## Checklist

### DIFF
--- a/spec/factories/legal_aid_applications.rb
+++ b/spec/factories/legal_aid_applications.rb
@@ -9,7 +9,7 @@ FactoryBot.define do
       transient do
         with_bank_accounts { 0 }
       end
-      applicant { create :applicant, with_bank_accounts: with_bank_accounts, date_of_birth: Date.new(2000, 1, 1) }
+      applicant { create :applicant, with_bank_accounts: with_bank_accounts }
     end
 
     trait :with_applicant_and_address do
--- a/spec/factories/legal_aid_applications.rb
+++ b/spec/factories/legal_aid_applications.rb
@@ -9,7 +9,7 @@ FactoryBot.define do
       transient do
         with_bank_accounts { 0 }
       end
-      applicant { create :applicant, with_bank_accounts: with_bank_accounts }
+      applicant { create :applicant, with_bank_accounts: with_bank_accounts, date_of_birth: Date.new(2000, 1, 1) }
     end
 
     trait :with_applicant_and_address do
--- a/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
+++ b/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
@@ -215,7 +215,7 @@ module CCMS # rubocop:disable Metrics/ModuleLength
               block = XmlExtractor.call(xml, entity, 'DATE_OF_BIRTH')
               expect(block).to be_present
               expect(block).to have_response_type 'date'
-              expect(block).to have_response_value '01-01-2000'
+              expect(block).to have_response_value(legal_aid_application.applicant.date_of_birth.strftime('%d-%m-%Y'))
             end
           end
         end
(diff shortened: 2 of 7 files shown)


app/models/legal_aid_application.rb
```diff
@@ -155,6 +155,10 @@ def submission_date
     (transaction_period_finish_at || Time.now.utc).to_date
   end
 
+  def passported?
+    benefit_check_result&.positive?
+  end
```
Kalagan: There's already `applicant_receives_benefit?` on line 104 so I suggest using that. Feel free to replace `.` by `&.`. My [PR 771](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/771/files#diff-f0eba5ecee5c3702fb86e094c801215eR112) also does it.

In my opinion, we don't need the term `passported?` but we can always use `applicant_receives_benefit?` instead.
It's clearer.

However, if you really want to use the term `passported?` you can add it as an alias of `applicant_receives_benefit?` :
```ruby
  def applicant_receives_benefit?
    benefit_check_result&.positive?
  end
  alias_method :passported?, :applicant_receives_benefit?
```

spec/factories/legal_aid_applications.rb
```diff
@@ -9,7 +9,7 @@
       transient do
         with_bank_accounts { 0 }
       end
-      applicant { create :applicant, with_bank_accounts: with_bank_accounts }
+      applicant { create :applicant, with_bank_accounts: with_bank_accounts, date_of_birth: Date.new(2000, 1, 1) }
```
Kalagan: I don't think you should fix this date_of_birth here.
See my next comment for an alternative

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -196,6 +196,66 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
       end
 
+      context 'applicant' do
+        context 'DATE_OF_BIRTH' do
+          it "inserts applicant's date of birth as a string" do
+            %i[global_means global_merits].each do |entity|
+              block = XmlExtractor.call(xml, entity, 'DATE_OF_BIRTH')
+              expect(block).to be_present
+              expect(block).to have_response_type 'date'
+              expect(block).to have_response_value '01-01-2000'
```
Kalagan:
```ruby
expect(block).to have_response_value(legal_aid_application.applicant.date_of_birth.strftime('%d-%m-%Y'))
```
would avoid setting `date_of_birth` to a specific date and work on any date

app/models/legal_aid_application.rb
stephenrichards: `applicant_receives_benefit?` is slightly misleading, in that an applicant may receive a benefit, but it is not a qualifying benefit, so he/she is not passported.  I would prefer `applicant_receives_qualifying_benefit?`, or using the term `passported?` which is jargon, but understood by everyone.

In AP-885, I have put a `#passported?` method on `LegalAidApplication` - so we probably need to standardise on one method and use it consistently.

PR #782, opened 2019-08-27
Ap 810 fix ccms document upload

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-810)

Documents (eg statement of case PDFs) aren't currently being uploaded to CCMS correctly. The current, incorrect, process is:

1. request_case_id
2. create_an_applicant
3. poll_applicant_creation
4. create_case
5. poll_case_creation_result
6. request_document_ids
7. upload_documents

This is incorrect as document ids need to be included in the
create_case payload.

This commit changes the order of CCMS requests to

1. request_case_id
2. create_an_applicant
3. poll_applicant_creation
4. request_document_ids
5. create_case
6. poll_case_creation_result
7. upload_documents

by amending the logic in `ccms_submission_state_machine`, and changing various CCMS `_service` classes appropriately.

It also amends `case_add_requestor.rb` so that `ccms_document_id`s are included in the `CaseDocs` section of the case creation payload.

## Checklist

### DIFF
--- a/app/models/ccms/submission.rb
+++ b/app/models/ccms/submission.rb
@@ -21,11 +21,11 @@ module CCMS
       when 'applicant_submitted'
         CheckApplicantStatusService.call(self)
       when 'case_created'
-        ObtainDocumentIdService.call(self)
-      when 'document_ids_obtained'
         UploadDocumentsService.call(self)
-      when 'applicant_ref_obtained'
+      when 'document_ids_obtained'
         AddCaseService.call(self, options)
+      when 'applicant_ref_obtained'
+        ObtainDocumentIdService.call(self)
       when 'case_submitted'
         CheckCaseStatusService.call(self)
       else
--- a/app/models/ccms/submission.rb
+++ b/app/models/ccms/submission.rb
@@ -20,14 +20,14 @@ module CCMS
         ObtainApplicantReferenceService.call(self)
       when 'applicant_submitted'
         CheckApplicantStatusService.call(self)
-      when 'case_created'
-        UploadDocumentsService.call(self)
-      when 'document_ids_obtained'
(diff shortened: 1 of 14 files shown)


app/models/ccms/submission.rb
```diff
@@ -21,11 +21,11 @@ def process!(options = {}) # rubocop:disable Metrics/MethodLength
...
         CheckApplicantStatusService.call(self)
       when 'case_created'
-        ObtainDocumentIdService.call(self)
-      when 'document_ids_obtained'
         UploadDocumentsService.call(self)
-      when 'applicant_ref_obtained'
+      when 'document_ids_obtained'
         AddCaseService.call(self, options)
+      when 'applicant_ref_obtained'
+        ObtainDocumentIdService.call(self)
```
stephenrichards: I would suggest these should be in the logical order of the events firing, in order to make it a bit more obvious to the reader, both here and in the list of states in `CCMSSubmissionStateMachine`, so something like this:
```case aasm_state
      when 'initialised'
        ObtainCaseReferenceService.call(self)
      when 'case_ref_obtained'
        ObtainApplicantReferenceService.call(self)
      when 'applicant_submitted'
        CheckApplicantStatusService.call(self)
      when 'applicant_ref_obtained'
        ObtainDocumentIdService.call(self)
      when 'document_ids_obtained'
        AddCaseService.call(self, options)
      when 'case_submitted'
        CheckCaseStatusService.call(self)
      when 'case_created'
        UploadDocumentsService.call(self)
      else
        raise CcmsError, "Unknown state: #{aasm_state}"
      end```

mpw5: Makes sense. I've pushed a change doing that.

PR #784, opened 2019-08-27
AP-877 generate CCM payload attributes from legal aid application

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-877)

- changed column name on `OtherAssetsDeclaration` from `money_assets_value` to `inherited_asssets_value` in order to better describe what the field holds
- Added memoize methods to return associated records `OtherAssetsDeclaration` and `SavingsAmount` in `AttributeValueGenerator`
- Rewrote existing method `AttributeValueGenerator#applicant_has_investments?` to return true if any investments are held (previously wrongly testing the old `money_assests_value`)
- Added methods and tests to generate the values for the attributes in the ticket

## Checklist


app/models/legal_aid_application.rb
```diff
@@ -212,6 +212,10 @@ def default_delegated_functions_cost_limitation
     lead_proceeding_type.default_cost_limitation_delegated_functions
   end
 
+  def passported?
+    benefit_check_result&.positive? || false
+  end
```
Kalagan: There's already `applicant_receives_benefit?` on line 112

bin/generate_webhint_reports.sh
```diff
@@ -14,7 +14,7 @@ no_files=$(ls -l ${HTMLS_DIR} | grep -v ^d | wc -l | sed -e 's/^[ ]*//')
 for filename in "$HTMLS_DIR"/*.html; do
     [ -e "$filename" ] || continue
     echo "${index}/${no_files} hint $filename"
-    result=$( yarn run hint "$filename" --tracking off )
+    result=$( yarn run hint "$filename" --tracking off --debug )
```
Kalagan: do we need this flag to be merged to master?

db/schema.rb
```diff
@@ -104,7 +104,6 @@
     t.datetime "updated_at", null: false
     t.boolean "substantive", default: true
     t.index ["legal_aid_application_id", "scope_limitation_id"], name: "scope_limitations_index", unique: true
-    t.index ["legal_aid_application_id"], name: "index_application_scope_limitations_on_legal_aid_application_id"
```
Kalagan: was this removed by mistake?

app/models/legal_aid_application.rb
stephenrichards: Have gone with this name.  It's not quite accurate, because the applicant could receive a benefit which still does not qualify them for a passported legal aid application, but will do for now.  Have repurposed `passported?` test to test the `applicant_receives_benefit?` method, and changed it to return false rather than nil if there is no benefit check record.  But prehaps raising an exeption is the right thing to do?

bin/generate_webhint_reports.sh
stephenrichards: ✅

db/schema.rb
stephenrichards: I guess.  Not sure how that happened

PR #785, opened 2019-08-28
AP-538 Add "Error: " prefix to page head title if error present

reggieb:
[Jira AP-538](https://dsdmoj.atlassian.net/browse/AP-538)

Modifies the way page the header is defined so that the header within `head` can be set differently to that used within the `body`. Then set the `head` title to be prefixed with "Error: " when an error occurs.

### DIFF
--- a/config/locales/en/errors.yml
+++ b/config/locales/en/errors.yml
@@ -1,6 +1,7 @@
 ---
 en:
   errors:
+    title_prefix: Error
     messages:
       already_confirmed: was already confirmed, please try signing in
       confirmation_period_expired: needs to be confirmed within %{period}, please request a new one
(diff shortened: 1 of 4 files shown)


config/locales/en/errors.yml
```diff
@@ -1,6 +1,7 @@
 ---
 en:
   errors:
+    title_prefix: Error
```
MalcolmVonMoJ: Should this have a colon in it `Error: ` so the title will be **Error: XYZ**

reggieb: If you look at the line where it's used in the `page_template_helper`:
```ruby
content_for(:head_title) { "#{prefix}: #{text}" }
```
The colon is added there. I think the colon is styling/formatting and therefore doesn't belong in the translation file.

PR #791, opened 2019-08-30
AP-876 CCMS Asset attributes

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-876)

Most of the attributes in the ticket had already been done.

Ones which were changes were:
- APPLICATION_CASE_REF
- GB_INPUT_B_2WP2_1A

Also changed the `AttributeValueGenerator` initializer to take CCMS::Submission record as a parameter rather than the `LegalAidApplication` - this is because we need the `case_ccms_reference` number from that record, and there could conceivably be more than one per application, so this ensures we get the right one.

## Checklist


spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -103,6 +105,36 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
+
+        context 'is a beneficiary' do
+          it 'inserts true into the attribute block' do
+            block = XmlExtractor.call(xml, :global_means, 'GB_INPUT_B_2WP2_1A')
+            expect(block).to have_boolean_response true
+          end
+        end
+      end
+
+      context 'GB_INPUT_B_3WP2_1A - applicant has financial interest in his main home' do
```
mpw5: Should there be tests here?

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb.orig
```diff
@@ -0,0 +1,1451 @@
+require 'rails_helper'
```
mpw5: I don't think this file is needed.

scruti:
I am confused about the couple "rebase" commits on this branch.
It looks like a "merge master" but with rebase title 😅, so not sure if I understood what is happening there.

Are we separately committing the rebase conflicts resolution? In that case shouldn't those conflict resolutions just amend the particular commit conflicting against master? 

I may be missing or may be a temporal thing pre-squashing & merging, just wanted to understand how those commits pop out :)

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
stephenrichards: Erm NO!

stephenrichards: Erm.  YES

stephenrichards:
> I am confused about the couple "rebase" commits on this branch.
> It looks like a "merge master" but with rebase title 😅, so not sure if I understood what is happening there.
> 
> Are we separately committing the rebase conflicts resolution? In that case shouldn't those conflict resolutions just amend the particular commit conflicting against master?
> 
> I may be missing or may be a temporal thing pre-squashing & merging, just wanted to understand how those commits did show up :)

I've no idea how this happened Marc.  I never merge master, but could it have been a rebase which required some changes?

PR #792, opened 2019-08-30
AP-886 printing output styling

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-886)

Added new HTML and CSS for printed output

## Checklist


app/views/providers/submitted_applications/show.html.erb
```diff
@@ -85,7 +85,7 @@
 
     <dl class="govuk-summary-list govuk-summary-list--no-border">
       <div class="govuk-summary-list__row">
-        <dt class="govuk-summary-list__key">
+        <dt class="govuk-summary-list__key" id="client-signature">
```
reggieb: If you are adding labels to elements purely so that you can style them a particular way, I think it is far better to use a class (so the styling can be reused), and make the name of that class meaningful to its purpose. So here something like a class `"print-fixed-bottom"`

app/views/shared/_applicant_declaration.html.erb
```diff
@@ -1,6 +1,6 @@
 <p class="govuk-body"><%= t('.submission_list_summary') %>:</p>
 
-<ul class="govuk-list govuk-list--bullet">
+<ul class="govuk-list govuk-list--bullet" id="declaration-agree-list">
```
reggieb: Again - rather than adding an `id` I think a descriptive class would be far better. Something like `"show-bullet-on-print"` or `"print-fix-show-bullet"`

app/views/shared/check_answers/_merits.html.erb
```diff
@@ -86,7 +86,7 @@
 
 <div class='govuk-body'><%= simple_format merits_assessment.details_of_proceedings_before_the_court %></div>
 
-<dl class="govuk-summary-list govuk-!-margin-bottom-2">
+<dl class="govuk-summary-list govuk-!-margin-bottom-2" id="statement-of-case-boolean">
```
reggieb: Again - use a descriptive class. Perhaps `"print-remove-border"` or `"remove-border-on-print"`

app/views/shared/check_answers/_merits.html.erb
```diff
@@ -97,11 +97,11 @@
       ) %>
 </dl>
 
-<div class='govuk-body'><%= simple_format statement_of_case.statement %></div>
+<div class='govuk-body' id="statement-of-case"><%= simple_format statement_of_case.statement %></div>
```
reggieb: See previous comment.

app/views/shared/check_answers/_merits.html.erb
```diff
@@ -97,11 +97,11 @@
...
 
 <%= render 'shared/check_answers/section_break' if respondent.bail_conditions_set %>
 
-<div class="govuk-summary-list govuk-!-margin-bottom-2">
+<dl class="govuk-summary-list govuk-!-margin-bottom-2">
```
reggieb: There has been some inconsistency as to whether `div` or `dl` tags should be used to define a "govuk-summary-list". To remove this inconsistency, there was a PR that went through and changed them to use `div` - so at least there was some consistency - especially as some of them contained inner elements that were constructed dynamically. This seems to reversing that change.

If we should be using `dl` tags for all `govuk-summary-list` lists, then that change needs to be done throughout the code base, and not just in isolated places like this.

MalcolmVonMoJ: In this instance, the opeining was a `div` and the closing was a `dl`.  

We should be using the `dl` `dt` `dd` structure as per the design system.  I went through changing them all to this structure as per the Webhint tickets.  I thought I was going mad seeing how many of them I seemed to have missed.  What was this PR?

reggieb: It was this PR: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/705

Looks like I jumped the wrong way. - should have made them all `dl`.

app/views/shared/check_answers/_merits.html.erb
MalcolmVonMoJ: I don't think this is needed at all anymore, so removed.

app/views/shared/check_answers/_merits.html.erb
MalcolmVonMoJ: Done

app/views/shared/_applicant_declaration.html.erb
MalcolmVonMoJ: Done

app/views/providers/submitted_applications/show.html.erb
MalcolmVonMoJ: Done

app/views/shared/check_answers/_merits.html.erb
MalcolmVonMoJ: I'll raise another ticket to get them moved back.

reggieb: Can you please reference #705 in the new ticket. I'll look out for it so I can correct my mistake.

MalcolmVonMoJ: Done (AP-924)  Thanks.

scruti:
Rebased against master (with no conflicts to solve) and Force Pushed in order to resolve Pipeline Deployment issues with this branch.

PR #806, opened 2019-09-03
AP-919: Normalize Cloud Platform resources between environments.

scruti:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-919)

This changes aim to:
1. Reduce over allocation of memory and specially CPU for our docker
   container in Cloud Platform.

2. Normalice our containers resource allocation between
   UAT/Staging/Production environments.

**RESOURCE ALLOCATION CHANGES**

The strategy we're taking for resource allocation is:
1. Be realistic with the Resource Requests (schedule allocation):
   Our containers CPU usage it stays well bellow 10m for all our
   containers, so we do not need to pre-allocate 200-250m per container.
   With memory the difference between usage and allocation is not so big,
   but still is good to adjust the schedule allocation of mem to numbers
   that reflect each container memory usage.

2. Be generous with the Resource Limits (running time limits):
   Containers going over their resource limits definitions will:
   - Be throtthled to the cpu limit if going over it.
   - Be marked as candidates to be killed if going over memory limit.
   Been told about Rails apps in Cloud Platform struggling when having
   low CPU limits on the web server as Rails require quite a bit of
   cpu on start time. So is good to leave the CPU limits quite high.
   Common Platform resources don't have problems with peaks/limits so
   far, but with over resource schedule allocation.

**NORMALIZATION BETWEEN ENVIRONMENTS**

Ideally we would have the same configuration in UAT, Staging and
Production namespaces. This simplify our configurations and also is
less prone to fall into situations where something was working in an
environment but didn't in other with a different configuration.

These changes extract individual resource allocation from
UAT/Staging/Product values specific files to set a single resource
definition for each service/container in the deployment template file.
So now "web", "worker", "clamav" and "metrics" resource allocation will
be the same for UAT, Staging and Production.

The specific namespace/environment files still contain the peculiarities
that change between each environment:

EG: - PostgreSQL definitions and resource allocation just for UAT.
       - 2 replicas set at Staging/Production but not at UAT.

## Checklist


scruti:
@mpw5 yes will definitely monitor the resources usage after releasing the changes.

Should be fine as the main reduction is on the amount of CPU that we schedule for each container & the namespace. The services should be able to go over the new scheduled CPU without issues as the CPU limits where they would be throttled are quite generous (even have been raised for some containers). 
But who knows, maybe some limit turns to not work as expected.

PR #814, opened 2019-09-04
Ap 875 hard coded CCMS attributes

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-875)

Added tests for existing attributes and created new attributes where required
## Checklist

### DIFF
--- a/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
+++ b/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
@@ -239,7 +239,11 @@ module CCMS # rubocop:disable Metrics/ModuleLength
             [:global_merits, 'MERITS_EVIDENCE_PROVIDED'],
             [:proceeding, 'SCOPE_LIMIT_IS_DEFAULT'],
             [:proceeding_merits, 'LEAD_PROCEEDING'],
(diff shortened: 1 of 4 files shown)


spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -133,7 +133,7 @@ module CCMS # rubocop:disable Metrics/ModuleLength
 
           it 'generates INJ_REASON_NO_WARNING_LETTER block with reason' do
```
mpw5: This could probably be tested in the test you added on line 824, then this one can be removed. Something like

```
        it 'should be hard coded with the correct notification' do
          attributes = [
            [:proceeding_merits, 'INJ_REASON_NO_WARNING_LETTER'],
            [:proceeding_merits, 'INJ_RECENT_INCIDENT_DETAIL'],
            [:global_merits, 'INJ_REASON_POLICE_NOT_NOTIFIED'],
            [:global_merits, 'INJ_REASON_NO_WARNING_LETTER']
          ]
          attributes.each do |entity_attribute_pair|
            entity, attribute = entity_attribute_pair
            block = XmlExtractor.call(xml, entity, attribute)
            expect(block).to have_text_response 'Apply Service application. See uploaded provider statement and report'
          end
        end
```

PR #815, opened 2019-09-04
AP880 CCMS payload general application attributes

Obsiye:
There are a number of CCMS attributes relating to the application that need to be generated

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-880)

Update values generation for the attributes:
- EMERGENCY_FC_CRITERIA
- URGENT_HEARING_DATE

Write and update tests for attributes:

- EMERGENCY_FC_CRITERIA
- URGENT_HEARING_DATE
- CONFIRMED_NOT_PASSPORTED

Set the attribute value for CONFIRMED_NOT_PASSPORTED to false and omit from payload and add to ccms_keys file

## Checklist

### DIFF
--- a/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
+++ b/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
@@ -105,6 +105,28 @@ module CCMS # rubocop:disable Metrics/ModuleLength
         end
       end
 
+      context 'EMERGENCY_FC_CRITERIA' do
+        it 'inserts emergency_fc criteria as hard coded string' do
+          block = XmlExtractor.call(xml, :global_merits, 'EMERGENCY_FC_CRITERIA')
+          expect(block).to have_text_response 'Apply Service application - see uploaded provider statement of case'
+        end
+      end
+
+      context 'URGENT_HEARING_DATE' do
+        before { allow(legal_aid_application).to receive(:calculation_date).and_return(Date.today) }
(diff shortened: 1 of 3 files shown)


spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -105,6 +105,28 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
+
+      context 'URGENT_HEARING_DATE' do
+        before { allow(legal_aid_application).to receive(:calculation_date).and_return(Date.today) }
+        it 'inserts ccms submission date as string' do
+          block = XmlExtractor.call(xml, :global_merits, 'URGENT_HEARING_DATE')
+          expect(block).to have_date_response legal_aid_application.calculation_date.strftime('%d-%m-%Y')
+        end
+      end
+
+      context 'PROCEEDING_NAME' do
```
mpw5: A test for this already exists on line 914 - I added it as part of another story but only merged it this morning so you might not have noticed it.

``` 
it 'populates PROCEEDING_NAME' do
  block = XmlExtractor.call(xml, :proceeding_merits, 'PROCEEDING_NAME')
  expect(block).to have_text_response application_proceeding_type.proceeding_type.ccms_code
end```

Delete one or the other, I don't mind which. If there are duplicate rows on the spreadsheet then you should probably update that too.

PR #818, opened 2019-09-06
AP 512 homepage table styling and functionality

reggieb:
[Jira AP-512](https://dsdmoj.atlassian.net/browse/AP-512)

The main purpose of this PR is to modify the appearance of the `providers/applications` page, and add sorting. However, to achieve that a number of prior steps were required.

**Note: the sorting applied in this PR does not work across pagination. This is a known issue and will be addressed in another ticket.**

# Tasks done include

- Fixed a bug where if you hit back and then save as draft after sending the client the email link, and the client then completed their journey, the provider couldn't then proceed beyond the `Providers::AboutTheFinancialAssessmentsController`
- Move the calculation of the emergency application deadline from the page first showing that deadline (where they choose where to submit a substantive application now or later), to the page where they enter the date when delegated functions were used (so that the deadline date can be shown if the provider saves as draft between those two points). Also extracted the code that makes the calculation so that it is easier to move to the best point in the flow.
- Added helper functions and refactored the CSS for tags so that they could be reused more easily.
- Added `summary_state` to the `LegalAidApplication` state machine, to manage the higher level state shown to the providers. At the moment, there are only two summary states, but this will probably develop with time.
- Refactored the sorting on the bank transactions pages so that it was easier to reuse the code on `providers/application` - mainly by wrapping up `th` and `td` tag generation into helper methods.
- Updated the `providers/applications` to match the prototype.

### DIFF
--- a/app/views/providers/legal_aid_applications/_legal_aid_applications.html.erb
+++ b/app/views/providers/legal_aid_applications/_legal_aid_applications.html.erb
@@ -5,9 +5,10 @@
       <thead class="govuk-table__head">
         <tr class="govuk-table__row">
           <%= sort_column_th type: :alphabetic, content: t('.applicant_name'),   currently_sorted: @initial_sort[:applicant_name] %>
-          <%= sort_column_th type: :date,       content: t('.created_at'),       currently_sorted: @initial_sort[:created_at] %>
+          <%= sort_column_th type: :date,       content: t('.created_at'),       currently_sorted: @initial_sort[:created_at],       combine_right: { at: 555, append: t('.col_and_ref') } %>
           <%= sort_column_th type: :alphabetic, content: t('.application_ref'),  currently_sorted: @initial_sort[:applicant_ref] %>
-          <%= sort_column_th type: :alphabetic, content: t('.certificate_type'), currently_sorted: @initial_sort[:certificate_type] %>
+          <th class="nullcell"></th>
+          <%= sort_column_th type: :alphabetic, content: t('.certificate_type'), currently_sorted: @initial_sort[:certificate_type], combine_right: { at: 555, append: t('.col_and_state') } %>
           <%= sort_column_th type: :alphabetic, content: t('.status'),           currently_sorted: @initial_sort[:status] %>
         </tr>
       </thead>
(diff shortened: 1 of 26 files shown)


config/locales/en/sorting.yml
```diff
@@ -0,0 +1,13 @@
+---
+en:
+  sorting:
+    oldest: oldest
+    newest: newest
+    a: A.
+    z: Zed
```
mpw5: Just curious - why is a `A.` but z not `Z.`?

MalcolmVonMoJ: Because a single A will be read out as "ah" (it is a word in itself), so we put the dot there to force it to be said in a correct way.

mpw5: Clever!

app/views/providers/legal_aid_applications/_legal_aid_applications.html.erb
```diff
@@ -1,30 +1,52 @@
...
-          <th class="govuk-table__header" scope="col"><%= t('.applicant_name') %></th>
-          <th class="govuk-table__header" scope="col"><%= t('.created_at') %></th>
-          <th class="govuk-table__header" scope="col"><%= t('.application_ref') %></th>
-          <th class="govuk-table__header" scope="col"><%= t('.status') %></th>
-          <th class="govuk-table__header" scope="col"></th>
+          <%= sort_column_th type: :alphabetic, content: t('.applicant_name'),   currently_sorted: @initial_sort[:applicant_name] %>
+          <%= sort_column_th type: :date,       content: t('.created_at'),       currently_sorted: @initial_sort[:created_at] %>
+          <%= sort_column_th type: :alphabetic, content: t('.application_ref'),  currently_sorted: @initial_sort[:applicant_ref] %>
+          <%= sort_column_th type: :alphabetic, content: t('.certificate_type'), currently_sorted: @initial_sort[:certificate_type] %>
+          <%= sort_column_th type: :alphabetic, content: t('.status'),           currently_sorted: @initial_sort[:status] %>
```
MalcolmVonMoJ: Can we have a _combine right_ value here, with associated stuff?
Probably combine `application_ref` and `certificate_type`.
At present the table goes off the right when narrow.

PR #820, opened 2019-09-06
Ap 883 last few attributes

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-883)

Add tests for various hard coded attributes
add method for the prospects of success value 

## Checklist

### DIFF
--- a/app/services/ccms/attribute_value_generator.rb
+++ b/app/services/ccms/attribute_value_generator.rb
@@ -84,10 +84,6 @@ module CCMS
       @legal_aid_application.used_delegated_functions? ? 'SUBDP' : 'SUB'
     end
 
-    def emergency_further_information(_options)
-      'Apply Service application - see uploaded provider statement of case'
-    end
-
     def provider_firm_id(_options)
       @legal_aid_application.provider.firm.id
     end
@@ -181,6 +177,24 @@ module CCMS
 
     def lead_proceeding_category_of_law_code(_options)
       lead_proceeding_type.ccms_category_law_code
+    end
+
+    def ccms_equivalent_prospects_of_success(_options) # rubocop:disable Metrics/MethodLength
+      check_value = @legal_aid_application.merits_assessment.success_prospect
+      case check_value
+      when 'likely'
+        'Good'
+      when 'marginal'
+        'Marginal'
+      when 'poor'
+        'Poor'
+      when 'borderline'
+        'Borderline'
+      when 'uncertain'
+        'Uncertain'
+      else
+        "Could not find match for #{check_value}"
+      end
     end
 
     private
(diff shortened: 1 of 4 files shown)


app/services/ccms/attribute_value_generator.rb
```diff
@@ -179,6 +179,24 @@ def lead_proceeding_category_of_law_code(_options)
...
+        'Marginal'
+      when 'poor'
+        'Poor'
+      when 'borderline'
+        'Borderline'
+      when 'uncertain'
+        'Uncertain'
+      else
+        "Could not find match for #{check_value}"
+      end
```
reggieb: A `case` statement is a little over-kill when each of the options simply return a small string. I'd suggest:

```ruby
PROSPECTS_OF_SUCCESS = {
  likely: 'Good',
  marginal: 'Marginal',
  poor: 'Poor',
  borderline: 'Borderline'
  uncertain: 'Uncertain'
}.freeze

attr_reader :legal_aid_application
delegate :merits_assessment, to: :legal_aid_application

def ccms_equivalent_prospects_of_success(_options)
  PROSPECTS_OF_SUCCESS[merits_assessment.success_prospect.to_sym] || "Could not find match for #{check_value}"
end
```

spec/services/ccms/case_add_requestor_spec.rb
```diff
@@ -210,6 +210,11 @@ module CCMS # rubocop:disable Metrics/ModuleLength
                capital_contribution: 0.0
       end
 
+      let(:merits_assessment) do
+        double 'MeritsAssesment',
+               success_prospect: 'Good'
```
reggieb: Why are you using a double, when there is a perfectly good real object to use? Also if this does represent a `MeritsAssessment` there is no "Good" `success_prospect`. I think the value should be "likely". So:
```ruby
let(:merits_assessment) { create :merits_assessment, success_prospect: 'likely' }
```

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -75,6 +77,32 @@ module CCMS # rubocop:disable Metrics/ModuleLength
         end
       end
 
+      context 'FAMILY_PROSPECTS_OF_SUCCESS' do
+        it 'returns the ccms equivalent prospect of success for likely' do
+          allow(legal_aid_application.merits_assessment).to receive(:success_prospect).and_return('likely')
```
reggieb: So a mock is being stubbed to return a value in a spec .... so what, if anything, is being tested here?

willc-work: I was aiming to test that the prospects of success assigned by the Apply service were translated to the correct output, e.g. when Apply assigns 'likely' this is translated to the CCMS equivalent and populates the block with 'Good'

stephenrichards: that seems reasonable to me, or am I missing something?

PR #840, opened 2019-09-13
Google tag manager bug fix

Obsiye:
Google tag manager bug fix

I altered the application layout view to render a partial for google tag manager instead of a normal layout view file.

## Checklist

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template">
   <head>
-    <%= render 'layouts/google_tag_manager' if google_tag_present? %>
+    <%= render partial: 'layouts/google_tag_manager' if google_tag_present? %>
     <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.erb
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template">
   <head>
-    <%= render 'layouts/google_tag_manager' if google_tag_present? %>
+    <%= render partial: 'layouts/google_tag_manager' if google_tag_present? %>
```
reggieb: I'm a little surprised that has any effect, but as it will do no harm ....

reggieb: Just to clarify, the default behaviour of `render` is to render a partial. As stated in the [docs](https://api.rubyonrails.org/classes/ActionController/Renderer.html#method-i-render): 

> If no options hash is passed or if :update is specified, the default is to render a partial and use the second parameter as the locals hash.

Obsiye: Interesting. Thanks.

PR #855, opened 2019-09-17
AP-939 Query CFE service - create capitals

stephenrichards:
## Query CFE service - create capitals

[Link to story](https://dsdmoj.atlassian.net/browse/AP-939)

In order for the Apply service to query the Check Financial Eligibility
service, it has to perform the  following 6 steps:

1. Create Assessment
2. Create Application
3. Create Capitals
4. Create Vehicles
5. Create Properties
6. Obtain Assessment Result

This PR covers step 3.

- Created `CreateCapitalsService`
- Modified `CFE::SubmissionManager` to call `CreateCapitalsService` as part of the workflow

## Checklist

### DIFF
--- a/app/services/cfe/base_service.rb
+++ b/app/services/cfe/base_service.rb
@@ -33,7 +33,7 @@ module CFE
 
     def query_cfe_service
       raw_response = post_request
-      parsed_response = JSON.parse(raw_response.body)
+      parse_json_response(raw_response.body)
       write_submission_history(raw_response)
       case raw_response.status
       when 200
@@ -43,6 +43,12 @@ module CFE
       else
         raise CFE::SubmissionError.new('Unsuccessful HTTP response code', raw_response.status)
       end
+    end
+
+    def parse_json_response(response_body)
+      JSON.parse(response_body)
+    rescue JSON::ParserError, TypeError
+      response_body || ''
     end
 
     def post_request
--- a/app/services/cfe/base_service.rb
+++ b/app/services/cfe/base_service.rb
@@ -37,7 +37,7 @@ module CFE
       write_submission_history(raw_response)
       case raw_response.status
       when 200
-        return parsed_response
+        return JSON.parse(raw_response.body)
       when 422
         raise CFE::SubmissionError.new('Unprocessable entity', 422)
       else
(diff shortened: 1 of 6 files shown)


app/models/cfe/submission.rb
```diff
@@ -0,0 +1,10 @@
+module CFE
+  class Submission < ApplicationRecord
+    include CFESubmissionStateMachine
+
+    self.table_name = :cfe_submissions
```
reggieb: I think if you add `CFE` as an acronym, I don't think you'll need to specify the table name. See:

- https://stackoverflow.com/questions/28990466/rails-autoloading-fully-capitalized-name-like-api
- https://borgs.cybrilla.com/tils/handling-acronyms-with-inflections/

app/services/cfe/base_service.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def call
+      @response = query_cfe_service
+      process_response
+    end
+
+    private
+
+    def cfe_url_host
+      'http://localhost:3001'
```
reggieb: This should be an environment variable

app/services/cfe/base_service.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+
+    def cfe_url
+      cfe_url_host + cfe_url_path
```
reggieb: Use `File.join` or `URI.join` here to avoid issues with leading and trailing slashes.

app/services/cfe/submission_error.rb
```diff
@@ -0,0 +1,10 @@
+module CFE
+  class SubmissionError < StandardError
+    attr_reader :http_status
+
```
reggieb: I'd suggest using [Nesty](https://github.com/skorks/nesty) here to preserve the original data error in the new error.

spec/factories/cfe_submissions.rb
```diff
@@ -0,0 +1,17 @@
+FactoryBot.define do
+  factory :cfe_submission, class: CFE::Submission do
+    legal_aid_application { create :legal_aid_application }
+
+    trait :initialised do
+      aasm_state { 'initialised' }
+    end
```
reggieb: I don't think we should be using traits for setting single attributes. They just complicate the factory and obscure the change. That is, in this example:
```ruby
let(:cfe_submission) { create :cfe_submission, aasm_state: :initialized }

let(:cfe_submission) { create :cfe_submission, :initialized }
```
I don't think the saving of 12 characters in the second example, is worth the loss of clarity in the first example.

PR #858, opened 2019-09-18
AP-888 show means result for passported

reggieb:
[AP-888](https://dsdmoj.atlassian.net/browse/AP-888)

This adds an "Capital Assessment Result" endpoint and a set of three partials - one each for the result: eligible, not eligible and contribution required.


reggieb:
I'm putting this on hold, because until we get results (or mocked results) back from the eligibility app, these changes break the app!

In addition, the prototype designs for the results pages are a best estimate. I will raise a ticket to get a better design.

PR #859, opened 2019-09-18
AP-940 Query Check Financial Elligibility service - create vehicles

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-940)

 
- Added `CreateVehiclesService` to send vehicles payload to CFE
- Moved common code for testing failed responses to shared example.
- Updated order of states `CFESubmissionStateMachine`
- Added the `CreateVehiclesService` step to `CFE::SubmissionManager`

## Checklist


spec/services/cfe/submission_manager_spec.rb
```diff
@@ -6,13 +6,14 @@ module CFE
     #   VCR.configure { |c| c.ignore_hosts 'localhost' }
     # end
 
-    let(:application) { create :legal_aid_application, :with_everything, :at_provider_submitted }
+    let(:application) { create :legal_aid_application, :with_everything, :with_vehicle, :at_provider_submitted, populate_vehicle: true }
```
reggieb: I think we need to stop relying on complicated FactoryBot trait constructions. They are very difficult to debug when problems occur.

This is much simpler and a lot less likely to come back and bite us later:
```ruby
let(:vehicle) { create :vehicle, :populated }
let(:application) { create :legal_aid_application, :with_everything, :at_provider_submitted, vehicle: vehicle
```

spec/services/cfe/create_applicant_service_spec.rb
```diff
@@ -1,143 +1,109 @@
...
-        expect(history.http_method).to eq 'POST'
-        expect(history.request_payload).to eq expected_payload
-        expect(history.http_response_status).to eq 200
-        expect(history.response_payload).to eq expected_response
-        expect(history.error_message).to be_nil
+      before do
+        allow(application).to receive(:applicant_receives_benefit?).and_return(true)
+        allow(application).to receive(:calculation_date).and_return(Date.today)
+        expect(Faraday).to receive(:new).with(url: 'http://localhost:3001').and_return(faraday_connection)
+        expect(faraday_connection).to receive(:post).and_yield(connection_param).and_return(faraday_response)
```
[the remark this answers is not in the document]
reggieb: Once we have the eligibility app up in staging we can point the apply test environment at that, and use VCR.

app/services/cfe/base_service.rb
```diff
@@ -20,7 +20,7 @@ def conn
     end
 
     def cfe_url_host
-      ENV['CHECK_FINANCIAL_ELIGIBILITY_URL']
+      Rails.configuration.x.check_finanical_eligibility_host
```
reggieb: :+1:

spec/services/cfe/create_applicant_service_spec.rb
```diff
@@ -1,30 +1,63 @@
...
+    let(:expected_payload) { expected_payload_hash.to_json }
+    let(:expected_response) { expected_response_hash.to_json }
+    let(:cfe_host) { Rails.configuration.x.check_finanical_eligibility_host }
+    let(:cfe_url) { "#{cfe_host}/assessments/#{submission.assessment_id}/applicant" }
+
+    context 'response received from CFE' do
+      before do
+        allow(application).to receive(:applicant_receives_benefit?).and_return(true)
+        allow(application).to receive(:calculation_date).and_return(Date.today)
+        allow_any_instance_of(described_class).to receive(:request_body).and_return(expected_payload)
```
reggieb: I strongly believe that we should only be mocking when there is no alternative.

- The only relevance of `applicant_receives_benefit?` in this spec is that the entry for the application matches that in the payload. So the default `false` is just as valid and the mocked true - so no need to mock.
- If `transaction_period_finish_on` is set on the application, that will be used as the `calculation_date` so that can be done in the factory call, without having to mock.
- `request_body` is the fundamental property of this service. As `expected_payload` gets passed to both the instance of the service and the webmock declaration ALL THAT IS BEING TESTED IS THAT MOCKING IS WORKING. Nowhere is the construction of `request_body` tested, and that is the most important part of the spec!!!!!

spec/services/cfe/create_applicant_service_spec.rb
```diff
@@ -1,30 +1,79 @@
...
+    describe '.call' do
+      context 'response received from CFE' do
+        before do
+          allow(application).to receive(:applicant_receives_benefit?).and_return(true)
+          allow(application).to receive(:calculation_date).and_return(Date.today)
+        end
+
+        describe 'successful post' do
+          before do
+            stub_request(:post, service.cfe_url)
```
reggieb: I'd recommend turning VCR off around the stub request call - then any problems raise WebMock errors rather than VCR errors. Something like this:
```ruby
    around do |example|
      VCR.turn_off!
      stub_request(:post, subject.cfe_url)
        .with(body: subject.request_body)
        .to_return(body: api_response)

      example.run
      VCR.turn_on!
    end
```
That is, if you change `subject.cfe_url` to `http://foo.com` and run the test, if VCR is on you'll just get a message along the lines that there isn't a VCR cassette for that call. If VCR is off, WebMock will return a much more meaningful error message, including an example of a `stub_request` configuration that would pass - which you can then compare with the one you have.

spec/services/cfe/create_applicant_service_spec.rb
```diff
@@ -1,30 +1,79 @@
...
           has_partner_opponent: false,
-          receives_qualifying_benefit: true
+          receives_qualifying_benefit: application.applicant_receives_benefit?
         }
-      }.to_json
+      }
     end
-    let(:expected_response) do
+
+    def expected_response_hash # rubocop:disable Metrics/MethodLength
```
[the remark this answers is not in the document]
stephenrichards: Good point.  I've changed the name to `dummy_response` to underline the fact that they don't represent a full response from the API

spec/services/cfe/create_applicant_service_spec.rb
stephenrichards: Turns out that there is a peculiarity with with `around` as used in your example above - it doesn't instantiate the variables defined with a `let!` first.  Solved by putting the `around` at the group level, but worth bearing in mind if you get similar problems in the future.

PR #860, opened 2019-09-19
AP-942 Get results from Check Financial Eligibility Service

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-942)

- Created `CFE::ObtainAssessmentResultsService` class to get and store results
- Updated `CFE::Submission manager` to call the above class

## Checklist


app/services/cfe/obtain_assessment_result_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      "/assessments/#{@submission.assessment_id}"
+    end
+
+    def query_cfe_service
+      make_get_request
+    end
+
+    def make_get_request
+      conn.get cfe_url_path
+    end
```
reggieb: Would it be simpler to just:
```ruby
def query_cfe_service
  conn.get cfe_url_path
end
```

app/services/cfe/obtain_assessment_result_service.rb
```diff
@@ -0,0 +1,35 @@
...
+      @submission.cfe_result = @response.body
+      if @response.status == 200
+        write_submission_history(@response, 'GET')
+        @submission.results_obtained!
+      else
+        @submission.fail!
+        raise_exception_error(
+          CFE::SubmissionError.new('Unsuccessful HTTP response code', @response.status),
+          'GET'
+        )
```
reggieb: That's results in a weird construction in `raise_exception_error` where another `CFE::SubmissionError` is created. Is there really not a simpler way to do this?

spec/services/cfe/obtain_assessment_result_service_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    context 'success' do
+      before do
+        stub_request(:get, service.cfe_url)
+          .to_return(body: expected_response)
+      end
+
+      it 'calls the expected URL' do
+        ObtainAssessmentResultService.call(submission)
+      end
```
reggieb: I don't think this is actually testing any expectation.

app/models/concerns/cfe_submission_state_machine.rb
```diff
@@ -35,7 +35,8 @@ module CFESubmissionStateMachine
       end
 
       event :results_obtained do
-        transitions from: :vehicles_created, to: :results_obtained
+        transitions from: :vehicles_created, to: :results_obtained # remove one CreatePropertiesService has been written
```
JulienSansot: Should it be "once" instead of "one"?
Also, I would add "TODO:" at the beginning of the comment

app/services/cfe/obtain_assessment_result_service.rb
stephenrichards: Yep.  I guess there was a need to that once which disappeared with the refactoring.

app/services/cfe/obtain_assessment_result_service.rb
stephenrichards: Open to suggestions....

spec/services/cfe/obtain_assessment_result_service_spec.rb
stephenrichards: Well, I think my thinking was that it was testing that the expected URL was called, bu that's tested elsewhere, so I've taken it out.

app/services/cfe/obtain_assessment_result_service.rb
reggieb: I think you need two methods - one takes an error, and the other builds an error from the input:
```ruby
def raise_exeception_error(message:, backtrace: nil, http_method: 'POST', http_status: nil)
  @submission.submission_histories.create!(
    url: cfe_url,
    http_method: http_method,
    request_payload: request_body,
    http_response_status: http_status
    error_message: messge,
    error_backtrace: backtrace
  )
  raise CFE::SubmissionError.new(message, http_status)
end

def catch_and_record_exception(error, http_method = 'POST')
  raise_exeception_error(
    message: formatted_error_message(error),
    backtrace: error.backtrace&.join("\n"),
    http_method: http_method,
    http_status: error.respond_to?(:http_status) ? err.http_status : nil
  )
end
```
Then in instances where an error has been caught:
```ruby
rescue SomeError => e
  catch_and_record_exception(e)
end
```
And instances like this where we need to raise a new exception:
```ruby
 raise_exeception_error(message: 'Unsuccessful HTTP response code', http_status: @response.status)
```

app/models/concerns/cfe_submission_state_machine.rb
stephenrichards: ✅

app/services/cfe/obtain_assessment_result_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def process_response
+      @submission.cfe_result = @response.body
+      if @response.status == 200
+        write_submission_history(@response, 'GET')
+        @submission.results_obtained!
+      else
+        @submission.fail!
+        error = CFE::SubmissionError.new('Unsuccessful HTTP response code', @response.status)
+        catch_and_record_exception(error, 'GET')
```
reggieb: My point was that here you wouldn't need to create an new error object!
```ruby
@submission.fail!
raise_exeception_error message: 'Unsuccessful HTTP response code', http_status: @response.status
```

PR #862, opened 2019-09-19
AP-949 Address Fixes

MalcolmVonMoJ:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-949)

- Examined the return from the address API and put different bits into different areas as needed.  Example: If Organisation is there, it goes on line 1, otherwise the street name and number does.  
- Removed the Organisation field (it now goes into line 1) as this was not editable.
- Added in the county field, so manually entered addresses with a county have that shew up on the CYA screen.
- Undid the wrapping of AP-813 - this should no longer be necessary.
- Changed the capitalisation code to keep hyphens and stop putting spaces between numbers and letters.

## Checklist

### DIFF
--- a/app/controllers/providers/address_selections_controller.rb
+++ b/app/controllers/providers/address_selections_controller.rb
@@ -46,12 +46,16 @@ module Providers
       end
     end
 
+    def address_titleize(sentence)
+      sentence.to_s&.split(' ')&.map(&:capitalize)&.join(' ')
+    end
+
     def titleize_addresses
       @addresses.each do |a|
-        a[:organisation] = a[:organisation]&.titleize
-        a[:address_line_one] = a[:address_line_one]&.titleize
-        a[:address_line_two] = a[:address_line_two]&.titleize
-        a[:city] = a[:city]&.titleize
+        a[:address_line_one] = address_titleize(a[:address_line_one])
+        a[:address_line_two] = address_titleize(a[:address_line_two])
+        a[:city] = address_titleize(a[:city])
+        a[:county] = address_titleize(a[:county])
       end
     end
   end
--- a/app/controllers/providers/address_selections_controller.rb
+++ b/app/controllers/providers/address_selections_controller.rb
@@ -46,16 +46,16 @@ module Providers
       end
     end
 
-    def address_titleize(sentence)
+    def hyphen_safe_titleize(sentence)
       sentence.to_s&.split(' ')&.map(&:capitalize)&.join(' ')
     end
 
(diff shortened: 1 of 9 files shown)


app/services/map_address_lookup_results.rb
```diff
@@ -1,19 +1,71 @@
...
+  end
+
+  def self.use_dependent_locality?(result)
+    (building_name(result).nil? && organisation_name(result).nil?) ||
+      ((building_name(result).nil? || organisation_name(result).nil?) && thoroughfare_name(result).nil?)
+  end
+
+  def self.get_line_one(result)
+    return combine_with_road(result, *NUMBER_NAME, ' ') if building_name(result).nil? && organisation_name(result).nil?
+    return combine(result, *FULL_NAME) if organisation_name(result).nil?
```
Obsiye: Duplication of condition with the line above e.g. organisation_name(result).nil?

mpw5: It's not quite duplication, the `organisation_name(result).nil?` needs to be present on both lines. 

The first `return` line checks whether the address has a building name or organisation name. If neither of those exists then the first line of the address is set as the building number and street name, and the method returns.

The second line will only be executed if one or both of the building name or organisation name are present. It checks to see if the organisation name is nil - if this is the case then we know that the building name is present, so the first line of the address is set to be the full building name, and the method returns.

If neither of those guards have returned then we know the address must have an organisation name present, and we always want the organisation on the first line, so the first line of the address is set to be the organisation name.

Obsiye: Makes sense thanks 👍

app/controllers/providers/address_selections_controller.rb
```diff
@@ -46,12 +46,16 @@ def build_addresses_from_form_data
       end
     end
 
+    def address_titleize(sentence)
+      sentence.to_s&.split(' ')&.map(&:capitalize)&.join(' ')
+    end
+
```
stephenrichards: I think the Rails helper method `#titleize` does this.

if you want to handle `nil`s then:

```sentence&.titleize || ''```

mpw5: `sentence.to_s&.split(' ')&.map(&:capitalize)&.join(' ')` handles hyphens better than `titleize` - `titleize` replaces them with a space, which can make addresses look odd.

Instead, renamed `address_titleize` to `hyphen_safe_titleize` to better describe what it does.

PR #874, opened 2019-09-23
Ap 952 create valid substantive case

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-952)

There are issues with substantive case payloads being injected in CCMS. This PR fixes the following issues:

* `ApplicationAmendmentType` is incorrectly being set as `SUBDP` for all cases. It should be `SUB` for substantive cases and `SUBDP` for delegated functions cases.
*  `CURRENT_CERT_SUBSTANTIVE`, `SUBSTANTIVE_APP` and `PROCEEDING_APPLICATION_TYPE` attributes are currently omitted. These should be included with values of `false`, `true` and `Substantive` for substantive cases, and `false`, `false` and `Both` for delegated functions cases.
* Fix a bug caused by a change to the `provider` model which was causing the `ccms_spec` to fail - add an email address to the provider double.

## Checklist

### DIFF
--- a/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
+++ b/spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
@@ -464,55 +464,59 @@ module CCMS # rubocop:disable Metrics/ModuleLength
 
       context 'APP_AMEND_TYPE' do
         context 'delegated function used' do
-          context 'in global_merits section' do
-            it 'returns SUBDP' do
-              allow(legal_aid_application).to receive(:used_delegated_functions?).and_return(true)
-              allow(legal_aid_application).to receive(:used_delegated_functions_on).and_return(Date.today)
-              block = XmlExtractor.call(xml, :global_merits, 'APP_AMEND_TYPE')
-              expect(block).to have_text_response 'SUBDP'
-            end
+          let(:legal_aid_application) do
+            create :legal_aid_application,
+                   :with_everything,
+                   :with_applicant_and_address,
+                   populate_vehicle: true,
+                   with_bank_accounts: 2,
+                   proceeding_types: [proceeding_type],
+                   provider: provider,
+                   used_delegated_functions: true,
+                   used_delegated_functions_on: Date.today
           end
-          context 'in global_means section;' do
-            it 'returns SUBDP' do
-              allow(legal_aid_application).to receive(:used_delegated_functions?).and_return(true)
-              allow(legal_aid_application).to receive(:used_delegated_functions_on).and_return(Date.today)
-              block = XmlExtractor.call(xml, :global_means, 'APP_AMEND_TYPE')
+
+          it 'returns SUBDP' do
+            %i[global_means global_merits].each do |entity|
+              block = XmlExtractor.call(xml, entity, 'APP_AMEND_TYPE')
               expect(block).to have_text_response 'SUBDP'
             end
           end
         end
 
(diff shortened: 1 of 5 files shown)


spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -519,6 +499,25 @@ module CCMS # rubocop:disable Metrics/ModuleLength
         end
       end
 
+      context 'EMERGENCY_FURTHER_INFORMATION' do
+        context 'delegated function used' do
+          it 'returns hard coded statement' do
+            allow(legal_aid_application).to receive(:used_delegated_functions?).and_return(true)
+            allow(legal_aid_application).to receive(:used_delegated_functions_on).and_return(Date.today)
```
reggieb: `legal_aid_application` is defined via a factory and `used_delegated_functions` and `used_delegated_functions_on` are both attributes of the model.

So override the factory rather than using mocking.

Also `block` is the same for both `it` declarations
```
    context 'delegated function used' do
      let(:block) { XmlExtractor.call(xml, :global_merits, 'EMERGENCY_FURTHER_INFORMATION') } 
      let(:legal_aid_application) do
        create :legal_aid_application,
               :with_everything,
               :with_applicant_and_address,
               populate_vehicle: true,
               with_bank_accounts: 2,
               proceeding_types: [proceeding_type],
               provider: provider,
               used_delegated_functions: true,
               used_delegated_functions_on: Date.today
      end

      it 'returns hard coded statement' do
        expect(block).to have_text_response 'Apply Service application - see uploaded provider statement of case'
      end
      ....
```

mpw5: Thanks. Good points. I didn't actually write those specs, I just moved them around a little to make them clearer (there was a problem with one `context` being nested inside an unrelated `context`) but your suggestions make sense and will bring the tests in line with the others in the file, so I've fixed them.

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -1049,6 +1052,20 @@ module CCMS # rubocop:disable Metrics/ModuleLength
           expect(block).to have_text_response application_proceeding_type.proceeding_type.ccms_code
         end
       end
+
+      context 'SUBSTANTIVE_APP populates correctly' do
+        it 'returns true when application is substantive' do
+          block = XmlExtractor.call(xml, :global_merits, 'SUBSTANTIVE_APP')
+          expect(block).to have_boolean_response true
+        end
+      end
```
stephenrichards: shouldn't we have tests for when it is a delegated functions application?

spec/services/ccms/case_add_requestor_xml_blocks_spec.rb
```diff
@@ -1049,6 +1052,20 @@ module CCMS # rubocop:disable Metrics/ModuleLength
...
+
+      context 'PROCEEDING_APPLICATION_TYPE populates correctly' do
+        it 'returns Substantive when application is substantive' do
+          block = XmlExtractor.call(xml, :proceeding_merits, 'PROCEEDING_APPLICATION_TYPE')
+          expect(block).to have_text_response 'Substantive'
+        end
+      end
```
stephenrichards: Ditto re tests for delegated functions application

PR #875, opened 2019-09-23
AP-941 Add service to pass property data to eligibility API

reggieb:
[Jira AP-941](https://dsdmoj.atlassian.net/browse/AP-941)

Adds the service `CFE::CreatePropertiesService` to manage the passing of property data to the Eligibility API. This service is also added to the CFE Submission manager


reggieb:
I'm putting this on hold while some of the other CFE service specs are updated, as fixing the coverage issues here will probably complicate the changes to the other specs that are currently being made.

PR #884, opened 2019-09-26
AP797 Delegated Functions Email Reminder

Obsiye:
Send an email reminder to the provider for delegated functions applications 5 days before and on the due date for the application

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&modal=detail&selectedIssue=AP-797)

- Add SubmitApplicationReminder service and mailer
- Test the reminder service and mailer
- Add smoke testing for this mailer

- Add a govuk notify template for :reminder_to_submit_an_application email

- Create a factory trait for :with_substantive_application_deadline_on instead of having this as part of the delegated functions trait because it calls an api and therefore will need vcrs in many tests that don't even require the substantive_application_deadline

- Reorder the ResendLinkRequestMailer arguments so that the smoke tests legal_aid_application method can be used by the SubmitApplicationReminderMailer in the smoke test

- Add a conditional to the mailer when their is no deadline date/ no delegated functions date 

To note:
The reminder emails are not sent if the application is already submitted or if it is not a delegated functions application (has no delegated functions date)

## Checklist


spec/factories/providers.rb
```diff
@@ -1,6 +1,7 @@
 FactoryBot.define do
   factory :provider do
     firm
+    id { SecureRandom.uuid }
```
reggieb: That shouldn't be necessary. Why did you need to do this? The database should be generating an `id` on create. If it isn't in your tests - that suggest something isn't right.

spec/services/smoke_test/test_emails_spec.rb
```diff
@@ -24,5 +24,11 @@
       subject
       ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper.drain
     end
+
+    it 'sends a notify_provider email' do
+      expect_any_instance_of(SubmitApplicationReminderMailer).to receive(:notify_provider).and_call_original
```
reggieb: Why `.add_call_original`? The expectation is fulfilled on calling the method, so what happens afterwards is not needed for the expectation to have been satisfied. And what you want to avoid is the system accidentally sending out emails. So I'd suggest doing the simpler `.and_return(true)`

app/services/working_day_calculator.rb
```diff
@@ -17,10 +17,18 @@ def initialize(date)
...
 
+  def working_days_from_date(number)
+    number.positive? ? add_working_days(number) : subtract_working_days(number)
+  end
+
   def add_working_days(number)
     calendar.add_business_days(date, number)
   end
 
+  def subtract_working_days(number)
```
reggieb: I think this new functionality needs it's own test. At the very least, I would suggest updating `spec/services/working_day_calculator_spec.rb` so that `number_to_target_dates` includes a negative number. So for example:

```ruby
let(:two_working_days_before) { Date.parse('13-Dec-2018') } # Weekend skip 
let(:number_to_target_dates) do
  {
    3 => three_working_days_later,
    5 => five_working_days_later,
    6 => six_working_days_later,
    -2 => two_working_days_before
  }
end
```

PR #899, opened 2019-09-30
AP-667 Bank selection page for OpenBanking

JulienSansot:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-667)

- Fetch and save list of banks the same way Rob did it for bank holidays
- Implement bank selection before TrueLayer
- Forward selected bank go TrueLayer

UAT: https://ap-667-replace-truelayer-scree-applyforlegalaid-uat.apps.live-1.cloud-platform.service.justice.gov.uk/

## Checklist

### DIFF
--- a/app/lib/omniauth/strategies/true_layer.rb
+++ b/app/lib/omniauth/strategies/true_layer.rb
@@ -14,7 +14,12 @@ module OmniAuth
              token_url: 'https://auth.truelayer.com/connect/token'
 
       def authorize_params
-        super.merge(enable_mock: enable_mock)
+        extra_params = { enable_mock: enable_mock }
+
+        extra_params[:provider_id] = session[:provider_id] if session[:provider_id].present?
+        extra_params[:consent_id] = consent_id if consent_id
+
+        super.merge(extra_params)
       end
 
       def token_params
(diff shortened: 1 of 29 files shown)


app/lib/omniauth/strategies/true_layer.rb
```diff
@@ -26,10 +31,15 @@ def token_params
...
       def enable_mock
-        env_setting = ENV['TRUE_LAYER_ENABLE_MOCK']
-        return false if env_setting.blank?
+        Rails.configuration.x.true_layer.enable_mock
+      end
 
-        ActiveModel::Type::Boolean.new.cast(env_setting)
+      # If consent_id is set, the flow will skip TrueLayer's consent page.
+      # consent_id is used by TrueLayer for tracking purposes and needs to be a static string specific to our TrueLayer account.
+      # It could be something different than true_layer.client_id but it makes sense to use this value
```
reggieb: I was a little concerned that this would expose the client id, but checking the Oauth2 documentation, I don't think this is a problem. For example, I found this quote: ["The client ID is considered public information, and is used to build login URLs"](https://aaronparecki.com/oauth-2-simplified/)

app/models/true_layer_bank.rb
```diff
@@ -0,0 +1,30 @@
...
+  MOCK_BANK = {
+    provider_id: 'mock',
+    display_name: 'Mock Bank',
+    logo_url: 'https://truelayer-client-logos.s3-eu-west-1.amazonaws.com/banks/banks-icons/mock-icon.svg'
+  }.freeze
+
+  def self.banks
+    TrueLayerBanksUpdateWorker.perform_in 10.seconds
+    instance = by_updated_at.last || create!
+    ([mock_bank] + instance.banks).compact
```
reggieb: Really minor: I think the only situation where there will be a `nil` in the array is if `mock_bank` is nil. By using `compact` that call operates over the whole array every time. Would this be better:
```
return instance.banks unless mock_bank

[mock_bank] + instance.banks
```

reggieb:
We'll need to be a little careful merging this or AP-966 as they both modify the true layer omniauth configuration.

app/models/true_layer_bank.rb
JulienSansot: That's much better, thanks

PR #905, opened 2019-10-03
Ap 965 ccms user login

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-965)

Added 2 new users 'NEETADESOR' and 'DAVIDGREYLLPTWO' to the mock_saml and set up hardcoded details for these providers which are required for CCMS

Set the CCMS submission to use the provider username linked to the application and remove the hardcoded value "NEETADESOR'

## Checklist

### DIFF
--- a/app/services/ccms/case_add_requestor.rb
+++ b/app/services/ccms/case_add_requestor.rb
@@ -133,8 +133,8 @@ module CCMS
 
     def generate_provider_details(xml)
       xml.__send__('ns2:ProviderCaseReferenceNumber', 'PC4') # TODO: insert @legal_aid_application.provider_case_reference_number when it is available in Apply
-      xml.__send__('ns2:ProviderFirmID', provider.firm_id)
-      xml.__send__('ns2:ProviderOfficeID', provider.selected_office_id)
+      xml.__send__('ns2:ProviderFirmID', provider.firm.ccms_id)
+      xml.__send__('ns2:ProviderOfficeID', @legal_aid_application.office.ccms_id)
       xml.__send__('ns2:ContactUserID') do
         xml.__send__('ns0:UserLoginID', provider.user_login_id)
       end
--- a/app/services/ccms/case_add_requestor.rb
+++ b/app/services/ccms/case_add_requestor.rb
@@ -22,7 +22,6 @@ module CCMS
       @transaction_time_stamp = Time.now.to_s(:ccms_date_time)
       @ccms_attribute_keys = YAML.load_file(File.join(Rails.root, 'config', 'ccms', 'ccms_keys.yml'))
       @attribute_value_generator = AttributeValueGenerator.new(@submission)
-      @provider = submission.legal_aid_application.provider
     end
 
     def call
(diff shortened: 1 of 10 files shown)


mpw5:
These changes seem to be working fine, but there's an issue somewhere else in the code which means that the `ProviderFirmID` and `ProviderOfficeID` aren't populating correctly. The xml contains the database uuid of the firm/office, not the actual id.

image

I think the problem is in these two lines:

image

Instead of `provider.firm.id` and `provider.selected_office_id` something like `provider.details_response['providerOffices'].first['providerfirmId']` and `p.details_response['providerOffices'].first['officeId']` would get the right values, assuming there's only one office. There might be a better way of doing that though.

config/initializers/mock_saml.rb
```diff
@@ -4,7 +4,9 @@
   users: [
     OpenStruct.new(username: 'test1', email: 'test1@example.com'),
     OpenStruct.new(username: 'test2', email: 'test2@example.com'),
-    OpenStruct.new(username: 'test3', email: 'really-really-long-email-address@example.com')
+    OpenStruct.new(username: 'test3', email: 'really-really-long-email-address@example.com'),
+    OpenStruct.new(username: 'NEETADESOR', email: 'neeta@desor.com'),
+    OpenStruct.new(username: 'DAVIDGRAYLLPTWO', email: 'dg@dgllp.com')
```
mpw5: These look like real email addresses. Can we change them to something fake to prevent emails being sent out accidentally to them? Something like `neeta@example.com` and `dg@example.com`

app/services/ccms/case_add_requestor.rb
```diff
@@ -133,7 +133,7 @@ def generate_client(xml)
 
     def generate_provider_details(xml)
       xml.__send__('ns2:ProviderCaseReferenceNumber', 'PC4') # TODO: insert @legal_aid_application.provider_case_reference_number when it is available in Apply
-      xml.__send__('ns2:ProviderFirmID', provider.firm.id)
+      xml.__send__('ns2:ProviderFirmID', provider.firm_id)
       xml.__send__('ns2:ProviderOfficeID', provider.selected_office_id)
```
stephenrichards: Firm id: As Mark said, the firm_id is the UUID of the record on our system.  What you need here is the  `ccms_id` on the `Provider`'s `Firm` record.

Office id:  I beleive that when the user logs in, the selected office_id is set on the `Provider` record, and this is used to populate the `office_id` field on the `LegalAidApplication`.  It's dangerous to use the `Provider#selected_office_id` because when the CCMS payload is generated (as a background task), the provider may have logged out and back in as another office.  Therefore you should use the `ccms_id` of the `LegalAidApplication`'s `Office` record.

Just check that my assumptions here are correct though

mpw5: @stephenrichards `provider.firm.ccms_id` and `@legal_aid_application.office.ccms_id` look good. I'll make the changes as Will is off today and it would be good to get this working so testing can start.

PR #918, opened 2019-10-07
AP-830 confirmation email on submission to CCMS

JulienSansot:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-830)

Send an email to the provider when the application has been submitted to CCMS

## Checklist


app/mailers/submission_confirmation_mailer.rb
```diff
@@ -0,0 +1,14 @@
+class SubmissionConfirmationMailer < GovukNotifyRails::Mailer
+  require_relative 'concerns/notify_template_methods'
+  include NotifyTemplateMethods
+
+  def notify(legal_aid_application, provider, applicant)
+    template_name :submission_confirmation
+    set_personalisation(
+      provider_name: provider['name'],
+      client_name: "#{applicant['first_name']} #{applicant['last_name']}".strip,
```
stephenrichards: Would this be better as a method on Applicant to ensure the same concatenation rules are followed everywhere?

JulienSansot: Yes, it would but in the case of smoke tests `applicant` is not an Applicant model but a hash, so this method is not available.
The params are hashes because otherwise, smoke tests (which are run every 30mn) would require to create new DB records or use existing ones (which don't always exist and I'm not sure it's a good idea to use real user data for smoke tests anyway)

stephenrichards: Ok, fair enough.  I think we should find a better way of doing smoke tests - perhaps using mock objects - but then we're not testing the real code.  A conundrum.  Let's leave it for now - I'll think about using perhaps a single dummy applicant in future.  I've approved this PR

JulienSansot: It's using mock objects generated by FactoryBot at the moment
(https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/services/smoke_test/test_emails.rb#L43)

PR #920, opened 2019-10-08
AP-965 CCMS user login in request Headers

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-965)

https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/905 made most of the changes required for this story. It did not include changing the `header_request` section of requests to use the correct provider user. 

This PR changes all CCMS `_requestor` classes so that they populate the correct provider username instead of the user hardcoded in the helm values yaml. It also removes the hardcoded value. 

## Checklist


spec/services/ccms/applicant_add_requestor_spec.rb
```diff
@@ -21,7 +21,9 @@ module CCMS
              date_of_birth: Date.new(1969, 1, 1))
     end
 
-    let(:requestor) { described_class.new(applicant) }
+    let(:provider_username) { 'my_login' }
+
+    let(:requestor) { described_class.new(applicant, provider_username) }
```
scruti: As the `provider_username` is only used in the requestor, couldn't we just hardcode it in the argument to avoid the indirection of the `let` declaration?
If we are worried on the argument meaning we could even use `"provider username"` as the  value itself.

mpw5: True. I'll change it in all the specs 👍

PR #921, opened 2019-10-08
AP-975 Call Check Financial Eligibility service and store result

stephenrichards:
## Call Check Financial Eligibility service for passported applicants at the end of the means journey

[Link to story](https://dsdmoj.atlassian.net/browse/AP-975)

- Add a step to call the `CFE::SubmissionManager` when continuing from the check passported answers page
- Create a new model `CFE::Result` to hold the result returned by the Check Financial Eligibility service 
- Add code to the `CFE::ObtainAssessmentResultService`to create the CFE::Result record

## Checklist


app/services/cfe/submission_manager.rb
```diff
@@ -16,9 +16,11 @@ def call
       CFE::CreateVehiclesService.call(submission)
       CFE::CreatePropertiesService.call(submission)
       CFE::ObtainAssessmentResultService.call(submission)
+      true
     rescue CFE::SubmissionError => e
       submission.error_message = e.message
       submission.fail!
+      false
```
JulienSansot: should we also log this error to Sentry so we'll get notified?
```ruby
 Raven.capture_exception(e)
```

spec/services/cfe/create_vehicles_service_spec.rb
```diff
@@ -3,7 +3,7 @@
 module CFE
   RSpec.describe CreateVehiclesService do
     let(:application) { create :legal_aid_application }
-    let!(:vehicle) { create :vehicle, :populated, legal_aid_application: application }
+    # let!(:vehicle) { create :vehicle, :populated, legal_aid_application: application }
```
JulienSansot: remove?

app/services/cfe/submission_manager.rb
stephenrichards: Yep, good idea.  Will add.

PR #923, opened 2019-10-08
Ap 847 privacy policy

Obsiye:
Create a page for the Privacy Policy statement and link it in the footer of the application view

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-847)

- Create page for the Privacy Policy statement and link it in the footer of the application view

- Create controller, view, translation and tests for the privacy_policy page

- Link the privacy policy in the footer layout of the application view

- Update list_from_translation_path method to handle translations where the journey is unknown to prefix the translation path with an empty string

To note:
There are embedded hyperlink texts in the privacy policy statement. I embedded the hyperlink in the translation file and appended _html at the end of the translation key. This meant that I was not able to iterate through a translation list if it had embedded links in it.

## Checklist

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -40,7 +40,7 @@ Rails.application.routes.draw do
 
   resources :status, only: [:index]
   resource :contact, only: [:show]
-  resource :privacy_policy, only: [:show]
+  resources :privacy_policy, only: [:index]
   resources :feedback, only: %i[new create show]
   resources :errors, only: [:show], path: :error
   resources :problem, only: :index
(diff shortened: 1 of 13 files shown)


config/routes.rb
```diff
@@ -20,6 +20,7 @@
 
   resources :status, only: [:index]
   resource :contact, only: [:show]
+  resource :privacy_policy, only: [:show]
```
JulienSansot: would
```ruby
resources :privacy_policy, only: [:index]
```
allow you to use `privacy_policy` everywhere? Rather than `privacy_policies`

PR #927, opened 2019-10-09
AP-968 Search applications

JulienSansot:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-968)

## Creating a new page to search for applications

I tried using PostreSQL's Full text searching (https://www.postgresql.org/docs/9.5/textsearch-intro.html) but it doesn't match what we want. It seems to work on words. For example, it doesn't find "Reese-Mogg" with "smog"

So I'm using PostreSQL's `regexp_replace` to remove all non-alphanumeric characters.

## Checklist

### DIFF
--- a/app/controllers/providers/legal_aid_applications_controller.rb
+++ b/app/controllers/providers/legal_aid_applications_controller.rb
@@ -12,6 +12,7 @@ module Providers
     def search
       if search_term.present?
         applications
+        log_search
       elsif search_term == ''
         @error = t('.error')
       end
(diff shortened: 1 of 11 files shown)


JulienSansot:
@stephenrichards I agree with you the performance is a concern. If it does become slow but we're happy with the matches, I suggest writing the cleaned values in extra indexed columns.

For example, when `application_ref` changes :
```ruby
application_ref_search = application_ref.gsub(/[^0-9a-z\\s]/i, '')
```

PR #928, opened 2019-10-10
AP-966 Allow admin users to log in via google

reggieb:
## What

[Jira AP-966](https://dsdmoj.atlassian.net/browse/AP-966)

Adds a link on the Admin User log in screen, which allow a person to log in via their Google account rather than by using user name and password.

There was no design for this - so I've just added a very simple link to the page.

Note, that for log in to work, there needs to be an `AdminUser` with the same email address as the google user. To add admin users go to `/support/admin_user`.

Most of the work involved refactoring the True Layer login so that it would work alongside the Google login - Device omniauth is a little complicated if two models both use OAuth2. See the Devise Wiki entry ["OmniAuth with multiple models"](https://github.com/plataformatec/devise/wiki/OmniAuth-with-multiple-models)


app/controllers/applicants/omniauth_callbacks_controller.rb
```diff
@@ -17,11 +17,6 @@ def true_layer
       redirect_to gather_citizens_accounts_path
     end
 
-    def failure
-      set_flash_message(:error, :failure, kind: 'TrueLayer', reason: 'Process cancelled')
-      redirect_to citizens_consent_path
```
JulienSansot: why we don't need this method anymore?

reggieb: With devise handing the omniauth failures, the failures are redirected back to the omniauth callbacks controller's `failure` action. 

When omniauth was stripped out of the Devise config failures were being sent back to `auth/failure` and I added a separate controller to handle that.

config/application.rb
```diff
@@ -67,13 +67,16 @@ class Application < Rails::Application
...
-    config.x.true_layer.enable_mock    = ENV['TRUE_LAYER_ENABLE_MOCK'] == 'true'
+    config.x.true_layer.client_id = ENV['TRUE_LAYER_CLIENT_ID']
+    config.x.true_layer.client_secret = ENV['TRUE_LAYER_CLIENT_SECRET']
+    config.x.true_layer.enable_mock = ENV['TRUE_LAYER_ENABLE_MOCK'] == 'true'
 
     require Rails.root.join 'app/lib/govuk_elements_form_builder/form_builder'
     ActionView::Base.default_form_builder = GovukElementsFormBuilder::FormBuilder
 
+    config.x.google_oauth2.client_id      = ENV['GOOGLE_CLIENT_ID']
+    config.x.google_oauth2.client_secret = ENV['GOOGLE_CLIENT_SECRET']
```
JulienSansot: line 77 has an extra space it seems

reggieb:
I've added the environment variables to the helm deploy code - but please note - at the moment I've used google client id and secret generated from my own google developer console. These will need to be changed for MoJ versions. See the omniauth-google-oauth2 gem's [README](https://github.com/zquestz/omniauth-google-oauth2) for info on getting these.

PR #930, opened 2019-10-10
AP-332 Add link to the feedback page when the application is submitted

JulienSansot:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-332)

Add link to the feedback page when the application is submitted

## Checklist

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -144,6 +144,7 @@ en:
         client_sign: get your client to sign it
         application_to_be_checked: We'll check your application to see if your client is entitled to legal aid.
         decision_in_ccms: We'll let you know our decision in CCMS.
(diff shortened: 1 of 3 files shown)


config/locales/en/providers.yml
```diff
@@ -144,6 +144,7 @@ en:
         client_sign: get your client to sign it
         application_to_be_checked: We'll check your application to see if your client is entitled to legal aid.
         decision_in_ccms: We'll let you know our decision in CCMS.
+        feedback_link_html: 'Please provide us with feedback to <a href="%{url}" class="govuk_link">help us improve this service</a>.'
```
reggieb: I started by thinking that this was interesting, but could be improved .... but then the more I looked at it, the harder it was to think of a better solution, and the more I appreciated what a great solution this is. Love it!

JulienSansot: Haha, this comment got better and better!

PR #933, opened 2019-10-11
AP-979 Setup Blazer

JulienSansot:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-979)

Adding Blazer to `/blazer` path
And adding tests to make sure `/blazer` and `/support` can only be accessed by admin users

## Checklist


app/controllers/saml_idp_controller.rb
```diff
@@ -1,3 +1,5 @@
+# TODO: add unit tests for this controller
+# :nocov:
 class SamlIdpController < SamlIdp::IdpController
```
JulienSansot: Adding the two spec files made this file being read by simplecov. Before that, it was just ignored

app/controllers/flow_base_controller.rb
```diff
@@ -1,4 +1,4 @@
-class BaseController < ApplicationController
+class FlowBaseController < ApplicationController
```
JulienSansot: It was conflicting with Blazer's BaseController 
https://github.com/ankane/blazer/blob/master/app/controllers/blazer/base_controller.rb

PR #945, opened 2019-10-14
AP792 Delegated confirmation page

Obsiye:
Add a delegated functions confirmation page after the substantive page when using delegated functions and clicking 'no, i'll start later' on the substantive page.

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-972)

- Add a controller, view, locales, unit tests and feature tests
- Update the provider flow
- Update existing tests

## Checklist


config/locales/en/providers.yml
```diff
@@ -129,6 +129,15 @@ en:
...
       show:
         h1-heading: Latest incident details
+    delegated_confirmation:
+      index:
+        page-heading: You notified us you've used delegated functions
+        sub_title: Your case reference number is
+        what_happens_next: What happens next
+        info: You can return to this application when you're ready to make a substantive application.
+        warning:
+          application_deadline_date: You must submit a substantive application by 10 November 2019.
```
mpw5: I don't think the date should be hardcoded here - won't it vary from case to case?

config/routes.rb
```diff
@@ -198,6 +198,7 @@
       resource :substantive_application, only: %i[show update]
       resource :end_of_application, only: %i[show update]
       resource :submitted_application, only: :show
+      resources :delegated_confirmation, only: :index
```
mpw5: Should this be `resource` not `resources`?

config/locales/en/providers.yml
Obsiye: 🤕 yes, bad mistake.

config/locales/en/providers.yml
```diff
@@ -136,7 +136,7 @@ en:
         what_happens_next: What happens next
         info: You can return to this application when you're ready to make a substantive application.
         warning:
-          application_deadline_date: You must submit a substantive application by 10 November 2019.
+          application_deadline_date: "You must submit a substantive application by "
```
mpw5: Are the `"`s necessary (Genuine question, I don't know! It looks like some translations use them, some don't. Is it for interpolation?)

PR #997, opened 2019-11-01
AP-989 Add Sentry alerting for postcode lookup

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-989)

We want to be alerted when calls to the Ordnance Survey Places API fail.

This commit adds Sentry calls to the `AddressLookupService` class in two places, so that alerts will occur if the service is unavailable and if the service is available but is unsuccessful in returning a response (for example if authentication fails).

## Checklist


mpw5:
> I think we should have a test for this

@stephenrichards Test added

PR #1022, opened 2019-11-07
AP-1035 fix bug where saved applications showing as submitted

stephenrichards:
## Bug fix - saved applications showing as submitted

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1035)

If the user pressed Save as Draft on the Merits Declaration page, the `submitted_at` field was being set on the `MeritsAssessment` which in turn caused the application to be shown as 'Submitted` on the index page.

- Amended  conditions under which `submitted_at` was set
- wrote specs to test the behaviour for both Save as Draft and Continue scenarios

## Checklist

### DIFF
--- a/app/controllers/providers/merits_declarations_controller.rb
+++ b/app/controllers/providers/merits_declarations_controller.rb
@@ -5,10 +5,11 @@ module Providers
     def update
       unless draft_selected?
         legal_aid_application.generate_reports! if legal_aid_application.may_generate_reports?
-        merits_assessment.update!(submitted_at: Time.current) unless merits_assessment.submitted_at?
+        merits_assessment.submit!
       end
       continue_or_draft
     end
+
 
     private
 
--- a/app/controllers/providers/merits_declarations_controller.rb
+++ b/app/controllers/providers/merits_declarations_controller.rb
@@ -3,8 +3,10 @@ module Providers
     def show; end
 
     def update
-      legal_aid_application.generate_reports! unless draft_selected? || !legal_aid_application.may_generate_reports?
-      merits_assessment.update!(submitted_at: Time.current) unless merits_assessment.submitted_at?
+      unless draft_selected?
+        legal_aid_application.generate_reports! if legal_aid_application.may_generate_reports?
+        merits_assessment.update!(submitted_at: Time.current) unless merits_assessment.submitted_at?
+      end
       continue_or_draft
     end
 
--- a/app/controllers/providers/merits_declarations_controller.rb
+++ b/app/controllers/providers/merits_declarations_controller.rb
@@ -10,7 +10,6 @@ module Providers
       continue_or_draft
     end
 
-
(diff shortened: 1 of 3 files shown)


app/controllers/providers/merits_declarations_controller.rb
```diff
@@ -3,8 +3,10 @@ class MeritsDeclarationsController < ProviderBaseController
     def show; end
 
     def update
-      legal_aid_application.generate_reports! unless draft_selected? || !legal_aid_application.may_generate_reports?
-      merits_assessment.update!(submitted_at: Time.current) unless merits_assessment.submitted_at?
+      unless draft_selected?
+        legal_aid_application.generate_reports! if legal_aid_application.may_generate_reports?
+        merits_assessment.update!(submitted_at: Time.current) unless merits_assessment.submitted_at?
+      end
```
scruti: :mega: This may be out of scope of this PR, but thought was worth to raise it here.

I can't but wonder if this original bug was somewhat caused by the complex branching logic caused by the controller calls `asking`, instead of `telling`.
In my point of view, having something like the following block would make the controller easy to understand/maintain:
```ruby
unless draft_selected?
  legal_aid_application.generate_reports! 
  merits_assessment.submit(Time.current)
end
```
The controller `tells` and does not `ask` for the other objects state. And in this case seems specially fine as is not collecting data for a third call, is just triggering actions that are not dependent to each other. 
The responsibility of knowing how these actions are carried on and if they can be executed should be on the objects receiving the messages, not in the caller.

In the case of `if legal_aid_application.may_generate_reports?` seems this is just a helper method provided from the State Machine that checks that the current state is the one needed to transition to the next state. 
I guess this condition was put in place in the controller to avoid a `AASM::InvalidTransition` exceptions when the state machine is not in the needed state. 
But this seems avoidable by wrapping the state machine transition definition inside a block:
```
  aasm whiny_transitions: false do
    ...
  end
```
So we could avoid the condition check in the controller and when the reports cannot be generated at this stage, `legal_aid_application.generate_reports! ` would just return `false`. 


For `unless merits_assessment.submitted_at?` seems this `update unless I already have the value defined` is a responsibility that belongs to the `merits_assessment` object, not to the controller.
So moving that check to a method at model definition (I used `submit` just as an example) seems the way to go.

What do you think?

stephenrichards: Good point.  Am going to change

stephenrichards: In the end, I left the `if legal_aid_application.may_generate_reports?` in the controller.

I couldn't find a way to turn off whiny transitions for just that one event, and doing it for all events would cause issues elsewhere I think - at the very least we would have to do a fair amount of investigation and refactoring.

But yet, the decision whether or not to submit belongs firmly in the `MeansAssessment` model.

PR #1098, opened 2019-12-04
AP-1126 Add new cronjob for database backups

willc-work:
## What

Added 2 cronjobs to deal with DB backups.

The first take an hourly snapshot and stores this in our RDS backup storage along side the daily snapshots taken by cloud platforms. Daily snapshots are retained for 7 days.

The second cronjob manages deleting the hourly snapshots. It deletes all but the most recent 30 snapshots every 4 days. We are currently taking 32 per day so this should mean we have approx 2 days of hourly snapshots available in case we need to do a restore.

## Checklist

### DIFF
--- a/bin/cleanup_manual_snapshots
+++ b/bin/cleanup_manual_snapshots
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+echo "Logging into RDS"
+
+regex_manual='apply-manual-snapshot'
+
+regex_automatic='rds:cloud-platform'
+
+unset AWS_PROFILE
+
+read AWS_ACCESS_KEY_ID name pass user addr endp AWS_SECRET_ACCESS_KEY url \
+      <<<$(kubectl -n laa-apply-for-legalaid-staging get secret apply-for-legal-aid-rds-instance-output -o json | jq -r '.data[] | @base64d')
+
+export AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
+
+
+# uncomment the line below to print a list of the current backups to the console
+aws rds describe-db-snapshots --db-instance-identifier cloud-platform-464651662c253592 | jq -r '.DBSnapshots[] | "\(.DBSnapshotIdentifier) created at \(.SnapshotCreateTime)"'
+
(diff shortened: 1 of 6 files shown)


bin/cleanup_manual_snapshots
```diff
@@ -0,0 +1,36 @@
...
+regex_automatic='rds:cloud-platform'
+
+unset AWS_PROFILE
+
+read AWS_ACCESS_KEY_ID name pass user addr endp AWS_SECRET_ACCESS_KEY url \
+      <<<$(kubectl -n laa-apply-for-legalaid-staging get secret apply-for-legal-aid-rds-instance-output -o json | jq -r '.data[] | @base64d')
+
+export AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
+
+# uncomment the line below to print a list of the current backups to the console
```
mpw5: Nit picky, but the line below isn't commented out, so I'm not sure if this comment really adds anything

bin/create_manual_snapshot
```diff
@@ -0,0 +1,19 @@
...
+
+unset AWS_PROFILE
+
+read AWS_ACCESS_KEY_ID name pass user addr endp AWS_SECRET_ACCESS_KEY url <<<$(kubectl -n laa-apply-for-legalaid-staging get secret apply-for-legal-aid-rds-instance-output -o json | jq -r '.data[] | @base64d')
+
+export AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY
+
+aws rds create-db-snapshot --region=eu-west-2 --db-instance-identifier cloud-platform-464651662c253592 --db-snapshot-identifier $new_fileName
+
+echo $new_fileName "has been created" # prints out even if snapshot creation fails, add if statement to print success/fail message
```
mpw5: Is the comment needed? Maybe add a 'TO DO' if it's a reminder to make improvements in future?

bin/cleanup_manual_snapshots
willc-work: now commented out

bin/create_manual_snapshot
willc-work: Have amended it as a TODO.

PR #1107, opened 2019-12-06
Ap 1094 journey specific errors

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1094)

Added separate errors to the citizen and journey paths for own_home and shared_ownership

## Checklist

### DIFF
--- a/app/controllers/citizens/own_homes_controller.rb
+++ b/app/controllers/citizens/own_homes_controller.rb
@@ -18,7 +18,7 @@ module Citizens
 
     def form_params
       merge_with_model(legal_aid_application, journey: :citizens) do
-        return {} unless params[:legal_aid_application]
+        next {} unless params[:legal_aid_application]
 
         params.require(:legal_aid_application).permit(:own_home)
       end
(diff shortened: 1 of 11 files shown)


app/controllers/citizens/own_homes_controller.rb
```diff
@@ -17,8 +17,8 @@ def update
     private
 
     def form_params
-      merge_with_model(legal_aid_application) do
-        return {} unless params[:legal_aid_application]
+      merge_with_model(legal_aid_application, journey: :citizens) do
+        next {} unless params[:legal_aid_application]
```
stephenrichards: I don't understand why `return` has been replaced with `next` here (or in the other controllers.

willc-work: When 'return' is used the translation is looking for the 'shared_ownership.blank' translation for the page and not the journey specific one, e.g shared_ownership.citizens.blank, this pattern is also used on vehicles_controller and success_likely_controller to achieve the same result. from what I read online , next exits the block and returns its argument as block return value, which in this case includes the 'journey'. I am not sure why return does not also include the 'journey' however.

PR #1134, opened 2019-12-20
AP590 Notify Provider When Citizen Completes Means

Obsiye:
AP590 Notify Provider When Citizen Completes Means

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&modal=detail&selectedIssue=AP-590)

- Add the relevant notify template, :client_completed_means, to govuk_notify_templates.yml 
- Create CitizenCompletedMeansMailer to notify a provider with the relevant notify_template
- Add ProviderEmailService to call a deliver_later! on CitizenCompletedMeansMailer
- Update Declaration Controller to call the ProviderEmailService on update

- Add tests
- Add cassettes
- Update SmokeTests

## Checklist


app/services/provider_email_service.rb
```diff
@@ -0,0 +1,41 @@
...
+      application,
+      provider.name,
+      applicant.full_name,
+      application_url
+    ] + with_provider_email
+  end
+
+  def with_provider_email
+    HostEnv.staging? ? [] : [provider.email]
+  end
```
stephenrichards: Can we turn this round so that it passes in the provider email for host env production, and not for everything else?  I feel that's a bit safer.

PR #1159, opened 2020-01-09
Ap 1170 scheduled mailings

willc-work:
## Replace delayed mails with ScheduledMailings model

The reminder emails were being generated with a delayed time to send.  However, if the application was subsequently submitted, the mails were not removed from the Sidekiq scheduled set, so users subsequently received mails asking them to submit applications that were already submitted.

Removing the mails from the Sidekiq Scheduled set is rather complicated, so a different approach was taken - to write records to a `scheduled_mailings` table, with all the details required to send them at the appropriate time.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1170)

- Created `scheduled_mailngs` table and model
- Replaced writing two delayed mails to the sidekiq scheduled set with writing two `ScheduledMailing` records
- Created `PostSubmissionProcessingJob` to send confirmation mail and cancel any outstanding 'scheduled_mailings`
- Created `ScheduledMailingsDeliveryJob` to scan the `scheduled_mailings` table for mails due to be sent, and send them
- Created a cron job to run the `ScheduledMailingsDeliveryJob` three times per day

## Checklist


app/mailers/submit_application_reminder_mailer.rb
```diff
@@ -2,7 +2,11 @@ class SubmitApplicationReminderMailer < GovukNotifyRails::Mailer
   require_relative 'concerns/notify_template_methods'
   include NotifyTemplateMethods
 
-  def notify_provider(application, name, to = support_email_address)
+  def notify_provider(application_id, name, to = support_email_address)
+    # in order to still work for scheduled sidekiq jobs, allow application rather than application_id to be passed in
+    # This can be changed to use just application id once all the scheduled mailer jobs in sidekiq have been delivered.
```
colinbruce: Should this comment be deleted?  I can't see Application objects being passed in anymore

willc-work: yes it should, i have removed it now👍

PR #1179, opened 2020-01-17
Auto screenshot cucumber failures

colinbruce:
## What

Updates cucumber to save screenshot when features fail
The screenshots will be stored in `tmp/capybara` so the screenshots will be ignored by .gitignore.
The file pattern is:
`folder`-`feature_file_name`-feature-`yearmonthday`T`time`.png
this allows you to see to the state of the page at failure

It will also upload the files in CircleCI, these will be available in the artifacts folder and means that you won't need to scroll through the list of jobs to find the failures

## Checklist

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -113,6 +113,11 @@ end
 World(FactoryBot::Syntax::Methods)
 World(Warden::Test::Helpers)
 
+After do |scenario|
+  name = scenario.location.file.gsub('features/','').gsub(/\.|\//, '-')
+  screenshot_image(name) if scenario.failed?
+end
+
 AfterStep do
   next unless ENV['SAVE_PAGES'] == 'true'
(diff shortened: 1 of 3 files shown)


features/support/env.rb
```diff
@@ -113,6 +113,11 @@
 World(FactoryBot::Syntax::Methods)
 World(Warden::Test::Helpers)
 
+After do |scenario|
+  name = scenario.location.file.gsub('features/', '').gsub(%r{/\.|\/}, '-')
+  screenshot_image(name) if scenario.failed?
```
Obsiye: Maybe wrap the name variable in the if scenario.failed? as well e.g. 
if scenario.failed?
name = ..
screenshot_image(name)
end

colinbruce: Good catch! Fixed in next commit.

PR #1206, opened 2020-01-31
AP-1182 CCMS attribute configuration merge

stephenrichards:
### ## Provide factility to merge CCMS attribute configs

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1182)

Created a class `CCMS::AttributeConfiguration` which will either load the standard CCMS attribute configuration, or load the standard and merge with the non_passported configuration, depending on the parameter passed in.

This is to enable us to start setting parameters for non_passported that are additional, or different in a separate file, but omit those that are the same for passported (standard) applications

## Checklist


app/services/ccms/attribute_configuration.rb
```diff
@@ -0,0 +1,24 @@
...
+    VALID_APPLICATION_TYPES = %i[standard non_passported].freeze
+
+    attr_reader :config
+
+    # Expects application type of :standard or :non_passported
+    def initialize(application_type:)
+      raise ArgumentError, 'Invalid application type' unless application_type.in?(VALID_APPLICATION_TYPES)
+
+      @config = YAML.load_file(configuration_files[:standard])
+      @config.deep_merge! YAML.load_file(configuration_files[application_type]) if application_type != :standard
```
Obsiye: is the `if application_type != :standard` required? since it's being checked in the raise error above `unless application_type.in?(VALID_APPLICATION_TYPES)`

PR #1211, opened 2020-02-03
BUGFIX

stephenrichards:
## Remove duplicate class

The `ObtainDocumentIdService` class was duplicated during a rebase to master resulting in the one in the parent directory being instantiated, and then failing because it couldn't find its base class.

## Checklist


app/services/ccms/submitters/obtain_document_id_service.rb
```diff
@@ -48,7 +52,7 @@ def request_document_ids
       end
 
       def document_id_requestor
-        @document_id_requestor ||= CCMS::Requestors::DocumentIdRequestor.new(submission.case_ccms_reference, submission.legal_aid_application.provider.username)
+        @document_id_requestor ||= Requestors::DocumentIdRequestor.new(submission.case_ccms_reference, submission.legal_aid_application.provider.username)
```
Obsiye: you could use the above #legal_aid_application method to get the provider username e.g. `legal_aid_application.provider.username`

PR #1223, opened 2020-02-07
Ap 1192 feedback link in confirmation email

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1192)

The feedback link is now added to the "successful submission to CCMS" email template in Notify. However when a user clicks the link, if they are not currently logged in to the application, they see a message that says "You have been logged out". As the feedback page is baked in to the signout process there is no way to bypass this message. I have therefore changed the wording to the more neutral "You are signed out" and after speaking with Jim I also added in a "Sign In" link in the header.

Also removes the smoke test as per discussion on AP-1209, this change broke the smoke test and as they were due to be removed I removed them on this ticket instead of fixing them and removing them later.

## Checklist


config/environments/production.rb
```diff
@@ -54,7 +54,7 @@
   config.action_mailer.perform_caching = false
 
   config.action_mailer.default_url_options = {
-    host: ENV.fetch('MAILER_DEFAULT_URL', 'localhost'),
+    host: ENV.fetch('MAILER_DEFAULT_URL', 'thisisatest'),
```
colinbruce: Should this be on the production settings?

PR #1232, opened 2020-02-11
AP-1169 Set APPLY_CASE_MEANS_REVIEW attribute according to business rules

stephenrichards:
## Set APPLY_CASE_MEANS_REVIEW attribute according to business rules

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1169)

The `APPLY_CASE_MEANS_REVIEW` attribute controls whether or not an appllication is routed to a case worker for review once it  reaches CCMS.

Normally, applications are not manually reviewed, however they must be if:
- There is a contribution to pay
- The capital assets are subject to a restriction

Additionally, a global setting is used to direct all applications to a case worker for review. This is for use during the interim period before CCMS routing has been updated.

 - updated the attribute confituration for APPLY_CASE_MEANS_REVIEW to set the value according to the output of a method
- created a class `CCMS::ManualReviewDeterminer` as a one-stop shop for decisions regarding whether or not the case needs to be reviewed
- Added a column `manually_review_all_cases` to the `Setting` model.

## Checklist


spec/services/ccms/requestors/case_add_requestor_xml_blocks/passported_spec.rb
```diff
@@ -563,19 +563,19 @@ module Requestors # rubocop:disable Metrics/ModuleLength
...
+        #      [:global_means, 'APPLY_CASE_MEANS_REVIEW'],
+        #      [:global_means, 'APPLY_CASE_MEANS_REVIEW']
+        #    ]
+        #    attributes.each do |entity_attribute_pair|
+        #      entity, attribute = entity_attribute_pair
+        #      block = XmlExtractor.call(xml, entity, attribute)
+        #      expect(block).to have_boolean_response true
+        #    end
+        #  end
+        # end
```
colinbruce: Can be deleted?

PR #1238, opened 2020-02-12
AP-1229 Hard-coded true attributes in non-passported payload

stephenrichards:
### This is branched from AP-1228, so best wait until that is merged before reviewing.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1229)

Editted configs and tests to generate all hard coded true attributes

## Checklist


spec/support/xml_entity_matcher.rb
```diff
@@ -0,0 +1,13 @@
+module XMLEntitykMatchers
```
colinbruce: Typo in XMLEntitykMatchers?

PR #1259, opened 2020-02-19
AP-1194 benefit Attribues

stephenrichards:
## Add attributes concerning benefits to the CCMS payload

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1194)

The attributes specified in the ticket describe whether or not the applicant recieves individual state benefits, and the amounts they receive.  We are not currently able to determine that within the system, so the boolean attributes are all coded as FALSE, and the value attributes omitted.

This may change once we identify individual benefits in the service.

## Checklist


spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_variable_attributes_spec.rb
```diff
@@ -319,6 +319,28 @@ module Requestors # rubocop:disable Metrics/ModuleLength
...
+                expect(block).not_to be_present
+              end
+            end
+          end
+
+          context 'applicant has vehicle' do
+            before { vehicle.update! estimated_value: 6500, payment_remaining: 3215.66, purchased_on: 5.years_ago.to_date, used_regularly: true }
+            context 'In regular use?' do
+              context 'in regular use' do
+              end
```
colinbruce: Should this have a test in it?

PR #1272, opened 2020-02-21
AP1197 Add legal_aid and maintenance_out attributes to CCMS payload

Obsiye:
AP1197 Add non-passported outgoing transaction attributes for legal_aid and maintenance_out to CCMS payload

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1197)

- Add two outgoing transactions attributes maintenance_out and legal_aid (criminal legal aid payments) into the non-passsported flow.

- Add `.for_outgoing_type?` to help with retrieval of required outgoing transactions data.

- Update tests

To note: The method calls for outgoing_* attributes should end in a '?' as it returns a boolean value. I've amended the case for OUTGOING attributes in the attribute_value_generator to use `.chomp('?'))`
## Checklist

### DIFF
--- a/app/models/transaction_type.rb
+++ b/app/models/transaction_type.rb
@@ -49,6 +49,11 @@ class TransactionType < ApplicationRecord
     I18n.t("transaction_types.names.#{journey}.#{name}")
   end
 
+  def self.for_outgoing_type?(transaction_type_name)
+    transaction_type = TransactionType.where(name: transaction_type_name, operation: :debit).first
+    transaction_type.present?
+  end
+
   def citizens_label_name
     label_name(journey: :citizens)
   end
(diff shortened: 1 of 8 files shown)


app/models/transaction_type.rb
```diff
@@ -44,6 +44,11 @@ def label_name(journey: :citizens)
     I18n.t("transaction_types.names.#{journey}.#{name}")
   end
 
+  def self.for_outgoing_type?(transaction_type_name)
+    transaction_type = TransactionType.where(name: transaction_type_name, operation: :debit).first
+    transaction_type.present?
+  end
```
stephenrichards: I like this, but it needs a specific test on the transaction_type_spec.rb

PR #1295, opened 2020-03-03
Ap-1199 CCMS Housing attributes

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1199)

Added the housing cost attributes for CCMS

## Checklist

### DIFF
--- a/spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
+++ b/spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
@@ -462,6 +462,13 @@ module CCMS
           end
         end
 
+        context 'GB_INPUT_C_13WP3_4A' do
+          it 'is omitted' do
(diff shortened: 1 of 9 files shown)


spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
```diff
@@ -462,6 +462,13 @@ module Requestors # rubocop:disable Metrics/ModuleLength
           end
         end
 
+        context 'GB_INPUT_C_13WP3_4A' do
+          it 'is omitted' do
+            block = XmlExtractor.call(xml, :global_means, 'GB_INPUT_C_13WP3_4A')
+            expect(block).not_to be_present, "Expected block for attribute GB_INPUT_C_13WP3_4A not to be generated, but was \n #{block}"
```
colinbruce: Is the error text supposed to be there? I've not seen that pattern on other tests.

willc-work: removed it as it is used for debugging (printing out the block)

PR #1334, opened 2020-03-17
AP-1300 Add Tranche 4 users

stephenrichards:
## Add tranche 4 users

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1300)

- added new rake task `tranche:maker` to read a file of usernames, check that they are on the provider API, and add them to the two config files.
- Ran it for the specified users.

## Checklist


lib/tasks/tranche_maker.rake
```diff
@@ -0,0 +1,38 @@
+# rake file to assist in adding beta users to the two config files
+#
+# Simply create a list of new users one per line in a text file, and put the filename in NEW_USERS_FILE
+#
+# After you run the rake taks, you will have two new config files with UPDATED in their name, so that you
+# can compare before replacing the original with the updated ones.
+#
+namespace :tranche do
+  task maker: :environment do
+    NEW_USERS_FILE = '/Users/stephenrichards/Library/Preferences/RubyMine2019.3/scratches/scratch_5.txt'
```
colinbruce: Would it be worth pointing to a placeholder text file in the root path and adding it to git ignore?
e.g. `File.join(Rails.root, 'add_tranche_users.txt')`

PR #1342, opened 2020-03-19
AP-1305: Non passported apply case means review

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1305)

Add extra checks to ensure that, no matter the state, non-passported claims are always marked for manual review in CCMS

## Checklist


app/services/ccms/manual_review_determiner.rb
```diff
@@ -14,13 +14,19 @@ def initialize(legal_aid_application)
     def call
       return true if Setting.manually_review_all_cases?
 
+      return true if application_is_non_passported?
```
stephenrichards: There are convenience methods on `LegalAidApplication`, so we can write here:

`return true if @legal_aid_application.non_passported?`

I think this is better because we want to know if it's passported or not in several places, and it's good to have the rules just in once place in case they ever change.

colinbruce: Absolutely, thanks! 👍

colinbruce: I've left this as a separate method because the logic _will_ change in the future and this may be easier to re-use in more complex logic paths and avoid the need to re-extract the method

PR #1349, opened 2020-03-24
AP-1304 extract errors doc upload

stephenrichards:
## Extract errors from response XML for doc registration and doc upload

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1304)

Two of the calls to CCMS - GetDocumentID and DocumentUpload were not extracting error information from the response XML and storing in the `CCMS::SubmissionHistory` record.

This ticket fixes that problem.

Note that the xpath to get the status out of the XML is different for these two calls compared with the other calls, so the standard path is defined in the base response parser, and the modified path in these two classes.

## Checklist

### DIFF
--- a/app/models/ccms/submission.rb
+++ b/app/models/ccms/submission.rb
@@ -12,7 +12,7 @@ module CCMS
       ActiveSupport::Notifications.instrument 'dashboard.ccms_submission_saved', id: id, state: aasm_state
     end
 
-    POLL_LIMIT = 10
+    POLL_LIMIT = Rails.env.development? ? 99 : 10
 
     def process!(options = {}) # rubocop:disable Metrics/MethodLength
       case aasm_state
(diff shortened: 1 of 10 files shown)


app/models/ccms/submission.rb
```diff
@@ -27,6 +27,7 @@ def process!(options = {}) # rubocop:disable Metrics/MethodLength
       when 'document_ids_obtained'
         CCMS::Submitters::AddCaseService.call(self, options)
       when 'case_submitted'
+        puts ">>>>>>>>>>>>  #{__FILE__}:#{__LINE__} <<<<<<<<<<<<\n"
```
willc-work: assume this is left in from testing

app/services/ccms/submitters/check_case_status_service.rb
```diff
@@ -2,6 +2,7 @@ module CCMS
   module Submitters
     class CheckCaseStatusService < BaseSubmissionService
       def call
+        puts ">>>>>>>>>>>> calling CheckCaseStatusService #{__FILE__}:#{__LINE__} <<<<<<<<<<<<\n"
```
willc-work: left over from testing?

PR #1389, opened 2020-04-12
AP-1242 non passported means results header blocks

stephenrichards:
## Display appropriate header blocks on non-passported means assessment results page

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1242)

- New controller `CapitalIncomeAssessmentResultsController` to display results
- Changed `#overview` method in `CFE::Result` to always use `CCMS::ManualReviewDeterminer` to determine whether or not a manual review is required
- Updated `CCMS::ManualReviewDeterminer` to cater for non-passported applications
- Added partials and translations for `CapitalIncomeAssessmentResultsController` 
- Added new set of mock true layer bank transactions
- Added a new field on `Setting` to say which set of bank transactions to use
- Added new test data for `CFE::V2::Result` in factories
- Added test for `CFE::V2::Result` in factories to ensure test data being set up correctly

## Checklist


app/models/cfe/v2/result.rb
```diff
@@ -7,16 +7,32 @@ def assessment_result
         result_hash[:assessment][:assessment_result]
       end
 
+      def capital_assessment_result
+        capital[:assessment_result]
+      end
+
+      def capital_contribution_required?
```
willc-work, line 14: Could this method use the one above?
`def capital_contribution_required?
    capital_assessment_result  == 'contribution_required'
 end`

app/models/cfe/v2/result.rb
```diff
@@ -7,16 +7,32 @@ def assessment_result
...
-        result_hash[:assessment][:capital]
+      def income_assessment_result
+        disposable_income[:assessment_result]
       end
 
-      def property
-        capital[:capital_items][:properties]
+      def income_contribution_required?
+        disposable_income[:assessment_result] == 'contribution_required'
+      end
```
willc-work, line 28: Same as other comment, could this method use ```income_assessment_result``` from the method above?

config/locales/en/providers.yml
```diff
@@ -106,6 +109,47 @@ en:
...
+        detail: 'This is because they have too much disposable capital.'
+      capital_contribution_required:
+        heading: '%{name} may need to pay towards legal aid'
+        details:
+          - 'We’ve calculated that your client should pay %{amount} from their disposable capital.'
+          - 'You still have to provide details of the case before we can grant legal aid.'
+      capital_and_income_contribution_required:
+        heading: '%{name} may need to pay towards legal aid'
+        line_1: "We've calculated that your client should pay:"
+        line_2: You still need to provide details of the case before we can grant legal aid.
```
willc-work, line 137: this line says ```You still need to...``` other lines use ```You still have to ...```

app/models/cfe/v2/result.rb
stephenrichards: Good catch

PR #1395, opened 2020-04-16
Ap 1350 fix feedback

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1350)

Feedback could be submitted without a response for 'satisfaction; which subsequently broke the geckoboard stats. This adds validation so that both difficulty and satisfaction now require responses before it can be submitted

## Checklist


config/locales/en/feedback.yml
```diff
@@ -1,11 +1,13 @@
 ---
 en:
   feedback:
+    errors:
+      attributes:
+        satisfaction:
+          blank: this is my meeesage
```
Obsiye, line 7: is this leftover code?

willc-work: yes, yes it is 🤦‍♂️

PR #1420, opened 2020-04-29
AP1340 Validation for categorising bank transactions

Obsiye:
AP1340 Validation for categorising bank transactions

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1340)

The BankTransaction model belongs_to a TransactionType, optionally (optional: true). Therefore, on the income summary page, you could have transaction_types on the model such as '1. Benefits' but no bank transactions were added or are associated with it. I added validation on the legal aid application model to check that those transaction types on the model have one or more bank transactions. If not then add relevant errors.

**I've listed each step in the commit messages instead of overdoing this PR message.** Please have a look at them [here](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/1420/commits).

Challenges:

- I was finding it difficult to create a form that wasn't disturbing the routing paths. I would get 'page not found' a lot. Instead, I added this validation into the legal_aid_application model.

- Testing the feature tests was hacky. The categorising bank transactions process wasn't in it so I added it.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -133,9 +133,9 @@ class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLen
   end
 
   def uncategorised_income_transactions?
-    @bank_transactions = bank_transactions.credit.order(happened_at: :desc).by_type
+    grouped_transactions = bank_transactions.credit.order(happened_at: :desc).by_type
     transaction_types.credits.each do |transaction_type|
-      uncategorised_transactions_errors(transaction_type.name) if @bank_transactions[transaction_type].blank?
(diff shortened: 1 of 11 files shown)


app/models/legal_aid_application.rb
```diff
@@ -132,6 +132,14 @@ def set_transaction_period
     )
   end
 
+  def uncategorised_income_transactions?
+    @bank_transactions = bank_transactions.credit.order(happened_at: :desc).by_type
+    transaction_types.credits.each do |transaction_type|
```
stephenrichards, line 137: Why the instance variable here?  It isn't used outside this method is it?  Would prefer to see a local variable called `grouped_transactions` or something similar.

PR #1429, opened 2020-05-05
AP1385 Bugfix Incorrect result content in staging

Obsiye:
AP1385 Bugfix Incorrect result content in staging

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1385)

### First bug fix
- Set the default value of the column manually_review_all_cases in the Settings table to **false**.

This allows the application to go through the CFE result and it returns valid assessment results. Before, it was set to true and thus required a case review no matter what.

## Checklist


app/models/concerns/legal_aid_application_state_machine.rb
```diff
@@ -48,6 +48,7 @@ def summary_state
         transitions from: :delegated_functions_used, to: :checking_passported_answers
         transitions from: :client_details_answers_checked, to: :checking_passported_answers
         transitions from: :provider_assessing_means, to: :checking_passported_answers
+        transitions from: :provider_checked_citizens_means_answers, to: :provider_checking_citizens_means_answers
```
willc-work, line 51: Should this transition from provider_checking_citizens_means_answers to provider_checked_citizens_means_answers, instead of the other way around as it is here?

Obsiye: Thanks for noticing! We're going to the check_passported_answers page instead of the means_summaries page. I'm on the fix now.

Obsiye: Removed now thanks.

PR #1445, opened 2020-05-14
AP-1424 Add tranche 6 users

stephenrichards:
## Add tranche 6 users

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1424)

- Updated `tranche:maker` rake task to cope with users with spaces in their logins.
- updated config files with new users

## Checklist


lib/tasks/tranche_maker.rake
```diff
@@ -13,17 +13,18 @@ namespace :tranche do
...
       uri = URI("https://ccms-pda.legalservices.gov.uk/api/providerDetails/#{encoded_user}")
+      puts uri.to_s
       response = Net::HTTP.get(uri)
       hash = JSON.parse(response)
       if hash.key?('error')
         puts ">>>>>>>> User #{user} / #{encoded_user} not found <<<<<<<<<"
       else
-        contact_hash = hash['contacts'].detect { |h| h['name'] == encoded_user.upcase }
-        whitelisted_users << encoded_user
+        contact_hash = hash['contacts'].detect { |h| h['name'] == user.upcase }
```
colinbruce, line 24: I can't see how this `==` works?
On line 16 we find spaces and replace them with the encoed space `%20` is the name value in contacts with a normal, unencoded, space? 😕

lib/tasks/tranche_maker.rake
```diff
@@ -13,17 +13,18 @@ namespace :tranche do
...
+        whitelisted_users << user
```
colinbruce, line 25: Does this mean that a user with a space in their name will be in the whitelist as 'TEST USER' but will be in the portal as TEST%20USER?

Will they match on login?

lib/tasks/tranche_maker.rake
stephenrichards: If a user has a space in their username, then the url has to be encoded with %20.  Then in the payload, we look for the username in upper case with a space, not %20 in the list of contacts to get their contact Id.

lib/tasks/tranche_maker.rake
stephenrichards: Exactly.  Take a look at the user that was causing all this trouble: RUFIA ADALAT.

He or she logs in apparently as `RUFIA ADELAT`.  We have to look him/her up as `RUFIA%20ADELAT`.  The name in the list of contacts is retrurned as `RUFIA ADELAT`.  That all makes sense actually if you're dumb enough to allow spaces in user names.

PR #1447, opened 2020-05-15
Ap 1384 select disregard benefits

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1384)

Removes the label from the transactions selected as relating to disregarded benefits.

Adds in a partial to display on the transaction page to show which benefits are disregarded, it calls CFE to get this list

## Checklist


app/controllers/providers/transactions_controller.rb
```diff
@@ -41,6 +42,14 @@ def bank_transactions
         .order(happened_at: :desc, description: :desc)
     end
 
+    def disregarded_state_benefits_list
+      benefits_list = CFE::ObtainStateBenefitTypesService.call
```
Obsiye, line 46: Since this "CFE::ObtainStateBenefitTypesService.call" does a HTTP request to CFE, we should rescue its errors and use Raven.capture_exception.

app/controllers/providers/transactions_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Providers
   class TransactionsController < ProviderBaseController
     def show
+      disregarded_state_benefits_list
```
Obsiye, line 4: Do something similar to what we already do and `redirect_to(problem_index_path) && return unless disregarded_state_benefits_list`

PR #1483, opened 2020-06-02
AP1115 Fix screen reader issues with multiple H1s on the Respondants page

Obsiye:
AP1115 Fix screen reader issues with multiple H1s on the Respondants page

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1115)

- Allow the use of a h2 tag for govuk_collection_radio_buttons
- Add htag param to fieldset_legend method on form builder
- Update the respondent page to use this instead
- Update tests

**Another bugfix:**

- Fix voiceover bug with using class 'govuk-visually-hidden' because the screen reader reads whatever is inside it aloud even if it's hidden. Instead use class 'govuk-radios__conditional--hidden' for this, which hides it from the screen reader too if it's not showing on the page.

e.g. visually hidden (visually hides) but doesn't hide from the screenreaders/the ear.

Co-authored-by: @stephenrichards 

## Checklist

### DIFF
--- a/app/lib/govuk_elements_form_builder/form_builder.rb
+++ b/app/lib/govuk_elements_form_builder/form_builder.rb
@@ -95,8 +95,8 @@ module GovukElementsFormBuilder
     # You can pass an error with the :error parameter.
     # e.g., <%= form.govuk_collection_radio_buttons(:gender, ['f', 'm'], error: 'Please select a gender') %>
     #
-    # If you wish to specify the size of the heading pass a hash into title with text and size:
-    # <%= form.govuk_collection_radio_buttons(:gender, ['f', 'm'], title: { text: 'What is your gender?', size: :m } ) %>
+    # If you wish to specify the size of the heading and/or which header tag to use, pass a hash into title with text and size:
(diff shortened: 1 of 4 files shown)


app/lib/govuk_elements_form_builder/form_builder.rb
```diff
@@ -95,8 +95,8 @@ def govuk_radio_button(attribute, value, *args)
     # You can pass an error with the :error parameter.
     # e.g., <%= form.govuk_collection_radio_buttons(:gender, ['f', 'm'], error: 'Please select a gender') %>
     #
-    # If you wish to specify the size of the heading pass a hash into title with text and size:
-    # <%= form.govuk_collection_radio_buttons(:gender, ['f', 'm'], title: { text: 'What is your gender?', size: :m } ) %>
+    # If you wish to specify the size of the heading and/or which header tag to use, pass a hash into title with text and size:
+    # <%= form.govuk_collection_radio_buttons(:gender, ['f', 'm'], title: { text: 'What is your gender?', size: :m, htag: :h2 } ) %>
```
stephenrichards, line 99: Actually this is my bad - but perhaps add that the default if no htag is supplied is h1

PR #1492, opened 2020-06-04
AP-1442 Don't send reminder mails when using CCMS

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1442)

- Added new base mailer class with class method `.eligible_for_delivery?`.  This enables scheduled mailing to check that the mail still needs to be sent at the time it is due to be sent.  This is better than deleting mails earlier in the process, because the user may go back and change answers that mean the mail stilll needs to be sent, for example, a non-passported result from the DWP check on a delegated functions case means the submission reminder mails do not need to be sent.  If they were deleted at that point, and the user went back, changed some details and then got a positive DWP result, those mails wouldn't exist.  This way, the mails stay scheduled, and at the time of delivery, we can call `.eligible_for_delivery?` on the mailer class and delete them at that point if the response is false.
- set state to `use_ccms` if the SWP check fails and the setting to handle non-passported case is false
- set state to `use_ccms` if the applicant doesn't agree to use open banking
- Write an '.eligible_for_delivery?` method for SubmitApplicationReminderMailer to ensure that the application is not in a state where the application has either been abandoned or submitted.
- removed the cancellation of the submission reminder emails from the PostSubmissionProcessingJob - the changes in ScheduledMailing described above means this is no longer needed.

## Checklist


app/models/concerns/legal_aid_application_state_machine.rb
```diff
@@ -127,6 +130,10 @@ def provider_checking_or_checked_citizens_means_answers?
       event :submitted_assessment do
         transitions from: :submitting_assessment, to: :assessment_submitted
       end
+
+      event :reset_from_use_ccms do
+        transitions from: :use_ccms, to: :initiated
```
colinbruce, line 135: My only concern is around this bit at the moment...
We transition _to_ use_ccms from a number of states, see line 68 above for another example.
Block reversion to initiated may cause issues with other states.

app/mailers/submit_application_reminder_mailer.rb
```diff
@@ -1,8 +1,14 @@
-class SubmitApplicationReminderMailer < GovukNotifyRails::Mailer
+class SubmitApplicationReminderMailer < BaseApplyMailer
   self.delivery_job = GovukNotifyMailerJob
   require_relative 'concerns/notify_template_methods'
   include NotifyTemplateMethods
 
+  INELIGIBLE_STATES = %w[use_ccms generating_reports submitting_assessment assessment_submitted].freeze
```
Obsiye, line 6: generating_reports makes use of PostSubmissionProcessingJob to cancel its jobs so is there a duplication here?

app/models/concerns/legal_aid_application_state_machine.rb
colinbruce: No, that's cool :+1:

PR #1508, opened 2020-06-11
AP-1446 Flow around offline bank accounts

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1446)

When the client arrives on the 'citizens/additional_accounts' page, the state is reverted to 'provider_submitted' and the boolean 'has_offline_accounts' is set to nil. This requires the client to make a selection before leaving the page (particularly relevant on a re-visit, as currently the initial response is stored and not changed). This allows the user to change their mind about additional accounts at any point in the subflow and to use the back buttons to amend their answers.

### DIFF
--- a/app/controllers/citizens/additional_accounts_controller.rb
+++ b/app/controllers/citizens/additional_accounts_controller.rb
@@ -1,6 +1,9 @@
 module Citizens
   class AdditionalAccountsController < CitizenBaseController
-    def index; end
+    def index
+      legal_aid_application.update!(has_offline_accounts: nil)
+      legal_aid_application.update!(state: 'provider_submitted')
+    end
 
     def create
       case params[:additional_account]
--- a/app/controllers/citizens/additional_accounts_controller.rb
+++ b/app/controllers/citizens/additional_accounts_controller.rb
@@ -2,7 +2,7 @@ module Citizens
   class AdditionalAccountsController < CitizenBaseController
     def index
       legal_aid_application.update!(has_offline_accounts: nil)
-      legal_aid_application.update!(state: 'provider_submitted')
+      legal_aid_application.provider_submit! unless legal_aid_application.provider_submitted?
     end
 
     def create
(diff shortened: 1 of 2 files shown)


app/controllers/citizens/additional_accounts_controller.rb
```diff
@@ -1,6 +1,9 @@
 module Citizens
   class AdditionalAccountsController < CitizenBaseController
-    def index; end
+    def index
+      legal_aid_application.update!(has_offline_accounts: nil)
+      legal_aid_application.update!(state: 'provider_submitted')
```
stephenrichards, line 5: WE shouldn't be changing the state like this - we should use a transition in the state machine to do it, that way we can keep control of what state changes are legal.

You may need to create a new transition inside the `provider_submitted` event

spec/requests/citizens/additional_accounts_spec.rb
```diff
@@ -16,6 +16,19 @@
     end
   end
 
+  context 'when an applicant revisits the page to change their answer' do
+    before do
+      application.update!(has_offline_accounts: true)
+      application.update!(state: 'use_ccms')
+    end
```
stephenrichards, line 23: it's OK to update the state directly in tests if there is no other way.  this could be combined into one statement, e.g. 
`application.update!(has_offline_accounts: true, state: 'use_ccms')`

spec/requests/citizens/additional_accounts_spec.rb
```diff
@@ -16,6 +16,19 @@
...
+
+    it 'checks that offline account is reset to nil' do
+      get citizens_additional_accounts_path
+      expect(application.reload.has_offline_accounts).to be_nil
+      expect(application.reload.state).to eq 'provider_submitted'
+    end
```
stephenrichards, line 29: You don't need the second reload here - it was reloaded after the controller action in line 27.

PR #1509, opened 2020-06-11
AP-1474 create new irregular payment table and model

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1471)

Added a new table with and a model with validation, linking it to legal_aid_applications

## Checklist

### DIFF
--- a/spec/models/irregular_income_spec.rb
+++ b/spec/models/irregular_income_spec.rb
@@ -1,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe IrregularIncome, type: :model do
+  let(:application) { create :legal_aid_application }
+
+  describe 'on validation' do
+    subject { described_class.new }
+
+    before do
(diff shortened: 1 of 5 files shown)


spec/models/irregular_income_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    it 'is valid with all valid attributes' do
+      expect(subject).to be_valid
+    end
+
+    context 'invalid attributes' do
+      before do
+        subject.income_type = 'medical_loan'
+        subject.frequency = 'weekly'
+      end
```
stephenrichards, line 23: if you substitute these values with 'xxxstudent_loanzz' and 'annually', it will still pass validation.

willc-work: added start/end of string characters

PR #1511, opened 2020-06-12
Ap 1466 student finance cya

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1466)

Added a `student_finance` module to the check your answers page. If the client has student finance it will also show the value, otherwise this section is hidden. Created a factory for irregular incomes.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -266,12 +266,11 @@ class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLen
   end
 
   def receives_student_finance?
-    irregular_incomes.map{|ir| ir.income_type}.include?'student_loan'
+    irregular_incomes.map(&:income_type).include? 'student_loan'
   end
 
   def value_of_student_finance
-    irregular_incomes[0].amount
-    # irregular_incomes(&:amount).where(income_type: 'student_loan')
+    irregular_incomes.map { |ir| ir.income_type == 'student_loan' ? ir.amount : nil }.first
   end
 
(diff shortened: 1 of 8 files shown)


app/views/citizens/check_answers/index.html.erb
```diff
@@ -51,6 +51,8 @@
           ),
           read_only: false) %>
 
+  <%= render 'student_finance' %>
```
stephenrichards, line 54: We only want to do this if the feature flag on the settings table is set to use new student loan - otherwise we'll get a no to the answer for "do you get student finance?" even though it's specified as part of other income.

app/models/legal_aid_application.rb
```diff
@@ -264,6 +264,14 @@ def reset_proceeding_types!
     reset_delegated_functions
   end
 
+  def receives_student_finance?
+    irregular_incomes.map(&:income_type).include? 'student_loan'
+  end
+
+  def value_of_student_finance
+    irregular_incomes.map { |ir| ir.income_type == 'student_loan' ? ir.amount : nil }.first
```
stephenrichards, line 272: I think the responsibility for extracting student loan records from all the other types of irregular income properly lies with the `IrregularIncome` model with a scope

`scope :student_finance, -> { where(income_type: 'student_loan) }`

then these two methods become:

```
def receives_student_finance?
  irregular_incomes.student_finance.any?
end

def value_of_student_finance
  receives_student_finance? ? irregular_incomes.student_student_finance.first.amount : nil
end
```

PR #1518, opened 2020-06-15
AP-1486: true layer date formatting

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1486)

TrueLayer have confirmed that it is because we are asking for transactions in the future, i.e. our end date isn't in the past.

Reverted the calls to truelayer to be `DateTime` instead of simply `Date`
According to #771 
> because TrueLayer's filter is only granular to the date anyway, not the time
This reverts the `to_date` to the beginning of the chosen day, either today or the delegated_permission date, depending.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -129,8 +129,8 @@ class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLen
     date_from = date_to - 3.months
 
     update!(
-      transaction_period_start_on: date_from,
-      transaction_period_finish_on: date_to
+      transaction_period_start_on: date_from.beginning_of_day,
+      transaction_period_finish_on: date_to.beginning_of_day
     )
   end
 
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -129,8 +129,8 @@ class LegalAidApplication < ApplicationRecord # rubocop:disable Metrics/ClassLen
     date_from = date_to - 3.months
 
     update!(
-      transaction_period_start_on: date_from.beginning_of_day,
-      transaction_period_finish_on: date_to.beginning_of_day
+      transaction_period_start_on: date_from,
+      transaction_period_finish_on: date_to
     )
   end
 
--- a/app/services/true_layer/api_client.rb
+++ b/app/services/true_layer/api_client.rb
@@ -21,8 +21,8 @@ module TrueLayer
 
     def transactions(account_id, date_from, date_to)
       params = {
-        from: date_from.to_datetime.iso8601,
-        to: date_to.to_datetime.iso8601
+        from: date_from.to_datetime.utc.iso8601,
+        to: date_to.to_datetime.utc.iso8601
       }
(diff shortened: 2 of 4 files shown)


app/models/legal_aid_application.rb
```diff
@@ -129,8 +129,8 @@ def set_transaction_period
     date_from = date_to - 3.months
 
     update!(
-      transaction_period_start_on: date_from,
-      transaction_period_finish_on: date_to
+      transaction_period_start_on: date_from.beginning_of_day,
+      transaction_period_finish_on: date_to.beginning_of_day
     )
```
stephenrichards, line 134: This won't have any effect because the `transaction_period_start_on` and `transaction_period_finish_on` columns are defined as dates, not datetimes.

app/services/true_layer/api_client.rb
```diff
@@ -21,8 +21,8 @@ def accounts
 
     def transactions(account_id, date_from, date_to)
       params = {
-        from: date_from.to_date.iso8601,
-        to: date_to.to_date.iso8601
+        from: date_from.to_datetime.iso8601,
+        to: date_to.to_datetime.iso8601
       }
```
stephenrichards, line 26: This will give a string like `2020-06-16T00:00:00+01:00`, whereas the TrueLayer documentation referenced in the ticket asks for `2018-02-18T00:00:00`.  I was actually wondering whether or not we needed to take timezones into account - but I guess the answer is no - TrueLayer always assumes London timezone with BST where applicable.

spec/services/true_layer/importers/import_transactions_service_spec.rb
```diff
@@ -57,8 +57,8 @@
 
         expect(WebMock).to have_requested(:get, expected_request_url)
           .with(query: {
-                  from: now_minus_3_month.iso8601,
-                  to: now.iso8601
+                  from: now_minus_3_month.beginning_of_day.iso8601,
+                  to: now.beginning_of_day.iso8601
                 })
```
stephenrichards, line 62: As we have fixed dates  in the test, can we test here for the actual strings - then that would be a test that the right formatting is taking place

app/models/legal_aid_application.rb
colinbruce: I have reverted this change... we only need to store the date and send a datetime to true-layer
As the time element is always defaulted to 00:00:00 storing it
seems superfluous

app/services/true_layer/api_client.rb
colinbruce: I have changed the date procuded to `.to_datetime.utc` this results in `yyyy-mm-ddThh:mm:ssZ` without additional timezone data.
It doesn't specifically match the true layer documentation but they return values suffixed with `Z` so I have made the assumption that they will accept them on the inputs.
I have tested this against the TrueLayer API and it successfully returns transaction data.

spec/services/true_layer/importers/import_transactions_service_spec.rb
colinbruce: I've updated the tests to ensure that the correct format is being produced and sent to true_layer

PR #1529, opened 2020-06-18
AP1485 BugFix - Return false after a Raven.capture_exception

Obsiye:
AP1485 BugFix - Return false after a Raven.capture_exception

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1485)

Return false after a Raven.capture_exception this is because our app gets the result and tries to dig into the hash result
but as it's a Raven event it raises another sentry error of "undefined method `dig' for #<Raven::Event"

Update tests for catching Savon::SOAPFault error and a StandardError

## Checklist

### DIFF
--- a/spec/services/benefit_check_service_spec.rb
+++ b/spec/services/benefit_check_service_spec.rb
@@ -56,11 +56,34 @@ RSpec.describe BenefitCheckService do
       end
     end
 
-    context 'when the API raises an error', vcr: { cassette_name: 'benefit_check_service/service_error' } do
+    context 'calling API raises a Savon::SOAPFault error', vcr: { cassette_name: 'benefit_check_service/service_error' } do
       let(:last_name) { 'SERVICEEXCEPTION' }
 
       it 'captures error' do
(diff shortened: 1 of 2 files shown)


spec/services/benefit_check_service_spec.rb
```diff
@@ -56,7 +56,7 @@
       end
     end
 
-    context 'when the API raises an error', vcr: { cassette_name: 'benefit_check_service/service_error' } do
+    context 'calling API raises a Savon::SOAPFault error', vcr: { cassette_name: 'benefit_check_service/service_error' } do
```
heidar, line 59: Curious how the `expect(subject.call).to eq false` in this context was passing before 🤔

Obsiye: Great Spot! I'm trying to find out now.

Obsiye: In the test / dev environment, we don't send to Sentry then Raven returns false. In production we send to sentry and so Raven returns its raven event object which lead to the sentry issue.

PR #1539, opened 2020-06-23
BugFix Run JavaScript tests in CircleCI as part of unit tests job

Obsiye:
BugFix Run JavaScript tests in CircleCI as part of unit tests job

The problem was that JavaScript tests weren't running as part of CircleCI.

- Reuse the javascript_tests rake command to run our javascript tests in circle ci job called unit_tests

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -172,6 +172,9 @@ jobs:
     - *install_js_packages
     - *save_js_packages_cache
     - run:
+        name: Run JavaScript tests
+        command: bin/rails javascript_tests
+    - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -171,6 +171,9 @@ jobs:
     - *restore_js_packages_cache
     - *install_js_packages
     - *save_js_packages_cache
+    - run:
```
colinbruce, line 174: I was going to suggest putting it in its own job but then I saw how few tests it is! :)

PR #1557, opened 2020-07-01
AP-1225: Improve dependant handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1225)

Refactor the dependant handling, remove the legacy code: forms, controllers, routes, tests, etc
Replace with a single DependantForm containing all of the previous validations and values.
Replace the controllers, update the views with the new designs
Extend the feature tests for the new routing, including return to check your answers after updating a dependant

 ## Checklist


app/views/providers/has_dependants/show.html.erb
```diff
@@ -1,6 +1,67 @@
...
+  <% if @legal_aid_application.has_dependants? %>
+    <%= govuk_fieldset_header(size: 'm') { t('.existing', count: "#{pluralize(@legal_aid_application.dependants.count, 'dependant')}")} %>
+    <div class="govuk-summary-list">
+      <% @legal_aid_application.dependants.each do |dependant| %>
+        <div class="govuk-summary-list__row">
+          <dd class="govuk-summary-list__value"><%= dependant.name %></dd>
+          <dd class="govuk-summary-list__actions">
+            <a class="govuk-link" href="<%= providers_legal_aid_application_dependant_path(@legal_aid_application, dependant) %>">Change<span class="govuk-visually-hidden"> name</span></a></dd>
+          <dd class="govuk-summary-list__actions">
+            <a class="govuk-link" href="remove_2">
```
Obsiye, line 11: Just curious what 'remove_2' in the href does?

colinbruce: Good catch, I thought I'd deleted this entire block!😢  
There should never be any dependants when viewing this page so it was moved to the has_other_dependants show :+1:

app/views/shared/check_answers/_dependants.html.erb
```diff
@@ -1,66 +1,70 @@
...
               question: t('providers.means_summaries.show.dependants.date_of_birth'),
               answer: dependant.date_of_birth,
               read_only: true,
-              align_right: false
+              align_right: true
           ) %>
       <%= check_answer_link(
               name: :dependant,
-              url: check_answer_url_for(:providers, :dependants_relationships, @legal_aid_application, dependant_id: dependant),
+              url: nil,
```
Obsiye, line 32: All looks good to me. Two clean up comments below:

1. Instead of having all the 'nil' values for the urls, could you use the #check_answer_no_link from the CheckAnswersHelper.

2. Also, on line 62 there's a comment "" i'm assuming that's no longer needed

colinbruce: Cheers @Obsiye I have removed the comment. You were correct, it was no longer needed! 😄 
I have completely re-written the view, it was always in the back of my mind but I got distracted, could you have a look at the subsequent commits and let me have your feedback?

app/views/shared/check_answers/_dependants.html.erb
```diff
@@ -1,66 +1,16 @@
...
-              align_right: false
-          ) %>
-    </dl>
+  <% @legal_aid_application.dependants.each_with_index do |dependant, index| %>
+    <%= check_answer_one_change_link(
+      name: "dependants_#{index + 1}",
+      url: providers_legal_aid_application_dependant_path(@legal_aid_application, dependant),
+      question: "Dependant #{index + 1}",
+      answer_hash: dependant_hash(dependant),
+      read_only: FALSE
```
Obsiye, line 8: I like what's done here - very clean. One thing, `FALSE` is a deprecated constant. Prefer `false`

colinbruce: 🤦 Stupid CAPSLOCK key! :)

PR #1562, opened 2020-07-03
Ap 717 accessibility

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-717)

Adds aria-labels for screen readers, removed label and replaces it with a "<p>"  to improve the experience for screen reader users.

The red items on the ticket all appear to have been addressed previously

For displaying the postcode with a space, I have just added the space in the view rather than change how we save or store the postcode. This was we uniformly show all postcodes in their correct format (which is with the space before the last 3 characters). It seemed to me a helper method for this bit of display was overkill but happy to discuss if people feel this sort of logic should not be in the view.

## Checklist


app/controllers/providers/address_selections_controller.rb
```diff
@@ -20,8 +20,18 @@ def update
...
+      string
+    end
+    helper_method :add_a_space_to_postcode
+
     private
 
+    # def add_a_space_to_postcode(string)
+    #   string.(insert(-3), " ")
+    # end
+
```
mpw5, line 34: Remove these comments?

app/views/providers/address_selections/show.html.erb
```diff
@@ -14,17 +14,17 @@
     <%= govuk_form_group show_error_if: @form.errors[:lookup_id].present? do %>
       <%= govuk_fieldset_header page_title %>
 
-      <div class="govuk-label">
+      <p class="govuk-label">
         <%= t('generic.address.postcode') %>
         <p class='govuk-body govuk-!-font-weight-bold'>
-          <%= @form.postcode %>
+          <%= @form.postcode.insert(-4, " ") %>
```
mpw5, line 20: You're inserting a spce both here and in the controller; is there any way of combining the two?

Also in the controller you do `insert(-3), ' '`, and here it's `insert(-4, " "` - shouldn't the space be inserted in the same place in both?

app/controllers/providers/address_selections_controller.rb
willc-work: apologies, I had not pushed my most recent changes, these are removed

PR #1292, opened 2020-03-02
Update ruby to 2.7.1

colinbruce:
## What
Update ruby to 2.7.1

### Dev instructions for installing
* Install ruby-2.7.1 `rvm install 2.7.1` (other ruby package managers are available)
* Checkout the new master
* Re-run `bundle install`

### Update - 17 Sep 2020
<details>
<summary>Second review</summary>

* Wasabi has published a 2.7 compatible version so that has now been updated using `bundle update --source wasabi` so it isn''t needed in the Gemfile
* Simplecommand has been updated but not published so that is being loaded from github until it gets pushed to rubygems

This makes rspec and cucumber look okay (there is a single deprecation waring regarding railtie whenever rails is initialised)
I will expand circle to confirm my suspicions, but I expect `yarn test` and `yarn hint` to cause issues next!
</details>

### Update - 8 Jul 2020
<details>
<summary>First review</summary>

There are ~90 occurrences of the `Wasabi` warning block
```
/Users/colinbruce/.rvm/gems/ruby-2.7.0/gems/wasabi-3.5.0/lib/wasabi/parser.rb:89: warning: URI.unescape is obsolete
/Users/colinbruce/.rvm/gems/ruby-2.7.0/gems/wasabi-3.5.0/lib/wasabi/parser.rb:90: warning: URI.escape is obsolete
```
This is being addressed in the wasabi gem https://github.com/savonrb/wasabi/pull/75 which is ready to merge

There are multiple occurrences of warnings about simple command this has been addressed in https://github.com/nebulab/simple_command/pull/28 but the repo doesn't seem to be regularly maintained 😞 

Otherwise, as at this date, the following warnings still occur that make the rspec output difficult to read
<details>
<summary>Click to expand full warning list</summary>

```ruby
/Users/colinbruce/Projects/laa-apply/app/helpers/page_template_helper.rb:53: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/colinbruce/Projects/laa-apply/app/helpers/application_helper.rb:29: warning: The called method `back_link' is defined here

/Users/colinbruce/Projects/laa-apply/app/views/shared/page_templates/_default_page_template.html.erb:15: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/colinbruce/Projects/laa-apply/app/helpers/page_template_helper.rb:69: warning: The called method `page_heading' is defined here

[...]

PR #1569, opened 2020-07-08
Tech-debt: Prep work for ruby 2.7.x update

colinbruce:
## What

Ruby 2.7 has been out since Christmas 2019, however it introduced a number of significant changes to the reporting of deprecated changes.  Due to those changes the tests are virtually unreadable due to the level of alerts that occur

The system itself works in Ruby 2.7 and there is a separate PR to update ruby itself (#1292), I would recommend not switching yet as there are two gems in use that have PRs awaiting merge to remove their deprecation warnings.

Making these changes now reduces the footprint of #1292 as, and when, we can merge it

## Checklist


.circleci/config.yml
```diff
@@ -65,7 +65,9 @@ references:
   install_js_packages: &install_js_packages
     run:
       name: Install Yarn packages
-      command: yarn --frozen-lockfile
+      command: |
+        npm install
```
Obsiye, line 69: We use yarn so it'd be `yarn install` . Just curious to know what issue this npm install command resolved?

Obsiye: Ah gotya 👍 . 
How about trying the below (i'm still learning front-end stuff but the below should help)
```suggestion
    yarn install --frozen-lockfile
```

colinbruce: Cheers. I'll try removing the commit from this branch anyway as I don't think we need it for the ruby 2.6  deploys on circle and I'll play with it on #1292 :+1:

PR #1583, opened 2020-07-13
AP-1505 Pundit policy to restrict provider permissions

stephenrichards:
## Restrict permissions to manage passported/non-passported applications

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1505)

Now that we have permission set up for each user, we want to restrict access to passported applications to only those providers that have that permission, and ditto for non-passported.  All providers have access to applications which have not yet got to the benefit check stage.

All of these new permission are within the existing restriction that providers can only see applications submitted by themselves, or another provider from the same firm.

The passported/non-passported checks are not carried out if the controller is one of those that is used before the DWP check.  This is done by inlcuding the `PreDWPCheckVisible` concern into those  controllers from which will respond true to the `#pre_dwp_check?` method, and can be tested by the policy (simply checking if the application has a benefit check result does not work because it does not allow the user to go back after getting a negative result if they don't have the right permissions).

### Describe what you did and why.

* Added two methods `#passported_permissions?` and `#non_passported_permissions?`  to the `Provider` model
* Created a new concern `Providers::PreDWPCheckvisible` which always returns true to `#pre_dwp_check?`
* Changed all controllers up to and including `Providers::CheckBenefitsController` to be include the `Providers::PreDWPCheckVisible` concern
* Modified the `Authorizeable` module to instantiate the `LegalAidApplicationPolicy` with an `AuthorizationContext` object (which comprises just the provider and the controller) instead of just the provider
* Renamed the `.authorize_with_policy` method to `.authorize_with_policy_method` in order to better describe what it actually does.
* Added method to `LegalAidApplicationPolicy` to check that the provider had the correct permissions if the controller does not respond `true` to `#pre_dwp_check?`
* Updated the provider factory to be able to create providers with different permissions

## Checklist


app/controllers/providers/address_lookups_controller.rb
```diff
@@ -1,5 +1,5 @@
 module Providers
-  class AddressLookupsController < ProviderBaseController
+  class AddressLookupsController < PreDWPCheckProviderBaseController
```
Obsiye, line 2: Instead of having to change the class name, we could have it in the application_dependable.rb just like legal_aid_application_not_required! and then have that at the top of the controller like pre_dwp_check! 

it'll look like this in the end:
```
module Providers
  class AddressLookupsController < ProviderBaseController
    pre_dwp_check!
```

I did the above by coding the below:

```

    class_methods do
      def legal_aid_application_not_required!
        @legal_aid_application_not_required = true
      end

      def legal_aid_application_not_required?
        @legal_aid_application_not_required
      end

      def pre_dwp_check!
        @pre_dwp_check = true
      end

      def pre_dwp_check?
        @pre_dwp_check
      end
    end
```

Then in the LegalAidApplicationPolicy have:
( where klass = authorization_context.klass in ApplicationPolicy and klass = klass.class in AuthorizationContext )
 ```
 def authorized_passported_permissions?
    return true if @klass.pre_dwp_check?

    record.passported? ? provider.passported_permissions? : provider.non_passported_permissions?
  end
```

Obsiye: What's here works well. I was just apprehensive about having to change the class name because of DWP. It feels odd to me that's why. It feels more entangled. Your thoughts?

stephenrichards: That's not a bad idea! Let me have a play around with it tomorrow.

PR #1588, opened 2020-07-16
AP-1518 Gracefully handle True Layer auth errors

stephenrichards:
## Gracefully handle True Layer auth errors

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1518)

The applicant authentication at the bank is handled by OmniAuth through True Layer.  If the applicant fails to authenticate, then the default action in development mode is to raise an exception, and in production mode to redirect the callback to `/auth/failure`.  This controller action handles failed provider logins to the Apply system as well as failed applicant bank logins..

### What we did:
- added a configuration directive to the OmniAuth initializer config to keep the same behaviour as in development as in production,  (i.e. to redirect to `/auth/failure`)
- in the `AuthController#failure` action, detect those requests which are as a result of a TrueLayer auth fail (the origin parameter will be `/citizens/banks`) and redirect them to `/citizens/consents` with a parameter to indicate a fail
- display the error on the citizen consents page.

## Checklist

### DIFF
--- a/config/locales/en/citizens.yml
+++ b/config/locales/en/citizens.yml
@@ -59,6 +59,7 @@ en:
         title: Do you agree to share your bank account information with the LAA?
         inset_text: We only use your financial information to support your legal aid application. We do not use it for anything else.
         summary_heading: What information will the LAA be able to see?
+        true_layer_auth_error: Select yes if you agree to share your bank account information with the LAA
         able_to_see: 
           heading: "We'll be able to see your:"
           list: |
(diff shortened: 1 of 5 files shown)


config/locales/en/citizens.yml
```diff
@@ -59,6 +59,7 @@ en:
         title: Do you agree to share your bank account information with the LAA?
         inset_text: We only use your financial information to support your legal aid application. We do not use it for anything else.
         summary_heading: What information will the LAA be able to see?
+        true_layer_auth_error: Select yes if you agree to share your bank account information with the LAA
```
colinbruce, line 62: Is it worth getting Jim to review the message?  If they have selected `yes` and authentication fails it could be very confusing for citizen users?

PR #1589, opened 2020-07-17
AP-1461 Rearchitect state machine

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1461)

The `LegalAidApplicationStateMachine` has become too big and is doing two things - managing both the passported and non-passported flows.

This PR breaks the state machine into three parts, a `BaseStateMachine` which includes states used by both flows, and two subclasses, `PassportedStateMachine` and `NonPassportedStateMachine`, to handle the relevant states for those flows.

It adds functionality to the `LegalAidApplication` to create and manage the state machine.

The majority of changes in the commit are to specs and feature tests to ensure that test data is created properly using this new pattern. Traits have been added and amended in the `LegalAidApplication` factory to support this.

Other changes include - 
* removal of the old `LegalAidApplicationStateMachine`
* addition of logic to switch from a `PassportedStateMachine` to `NonPassportedStateMachine` after a DWP benefit check if the applicant does not receive benefits
* a migration to move all pre-existing applications from the old state machine to the new state machines
* removed `migration_helpers/state_changer_spec.rb` and excluded `migration_helpers/state_changer.rb` from code covereage, as these tests will always fail after this change

## Checklist


spec/factories/legal_aid_applications.rb
```diff
@@ -27,80 +27,142 @@
     #        TRAITS TO SET STATE                          #
     #######################################################
 
+    trait :with_non_passported_state_machine do
+      before(:create) do |application|
+        application.state_machine = NonPassportedStateMachine.new
```
colinbruce, line 32: These are going to be quite resource-intensive, can we replace it with another factory?

Creating 'real' objects in the database is slow and could slow down the test runs as it will be hit hundreds of times :/

PR #1609, opened 2020-07-27
Update ProviderDetailsRetriever to replace space with %20 in url

Obsiye:
Bug Fix this sentry issue [ProviderDetailsRetriever::ApiError](https://sentry.service.dsd.io/mojds/apply-for-legal-aid/issues/40113/)

In the ProviderDetailsRetriever the method [encode_www_form_component](https://apidock.com/ruby/v1_9_3_125/URI/encode_www_form_component/class) method converts a space to a + which fails to retreive information for a provider username that has a space in it "John Lang" > "John+Lang" but instead use %20 which works "John Lang" > "/John%20Lang"

Update tests

## Checklist

### DIFF
--- a/app/services/provider_details_retriever.rb
+++ b/app/services/provider_details_retriever.rb
@@ -31,7 +31,11 @@ class ProviderDetailsRetriever
   end
 
   def url # rubocop:disable Lint/UriEscapeUnescape
-    File.join(Rails.configuration.x.provider_details.url, URI.encode_www_form_component(username))
+    File.join(Rails.configuration.x.provider_details.url, encoded_uri)
+  end
+
+  def encoded_uri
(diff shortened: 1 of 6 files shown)


app/services/provider_details_retriever.rb
```diff
@@ -31,7 +31,11 @@ def response
   end
 
   def url # rubocop:disable Lint/UriEscapeUnescape
-    File.join(Rails.configuration.x.provider_details.url, URI.encode_www_form_component(username))
+    File.join(Rails.configuration.x.provider_details.url, encoded_uri)
+  end
+
+  def encoded_uri
+    URI.encode_www_form_component(username).gsub('+', '%20')
```
colinbruce, line 38: We could use 
```suggestion
    URI.escape(username)
```

PR #1613, opened 2020-07-28
Bugfix 2807 add permissions to new firms as they are created

stephenrichards:
## What
We now use permissions to decide whether or not provider firms can process passported or non-passported applications beyond the DWP check, but we forgot to assign passported permissions as a default to all new firms being created (which happens when a provider user from that firm logs in for the very first time).

This bugfix rectifies that.

## Checklist

### DIFF
--- a/app/services/provider_details_creator.rb
+++ b/app/services/provider_details_creator.rb
@@ -36,7 +36,7 @@ class ProviderDetailsCreator
       firm.update!(name: firm_name)
       firm.offices << offices
     end
-    # TODO For the timebeing, we add passported permissions to every new firm, but this may change in the future.
+    # TODO: For the timebeing, we add passported permissions to every new firm, but this may change in the future.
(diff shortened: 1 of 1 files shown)


app/services/provider_details_creator.rb
```diff
@@ -31,10 +32,12 @@ def should_clear_selected_office?
   end
 
   def firm
-    Firm.find_or_create_by!(ccms_id: firm_id).tap do |firm|
+    current_firm = Firm.find_or_create_by!(ccms_id: firm_id).tap do |firm|
       firm.update!(name: firm_name)
       firm.offices << offices
     end
+    current_firm.permissions << @passported_permission unless current_firm.permissions.include?(@passported_permission)
```
Obsiye, line 39: To me this says 'every provider has a a firm with passported permissions.'

This may cause a problem for when we move to non-passported. Maybe have a TODO comment or refactor to say
if there are no permissions on the firm then default to passported_permission 

`current_firm.permissions << @passported_permission if current_firm.permissions.empty?`

PR #1634, opened 2020-08-06
AP1555 Show expanded detail only for citizen on the types of income and outgoing pages

Obsiye:
AP1555 Show expanded detail only for citizen on the types of income and outgoing pages

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1555)

Update types of income and outgoing pages for provider and citizen by showing expanded details for the
Citizen but not for the provider

Test the above logic

## Checklist

### DIFF
--- a/app/views/shared/forms/_identify_types_of_outgoings_form.html.erb
+++ b/app/views/shared/forms/_identify_types_of_outgoings_form.html.erb
@@ -28,20 +28,22 @@
     </fieldset>
   <% end %>
 
-  <details class="govuk-details govuk-!-margin-top-4" data-module="govuk-details">
-    <summary class="govuk-details__summary">
-          <span class="govuk-details__summary-text">
-            <%= t('.expanded_explanation.heading') %>
-          </span>
-    </summary>
+  <% if journey == :citizen %>
+    <details class="govuk-details govuk-!-margin-top-4" data-module="govuk-details">
+      <summary class="govuk-details__summary">
+            <span class="govuk-details__summary-text">
+              <%= t('.expanded_explanation.heading') %>
+            </span>
+      </summary>
 
-    <div class="govuk-details__text">
-      <% t('.expanded_explanation.list').each_line do |para| %>
-        <p>
-          <%= para %>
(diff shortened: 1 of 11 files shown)


app/views/shared/forms/_identify_types_of_outgoings_form.html.erb
```diff
@@ -28,20 +32,22 @@
     </fieldset>
   <% end %>
 
-  <details class="govuk-details govuk-!-margin-top-4" data-module="govuk-details">
-    <summary class="govuk-details__summary">
-          <span class="govuk-details__summary-text">
-            <%= t('.expanded_explanation.heading') %>
-          </span>
-    </summary>
+  <% if citizen %>
```
mpw5, line 35: Could you just do `<% if journey == :citizen %>` here, and then the remove the code on lines 1-4? You're not using the `citizen` variable anywhere else so it seems a bit unncessary.

app/views/shared/forms/_types_of_income_form.html.erb
```diff
@@ -33,20 +37,22 @@
     </fieldset>
   <% end %>
 
-  <details class="govuk-details govuk-!-margin-top-4" data-module="govuk-details">
-    <summary class="govuk-details__summary">
-          <span class="govuk-details__summary-text">
-            <%= t('.expanded_explanation.heading') %>
-          </span>
-    </summary>
+  <% if citizen %>
```
mpw5, line 40: Could you just do `<% if journey == :citizen %>` here, and then the remove the code on lines 1-4? You're not using the `citizen` variable anywhere else so it seems a bit unncessary.

app/views/shared/forms/_identify_types_of_outgoings_form.html.erb
Obsiye: True, I just wanted to make it more visible by placing the variable at the top. Since I have tests in, It'll probably fail if the journey variable wasn't defined or deleted somehow.

I'll change it, thanks 👍

PR #1681, opened 2020-08-19
AP1587 geckoboard nonpassported

Obsiye:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1587)

Show the total number per day split by passported vs. non-passported submitted applications on geckoboard

- Update the existing /widget_data_providers/applications.rb with nonpassported and passported data

- Create tests

The dataset is just added ontop of our existing 'Applications' widget 
image

## Checklist


app/services/test_application_creation_service.rb
```diff
@@ -1,13 +1,6 @@
 class TestApplicationCreationService
   APPLICATION_TEST_TRAITS = %i[
-    at_initiated
     at_checking_applicant_details
-    at_checking_passported_answers
-    at_applicant_details_checked
-    awaiting_applicant
-    applicant_entering_means
-    provider_entering_merits
-    at_checking_merits_answers
```
mpw5, line 10: I might be missing something but I'm not sure I understand why these are being removed? It seems like it's removing most of the purpose of this class.

Also it doesn't look like `TestApplicationCreationService` is used by the geckoboard tests, so should these changes even be part of this PR?

app/services/test_application_creation_service.rb
```diff
@@ -31,9 +24,14 @@ def providers
...
       trait,
-      provider: provider
+      provider: provider,
+      applicant: passported_applicant
     )
   end
+
+  def passported_applicant
+    FactoryBot.create :applicant, date_of_birth: '10-01-1980', last_name: 'Walker', national_insurance_number: 'JA293483A'
+  end
```
mpw5, line 36: Similar to above, I'm not sure why this change is necessary here (sorry if I'm missing something obvious 😁)

mpw5 (commented):
Looks mostly good to me. I've left a comment about the changes to `test_application_creation_service`, as I'm not sure I understand the changes you've made there.

It's probably a good idea to get @colinbruce (sorry Colin) to take a look at it too as I have little experience with GeckoBoard and might be missing something

app/services/test_application_creation_service.rb
Obsiye: sorry that was leftover stuff. i'll remove all changes to that file

colinbruce (changes requested):
Couple of possible updates needed... happy to discuss if I've missed some reasoning

colinbruce (approved):
:+1:

PR #1682, opened 2020-08-20
Bugfix: move flags in curl command

willc-work:
## What

The database backup cronjob was failing due to `kubectl` not being installed on the docker image. The install was failing because of an incorrectly formatted curl command.

Moved the `--retry` and `--retry-delay `flags to after the filename.b

## Checklist


mpw5 (approved):
LGTM

colinbruce (approved):
Nice!  We only added it because the download kept failing

PR #1683, opened 2020-08-20
AP-1588: Geckoboard client emails

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1588)

Create a new single_object geckoboard data set, job, dashboard_handler event and calling method

This needed to be a single object because we don't record a database value for emails sent so it must incrementally increase on each send

## Checklist

### DIFF
--- a/spec/services/dashboard_event_handler_spec.rb
+++ b/spec/services/dashboard_event_handler_spec.rb
@@ -116,11 +116,6 @@ RSpec.describe DashboardEventHandler do
 
   context 'applicant_emailed' do
     let(:legal_aid_application) { create :legal_aid_application, :with_applicant }
-    before do
-      # allow_any_instance_of(DashboardEventHandler).to receive(:firm_created)
-      # allow_any_instance_of(DashboardEventHandler).to receive(:provider_created)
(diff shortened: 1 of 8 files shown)


spec/services/dashboard_event_handler_spec.rb
```diff
@@ -113,4 +113,18 @@
       merits_assessment.submit!
     end
   end
+
+  context 'applicant_emailed' do
+    let(:legal_aid_application) { create :legal_aid_application, :with_applicant }
+    before do
+      # allow_any_instance_of(DashboardEventHandler).to receive(:firm_created)
+      # allow_any_instance_of(DashboardEventHandler).to receive(:provider_created)
+      # allow_any_instance_of(DashboardEventHandler).to receive(:application_created)
```
mpw5, line 122: Can these be removed?

mpw5 (approved):
Looks good. I've left one comment about some commented-out code

spec/services/dashboard_event_handler_spec.rb
colinbruce: Good catch... they seemed pointless and the passing tests confirmed it! 😄

PR #1694, opened 2020-08-26
AP-1634 Replace staging true layer config with live

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1634)

replaced staging true layer credentials with live in order to test that we can actally access live true layer.

## Checklist


colinbruce (approved):
👍

PR #1711, opened 2020-08-31
AP-1643 Clear TrueLayer errors

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1643)

After receiving an error authenticating with TrueLayer, if a user tries again and successfully authenticates, the same error is still displayed. This change fixes this problem by setting `session[:worker_id] = nil` in `accounts_controller` after the error has been handled. This removes the worker from the session and when the second attempt is made a new worker
is created. This prevents the error from being replayed.

## Checklist


colinbruce (approved):
LGTM :+1:

PR #1712, opened 2020-08-31
AP-1641 Improve display of TrueLayer errors

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1641)

If errors are received when authenticating with a bank through TrueLayer an ugly error string is displayed on the page. This change improves that by extracting the description of the error and displaying it, or a default error if no description can be found.

It does several things:

- Add methods to the `accounts_controller` to parse the error and extract the relevant description or provide a default error.
- Amends the error generated by the `api_client` to make it easier to parse.
- Removes an error status that was being set by `consents_controller` if authentication failed. This was having the effect of making it look as if the error was caused by the user not giving consent, when this was not the case.
- Wraps the error in a `page_template` so that a back button is generated to allow the user to navigate away from the error.

## Checklist

### DIFF
--- a/app/controllers/citizens/accounts_controller.rb
+++ b/app/controllers/citizens/accounts_controller.rb
@@ -2,6 +2,7 @@ module Citizens
   class AccountsController < CitizenBaseController
     class TrueLayerWorkerError < StandardError; end
     skip_back_history_for :gather
+    helper_method :error_description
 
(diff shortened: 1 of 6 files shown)


colinbruce (approved):
Looks good! :+1:

PR #1719, opened 2020-09-02
AP1512 Subclass oauth2

Obsiye:
Subclass oauth2

This is to better test the issue that is caused by omniauth raising a CSRF detected error

Send a sentry message with raven containing the client_id , request params and session 'state' value

This is an almost exact copy of the oauth2 https://github.com/omniauth/omniauth-oauth2/blob/master/lib/omniauth/strategies/oauth2.rb
and it was subclassed to allow for better logging.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1512)

**To note:**

I've added a rubocop disable # rubocop:disable Lint/MissingSuper
for the  callback error class which subclasses standard error

## Checklist

### DIFF
--- a/app/lib/omniauth/strategies/laa_oauth2.rb
+++ b/app/lib/omniauth/strategies/laa_oauth2.rb
@@ -1,9 +1,11 @@
-require "oauth2"
-require "omniauth"
-require "securerandom"
-require "socket"       # for SocketError
-require "timeout"      # for Timeout::Error
+require 'oauth2'
+require 'omniauth'
+require 'securerandom'
+require 'socket'       # for SocketError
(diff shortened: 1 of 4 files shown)


mpw5 (approved):
Approved, but it looks like your code coverage is < 100%, so you might need to revisit the specs. Or, as this is just temporary, maybe consider excluding the omniauth code from simplecov, eg add `app/lib/omniauth` to 

```
  SimpleCov.start do
    add_filter 'config/initializers/'
    add_filter 'spec/'
    add_filter 'services/migration_helpers/'
  end
``` 

in `spec/spec_helper.rb`.

PR #1734, opened 2020-09-07
AP-1612 Make screen readers ignore null table cells

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1612)

In two places there are tables that contain deliberately empty cells - the provider's applications list and the transaction selection screens.

These are read out by screen readers which is confusing for users. The null cells cannot be easily removed as they are used to rearrange the layout of the pages on small screens.

Instead to prevent the cells being announced by screen readers, mark them as `aria-hidden="true"`.

## Checklist


Obsiye (approved):
Nice

willc-work (approved):

PR #1737, opened 2020-09-08
AP-1558 Fix back button behaviour in merits screens

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1558)

There is inconsistent behaviour when using the back links in the merits screens. It often takes the user back much further than required, into the means screens, and results in a lot of clicking to get back to the correct place.

This commit hardcodes the correct back link path into each view in the merits flow.

## Checklist


Obsiye (approved):
LGTM

PR #1748, opened 2020-09-10
AP-1644 Prevent duplicate bank providers

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1644)

If a user authenticates twice with the same bank through TrueLayer, a duplicate `bank_provider` record is created resulting in duplicate `bank_account`, `bank_account_holder`, and `bank_transaction` records.

This duplication occurs because `ImportProviderService` checks for existing `bank_provider` based on `credentials_id`, however `credentials_id` is always unique and so no duplicate will ever be found.

This PR changes this functionality so that it instead checks for duplicates based on `true_layer_provider_id`, which will be consistent for all bank details retrieved from a given bank.

## Checklist

### DIFF
--- a/app/services/true_layer/importers/import_provider_service.rb
+++ b/app/services/true_layer/importers/import_provider_service.rb
@@ -24,7 +24,7 @@ module TrueLayer
 
       def import_provider
         bank_provider = applicant.bank_providers.find_or_create_by!(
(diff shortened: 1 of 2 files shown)


willc-work (approved):
LGTM

PR #1755, opened 2020-09-11
Ap 1658 feedback question

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1658)

Add in an alternative to the provider/citizen option on feedback to cover `unknown` results.
Reword the Provider question so all visitors see the same questions but we retain the ability to check where a visitor came from
Adds a test to make sure the provider sees the correct question

## Checklist

### DIFF
--- a/config/locales/en/feedback.yml
+++ b/config/locales/en/feedback.yml
@@ -2,10 +2,7 @@
 en:
   feedback:
     new:
-      difficulty: 
-        provider: How easy or difficult was it to use this service?
-        citizen: How easy or difficult was it to use this service?
-        unknown: How easy or difficult was it to use this service?
+      difficulty: How easy or difficult was it to use this service?
       done_all_needed: Were you able to do what you needed today?
       satisfaction: Overall, how satisfied were you with this service?
       signed_out: You are signed out
--- a/config/locales/en/feedback.yml
+++ b/config/locales/en/feedback.yml
@@ -3,8 +3,9 @@ en:
   feedback:
     new:
       difficulty: 
-        Provider: How easy or difficult was it to apply for legal aid using the new service?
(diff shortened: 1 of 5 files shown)


config/locales/en/feedback.yml
```diff
@@ -3,8 +3,9 @@ en:
   feedback:
     new:
       difficulty: 
-        Provider: How easy or difficult was it to apply for legal aid using the new service?
-        Citizen: How easy or difficult was it to use this service?
+        provider: How easy or difficult was it to use this service?
+        citizen: How easy or difficult was it to use this service?
+        unknown: How easy or difficult was it to use this service?
```
colinbruce, line 8: If all three questions are now the same, can we not just use a single key?
It was only split because there was a need for a different text for applicants and professional users 🤷‍♂️

willc-work: We want to keep the functionality for identifying providers for this ticket https://dsdmoj.atlassian.net/browse/AP-1545.

colinbruce: :+1: Sorry if I wasn't clear, I just meant that if the text is the same, do we need to separate the three strings ,provider, citizen and unknown? Wouldn't a single string be okay, e.g.
```ruby
    new:
      difficulty: How easy or difficult was it to use this service?
      done_all_needed: Were you able to do what you needed today?
      satisfaction: Overall, how satisfied were you with this service?
```
The source (:provider, :citizen or :unknown) is still set in the controller at https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/master/app/controllers/feedback_controller.rb#L59

willc-work: 👍🏻

colinbruce (approved):
:+1: LGTM

PR #1757, opened 2020-09-11
Fix TrueLayer access denied issue

mpw5:
## What

Requests to TrueLayer are intermittently failing due to CORS errors. This is occurring because the `state` value returned in their responses does not match the `state` value in our session.

https://github.com/omniauth/omniauth-oauth2/issues/58 suggests various solutions for this.

The one that appears to work involves adding `config/initializers/session_store.rb` configured to use `cookie_store`.  This should not be necessary as since Rails 5 this config has been handled implicitly by Rails, but it appears to work 🤷 

## Checklist

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,0 +1,1 @@
+Rails.application.config.session_store :cookie_store, key: '_app_session'


stephenrichards (approved):

mpw5:
No longer needed after changes to reduce session/cookie size

PR #1763, opened 2020-09-15
AP-1667 Reset sessions on login

stephenrichards:
## Reset session on login

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1667)

The purpose of this ticket is to stop sessions (which are stored in the cookie) from getting too large.  We suspect that this is the reason we are occasionally getting Access Denied error on the callback from TrueLayer, because the session which holds a state variable has been truncated.

The session contains a history of pages visited, which is used by the back buttons to determine where to go back to.

This ticket does two things:
1) on login by the applicant (i.e. clicking on the link that has been sent them in the mail), it resets the session
2) On logging in by the provider, resetting the session isn't an option because devise has already put something it needs there, so we simply reset the page history to an empty array

## Checklist

### DIFF
--- a/spec/requests/applicants/omniauth_callbacks_spec.rb
+++ b/spec/requests/applicants/omniauth_callbacks_spec.rb
@@ -47,7 +47,7 @@ RSpec.describe 'applicants omniauth call back', type: :request do
 
     it 'does not add its url to page history' do
       subject
-      expect(session[:page_history]).not_to include(applicant_true_layer_omniauth_callback_path)
+      expect(session.keys).not_to include(:page_history)
     end
 
     context 'with a string time' do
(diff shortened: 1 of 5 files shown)


spec/requests/applicants/omniauth_callbacks_spec.rb
```diff
@@ -47,7 +47,7 @@
 
     it 'does not add its url to page history' do
```
mpw5, line 48: Given the expectation of this test has changed, it might be clearer if this was `it 'has no page history' do`.

mpw5 (approved):
LGTM. Minor comment on one of the specs but not a blocker

PR #1774, opened 2020-09-16
AP-938 Remove provider roles from profile screen

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-938)

The profile page shows the role codes for the logged in user, which looks confusing -  the codes are meaningless to the user. This commit removes the roles from the view so they are no longer displayed to the user.

## Checklist


colinbruce (approved):
:+1:

PR #1779, opened 2020-09-17
AP-1616 Make continue link styled as button accessible

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1616)

On `application_confirmations/show.html.erb`, the 'Back to your application' link has been styled as a button but does not have a role of `button` assigned or a keyboard event handler to allow keyboard-only users to activate the element with the space bar.

This change adds a role of `button` to the link and create an event handler to allow a user to activate the `button` with the space bar.

It also fixes a couple of issues with the similar event handler for `start` buttons, 
- an error is firing an error on all pages where there is no `start` button. Fix this by adding a check for `null`s
- the default space bar behaviour of scrolling is firing when a user uses the keyboard to activate the button, which looks a bit strange. Add a `preventDefault` to stop this happening

## Checklist

### DIFF
--- a/app/javascript/src/continue_button.js
+++ b/app/javascript/src/continue_button.js
@@ -1,0 +1,12 @@
+$(document).ready(function() {
+  const link = document.getElementById("continue");
+
+  if (link != null) {
+    link.onkeydown = function(e) {
+      if (e.keyCode == 32) {
+        e.preventDefault();
+        link.click();
+      }
+    };
+  }
+});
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM :+1:
Not a blocker but the two button-handlers look quite similar, could they be DRYed?

mpw5:
I knew someone was going to mention that @colinbruce 😁 I did try to DRY them out earlier but my javascript skills possibly aren't good enough and I couldn't get it to work properly...

I might hold off merging this for now and take another look at it later.

colinbruce:
I really don't think it's that urgent.  It's a minor code smell and something that a FE dev will fix in two minutes rather than us sweating and swearing over it!

mpw5:
Ok, I'm going to put it into UAT and let our mythical FE dev fix it instead 🦄

PR #1783, opened 2020-09-17
AP-1668 List firm permissions

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1668)

Amend the index view for /admin/firms to display which permissions a firm holds.

Also adjusts the table layout to account for the extra column and tidies up the existing code a little.

## Checklist

### DIFF
--- a/spec/requests/admin/firms_spec.rb
+++ b/spec/requests/admin/firms_spec.rb
@@ -3,9 +3,13 @@ require 'rails_helper'
 module Admin
   RSpec.describe FirmsController, type: :request do
     let(:admin_user) { create :admin_user }
+    let(:permission) { create :permission }
     let!(:firms) { create_list :firm, 3 }
 
-    before { sign_in admin_user }
+    before do
+      firms.first.update!({ permission_ids: [permission.id] })
+      sign_in admin_user
+    end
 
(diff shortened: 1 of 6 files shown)


spec/requests/admin/firms_spec.rb
```diff
@@ -3,9 +3,13 @@
 module Admin
   RSpec.describe FirmsController, type: :request do
     let(:admin_user) { create :admin_user }
+    let(:permission) { create :permission }
     let!(:firms) { create_list :firm, 3 }
 
-    before { sign_in admin_user }
+    before do
+      firms.first.update!({ permission_ids: [permission.id] })
```
colinbruce, line 10: Reviewing this after #1782 should we update the factory to have a default permission too?

colinbruce (approved):
One observation that is not a blocker (and may not even be for this ticket! 😄 )
:+1:

spec/requests/admin/firms_spec.rb
mpw5: Good point. I've updated the factory so that it creates a permission by default, and fixed some tests.

colinbruce (approved):

PR #1789, opened 2020-09-18
move the to_s to fix tests

willc-work:
## What

Bugfix

If only one type of account (current or savings) is created by the test, then the other account type had a value of 0.  Zero is handled as a number in the test whereas any positive or negative value is a string. The fix moves the creation of the string to the method on legal_aid_application instead of in the test.

## Checklist


mpw5 (approved):
LGTM

PR #1797, opened 2020-09-22
AP-1648 Add CCMS applicant if no applicant ref returned

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1648)

A recent production incident highlighted that it's possible for an applicant to exist in the CCMS database but be disabled, in which case a non-zero `record_count` is returned to Apply but no `ClientReferenceNumber` is included. This causes failures later in the submission process

This PR changes that functionality so that in this situation Apply makes a call to CCMS to create a new applicant.

## Checklist


Obsiye (approved):
Cool work. LGTM

PR #1806, opened 2020-09-24
AP-1625 More descriptive links for screenreaders

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1625)

There are a number of links in the application where the intent is clear for a sighted user but not obvious for a screenreader user, especially when multiple links exist on a page. This change adds a helper method to create links with text that is partially hidden using a `govuk-visually-hidden` span.

This allows us to keep to our current design but also add extra context for screenreader users - for example a link can be displayed as 'Change' but read out as 'Change postcode'.

## Checklist


stephenrichards (approved):

colinbruce (approved):
Awesome! 👍

colinbruce (approved):

PR #1808, opened 2020-09-25
Reduce calls to Geckoboard

mpw5:
## What

We're seeing lots of Sentry alerts due to exceeding the number of permitted calls to Geckoboard. This seems to be linked (at least partially) to the `ProviderDataJob`, which is triggered by a save to the `LegalAidApplication model`. The call is guarded by an if statement that always resolves to true:

`if proc { |laa| laa.state.eql?(:assessment_submitted) }`

It returns the proc, which is truthy, so a Geckoboard request is sent every time an application is saved. This corrects it so that the request is only made when the application is submitted:

`if proc { |laa| laa.state }.eql?(:assessment_submitted)`

## Checklist


colinbruce (approved):
👏 
🤦

PR #1811, opened 2020-09-25
AP-1626: Update result details header

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1626)

This was being called as a partial from capital_assessment_results
and capital_income_assessment_results.  In both cases it was being
called below the blue panel that already contained an H1.

Therefore downgrading this to an H2 should have no effect on the
display for users and will improve experince for screenreader
users

## Checklist


mpw5 (approved):
LGTM

PR #1812, opened 2020-09-25
AP-1709 Don't send data to Geckoboard in Dev, Test, UAT environments

stephenrichards:
## Don't send data to Geckoboard in non-live environments

Allows devs to specify class names of jobs which should do nothing if run in development, test, staging or UAT.

- Added a Rails configuration variable `Rails.configuration.x. suspended_geckoboard_updater_jobs `, which is hash keyed by `HostEnv` environment, (`:development`, `:test`, `:staging`, `:uat`, `:production`) each containing a list of jobs which should **NOT** be run in that host environment (the values for production is an empty array, meaning all jobs are to be run.
- Added a module `Dashboard::SuspendableJob` with one method `#job_suspended?` which returns true if the class name is in the config variable for the current `HostEnv`
- Exit immediately from the job if job is suspended

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1709)

## Checklist

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -84,5 +84,34 @@ module LaaApplyForLegalAid
     config.x.db_url = Rails.env.production? ? "postgres://#{ENV['POSTGRES_USER']}:#{ENV['POSTGRES_PASSWORD']}@#{ENV['POSTGRES_HOST']}:5432/#{ENV['POSTGRES_DATABASE']}" : 'postgres://localhost:5432/apply_for_legal_aid_dev' # rubocop:disable Layout/LineLength
 
     config.active_job.queue_adapter = :sidekiq
+
+    # list geckoboard updater jobs which are suspended by host environment
+    config.x.suspended_geckoboard_updater_jobs = {
+      development: %w[
+        Dashboard::FeedbackItemJob
+        Dashboard::ApplicantEmailJob
+        Dashboard::ProviderDataJob
+        Dashboard::UpdaterJob
+      ],
+      test: %w[
+        Dashboard::FeedbackItemJob
+        Dashboard::ApplicantEmailJob
+        Dashboard::ProviderDataJob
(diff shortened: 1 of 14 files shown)


config/application.rb
```diff
@@ -84,5 +84,34 @@ class Application < Rails::Application
...
+        Dashboard::ProviderDataJob
+        Dashboard::UpdaterJob
+      ],
+      uat: %w[
+        Dashboard::FeedbackItemJob
+        Dashboard::ApplicantEmailJob
+        Dashboard::ProviderDataJob
+        Dashboard::UpdaterJob
+      ],
+      staging: %w[
```
colinbruce, line 108: I wouldn't block staging by default... 
It shouldn't be as noisy as UAT and it's often used to test functionality

colinbruce (commented):
Looks good. It might be worth dropping something on the README on testing/adding new dashboard features?

config/application.rb
```diff
@@ -84,5 +84,34 @@ class Application < Rails::Application
...
+        Dashboard::FeedbackItemJob
+        Dashboard::ApplicantEmailJob
+        Dashboard::ProviderDataJob
+        Dashboard::UpdaterJob
```
mpw5, line 112: We might want to leave the staging jobs enabled - I know Alice sometimes tests changes there before deploying to production. With UAT it should be fairly easy to enable/disable them per environment by pushing a commit to a PR, but with staging we would have to merge and go through nearly the whole pipeline which feels time consuming/risky.

config/application.rb
stephenrichards: Yep, areed.  Have fixed.

config/application.rb
stephenrichards: Agreed

mpw5 (approved):
LGTM

PR #1814, opened 2020-09-25
Fix flaky postcode test

mpw5:
## What

The OS postcode checker previously didn't return any addresses for `SW1H 9AJ`, so that postcode is used in tests for non-existent addresses. However 102 Petty France is now (correctly) returned for that postcode. This will cause tests to fail next time VCR cassettes are recorded.

This changes `SW1H 9AJ` to `XX1X 1XX` (which is vaild postcode that is unlikely to have any addresses associated with it the near future) in one spec and two feature tests, and re-records the cassettes.

## Checklist


colinbruce (approved):
LGTM :+1:

PR #1823, opened 2020-09-30
AP1622 Add accessible message for loading transactions page

Obsiye:
AP1622 Add accessible message when page is loading to import bank transactions

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1622)

Update the page belonging to the gather route on the AccountsController page with an accessible message describing that 'Transactions are loading, you will be redirected once complete.'

## Checklist

### DIFF
--- a/app/views/citizens/accounts/gather.html.erb
+++ b/app/views/citizens/accounts/gather.html.erb
@@ -5,8 +5,9 @@
     </div>
   <% end %>
 <% else %>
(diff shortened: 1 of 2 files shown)


colinbruce (commented):
I know it was the DAC recommendation but is that wording right?

The page isn't technically loading, would we get a better steer from Content?
e.g. "Importing transactions" and "Transactions are downloading, you will be redirected once complete"

PR #1871, opened 2020-10-14
REVERT: remove hold-uat step from circle

colinbruce:
## What

Revert the circle hold-uat changes.  There will be a third DAC review so only commented out for now

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -391,19 +391,21 @@ workflows:
       - build_and_push:
           requires:
           - lint_checks
-      - hold_master_uat_notification:
-          requires:
-            - lint_checks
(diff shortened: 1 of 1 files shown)


willc-work (approved):
Should we also comment out lines 354 to 358 which is the `hold_master_uat_notification` method? As this is temporary I have approved anyway.

colinbruce:
> Should we also comment out lines 354 to 358 which is the `hold_master_uat_notification` method? As this is temporary I have approved anyway.

I thought about it, but as we just don't call it, I think it just sits there 🤷‍♂️ 
As AC mentioned another review, I thought I'd leave it there until we can finally remove it

PR #1873, opened 2020-10-14
AP-1762: Accessibility updates

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1727)

Following the DAC re-review, the following changes were required.
1. Form Controls - Grouped (A)
"do you want to make a substantive application now?"
"does your client own their home with anyone else?"
These views were missing fieldsets that enclosed the radio buttons

2. Sortable table (A)
The initial issue has been fixed and has been approved by DAC as a workable solution. It now needs to be applied to the transaction tables in the service:

3. Status message (AA)
The issue  with he proceeding type view has been initially resolved but have resulted in a new issue:
The alert is not being read out to JAWS users when browsing in Internet Explorer. The role has changed from Alert to Region which will prevent the live region from being read correctly. Change the role back to ‘alert’ or ‘status’  

This PR addresses the points raised

## Checklist

### DIFF
--- a/app/helpers/table_sort_helper.rb
+++ b/app/helpers/table_sort_helper.rb
@@ -57,27 +57,7 @@ module TableSortHelper
   #   where at: is the width below which the column will be combined (options are 470 and 555)
   #   and append: is the contented that will be appended to the main content when the columns are combined
   #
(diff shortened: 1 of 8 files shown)


mpw5 (approved):
LGTM 🎉

PR #1875, opened 2020-10-14
AP-1751 Fix HTML injection

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1751)

Fixes issue raised in Section 4.6 of the security testing report.

* Removes the use of `simple_format` to display user-entered text, which is causing HTML render instead of being escaped.
* Fixes a minor layout issue caused by the removal of `simple_format` from the `restrictions` partial.
* Fixes a pre-existing issue where a section break on the merits check your answers page is displayed incorrectly when 'Have the police been notified?' is answered 'Yes'

## Checklist

### DIFF
--- a/app/views/shared/check_answers/_merits.html.erb
+++ b/app/views/shared/check_answers/_merits.html.erb
@@ -39,7 +39,7 @@
           no_border: !respondent.understands_terms_of_court_order
         ) %>
   </dl>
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
LGTM 👍

PR #1878, opened 2020-10-15
AP-1727 Add application detail to Feedback email

Obsiye:
## AP-1727 Add application detail to Feedback email

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1727)

- Update feedback process to include legal aid application infromation: application_ref, state, user_type (citizen / provider).

- Give the application_id to the feedback mailer to gather application information for the email.

#### When the provider clicks on feedback/signout when on an application then:
-  Use the PageHistoryService / page_history to get the application id

#### When the provider clicks on feedback/signout on index page or non appliction page then:
- Add provider email from :current_provider 
- Application information is left empty on the feedback mailer

#### When the citizen clicks on feedback then:
- Use the existing session['current_application_id'] to get the provider email
- Pass the session['current_application_id'] as a param to the feedback mailer

--- 
- Update tests

- Test when a provider clicks feedback whilst not on an application e.g. on the index page / application dashboard

## Citizen feedback email

image

## Provider feedback email

image

## Checklist

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -4,18 +4,16 @@ class FeedbackController < ApplicationController
   def new
     @journey = source
     @feedback = Feedback.new
-    @signed_out_provider_id = session.delete('signed_out_provider_id')
+    @signed_out = session.delete('signed_out')
+    @application_id = application_id
   end
 
   def create
-    @feedback = Feedback.new(feedback_params)
-    @feedback.originating_page = originating_page
-    @feedback.email = provider_email
-    @display_close_tab_msg = params[:signed_out_provider_id].present?
-
+    initialize_feedback
+    @display_close_tab_msg = params['signed_out'].present?
     # must use bang version `deliver_later!` or failures won't be retried by sidekiq
     if @feedback.save
-      FeedbackMailer.notify(@feedback).deliver_later!
+      FeedbackMailer.notify(@feedback, application_id).deliver_later!
       render :show
     else
       render :new
@@ -28,31 +26,40 @@ class FeedbackController < ApplicationController
 
   private
 
+  def initialize_feedback
+    @feedback = Feedback.new(feedback_params)
+    @feedback.originating_page = originating_page
+    @feedback.email = provider_email
+  end
+
   def provider_email
-    if params[:signed_out_provider_id].present?
-      signed_out_provider_email
-    else
-      signed_in_provider_email
+    # citizen side won't have access to current_provider but can get provider_email via current_application_id
+    provider&.email || current_provider&.email
+  end
+
+  def provider
+    LegalAidApplication.find_by_id(application_id)&.provider
+  end
+
+  def application_id
+    return params['application_id'] if params['application_id']
+    return session['current_application_id'] if source == :citizen
+
+    application_id_from_page_history || nil
+  end
+
+  def application_id_from_page_history
+    page_history_service = PageHistoryService.new(page_history_id: session[:page_history_id])
+    page_history = JSON.parse(page_history_service.read)
+    previous_page = page_history[-2]
+
+    previous_page&.split('/')&.each_with_index do |section, i|
+      return previous_page.split('/')[i + 1] if ['applications'].include?(section)
     end
   end
 
-  def signed_out_provider_email
-    provider_id = params[:signed_out_provider_id]
-    Provider.find(provider_id).email
-  end
-
-  def signed_in_provider_email
-    provider_struct = session['warden.user.provider.key']
-    return nil unless provider_struct
-
-    provider_id = provider_struct.first&.first
-    return nil unless provider_id
-
-    Provider.find(provider_id).email
-  end
-
   def originating_page
-    params[:signed_out_provider_id].present? ? destroy_provider_session_path : session['feedback_return_path']
+    params['signed_out'].present? ? destroy_provider_session_path : URI(session['feedback_return_path']).path.split('/').last
   end
 
   def feedback_params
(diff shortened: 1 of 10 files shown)


config/application.rb
```diff
@@ -53,7 +53,7 @@ class Application < Rails::Application
       env: ENV.fetch('GOVUK_NOTIFY_ENV', 'development')
     ).symbolize_keys
 
-    config.x.support_email_address = 'apply-for-legal-aid@digital.justice.gov.uk'.freeze
+    config.x.support_email_address = 'abdirizak.obsiye@digital.justice.gov.uk'.freeze
```
mpw5, line 56: You should probably change this back, unless you want to receive all the support emails? 😁

config/initializers/devise.rb
```diff
@@ -21,7 +21,7 @@
   # note that it will be overwritten if you use your own mailer class
   # with default "from" parameter.
   # This should be over-written if we configure Devise to send via Notify
-  config.mailer_sender = 'apply@digital.justice.gov.uk'
+  config.mailer_sender = 'abdirizak.obsiye@digital.justice.gov.uk'
```
mpw5, line 24: You probably want to change this email back too 😁

app/controllers/feedback_controller.rb
```diff
@@ -4,18 +4,18 @@ class FeedbackController < ApplicationController
   def new
     @journey = source
     @feedback = Feedback.new
-    @signed_out_provider_id = session.delete('signed_out_provider_id')
+    @signed_out = session.delete('signed_out')
+    @application_id = application_id
   end
 
-  def create
+  def create # rubocop:disable Metrics/AbcSize
```
mpw5, line 11: I'm not sure rubocop needs disabling here, I've just tried running it without and it doesn't give me a warning.

If you do do still get a warning then it might be worth refactoring some of the code into a new method to reduce the complexity rather than just disabling. You could do something like:
```
def initialize_feedback
  @feedback = Feedback.new(feedback_params)
  @feedback.originating_page = originating_page   
  @feedback.email = provider_email
end
```
and then call `initialize_feedback` in `create`

app/controllers/feedback_controller.rb
```diff
@@ -29,30 +29,31 @@ def show
...
-    provider_id = params[:signed_out_provider_id]
-    Provider.find(provider_id).email
+  def provider
+    LegalAidApplication.find_by_id(application_id)&.provider
   end
 
-  def signed_in_provider_email
-    provider_struct = session['warden.user.provider.key']
-    return nil unless provider_struct
+  def application_id # rubocop:disable Metrics/AbcSize
```
mpw5, line 40: Same here, I think this will pass rubocop without the disable statement

app/controllers/feedback_controller.rb
```diff
@@ -29,30 +29,31 @@ def show
...
+    return params['application_id'] if params['application_id']
+    return session['current_application_id'] if source == :citizen
+
+    application_id_from_page_history || nil
+  end
 
-    provider_id = provider_struct.first&.first
-    return nil unless provider_id
+  def application_id_from_page_history
+    previous_page = session['page_history'][-2]
```
mpw5, line 48: We're not storing page history in the session any more. You'll need to get it from redis instead. (It looks like you're on an old branch, you'll need to rebase to pick up the changes).

We store a `page_history_id` in the session now, which can be used by the `PageHistoryService` to get the history, so this code would now be something like:
```
page_history_service = PageHistoryService.new(page_history_id: session[:page_history_id])
page_history = page_history_service.read
previous_page = page_history[-2]
```
(you might be able to refactor that a bit)

app/controllers/feedback_controller.rb
```diff
@@ -29,30 +29,31 @@ def show
...
 
-    Provider.find(provider_id).email
+    previous_page&.split('/')&.each_with_index do |section, i|
+      return session['page_history'][-2].split('/')[i + 1] if ['applications'].include?(section)
```
mpw5, line 51: This line will also need to change because of the new way of handling page histories.

spec/requests/feedbacks_spec.rb
```diff
@@ -7,10 +7,16 @@
     let(:feedback) { Feedback.order(created_at: :asc).last }
     let(:feedback_params) { params[:feedback] }
     let(:provider) { create :provider }
-    let(:session_vars) { {} }
-    let(:address_lookup_page) { 'http://localhost:3000/providers/applications/fa580023-5b07-493d-bb64-49b6d97c2a97/address_lookup' }
+    let(:session_vars) do
+      {
+        'page_history' => [address_lookup_page, '/feedback']
+      }
+    end
```
mpw5, line 14: This will need to change because of the new way of handling page histories. You can use the `PageHistoryService` to set the history:

```
  let(:page_history_service) { PageHistoryService.new(page_history_id: session['page_history_id']) }
  before { page_history_service.write([address_lookup_page, '/feedback']) }
```

mpw5 (changes requested):
A few changes needed:
 - this seems to be quite an old branch, we've changed the way the page history is handled so you'll need to rebase and then make some changes to account for that.
 - you've left your own email address in a couple of places (I suspect you were using it for testing?)
 - a couple of minor rubocop comments

spec/requests/feedbacks_spec.rb
mpw5: Perfetct, better than my suggestion. You're right that using a stub is the best solution in a spec.

mpw5 (approved):
Great work 🎉

PR #1880, opened 2020-10-15
Enable cucumber to cope with Faker names with apostrophes

stephenrichards:
## Enable cucumber to cope with Faker names with apostrophes

Occasionally, feature tests fail because Faker has generated names like O'Keefe which then cause the cukes to fail because it's trying to find `O'Keefe` on the page, and it is rendered as `O&apos;Keefe`. This fixes that issue (in at least one place - there may be more).

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -101,6 +101,7 @@ group :development, :test do
   gem 'dotenv-rails', '>= 2.7.5'
   gem 'erb_lint', require: false
   gem 'i18n-tasks', '>= 0.9.31'
+  gem 'htmlentities'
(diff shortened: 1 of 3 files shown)


colinbruce:
Sorry @stephenrichards I didn't see this PR when I created mine #1883  😬 
I have just removed the applicant name from the HTML.  It was a belt and braces addition when I added the delete button and, it turns out, wasn't needed anyway

willc-work (approved):

colinbruce (changes requested):
This has been superceded and can be deleted

PR #1881, opened 2020-10-15
AP-1712 Do not show admin login form when in production environment

Obsiye:
AP-1712 Do not show admin login form when in production but allow for the google login.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1712)

- Update the staging and production admin pages to not show the admin_users login form but show a google login link
- Add testing

## Checklist


app/controllers/admin_users/sessions_controller.rb
```diff
@@ -0,0 +1,8 @@
+module AdminUsers
+  class SessionsController < Devise::SessionsController
+    def new
+      @show_sign_in = !Rails.env.production?
```
willc-work, line 4: There is an environment variable `allowAdminPassword` in `helm_deploy/apply-for-legal-aid/values-uat.yaml:102` that may be more appropriate for this check. It is set as `true` for UAT and `false` for staging and prod.

willc-work (approved):

PR #1887, opened 2020-10-17
AP-1746 order bank transactions

stephenrichards:
## Sort bank transactions so that the balances make sense

Bank transaction that we receive from TrueLayer need to be ordered so that when output to a CSV, the balance on one line plus the credit or debit transaction on the next line equals the balance on the next line.

Not all banks put the correct `happened_at` date on the transactions (e.g. Royal Bank of Scotland times every transaction at midnight UTC), and retrieving the transactions in `happened_at` sequence does not pull them back in the correct order.

Additionally, not all banks (e.g Monzo) put a running_balance on each transaction

This PR addresses those issues so that a CSV file of transactions is produced which orders the transactions so that the balances reconcile.

- moved the `BankTransactionReportCreator` class and all new classes involved in the creation of this report into the `Reports::BankTransactions` namespace
- modified `BankTransactionReportCreator` to call `BankTransactionSorter` before generating the CSV
- Created class `BankTransactionSorter` to iterate through the bank accounts, retrieve transactions for that account grouped by `happened_at` and then call `FirstDayTransactionSorter` on the first group, and `SubsequentDayTransactionSorter` on all other groups
- Created class `FirstDayTransactionSorter` to order the first set of transactions for the account (there is no running balance from the previous day, to the algorithm for finding the correct order is slightly different)
- Created class `SubsequentDayTransactionSorter` to order all other sets of transactions, using the balance from the last transaction in the previous set.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1746)

## Checklist


spec/services/reports/bank_transactions/bank_transaction_report_creator_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        end
+
+        context 'with local_csv param' do
+          subject { creator.call(local_csv: true) }
+
+          let(:creator) { BankTransactionReportCreator.new(legal_aid_application) }
+          let(:tempfile) { Rails.root.join('tmp', 'bank_transactions.csv') }
+
+          it 'creates a local file' do
+            File.unlink(tempfile) if File.exist?(tempfile)
```
colinbruce, line 49: Cool :+1: 
Not seen `unlink` before, I've always used a block and hoped it deleted it properly! 😄

colinbruce (approved):
:+1:
Looking good but will need lots of UAT!

PR #1892, opened 2020-10-19
AP-1423 Improve errors

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1423)

* Updates copy for a number of errors.

* Prevents an error being displayed if no disregarded benefits are categorised.

## Checklist

### DIFF
--- a/spec/requests/providers/vehicles/ages_spec.rb
+++ b/spec/requests/providers/vehicles/ages_spec.rb
@@ -92,7 +92,7 @@ RSpec.describe Providers::Vehicles::AgesController, type: :request do
       it 'displays error' do
         subject
         expect(response.body).to include('govuk-error-summary')
-        expect(response.body).to include('Select Yes if your client bought the car over three years ago')
+        expect(response.body).to include('Select yes if your client bought the vehicle over 3 years ago')
       end
     end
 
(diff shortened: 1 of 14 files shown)


spec/requests/providers/vehicles/ages_spec.rb
```diff
@@ -92,7 +92,7 @@
       it 'displays error' do
         subject
         expect(response.body).to include('govuk-error-summary')
-        expect(response.body).to include('Select Yes if your client bought the car over three years ago')
+        expect(response.body).to include('Select yes if your client bought the vehicle over 3 years ago')
```
willc-work, line 95: If we expect the translation rather than the actual text it could eliminate the need for amending the tests when content is updated.

willc-work (approved):
LGTM - I left a comment re: translations in test specs which could be applied to any of the updated tests, but it is not a blocker at all.

spec/requests/providers/vehicles/ages_spec.rb
mpw5: Fair point. I've pushed an extra commit that fixes this in a few places (all that I could find, although there might be some that I've missed) if you fancy giving it a quick re-review.

willc-work (approved):
LGTM

PR #1896, opened 2020-10-20
AP-1773: update benefit selection

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1773)

Update the transaction handling.  
Previously transactions that had been identified as benefits would be displayed, ticked, on the add new transactions page. If a user submitted the page (regardless of adding new ticked rows) all rows would identified as `manually_chosen`.

This affected the data sent to CFE.

This removes previously ticked rows.  The main screen says 
image

So I have taken this to mean that the `remove` link is the sole way of de-selecting a line-item.  Now a provider can only tick new rows on the incoming_transactions page
image

## Checklist


mpw5 (approved):
LGTM

PR #1899, opened 2020-10-20
AP-1753 Whitelist IP ranges that can access admin pages

stephenrichards:
## Whitelist IP ranges that can access admin pages

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1753)

- created encrypted file `config/ipaddrs.crypt.yml` to list ranges that can access admin pages
- Created `AdminBaseController` that has a before action to call `AuthorizedIpRanges` to check that the request ip is in one of those ranges
- Derived all admin controllers from the `AdminBaseController`

## Checklist


Gemfile
```diff
@@ -100,6 +100,7 @@ group :development, :test do
   gem 'byebug', platforms: %i[mri mingw x64_mingw]
   gem 'dotenv-rails', '>= 2.7.5'
   gem 'erb_lint', require: false
+  gem 'htmlentities'
```
mpw5, line 103: This doesn't seem to be related to this PR. Has it crept in from another change?

spec/requests/citizens/check_answers_spec.rb
```diff
@@ -60,7 +60,7 @@
 
     it 'displays the name of the firm' do
       subject
-      expect(response.body).to include(firm.name)
+      expect(response.body).to include(HTMLEntities.new.encode(firm.name))
```
mpw5, line 63: This doesn't seem to be related to this PR. Has it crept in from another change?

mpw5 (commented):
Looks good, but there appear to be some changes in the PR which don't relate to this change. Have then been added by mistake?

Gemfile
stephenrichards: No - I included it because I was getting spec failures when Faker generates a name like "O'Keefe" - so this gem enables us to test against those names when they crop up as well.

spec/requests/citizens/check_answers_spec.rb
stephenrichards: See above

mpw5 (approved):
LGTM

PR #1900, opened 2020-10-20
AP-1748 Sanitise CSV files

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1748)

CSV files produced by Apply are vulnerable to csv injection exploits. This adds a mixin to provide functionality to sanitise the data by prefixing an apostrophe to any fields that begin with an exploitable character (+, -, =, @ or %).

This has been added to the `ApplicationDetailCsvLine` and `NonPassportedApplicationCsvLine` classes to sanitise both reports (the non-passported application report doesn't contain any user-entered data so is not particularly vulnerable but has been included for consistency).

## Checklist

### DIFF
--- a/app/services/reports/mis/application_detail_csv_line.rb
+++ b/app/services/reports/mis/application_detail_csv_line.rb
@@ -1,6 +1,8 @@
 module Reports
   module MIS
     class ApplicationDetailCsvLine
+      include Sanitisable
+
       attr_reader :laa
 
       delegate :application_ref,
@@ -156,7 +158,7 @@ module Reports
         restrictions
         respondent_details
         merits
-        @line
+        sanitise
       end
 
       private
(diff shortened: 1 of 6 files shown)


app/services/reports/mis/securable.rb
```diff
@@ -0,0 +1,25 @@
...
+      def starts_with_special_character?(str)
+        %w[= + - @ %].include?(str.first)
+      end
+
+      def prefix(field)
+        "'#{field}"
+      end
+    end
+  end
+end
```
stephenrichards, line 25: Would this module be better named `Sanitizable`  to match up with the method name?

stephenrichards (changes requested):

app/services/reports/mis/securable.rb
mpw5: Good point.  I'll change it

mpw5: @stephenrichards I've changed this, can you re-review at some point please?

stephenrichards (approved):

PR #1902, opened 2020-10-21
AP-1674 Enable Welsh locale

stephenrichards:
## Enable Welsh locale

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1674)

- Configured Rails to enable locales `:en` and `:cy` (and additionally `:en-GB` in test mode as it is required by Faker)
- Configured Rails to use `:en` as the default locale
- Configured url_helpers always to include a locale parameter set to the current locale for every URL
- Added an around_action on the `CitizenBaseController` to switch locales based on the `locale` parameter in the URL
- Added the locale parameter into all the procs on the Citizen Flow Service

All of the above work should have no effect on the user journey (except that the URLs with be suffixed with `?locale=en`).  Manually switching to the Welsh locale (by editing the URL to end `locale=cy` will display untranslated text until such time as the faux-Welsh translations are added in.

## Checklist


app/views/shared/check_answers/_client_details.html.erb
```diff
@@ -1,7 +1,7 @@
 <dl class="govuk-summary-list govuk-!-margin-bottom-9">
   <%= check_answer_link(
         name: :first_name,
-        url: providers_legal_aid_application_applicant_details_path(anchor: :first_name),
+        url: providers_legal_aid_application_applicant_details_path(@legal_aid_application, anchor: :first_name),
```
colinbruce, line 4: Why do we suddenly need to update this and pass in the `@legal_aid_application`?

spec/requests/admin/reports_spec.rb
```diff
@@ -19,7 +19,7 @@
 
     it 'has a link to the download csv path' do
       subject
-      expect(response.body).to include(admin_reports_submitted_csv_path)
+      expect(response.body).to include(admin_reports_submitted_csv_path(format: :csv))
```
colinbruce, line 22: Again, what are we changing that we suddenly need to send the explicit format 😕

spec/requests/backable_spec.rb
```diff
@@ -31,7 +31,7 @@
     end
 
     it 'has a back link to the previous page' do
-      expect(response.body).to have_back_link("#{page_2}?back=true")
+      expect(response.body).to have_back_link("#{page_2}&back=true")
```
colinbruce, line 34: This format looks wrong, shouldn't urls be
`domain?query_params#anchor`
I can't visualise how the locale will fit in this example

colinbruce (commented):
This has raised a few questions for me. 
I'm not keen on so many tests having to be changed to make it pass.
What happens when some of the shared partials get called on the provider side?

app/views/shared/check_answers/_client_details.html.erb
stephenrichards: because rob was relying on @legal_aid_application being understood somehow, and it doesn't work when you have parameters added to the URL (locale=en).  Actually, I think it's better to explicitly give the application id rather than rely on magic anyway.

spec/requests/admin/reports_spec.rb
stephenrichards: because previously, the repsonse body contained `/admin/admin_report_submitted.csv` and we were seeing if it included  `admin_reports_submitted_csv_path`, which was `/admin/admin_report_submitted`.

Now, the response_body contains `/admin/admin_report_submitted.csv?locale=en` and the `admin_reports_submitted_csv_path` without the format param equates to `/admin/admin_report_submitted?locale=en`, hence the need to add the format param to get this test to pass

spec/requests/backable_spec.rb
stephenrichards: Remember that all urls are now generated with a locale parameter.

`page_2` in this case is the full url including the locale param, so something like `/citizens/own_home?locale=en`.  leaving it as it was would have resultted in /citizens/own_home?locale=en?back=true` which is obviously wrong.

willc-work (approved):
LGTM

PR #1904, opened 2020-10-22
AP-1672 Add fake Welsh translations

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1672)

Copies all translations for the applicant journey the `/cy` locale and converts them to fake-Welsh (by reversing each string) for use in developing the Welsh language version of the applicant journey.

* This only applies to text that appears on the current applicant journey - any translations for providers or for parts of the applicant journey that are no longer visible to applicants (eg dependants, vehicles, capital and equity) have not been included.

* Behaviour of `spec/i18n_spec.rb` has been changed - it now tests for all translations in the English locale but only translations on pages on the applicant journey for the Welsh locale.

* Rake task to generate fake-Welsh versions of all translation files is included for future use.

## Checklist

### DIFF
--- a/config/locales/cy/activemodel.yml
+++ b/config/locales/cy/activemodel.yml
@@ -1,0 +1,21 @@
+---
+cy:
+  accessibility:
+    problem_text: melborp a si erehT
+  activemodel:
+    errors:
+      models:
+        irregular_income:
+          attributes:
+            amount:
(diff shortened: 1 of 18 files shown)


Obsiye (approved):
Awesome!

stephenrichards (commented):
I think we should include the rake task to do this for the time being, as we may want to run it again at some stage before it gets translated into real Welsh.

PR #1911, opened 2020-10-23
AP-1778 Add rememberable Devise module to Applicant

stephenrichards:
## Allow for session being deleted when collecting bank transactions

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1778)

We have seen instances of applicants using mobile phones having the session destroyed between making the call to true layer and receiving the call back, which in turn causes an Omniauth CSRF error.  We believe this behaviour is caused by the applicant switching to another app on their mobile phone to look up their credentials and then switching back to the browser to continue.  This has been observed to happen.

Adding the `rememberable` Devise module is not enough to cure this problem because at the start of the applicant journey, the magic link is decoded to provide a `legal_aid_application_id` which is stored in the session.  So even with the `rememberable` module enabled, the new session will not contain the legal_aid_application_id.

The solution is during the request phase of the call to TrueLayer, to store the session in the redis database twice - once with a key of the state variable, and once with the key of the `applicant id`.  The is because it is needed during the callback phase of Omniauth, which only knows about the state variable, and again by the `gather_transactions_controller` which only knows about the `applicant_id`.  The records are written to redis with an expiry time of 10 minutes, and are in any case deleted once they have been successfully retrieved.

- Added `rememberable` to the `Applicant` model and to the relevant contollers.
- Called 'remember_me` in the `BanksController` to instantiate the devise cookie 
- Called `forget_me` in the `CheckAnswersController` to invalidate the devise cookie
- Replaced the before_action `authenticate_applicant!` in the `CitizenBaseController` with the method `authenticate_with_devise` which just calls `authenticate_applicant!`.  This enables it to be overridden in the `GatherTransactionsContoller` that have to restore the session before authentication
- Created `OauthSessionSaver` service to handle storage, retrieval and deletion of session data in the redis database
- Call `OauthSessionSaver.store` to store the session in the request phase of the omniauth strategy
- Call `OauthSessionSaver.get` and `OauthSessionSaver.destroy` in the callback phase of omniauth strategy
- Call `OauthSessionSaver.get` and `OauthSessionSaver.destroy` in `GatherTransactionController`

## Checklist

### DIFF
--- a/app/lib/omniauth/strategies/moj_oauth2.rb
+++ b/app/lib/omniauth/strategies/moj_oauth2.rb
@@ -71,7 +71,8 @@ module OmniAuth
         # where only the applicant_id will be available.
         #
         OauthSessionSaver.store(applicant_id, session)
-        OauthSessionSaver.store(state, session)
+        OauthSessionSaver.store(auth_params[:state], session)
+
         Debug.record_request(session, auth_params, callback_url, browser_details)
         redirect client.auth_code.authorize_url({ redirect_uri: callback_url }.merge(auth_params))
       end
--- a/app/lib/omniauth/strategies/moj_oauth2.rb
+++ b/app/lib/omniauth/strategies/moj_oauth2.rb
@@ -59,8 +59,19 @@ module OmniAuth
         hash
       end
 
-      def request_phase
+      def request_phase # rubocop:disable Metrics/AbcSize
         auth_params = authorize_params
+        applicant_id = session['warden.user.applicant.key'].first.first
+
+        # We save the session in the redis database here in case the user backgrounds his browser on a mobile
+        # device while looking up the bank credentials, which would result in the session being destroyed.
+        #
+        # We save it under the applicant_id key AND the state key, because it will be required in two different places,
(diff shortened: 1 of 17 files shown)


mpw5 (approved):
LGTM

mpw5 (approved):
LGTM! Great job 🎉

app/lib/omniauth/strategies/moj_oauth2.rb
```diff
@@ -59,8 +59,20 @@ def client
         hash
       end
 
-      def request_phase
+      def request_phase # rubocop:disable Metrics/AbcSize
```
mpw5, line 62: Just thinking about this a bit more. I think we initially added this class as a temporary fix to allow us to debug the access denied issue properly. If we start adding our own logic to it are we accepting that it's now a permanent part of app? 

If so, there's a TODO in`/app/lib/omniauth/strategies/true_layer.rb` which refers to reverting back to the standard OAuth2 strategy after we've finished debugging. That should probably be removed.

stephenrichards: Quite right.  Will remove it.

PR #1912, opened 2020-10-23
AP-1745 Error handling when Clamav is down or raises an error

Obsiye:
## AP-1745 Error handling when Clamav is down or raises an error

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1745)

- Add an error message when the clamadscan is down.
- Replace the raising of error :ClamdscanError with a raven capture message
- Update tests

This is a temporary fix until we turn this malware scanning into a worker/job that can be retried automatically.

## Checklist

### DIFF
--- a/app/services/malware_scanner.rb
+++ b/app/services/malware_scanner.rb
@@ -1,6 +1,4 @@
 class MalwareScanner
-  ClamdscanError = Class.new(StandardError)
-
   def self.call(**args)
     new(**args).call
   end
@@ -20,6 +18,7 @@ class MalwareScanner
       uploader: uploader,
       virus_found: virus_found?,
       scan_result: scan_result.result,
+      scanner_working: scanner_down?,
       file_details: file_details
     )
 
@@ -36,12 +35,16 @@ class MalwareScanner
     @virus_found ||= !scan_result.status.success?
   end
 
+  def scanner_down?
+    scan_result.status.exitstatus != 2
+  end
+
   def scan_result
     @scan_result ||= begin
       stdout, stderr, status = Open3.capture3(*(command + [file_path.to_s]))
       result = stdout.strip
 
-      raise ClamdscanError, (stderr.presence || stdout) if status.exitstatus == 2
+      Raven.capture_message("ClamdscanError, #{stderr.presence || stdout}") if status.exitstatus == 2
 
       OpenStruct.new(
         result: result,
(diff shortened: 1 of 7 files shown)


app/services/malware_scanner.rb
```diff
@@ -20,6 +18,7 @@ def call
       uploader: uploader,
       virus_found: virus_found?,
       scan_result: scan_result.result,
+      scanner_working: scanner_down?,
```
mpw5, line 21: I'm a bit confused by the logic here. If `scanner_down?` returns true when the scanner is _unavailable_ then wouldn't this always set `scanner_working` to the wrong value?

I think this would be better with the parameter renamed to `scanner_down` to match the method - `scanner_down: scanner_down?`.

EDIT: I think the confusion is caused by an error in `scanner_down?`. See my comments below. I think the right solution is to change the name of that method to `scanner_working?` and then this line becomes `scanner_working: scanner_working?`

app/services/malware_scanner.rb
```diff
@@ -36,12 +35,16 @@ def virus_found?
     @virus_found ||= !scan_result.status.success?
   end
 
+  def scanner_down?
+    scan_result.status.exitstatus != 2
+  end
+
```
mpw5, line 41: I think there's a logic issue here too. My understanding (which could be wrong!) is that an `exitstatus` of 2 indicates an error, so this method should return true if `exitstatus == 2`, not `!=2`.

Alternatively if you rename the method to `scanner_working?` (or `scanner_up?` maybe?) and make the same change on line 21 then there's no need to change the logic.

app/services/malware_scanner.rb
```diff
@@ -36,12 +35,16 @@ def virus_found?
...
   def scan_result
     @scan_result ||= begin
       stdout, stderr, status = Open3.capture3(*(command + [file_path.to_s]))
       result = stdout.strip
 
-      raise ClamdscanError, (stderr.presence || stdout) if status.exitstatus == 2
+      Raven.capture_message("ClamdscanError, #{stderr.presence || stdout}") if status.exitstatus == 2
```
mpw5, line 47: Could you reuse the method from line 38 here?

`Raven.capture_message("ClamdscanError, #{stderr.presence || stdout}") unless scanner_working?`

It would mean we're not explictly checking `exitstatus` in multiple places.

mpw5 (changes requested):
I've left some comments about the logic around checking whether the scanner is available. I'm happy to discuss them or pair with you on solutions if you want 👍

mpw5 (approved):
LGTM. Good job 🎉

PR #1928, opened 2020-10-29
AP-1772 Automatically create Provider records for new sign-ins

stephenrichards:
## Automatically create Provider records for new sign-ins

When a new user with the correct roles signs in for the first time, query the Provider Details API and automatically create the associated `Firm` and `Office` records (a minimal `Provider` record is created by Devise).

- Changed the after_action on `saml_sessions_controller` to call new service `ProviderAfterLoginService`
- Created `ProviderAfterLoginServices` that checks to see if the user has just been created by devise, and if so:
    - Checks that it has the `CCMS_Apply` role 
    - Calls the existing `ProviderDetailsRetriever` to get the details from the Provider Details Api and create `Firm` and `Office` records
    - Updates a new `invalid_login_details` field on `Provider` if not able to log the user in (because they don’t have the `CCMS_Apply` role, because they cannot be found in ProviderDetailsAPI or The ProviderDetailsAPi is down’
- Amended the `confirm_office controller` to redirect to the`InvalidLoginController` if the invalid_login_details on Provider is populated
- Created a new `InvalidLoginsController` to display the right page according to the contents of the invalid_login_details

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1772)

# IMPORTANT NOTE
## Checklist


config/application.rb
```diff
@@ -120,5 +123,7 @@ class Application < Rails::Application
     Rails.application.routes.default_url_options[:locale] = I18n.locale
     config.i18n.default_locale = :en
     config.i18n.available_locales = %i[en cy] # overriden in test to allow en-GB for Faker
+
+    config.x.team_email = 'apply-for-legal-aid@digital.justice.gov.uk'
```
colinbruce, line 127: Should this be in the open?  Does github get scraped for spam emails? Maybe it should be in the settings files, it will mean duplication but at least it would be encrypted

db/schema.rb
```diff
@@ -92,6 +92,7 @@
     t.datetime "locked_at"
     t.string "true_layer_secure_data_id"
     t.datetime "remember_created_at"
+    t.string "remember_token"
```
colinbruce, line 95: Has this crept in from another PR?

spec/requests/citizens/check_answers_spec.rb
```diff
@@ -60,7 +60,7 @@
 
     it 'displays the name of the firm' do
       subject
-      expect(response.body).to include(HTMLEntities.new.encode(firm.name))
```
colinbruce, line 63: Removing the encoding?

spec/requests/saml_sessions_spec.rb
```diff
@@ -70,20 +154,79 @@
...
+            expect(HostEnv).to receive(:staging_or_production?).and_return(false)
+            expect(provider).to receive(:update_details).and_call_original
+            expect(ProviderDetailsCreatorWorker).not_to receive(:perform_async).with(provider.id)
+            expect(ProviderDetailsCreator).not_to receive(:call).with(provider)
+            subject
+          end
         end
       end
     end
+
```
colinbruce, line 203: Large commented section... needed or deletable?

colinbruce (commented):
Looking good, a couple of queries that may need looking at before merging, but no reason to delay UAT :+1:

db/schema.rb
stephenrichards: yep.

spec/requests/citizens/check_answers_spec.rb
stephenrichards: yep - it doesn't work, becaue most entities get written out properly, e.g. `&amp;` or `&lt;`, for some reason apostrophes don't get written as `&apos;` but as `&#39;`. I've got another little PR that deals with that.

config/application.rb
stephenrichards: Good point.

db/schema.rb
```diff
@@ -91,7 +91,6 @@
     t.string "unlock_token"
     t.datetime "locked_at"
     t.string "true_layer_secure_data_id"
-    t.datetime "remember_created_at"
```
mpw5, line 94: Is this related to this change or has it crept in from somewhere else?

mpw5 (approved):
I think this mostly looks good, just one question about a change to schema.rb which looks like it might not belong here

PR #1929, opened 2020-10-30
Helper method to compare HTML in Rspec

stephenrichards:
## Helper method to compare HTML in Rspec

Rspec occasionally fails if Faker generates a firm name (or person's name) with an apostrophe.
the line

   `expect(response.body).to include(firm.name)`

will fail if the firm name is "O'Keefe & Sons" because the response body contains the special 
characters apostrophe and ampersand encoded into HTML entities.  

In a further little twist sent just to torment us, it encodes most special characters as the named enitity, e.g. the ampersand as `&amp;` but the apostrophe as a decimal entity `&#39;` rather than the more usual  `&apos;`.

This PR adds a method `html_compare` to `rails_helper.rb` which can be used to overcome that, and so the line above can be written

   `expect(response.body).to include(html_compare(firm.name))`

and always pass no matter what Faker throws at it.

## Checklist

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -111,3 +111,11 @@ def set_session(vars = {}) # rubocop:disable Naming/AccessorMethodName
     expect(session[var]).to be_present
   end
 end
+
+# encodes strings to HTML so that comparisons like "O'Keefe & Sons" don't fail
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
LGTM :+1:

PR #1930, opened 2020-10-30
AP-1744: Cucumber update

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1744)

We have previously locked cucumber at v3 because the gem update broke it...
Investigations matched @mpw5 previous work and highlighted a monkey-patch that could be applied to resolve the interaction between cucumber and vcr.

As a side test I tried moving the seed load outside of the `Before` block but this caused issues with the data being truncated and therefore not available for certain tests.

I switched the database strategy from `:trunaction` to `:transaction`.  This carries a warning about thread safety but it has reliably run on circle and locally.

## Checklist


Obsiye (approved):
Awesome!

PR #1933, opened 2020-10-30
AP-1758 Is the applicant employed

Obsiye:
## AP-1758 Is the applicant employed

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1758)

Add employed column to Applicants table
Add applicant_employed route, employed form and applicant_employed controller

Add UseCcmsEmployed page, route, controller and view which has a different view to the use_ccms page

Update feature tests

Create request tests for the applicant_employed and use_ccms_employed pages
Create form test for employed form

## Update flow for applicant_not_employed:

**When negative benefit check result and non passported with **used delegated function** then the page flow like:**
```
    Is your client employed? > NO
    Check if you can continue using this service > YES
    Do you want to make a substantive application now?
```
**When negative benefit check result and non passported and doesn't use **used delegated function** then the page flow     like:** ( no substantive application page )

    Is your client employed? > NO
    Check if you can continue using this service > YES
    What your client has to do ...

## The flow for applicant_employed:
```
   Is your client employed? > YES
   use_ccms_applicant_employed page
```
## Checklist


config/locales/cy/activemodel.yml
```diff
@@ -3,19 +3,465 @@ cy:
   accessibility:
     problem_text: melborp a si erehT
   activemodel:
+    attributes:
```
colinbruce, line 6: Do we need the extra translations? Only the citizen pages are being served in dual languages.

app/services/flow/flows/provider_start.rb
```diff
@@ -55,30 +55,44 @@ class ProviderStart < FlowSteps
...
+          forward: ->(application) do
+            application.applicant_employed? ? :use_ccms_employed : :open_banking_consents
+          end
+        },
         open_banking_consents: {
           path: ->(application) { urls.providers_legal_aid_application_open_banking_consents_path(application) },
-          forward: ->(application) { application.provider_received_citizen_consent? ? :non_passported_client_instructions : :use_ccms }
+          forward: ->(application) do
+            next_step = :non_passported_client_instructions
+            next_step = :substantive_applications if application.applicant_employed? == false && application.used_delegated_functions?
```
willc-work, line 92: could you use `if !application.applicant_employed? && ...`

spec/requests/providers/open_banking_consents_spec.rb
```diff
@@ -64,8 +64,43 @@
...
+
+          let(:applicant) { create :applicant, :not_employed }
+
+          it 'redirects to the client instructions page' do
+            expect(response).to redirect_to(providers_legal_aid_application_non_passported_client_instructions_path(application))
+          end
+        end
+      end
+
+      context 'positive beenfit check result' do
```
colinbruce, line 93: nitpick: benefit misspelled

colinbruce (approved):
A couple of observations, one v. minor! 
Otherwise, LGTM  👍

spec/forms/applicants/employed_form_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      expect(subject.save).to be false
+      expect(subject.errors[:employed]).to eq [I18n.t('activemodel.errors.models.applicant.attributes.employed.blank')]
+    end
+  end
+
+  describe '#save' do
+    let(:params) { { employed: 'false' } }
+    let(:form_params) { params.merge(model: applicant) }
+
+    it 'saves updates record with new value of employed attribute' do
```
willc-work, line 25: either saves or updates work but perhaps not both

app/services/flow/flows/provider_start.rb
Obsiye: i  said 'false' explicitly because applicant_employed could be nil e.g. if the benefit check result is positive then it wouldn't have employed info  , so i didn't know how else to ignore nil

spec/requests/providers/applicant_employed_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+        it 'returns http_success' do
+          expect(response).to have_http_status(:ok)
+        end
+
+        it 'does not update the record' do
+          applicant = legal_aid_application.reload.applicant
+          expect(applicant.reload.employed).to be_nil
+        end
+
+        it 'the response includes the error message' do
```
willc-work, line 92: ' includes the error message in the response' works better with the 'it'

willc-work (approved):
minor comments on test wording, nothing major. looks good

PR #1934, opened 2020-10-30
AP-1582 Valuable items content and validation

eloisetait3:
*What*
[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1582&assignee=5f968777632c6200712e443f)

Description
1/ Change check box text from ‘Any valuable items worth over £500’ to ‘Any valuable items worth £500 or more’.
Reason: our current text does not allow for items worth exactly £500.

2/ Change hint text from ‘Enter the estimated value, minus any sale costs’ to ‘Enter the estimated value’.
Reason: spoke to Dave about this and he agreed that asking for the estimated value is sufficient here, without the added burden of the provider or client having to guess or find out ‘sale costs’.

3/ Validation and Error message if value less than £500 entered:
Estimated value must be £500 or more

User Story
As a provider,
I want to understand what I am being asked for my client’s valuable items and I want to be told if there is an error if I have put an amount that is less than £500
so that I can be as accurate as possible with the means assessment

*Checklist*

### DIFF
--- a/app/forms/citizens/other_assets_form.rb
+++ b/app/forms/citizens/other_assets_form.rb
@@ -22,13 +22,16 @@ module Citizens
 
     OTHER_CHECKBOXES = %i[check_box_second_home check_box_valuable_items_value none_selected].freeze
 
+    INDIVIDUALLY_VALIDATED = %i[second_home_percentage valuable_items_value].freeze
+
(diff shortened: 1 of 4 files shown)


stephenrichards (approved):

stephenrichards:
Looks good Eloise.  This is your first PR isn't it?  Congratulations. many more to come.

PR #1940, opened 2020-11-02
AP-1636 updated content on means calculation result eligibility page

eloisetait3:
*What*

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1636&assignee=5f968777632c6200712e443f)

The following content is potentially misleading as it sounds like getting legal aid is a formality once the case details are entered:
You still have to provide details of the case before we can grant legal aid.

Change it to:
You still need to provide details of the case before we can decide if they’re eligible.

All 6 instances of this page should be updated.

*Checklist*

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -66,7 +66,7 @@ en:
         heading: '%{name} could be eligible for legal aid'
         details:
           - Based on their financial situation, your client may not have to pay towards legal aid.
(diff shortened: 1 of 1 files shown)


mpw5 (approved):
Looks good to me 🎉

PR #1950, opened 2020-11-04
AP-1673 Add language switcher

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1673)

Adds and enables a small component to allow applicants to switch language between English and Welsh. This will be present at the top-right of the page, aligned with the Back link (if present).

* Enabling the component on citizen pages (but not provider or admin pages) is done by setting a session variable `journey_type` as `:citizens` when an applicant accesses the service. This allows the language switcher to be displayed on screens that do not contain `/citizens/` in the path, for example errors, feedback, accessibility statement, contact.

* The language switcher is not displayed on `/citizens/gather_transactions` as switching there will cause bank data to be redownloaded.

* During testing it was found that by carrying out the applicant journey in Welsh and then returning to the provider journey could result in provider screens being displayed in Welsh. I think this only happens when both journeys are carried out on the same device, so is unlikely to ever happen in production, but to try and mitigate this some code has been added to force provider screens to update the locale.

## Checklist


stephenrichards (approved):

stephenrichards:
Nice.  I'd forgotten about the error pages and stuff.

mpw5:
> Nice. I'd forgotten about the error pages and stuff.

Thanks. Yeah, it started off as a really straightforward change and then seemed to spiral as I kept finding more and more pages that needed the switcher 🤦

PR #1955, opened 2020-11-05
AP-1781 BugFix for flow on delete path

Obsiye:
## AP-1781 BugFix for flow on delete path

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1781)

Add a FlowStep for the delete page to prevent the below error:

```
Flow::FlowError
:path of step :delete is not defined
```

The error occurred when clicking on the delete page as it was not a part of the flow and had no path for step :delete

## Checklist


stephenrichards (commented):
Were you able to replicate this before the change?  I couldn't.  We should really have a test to replicate the error which then passes when you apply the change.

stephenrichards (approved):

PR #1956, opened 2020-11-05
Remove webhint from Rake

mpw5:
## What

Webhint runs as part of the `rake` test suite. It usually fails. We've removed it from the CircleCI pipeline so it also makes sense to also remove it here, replacing it with `cucumber` so that the feature tests still run.

Webhint can still be run locally by running `rake webhint:generate_reports`.

## Checklist


Obsiye (approved):
LGTM

Issue #1468, opened 2020-05-27
Use Rails 6 default autoloader called zeitwerk


Obsiye:
There was an error during the upgrading of Rails 6 related to the autoloader, zeitwerk.

The temporary solution was to (https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/1240): 

- Revert to using classic as the rails autoloader instead of the rails 6 default of zeitwerk. This fixed issues relating to a loading in rails_helper require error :
```
File.expand_path('../../config/environment', __FILE__)
FrozenError:
  can't modify frozen
```

**TODO**:

Use the Rails 6 default autoloader - zeitwerk

PR #1984, opened 2020-11-11
Tech Debt: Enable zeitwerk autoloader

mpw5:
## What

Fixes https://github.com/ministryofjustice/laa-apply-for-legal-aid/issues/1468

Rails 6 now uses the [`zeitwerk`](https://github.com/fxn/zeitwerk) autoloader by default. When Apply was migrated to Rails 6 this caused problems so we reverted to the `classic` autoloader.

This change enables `zeitwerk` by removing the override to `classic`. `zeitwerk` requires strict camelcase and does not play nicely with acronyms. To resolve this I have added inflections for common acronyms used in Apply (`CCMS`, `CFE`, `DWP`, `MIS`) and ensured consistency in the way those acronyms are used in the code.

The alternative is to not add the inflections and instead change all instances of `CCMS` to `Ccms`, `CFE` to `Cfe`, etc, which results in a much greater number of changes.

## Checklist


colinbruce (approved):
Nice :+1:

colinbruce (approved):
🏆

PR #1985, opened 2020-11-11
Remove timecop gem

Obsiye:
## Remove timecop gem

Suggestion to remove timecop gem and replace it with rails [ActiveSupport::Testing::TimeHelpers ](https://api.rubyonrails.org/v5.2.4/classes/ActiveSupport/Testing/TimeHelpers.html)

Non blocker issue:

- The TimeHelper turns nanoseconds to 0 and this means tests for the #transaction_request_id will have 0 in place of the nanosecond making the test less accurate bot not wrong.
> Note that the usec for the time passed will be set to 0 to prevent rounding errors with external services, like MySQL (which will round instead of floor, leading to off-by-one-second errors).

## Checklist

### DIFF
--- a/spec/services/ccms/requestors/applicant_add_requestor_spec.rb
+++ b/spec/services/ccms/requestors/applicant_add_requestor_spec.rb
@@ -4,7 +4,7 @@ module CCMS
   module Requestors
     RSpec.describe ApplicantAddRequestor do
       let(:expected_xml) { ccms_data_from_file 'applicant_add_request.xml' }
-      let(:expected_tx_id) { '20190101121530123456' }
+      let(:expected_tx_id) { '20190101121530000000' }
 
       let(:address) do
         create(:address,
@@ -41,7 +41,7 @@ module CCMS
       end
       describe '#transaction_request_id' do
         it 'returns the id based on current time' do
-          Timecop.freeze(2019, 1, 1, 12, 15, 30.123456) do
+          travel_to Time.local(2019, 1, 1, 12, 15, 30) do
             expect(requestor.transaction_request_id).to start_with expected_tx_id
           end
         end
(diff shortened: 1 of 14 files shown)


spec/services/ccms/requestors/applicant_add_requestor_spec.rb
```diff
@@ -4,7 +4,7 @@ module CCMS
   module Requestors
     RSpec.describe ApplicantAddRequestor do
       let(:expected_xml) { ccms_data_from_file 'applicant_add_request.xml' }
-      let(:expected_tx_id) { '20190101121530123456' }
+      let(:expected_tx_id) { '20190101121530000000' }
```
Obsiye, line 7: As you can see the nanoseconds of 123456 is now 000000 because of TimeHelper

spec/services/reports/mis/non_passported_applications_report_spec.rb
```diff
@@ -5,9 +5,18 @@ module MIS
     RSpec.describe NonPassportedApplicationsReport do
       before do
         create_early_non_passported_application
-        Timecop.freeze(8.minutes.ago) { create_non_passported_application }
-        Timecop.freeze(6.minutes.ago) { create_non_passported_application_use_ccms }
-        Timecop.freeze(4.minutes.ago) { create_passported_application }
+        travel_back
+        travel(8.minutes) { create_non_passported_application }
```
colinbruce, line 9: Does this change the sequence? 
Previously we had 8.minutes.ago, now we have 9 minutes in the future.
Is this why the expected values changed at line 30?

Obsiye: Yes that's it XD. Great spot. I meant to use travel_to and 8.minutes ago to go to the past.

willc-work (approved):
LGTM and one less dependency👍

PR #1989, opened 2020-11-12
AP1425 Suppress Sentry alerts for postcode not found

Obsiye:
## AP1425 Suppress Sentry alerts for postcode not found

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1425)

- Don't send a sentry message if the user has typed a postcode with the correct format but was a mistype.
- test the changes

## Checklist

### DIFF
--- a/app/services/address_lookup_service.rb
+++ b/app/services/address_lookup_service.rb
@@ -2,7 +2,6 @@ class AddressLookupService
   prepend SimpleCommand
 
   ORDNANCE_SURVEY_URL = 'https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode'.freeze
-  POSTCODE_REGEXP = /\A([A-Z][A-HJ-Y]?[0-9][A-Z0-9]?[0-9][A-Z]{2}|GIR ?0A{2})\z/.freeze
 
   attr_reader :postcode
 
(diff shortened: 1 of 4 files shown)


app/services/address_lookup_service.rb
```diff
@@ -2,6 +2,7 @@ class AddressLookupService
   prepend SimpleCommand
 
   ORDNANCE_SURVEY_URL = 'https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode'.freeze
+  POSTCODE_REGEXP = /\A([A-Z][A-HJ-Y]?[0-9][A-Z0-9]?[0-9][A-Z]{2}|GIR ?0A{2})\z/.freeze
```
colinbruce, line 5: Does this need to be duplicated?  There is already a version in settings, can that be referenced?

app/services/address_lookup_service.rb
```diff
@@ -55,6 +56,6 @@ def parse_successful_response(response)
 
   def record_error(state, error)
     errors.add(:lookup, state)
-    Raven.capture_exception(error)
+    Raven.capture_exception(error) unless POSTCODE_REGEXP.match?(@postcode) && state == :unsuccessful
```
colinbruce, line 59: I think this would 'just work' without needing to copy the CONST?

colinbruce (approved):
Nice, one non-🚧  would be to move the comments about the REGEX from https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/1989/files#diff-4c2d4bccb73a3296911380789e5eb5e1bd37b277d07e743e6a6441879528cee6R1-R6 to the settings file, but otherwise :+1:

PR #1991, opened 2020-11-12
AP-1640 Remove feature flag code for non-passported applications

eloisetait3:
**What**

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-1640&assignee=5f968777632c6200712e443f)

Feature flag code is removed for non-passported, including the flag mechanism on the admin/settings page. This is so providers with non-passported rights can progress past DWP checks.

**Checklist**


features/step_definitions/civil_journey_steps.rb
```diff
@@ -707,11 +707,9 @@
 # rubocop:enable Lint/Debugger
 
 Given('I start the passported journey') do
-  LaaApplyForLegalAid::Application.config.x.allow_non_passported_route = false
```
colinbruce, line 710: If the step is empty it should probably be removed completely

features/step_definitions/civil_journey_steps.rb
```diff
@@ -707,11 +707,9 @@
...
 end
 
 Then('I complete the passported journey') do
-  LaaApplyForLegalAid::Application.config.x.allow_non_passported_route = true
```
colinbruce, line 714: If the step is empty it should probably be removed completely too, along with references to it in the features

colinbruce (changes requested):
Happy to discuss, but some changes may be needed?

features/step_definitions/civil_journey_steps.rb
eloisetait3: Yep can do! Will removing it only affect Cucumber if feature references aren't removed?

colinbruce: It should do! :)

colinbruce (approved):
Nice :+1

PR #1994, opened 2020-11-12
AP-1595 Enable ActiveStorage sidekiq queue

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1595)

ActiveStorage is enqueing `active_storage_analysis` and `active_storage_purge ` jobs which are never being processed because sidekiq does not know about the queues. There are currently ~6000 analysis and ~60 purge jobs queued in production. This change adds queue details to `config/sidekiq.yml` so that those jobs will be processed.

## Checklist

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -3,3 +3,5 @@
   - default
   - mailers
   - sidekiq_alive
+  - active_storage_analysis
+  - active_storage_purge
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
Nice :+1:
Waiting for the sidekiq 💣  to explode as it starts to run those! :)

PR #2002, opened 2020-11-13
Ap 1789 portal auth error

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1789)

When a CCMS user without Apply permissions tries to login they get an error due to a malformed SAML response from the portal. 

This change will capture the error and redirect the user to the access denied page

The `/invalid_login` page would appear to be the more appropriate page to redirect the user to, but this currently requires the user to have an account created for them when the login failed, however due to the SAML response this doesnt happen.

## Checklist


colinbruce (approved):
:+1:

PR #2008, opened 2020-11-16
AP-1804 Add MimeMagic gem to validate file types

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1804)

There is a security vulnerability in the way we check for valid statement of case file types using the file's `content_type`, which can be forged in HTTP requests.

Instead we can use the `MimeMagic` gem, which is able to examine the content of the file itself in order to identify the type more accurately.  This was chosen over the similar `Ruby-FileMagic` gem as it's easier to implement and better maintained.

ETCj

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -98,6 +98,9 @@ gem 'blazer', '>= 2.2.6'
 # Manage security headers
 gem 'secure_headers'
 
+# Identify file types before uploads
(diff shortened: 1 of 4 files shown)


Obsiye (approved):
LGTM

PR #2015, opened 2020-11-18
AP-1805 Add account details to transaction csv

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1805)

Adds sort code and account number to the bank transaction csv so that caseworkers are able to identify accounts more easily.

## Checklist

### DIFF
--- a/app/presenters/bank_transaction_presenter.rb
+++ b/app/presenters/bank_transaction_presenter.rb
@@ -9,6 +9,8 @@ class BankTransactionPresenter
     running_balance: 'balance/running total',
     account_type: 'account type',
     account_name: 'account name',
+    account_sort_code: 'sort code',
+    account_number: 'account number',
     category: 'category',
     selected_by: 'selected_by',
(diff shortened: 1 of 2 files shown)


app/presenters/bank_transaction_presenter.rb
```diff
@@ -9,6 +9,8 @@ class BankTransactionPresenter
     running_balance: 'balance/running total',
     account_type: 'account type',
     account_name: 'account name',
```
Alex-Swann, line 11: For a refactor, couldn't CELLS just be an array of syms instead and then you have the following methods. Saves the duplication of keys and values.
```
  def self.headers
    CELLS.map(&:to_s)
  end

  def build_transaction_hash
    values = {}
    CELLS.map { |k| values[k.to_sym] = send("transaction_#{k}") }
    values
  end
```

Alex-Swann (approved):

PR #2020, opened 2020-11-19
AP-1726 Standardise usage of radio buttons

Obsiye:
## AP-1726 Standardise usage of radio buttons

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1726)

- Update form builder method #yes_no_radio_button_array to use the value true or false as true or false has been used in many more places as the radio button value than yes or no.
- Update relevant controllers to expect true or false instead of yes or no

- Update views to pass in (form.yes_no_radio_button_array, :value, :label,...) to the form.govuk_collection_radio_buttons method

- Update tests

## Checklist

### DIFF
--- a/app/views/providers/confirm_offices/show.html.erb
+++ b/app/views/providers/confirm_offices/show.html.erb
@@ -19,23 +19,22 @@
     <%= govuk_form_group(
           show_error_if: @error,
         ) do %>
+      <fieldset class="govuk-fieldset">
+        <%= govuk_fieldset_header page_title, padding_below: 4 %>
 
-      <%= form.govuk_collection_radio_buttons(
-            :correct,
-            [
-              { value: :yes, label: t('generic.yes') },
-              { value: :no,  label: t('.no_another_office') }
-            ],
-            :value,
-            :label,
-            error: @error,
-            title: {
-                      text: t('.h1-heading', office_code: current_provider.selected_office.code),
-                      htag: :h1,
-                      padding_below: 6
-            }
-          ) %>
+        <%= form.govuk_radio_button(
+                :correct,
+                true,
+                label: t('generic.yes')
+            ) %>
 
+        <%= form.govuk_radio_button(
+                :correct,
+                false,
+                label: t('.no_another_office')
+            ) %>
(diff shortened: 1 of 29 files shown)


app/views/providers/confirm_offices/show.html.erb
```diff
@@ -19,23 +19,22 @@
     <%= govuk_form_group(
           show_error_if: @error,
         ) do %>
+      <fieldset class="govuk-fieldset">
+        <%= govuk_fieldset_header page_title, padding_below: 4 %>
```
colinbruce, line 23: Question: Is 4 the standard padding_below?  The deleted one was 4
Not a blocker, hopefully a move for standardising!

app/views/providers/no_outgoings_summaries/show.html.erb
```diff
@@ -36,20 +36,12 @@
...
-              ) %>
-
-          <%= form.govuk_radio_button(
-                  :confirm_no_outgoings,
-                  'no',
-                  label: t('generic.no'),
-                  error: @error&.values&.first
-              ) %>
+        <div class="govuk-radios govuk-radios--conditional" data-module="govuk-radios">
+          <%= form.govuk_collection_radio_buttons :confirm_no_outgoings,
```
colinbruce, line 40: Could this pattern be used on the previous change, `no_income_summaries`, too?

colinbruce (commented):
Looking good, one possible further change and an observation 👍

app/views/providers/confirm_offices/show.html.erb
Obsiye: Yep, does the same thing. It was adding a padding below the title. So the govuk_collection_radio_buttons was also creating the page title with a padding below it. It's doing too much 😂

app/views/providers/no_outgoings_summaries/show.html.erb
Obsiye: Good spot! Updated now.

colinbruce (approved):
LGTM 👍

PR #2022, opened 2020-11-19
Allow microsoft word documents to be uploaded on the statement of case page

Obsiye:
## Allow microsoft word documents to be uploaded on the statement of case page

- Allow docx and doc file types

mimemagic gem returns content type application/zip for docx or doc type files and zip files are not allowed to be uploaded
https://github.com/minad/mimemagic#extra-magic-overlay

## Checklist


colinbruce (approved):
👍

PR #2032, opened 2020-11-23
AP-1553: Remove todos

Alex-Swann:
## What

## Checklist

### DIFF
--- a/spec/requests/providers/property_value_spec.rb
+++ b/spec/requests/providers/property_value_spec.rb
@@ -62,7 +62,6 @@ RSpec.describe Providers::PropertyValuesController, type: :request do
           context 'property is owned outright' do
             let(:legal_aid_application) { create :legal_aid_application, :with_applicant, :with_own_home_owned_outright }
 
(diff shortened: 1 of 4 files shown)


spec/requests/providers/property_value_spec.rb
```diff
@@ -62,7 +62,6 @@
           context 'property is owned outright' do
             let(:legal_aid_application) { create :legal_aid_application, :with_applicant, :with_own_home_owned_outright }
 
-            # TODO: replace with correct path once other controllers are ready
```
Alex-Swann, line 65: Wasn't sure about this one. It appears to be a bit superfluous, mainly because if there was a change, it sounds like this test would break and need changing anyway.

colinbruce: This looks fine to delete. The test below has been updated 12 times so I'd assume that, in one of those, the URL was updated 🤷‍♂️

willc-work (approved):
LGTM

PR #2044, opened 2020-11-25
ap-1875 display the student loan value

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1875)

We are no longer using the monthly_equivalent calculator to calculate the student loan value. Student loan is now under irregular income instead of other_income and the annual sum is divided by 12. 

Changed `mei_student_loan` to take the value from cfe_result.gross_income[:monthly_student_loan] This ensures that wherever `mei_student_loan` is used (particularly to display results) we now access the correct value.

The `monthly_student_loan` value is already being included in the calculations for total monthly/disposable income etc. So the totals and calculations are all still correct it is only on pages where we breakdown by income type that the values are not being displayed.

## Checklist

### DIFF
--- a/app/models/cfe/v2/result.rb
+++ b/app/models/cfe/v2/result.rb
@@ -194,7 +194,7 @@ module CFE
       end
 
       def mei_student_loan
-        monthly_income_equivalents[:student_loan].to_d
+        gross_income[:monthly_student_loan].to_d
       end
 
       def mei_pension
--- a/spec/factories/cfe_results.rb
+++ b/spec/factories/cfe_results.rb
@@ -44,6 +44,10 @@ FactoryBot.define do
       result { CFEResults::V2::MockResults.with_maintenance_received.to_json }
     end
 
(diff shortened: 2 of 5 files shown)


app/models/cfe/v2/result.rb
```diff
@@ -194,7 +194,7 @@ def mei_property_or_lodger
       end
 
       def mei_student_loan
```
Obsiye, line 196: This could do with a test

app/models/cfe/v2/result.rb
```diff
@@ -194,7 +194,7 @@ def mei_property_or_lodger
...
-        monthly_income_equivalents[:student_loan].to_d
+        gross_income[:monthly_student_loan].to_d
```
Obsiye, line 197: This could do with a test

Obsiye (approved):
LGTM 💯

Obsiye (approved):

PR #2051, opened 2020-11-26
Ap 1807 applicant email

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1807)

Added in new parameter for client emails to include the date on which the url expires. 

URL is set to expire 7 days from the date it is issued. Reminder emails have the date set on the day they are scheduled

For UAT testing this has 3 new notify templates. Once it has passed UAT these will need to be removed and the original templates updated to include the new wording and date. This will then need to be tested in staging before release to production.

## Checklist

### DIFF
--- a/app/mailers/concerns/notify_template_methods.rb
+++ b/app/mailers/concerns/notify_template_methods.rb
@@ -15,4 +15,8 @@ module NotifyTemplateMethods
   def safe_nil(value)
     value || ''
   end
+
+  def url_expiry_date
+    (Date.today + 7.days).strftime('%d %B %Y')
+  end
 end
--- a/app/mailers/concerns/notify_template_methods.rb
+++ b/app/mailers/concerns/notify_template_methods.rb
@@ -17,6 +17,6 @@ module NotifyTemplateMethods
   end
 
   def url_expiry_date
-    (Date.today + 7.days).strftime('%d %B %Y')
+    (Date.today + 7.days).strftime('%-d %B %Y')
   end
 end
--- a/app/mailers/notify_mailer.rb
+++ b/app/mailers/notify_mailer.rb
@@ -16,10 +16,4 @@ class NotifyMailer < BaseApplyMailer
     )
     mail(to: email)
   end
-
-  private
-
-  def url_expiry_date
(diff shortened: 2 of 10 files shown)


app/mailers/notify_mailer.rb
```diff
@@ -11,8 +11,15 @@ def citizen_start_email(app_id, email, application_url, client_name, provider_fi
...
-      ref_number: app_id
+      ref_number: app_id,
+      expiry_date: url_expiry_date
     )
     mail(to: email)
   end
+
+  private
+
+  def url_expiry_date
```
mpw5, line 22: Not a blocker, but I wonder if you could add this method to the `NotifyTemplateMethods` module (which is included in each of the mailers) instead of defining it in each?

mpw5 (approved):
LGTM. One suggestion about a possible refactor but not a blocker

app/mailers/notify_mailer.rb
willc-work: Yes, good point.

app/mailers/concerns/notify_template_methods.rb
```diff
@@ -15,4 +15,8 @@ def support_email_address
   def safe_nil(value)
     value || ''
   end
+
+  def url_expiry_date
+    (Date.today + 7.days).strftime('%d %B %Y')
```
Alex-Swann, line 20: Do we know if the GDS style guide specifies zero padding or not. For some reason I thought it is against zero padding as when months are numeric they appear as `3` not `03`. As per the date hint text on this page https://design-system.service.gov.uk/patterns/dates/. If so, then this needs to be `%-d`

Alex-Swann: Yeah it looks like it does in this example. Notice it is `4` and not `04` https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style 

- Dates
- use upper case for months: January, February
- do not use a comma between the month and year: 4 June 2017

willc-work: Updated thanks, good catch

PR #2058, opened 2020-11-27
AP-1496 Accessible Statement of Case delete links

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1496)

Make deleting statement of case files more accessible for screen reader users by including the filename in the button name, but hidden so that the name can be read out without changing the button's appearance. 

Reuses the pattern already established for doing this with links, adding an additional helper method so that this can be used elsewhere in future.

## Checklist


colinbruce (approved):
👍 LGTM!

app/helpers/accessible_link_helper.rb
```diff
@@ -4,4 +4,10 @@ def link_to_with_hidden_suffix(path:, text:, suffix:, klass: nil)
       "#{text}<span class='govuk-visually-hidden'> #{suffix}</span>".html_safe
     end
   end
+
+  def button_to_with_hidden_suffix(path:, text:, method:, params:, suffix:, klass: nil)
```
Alex-Swann, line 8: @mpw5 don't need a test for this method do we?

mpw5: 🤦 Excellent point. It has one now.

Feel free to re-review 👍

PR #2080, opened 2020-12-02
AP-1881 Add debugging and reset ccms_reason

stephenrichards:
## Add debugging and reset ccms_reason

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1881)

This PR fixes two issues, and introduces debugging to track down another

Fixed: 
 - the ccms_reason wasn't being reset to nil in all cases when transitioning from use_ccms to another state.
 - adds `expire_now` to the `Admin::ReportsController` actions to produce the CSV files to prevent caching and responding with stale data.

Debugging:
 - in all cases where we transition to use_ccms state with a reason of unknown, print a back trace and the referer URL

## Checklist

### DIFF
--- a/app/models/concerns/base_state_machine.rb
+++ b/app/models/concerns/base_state_machine.rb
@@ -83,6 +83,9 @@ class BaseStateMachine < ApplicationRecord  # rubocop:disable Metrics/ClassLengt
                   after: proc { |reason|
                     raise 'Invalid ccms_reason' unless reason.in?(VALID_CCMS_REASONS)
 
+                    # logging to find out where unknonw ccms reasons are coming from
+                    Raven.capture_message "UNKNOWN CCMS REASON\n#{caller}" if reason.to_sym == :unknown
+
                     update!(ccms_reason: reason)
                   }
     end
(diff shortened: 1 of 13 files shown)


app/models/concerns/base_state_machine.rb
```diff
@@ -83,6 +83,9 @@ class BaseStateMachine < ApplicationRecord  # rubocop:disable Metrics/ClassLengt
                   after: proc { |reason|
                     raise 'Invalid ccms_reason' unless reason.in?(VALID_CCMS_REASONS)
 
+                    # logging to find out where unknonw ccms reasons are coming from
+                    Raven.capture_message "UNKNOWN CCMS REASON\n#{caller}" if reason.to_s == 'unknown'
```
Alex-Swann, line 87: couldn't you just do `reason == :unknown` to save having to create strings?

Alex-Swann (approved):
LGTM bar one minor comment

app/models/concerns/base_state_machine.rb
stephenrichards: Well, I don't know if the reason is being passed in as a symbol or a string, so this makes sure it captures both possibilities

Alex-Swann: `reason.to_sym == :unknown` then?

Alex-Swann (approved):
Thanks for the explanation. It makes it clear what the additional bit you've added is doing.

PR #2085, opened 2020-12-04
Record CFE API interaction

stephenrichards:
## Record CFE API interaction

Sometimes it is necessary to replay what apply sent to CFE in order to debug something in CFE.  This PR introduces that ability with a rake task that will read the CFE::SubmissionHistory records for an application, and output the details as YAML which can then be replayed locally to CFE with the `rake replay` task in the CFE repo.

### Usage

  `rake cfe:record_payloads[L-A12-B34]`

This will print out YAML which can be copied and pasted into the `tmp/api_replay.yml` file in the CFE project ready for the `rake replay` task.

## Checklist

### DIFF
--- a/db/sample_data/bank_transactions_3.csv
+++ b/db/sample_data/bank_transactions_3.csv
@@ -1,0 +1,3 @@
+date,description,amount_out,amount_in
+06/11/2020,DWP AA123456A JSA,,148.70
+23/10/2020,DWP AA123456A JSA,,43.49
--- a/db/sample_data/bank_transactions_3.csv
+++ b/db/sample_data/bank_transactions_3.csv
@@ -1,3 +1,0 @@
-date,description,amount_out,amount_in
-06/11/2020,DWP AA123456A JSA,,148.70
-23/10/2020,DWP AA123456A JSA,,43.49
--- a/lib/tasks/helpers/cfe_payload_recorder.rb
+++ b/lib/tasks/helpers/cfe_payload_recorder.rb
@@ -1,0 +1,36 @@
+class CfePayloadRecorder
+
+  def self.call(application_ref)
+    new(application_ref).call
+  end
+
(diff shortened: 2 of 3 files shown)


db/sample_data/bank_transactions_3.csv
```diff
@@ -0,0 +1,3 @@
+date,description,amount_out,amount_in
+06/11/2020,DWP AA123456A JSA,,148.70
+23/10/2020,DWP AA123456A JSA,,43.49
```
Obsiye, line 3: Do we still need this?

Obsiye (approved):
LGTM! Left a minor comment/question

lib/tasks/helpers/cfe_payload_recorder.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def call
+    submission = @legal_aid_application.most_recent_cfe_submission
+    submission.submission_histories.each do |history|
+      record_history(history)
+    end
+    puts '#'
+    puts "# CFE API calls for applicaiton #{@legal_aid_application.application_ref}"
+    puts '# Copy and paste this output in to tmp/api_replay.yml in the CFE repo, and '
```
Obsiye, line 20: Is there a way to make this BOLD so that we don't miss the fact that it says " `in the CFE repo` "

db/sample_data/bank_transactions_3.csv
stephenrichards: Nope, I'll remove it.

lib/tasks/helpers/cfe_payload_recorder.rb
stephenrichards: I don't think so - it is just output at the top of the YAML on the screen.  I guess we can make it in red. Let me try that.

PR #2091, opened 2020-12-10
AP-1907 Fix unknown ccms reason

stephenrichards:
## Fix bug: ccms reason set to unknown

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1907)

The `Providers::UseCCMSController` sets the state to `use_ccms` and the `ccms_reason` to a value depending on what page the request came from.  This does not work well if an application is already in `use_ccms` state, and it is viewed from the applications list page.  This bug is fixed by not calling '#use_ccms!' if the state is already `use_ccms`

## Checklist

### DIFF
--- a/app/controllers/providers/use_ccms_controller.rb
+++ b/app/controllers/providers/use_ccms_controller.rb
@@ -1,7 +1,7 @@
 module Providers
   class UseCCMSController < ProviderBaseController
     def show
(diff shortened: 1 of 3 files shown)


Alex-Swann (approved):
Thanks for explanation. It makes it clear what the additional bit you've added is doing. 👍

PR #2098, opened 2020-12-11
Ap 1854 refactor error summaries

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1894)

Replaced error summary sections in views to use the same shared partial.
Added an if/else to the error summary partial to deal with the 2 different types of errors that the summary has to handle
Updated the styling of the error summary to match what we have used elsewhere

## Checklist


mpw5 (approved):
Looks good to me 👍

PR #2107, opened 2020-12-15
AP-1893: Add cash income page

Alex-Swann:
## Add Cash Income pages

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1893)

This PR implements a Cash income page behind a feature flag on the settings page.  the Cash outgoings page is implemented in a separate ticket.

Cash transactions are stored on the cash_transactions table linked to the legal aid application and the transaction type.  Each transaction type for which values exist have three records, representing amounts paid in each of the last three months.

An `AggregatedCashIncome` model represents all three payments for all Transaction Types and is a facade object to facilitate displaying and capturing data on the form, validation and writing and reading the `CashTransaction` records.  Most of the code is in a base class that will be used by the cash outgoings page as well.

## Checklist

### DIFF
--- a/app/models/aggregated_cash_income.rb
+++ b/app/models/aggregated_cash_income.rb
@@ -1,0 +1,183 @@
+class AggregatedCashIncome # rubocop:disable Metrics/ClassLength
+  include ActiveModel::Model
+  include MoneyHelper
+
+  CATEGORIES = %i[
+    benefits
+    friends_or_family
+    maintenance_in
+    property_or_lodger
+    pension
+  ].freeze
+
+  NUMBER_OF_FIELDS = 3
+
+  OPERATION = :credits
+
+  MONTH_RANGE = (1..NUMBER_OF_FIELDS).freeze
+
+  def self.category_attributes(categories)
+    categories.each do |category|
+      attr_accessor "check_box_#{category}".to_sym
+
+      MONTH_RANGE.each do |i|
+        attr_accessor "#{category}#{i}".to_sym
+      end
+    end
+  end
+
+  def self.populate_attribute(model, trx)
+    month_no = month_number(trx)
+    value_method = "#{trx.transaction_type.name}#{month_no}="
+    checkbox_method = "check_box_#{trx.transaction_type.name}="
+    model.__send__(value_method, trx.amount)
+    model.__send__(checkbox_method, 'true')
+  end
+
+  def self.month_number(trx)
+    current_month = trx.updated_at.to_date.at_beginning_of_month
+    MONTH_RANGE.each do |i|
+      return i if (current_month - i.months) == trx.transaction_date
+    end
+  end
+
+  def self.find_transactions(id)
+    transaction_type_ids = TransactionType.credits.map(&:id)
+
+    CashTransaction.where(
+      legal_aid_application_id: id,
+      transaction_type_id: transaction_type_ids
+    )
+  end
+
+  attr_accessor :legal_aid_application_id,
+                :none_selected
+
+  category_attributes CATEGORIES
+
+  validate :validate_none_selected
+  validate :validate_at_least_one_selected
+  validate :checkbox_integrity
+
+  def self.find_by(legal_aid_application_id:)
+    transactions = find_transactions(legal_aid_application_id)
+    model = new
+    transactions.each { |trx| populate_attribute(model, trx) }
+    model
+  end
+
+  def update(params)
+    update_attributes(params)
+
+    return false unless valid?
+
+    save_cash_transaction_records
+  end
+
+  private
+
+  def checkbox_integrity
+    CATEGORIES.each do |category|
+      checkbox_attr = "check_box_#{category}".to_sym
+
+      if __send__(checkbox_attr) == 'true'
+        check_all_values_set_for category
+      else
+        erase_values_for category
+      end
+    end
+  end
+
+  def validate_at_least_one_selected
+    return if none_selected.present?
+
+    CATEGORIES.each do |category|
+      checkbox_attr = "check_box_#{category}".to_sym
+      return false if __send__(checkbox_attr) == 'true'
+    end
+
+    errors.add(:none_selected, model_error(:blank))
+  end
+
+  def validate_none_selected
+    return unless none_selected.present?
+
+    CATEGORIES.each do |category|
+      checkbox_attr = "check_box_#{category}".to_sym
+      next unless __send__(checkbox_attr) == 'true'
+
+      errors.add(:none_selected, model_error(:others_present))
+      break
+    end
+  end
+
+  def check_all_values_set_for(category) # rubocop:disable Metrics/AbcSize
+    MONTH_RANGE.each do |i|
+      value_attr = "#{category}#{i}".to_sym
+      value = __send__(value_attr)
+
+      errors.add(value_attr, amount_error(:blank)) unless value.present?
+      errors.add(value_attr, amount_error(:invalid_type)) unless number? value
+      errors.add(value_attr, amount_error(:negative)) unless valid_amount? value
+      errors.add(value_attr, amount_error(:too_many_decimals)) unless CurrencyValidator::ONLY_2_DECIMALS_PATTERN.match? value
+    end
+  end
+
+  def model_error(type)
+    I18n.t("activemodel.errors.models.aggregated_cash_income.#{OPERATION}.attributes.none_selected.#{type}")
+  end
+
+  def amount_error(type)
+    I18n.t("errors.cash_amount.#{type}")
+  end
+
+  def erase_values_for(category)
+    MONTH_RANGE.each do |i|
+      method = "#{category}#{i}=".to_sym
+      __send__(method, nil)
+    end
+  end
+
+  def update_attributes(params)
+    params.each do |key, value|
+      __send__("#{key}=", value)
+    end
+  end
+
+  def save_cash_transaction_records
+    CATEGORIES.each do |category|
+      CashTransaction.where(
+        legal_aid_application_id: legal_aid_application_id,
+        transaction_type_id: transaction_type_id(category)
+      ).destroy_all
+
+      save_category(category) if checkbox_for(category) == 'true'
+    end
+  end
+
+  def save_category(category)
+    MONTH_RANGE.each do |i|
+      amount = __send__("#{category}#{i}".to_sym)
+      date = calculated_date(i)
+
+      CashTransaction.create!(
+        legal_aid_application_id: legal_aid_application_id,
+        transaction_type_id: transaction_type_id(category),
+        transaction_date: date,
+        amount: amount
+      )
+    end
+  end
+
+  def transaction_type_id(category)
+    TransactionType.__send__(OPERATION).find_by(name: category)&.id
+  end
+
+  def calculated_date(month_number)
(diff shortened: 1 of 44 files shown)


app/models/cash_transaction.rb
```diff
@@ -0,0 +1,3 @@
+class CashTransaction < ApplicationRecord
+  has_one :cash_transaction_type
```
stephenrichards, line 2: I think this should be belongs_to: cash_transaction_type

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,75 @@
...
+  attr_accessor :legal_aid_application_id, :amount, :transaction_date
+
+  category_attributes CATEGORIES
+
+  validate :checkbox_integrity
+
+  def checkbox_integrity
+    CATEGORIES.each do |category|
+      if "check_box_#{category}".present?
+        errors.add(":#{category}_month_1", "Cannot be empty") unless ":#{category}_month_1".present?
```
Alex-Swann, line 36:
```
(1..3).to_a.each do |i|
    errors.add(":#{category}_month_#{i}", "Cannot be empty") unless ":#{category}_month_#{i}".present? 
```
maybe?

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,75 @@
...
+  #     attr_accessor "check_box_#{category}".to_sym
+  #     attr_accessor "#{category}_value_1".to_sym
+  #     attr_accessor "#{category}_value_2".to_sym
+  #     attr_accessor "#{category}_value_3".to_sym
+  #   end
+  # end
+
+  def self.category_attributes(categories)
+    categories.each do |category|
+      attr_accessor "check_box_#{category}".to_sym
```
Alex-Swann, line 23: This is mad but I love it 🤯

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    case month
+    when 1
+      Date.today.at_beginning_of_month - 1.month
+    when 2
+      Date.today.at_beginning_of_month - 2.month
+    when 3
+      Date.today.at_beginning_of_month - 3.month
+    end
+  end
```
stephenrichards, line 78: perhaps:
```
MONTH_NUMBER_CONVERSION = { 1 => 3, 2 => 2, 3 => 3}.freeze

def calculated_date(month_number)
  Date.today.at_beginning_of_month - (MONTH_NUMBER_CONVERSION[month_ number]).months
end
```

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,85 @@
...
+  def save_category(category)
+    # for each monthly amount attribute we need a transaction to be created
+    # each transaction needs to have a different calculated date based on the
+    # numbered suffix - _month_1 is the previous month
+    # need to pass the attr_accessor into calculated date
+    transaction_type = TransactionType.credits.find_by(name: category)
+    CashTransaction.create!(legal_aid_application_id: legal_aid_application_id,
+                            transaction_type_id: transaction_type.id,
+                            transaction_date: calculated_date,
+                            amount:  __send__(:amount))
```
stephenrichards, line 61: This `CashTransaction.create!` statement needs to be wrapped in a `(1..3).each do |i|` loop and the variable `i` needs to be passed in to the `calculated_date` method.

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,106 @@
...
+        args = search_args(category)
+        CashTransaction.where(**args).destroy_all
+      end
+    end
+
+    anything_selected?
+  end
+
+  def save(params)
+    params.each do |key, value|
```
stephenrichards, line 54: I think this should be `def update(params)`.  We want to try and imitate the ActiveRecord API as much as possible, and this is doing what the `#update` method does.

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,106 @@
...
+      __send__("#{key}=",value) if value != ''
+    end
+
+    checkbox_integrity(params)
+
```
stephenrichards, line 59: I think rather than call a specific validation method here, we should just call `valid?` which will call ALL the validations specified in the model.  So here we could write 
``` 
return false unless valid? 
```
instead of lines 53 and 55.

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,106 @@
...
+  def search_args(category)
+    return {
+      legal_aid_application_id: legal_aid_application_id,
+      transaction_type_id: TransactionType.credits.find_by(name: category).id
+    }
+  end
+
+  def calculated_date(month_number)
+    date = Date.today.at_beginning_of_month - (month_number).months
+    date.strftime('%-d %B %Y')
```
stephenrichards, line 105: we can save a `Date` object to the database, so we don't have to convert it into a string here.

app/models/aggregated_cash_income.rb
```diff
@@ -0,0 +1,106 @@
...
+
+  def save_category(category)
+    RANGE.each do |i|
+      args = search_args(category)
+      amount = __send__("#{category}_month_#{i}".to_sym)
+      date = calculated_date(i)
+
+      CashTransaction.where(**args, transaction_date: date).destroy_all
+      CashTransaction.create!(**args, transaction_date: date, amount: amount)
+    end
```
stephenrichards, line 81: I don't think I like the use of **args here - it over complicates things.  And I think I'd rather see existing records being updated rather than deleted and re-created.  so how about something like this:
```
CashTransaction.update_or_create(legal_aid_application_id: legal_aid_application_id,
                                                             transaction_type_id: TransactionType.credits.find_by(name: category).id,
                                                             date: date,
                                                             amount: amount)
```

Then writing a method  the `.update_or_create` method on CashTransaction to find the record and update it or create it if it doesn't exist.

I think the intent is a lot clearer.

Also, don't forget that we have to delete all records belonging to categories that are not entered if they exist.

app/models/aggregated_cash_income.rb
Alex-Swann: 👍

app/models/aggregated_cash_income.rb
Alex-Swann: So interestingly I tried that and it didn't work, for now I suck in errors.empty? but once I've tidied up, if I can't get that working I'll run it by you. I think valid? might reset itself after adding the first failed validation or something as the frontend was only displaying 1 error instead of 3 for each field. Anyway I'll give the above a try. It might be we are missing something.

app/models/aggregated_cash_income.rb
Alex-Swann: So we had a discussion about updating records, we thought if a user saves and returns that then, not only amount, but date would have to be updated as well. Due to that we thought it easier to wipe the records and recreate them.

`Also, don't forget that we have to delete all records belonging to categories that are not entered if they exist.`

Yes we are doing that on line 46.

app/models/aggregated_cash_income.rb
Alex-Swann: So we did this because without explicitly setting it with strftime, the date is formatted as ` 1 November 2020` (with a prefixed space). I wasn't sure if there was a default helper doing this when it is saved to the DB, but by doing this, when we query records using the date as a parameter when doing `CashTransactions.where(...etc` it meant we could reassured there was consistency in it being stored and when being queried. Again, it may be we might have missed something so we can remove this, but currently if it does then we have to convert it like this anyway when querying.

stephenrichards (commented):
The architecture you've put together here is a bit strange to a Rubyist's eyes.  And the nomenclature is a bit confusing.  There are basically two ways of doing this - inheritance or including a module as a concern.

With inheritance, you'd have a base class, let's call it `BaseAggregatedCash` with all the common methods in it, and from that derive two other classes `AggregatedCashIncome` and `AggregatedCashOutgoings`.

The other way is to put all the common code into a module in the `app/models/concerns` directory, and then include that module into your `AggregaredCashIncome` and `AggregatedCashOutgoings` models.

The way you instantiate the AggregatedCashIncome in the CashIncomesController is not really idiomatic ruby.  I'd be more inclined to say
```
def aggregated_cash_income
  @aggregated_cash_income ||= AggregatedCashIncome.find_by_legal_aid_application_id(params[:legal_aid_application_id])
end
```

And then have that `find_by_legal_aid_application_id` method on your model to find the CashTransaction records if they exist, and populate an AggregatedCashIncome record accordingly and return it.

config/locales/en/shared.yml
```diff
@@ -1,6 +1,18 @@
...
+        dates:  "%{date_start} to %{date_end}"
+        hint: Enter the total cash amount you received in each calendar month, or enter 0 if you received none.
+        citizens:
+          cash_incomes:
+            date_goes_here: Dates go here
+            check_box_benefits: Benefits
+            check_box_friends_or_family: Financial help from friends or family
+            check_box_maintenance_in: Maintenance payments
+            check_box_property_or_lodger: Income from a property or lodger
+            check_box_pension: Pension
```
Obsiye, line 15: I believe the welsh translations for the above are missing.

config/locales/en/activemodel.yml
```diff
@@ -71,6 +71,12 @@ en:
             postcode:
               blank: Enter a postcode
               invalid: Enter a postcode in the right format
+        aggregated_cash_income:
+          credits:
+            attributes:
+              none_selected:
+                blank: Select if you make these payments in cash
+                others_present: Select payments in cash or None of the above
```
Obsiye, line 79: missing welsh translations for aggregated_cash_income undeer cy/activemodel.yml

config/locales/en/errors.yml
```diff
@@ -1,6 +1,11 @@
 ---
 en:
   errors:
+    cash_amount:
+      blank: Enter the total cash amount you received in each calendar month
+      invalid_type: Amount must be an amount of money, like 1,000
+      negative: Amount must be more than or equal to zero
+      too_many_decimals: Amount must not include more than 2 decimal numbers
```
Obsiye, line 8: welsh translations for the above changes is missing (cash_amount)

Obsiye (changes requested):
LGTM! Awesome! Just left a few nitpicky comments around welsh transltions

config/locales/en/activemodel.yml
Obsiye, line 79: These errors > blank and others_present say payments

aren't these income so it should be received ?

app/models/base_aggregated_cash_transaction.rb
```diff
@@ -0,0 +1,169 @@
...
+  end
+
+  def validate_none_selected
+    return unless none_selected.present?
+
+    self.class.cash_transaction_categories.each do |category|
+      checkbox_attr = "check_box_#{category}".to_sym
+      next unless __send__(checkbox_attr) == 'true'
+
+      errors.add(:none_selected, model_error(:others_present))
```
Obsiye, line 134: not sure what others_present does? I might be misunderstanding it.

my belief is that we have javascript around none_selected checkboxes to unselect other checkboxes and clear their values.

stephenrichards: Yes, it's belt and braces for non-js

Obsiye (approved):

PR #2109, opened 2020-12-17
Ap 1901 policy disregards

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1901)

## Checklist


app/helpers/check_answers_helper.rb
```diff
@@ -72,4 +72,21 @@ def check_long_question_no_link(question:, answer:, name:, no_border: false)
   def number_to_currency_or_na(number)
     number.to_d == 999_999_999_999.0.to_d ? 'N/a' : gds_number_to_currency(number)
   end
+
+  def safe_yes_or_no(value)
```
Alex-Swann, line 76:
```
def safe_yes_or_no(value)
  return unless !!value == value
  value ?  'Yes' : 'No'
end
```
Can be refactored even if you want to keep `boolean?` for clarity

app/helpers/policy_disregards_helper.rb
```diff
@@ -0,0 +1,78 @@
...
+    criminal_injuries_compensation_scheme
+    national_emergencies_trust
+    we_love_manchester_emergency_fund
+    london_emergencies_trust
+  ].freeze
+
+  def policy_disregards_hash(policy_disregards)
+    items = policy_disregards_as_array(policy_disregards)
+    items&.compact!
+    return nil unless items.present?
```
Alex-Swann, line 17: `return nil unless items.present?` 
can be
`return unless items.present?` which returns nil without needing to specify it

app/helpers/policy_disregards_helper.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    {
+      items: items
+    }
+  end
+
+  def policy_disregards_as_array(policy_disregards)
+    ATTRIBUTES&.map do |attribute|
```
Alex-Swann, line 25: why is the `&` needed here?

app/models/legal_aid_application.rb
```diff
@@ -8,6 +8,7 @@ class LegalAidApplication < ApplicationRecord
   SECURE_ID_DAYS_TO_EXPIRE = 7
   WORKING_DAYS_TO_COMPLETE_SUBSTANTIVE_APPLICATION = 20
   CCMS_SUBMITTED_STATES = %w[generating_reports submitting_assessment assessment_submitted].freeze
+  POLICY_DISREGARDS_START_DATE = Date.new(2021, 1, 8)
```
Alex-Swann: 💥

spec/helpers/policy_disregards_helper_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'rails_helper'
+
+RSpec.describe PolicyDisregardsHelper, type: :helper do
+  include ApplicationHelper
+
+  describe '#policy_disregards_hash' do
+    # subject(:policy_disregards_helper) { policy_disregards_hash(policy_disregards) }
+
+    # it { is_expected.to be_a(Hash) }
```
Alex-Swann, line 9: Remove comments

Alex-Swann (approved):
Approved pending very minor remarks

app/helpers/policy_disregards_helper.rb
eloisetait3: It's syntax to make sure there are values there before mapping over (Colin says it's called the safe navigation operator)

Alex-Swann: Ah cool, I know it's used for digging, but good to know 👍

stephenrichards: But if it's a constant, I think you can be pretty sure it's there, no?

PR #2118, opened 2020-12-22
AP-1918: Policy disregards caseworker review

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1918)

## Checklist

### DIFF
--- a/app/models/cfe/remarks.rb
+++ b/app/models/cfe/remarks.rb
@@ -27,6 +27,8 @@ module CFE
     def populate_review_transactions
       collection = RemarkedTransactionCollection.new
       @hash.each do |category, reason_hashes|
+        break if reason_hashes.is_a?(Array)
+
         reason_hashes.each do |reason, tx_ids|
           tx_ids.each { |tx_id| collection.update(tx_id, category, reason) }
         end
@@ -36,8 +38,13 @@ module CFE
 
     def populate_review_reasons
       reasons = []
-      @hash.each do |_category, reason_hashes|
-        reasons += reason_hashes.keys
+      @hash.each do |category, reason_hashes|
+        reasons += case reason_hashes
+                   when Array
+                     reasons << category
+                   else
+                     reason_hashes.keys
+                   end
       end
       reasons.flatten.uniq.sort
     end
@@ -45,8 +52,15 @@ module CFE
     def populate_review_categories_by_reason
       result = Hash.new([])
       @hash.each do |category, reason_hash|
-        reason_hash.each_key do |reason|
-          result[reason] += [category] unless REASONS_WITHOUT_CATEGORIES.include? reason
+        case reason_hash
+        when Array
+          reason_hash.each do |reason|
+            result[category] += [reason.to_sym] unless REASONS_WITHOUT_CATEGORIES.include? reason
+          end
+        else
+          reason_hash.each_key do |reason|
+            result[reason] += [category] unless REASONS_WITHOUT_CATEGORIES.include? reason
+          end
         end
       end
       result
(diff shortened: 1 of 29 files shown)


features/accessibility/provider.feature
```diff
@@ -188,10 +188,10 @@ Feature: Provider accessibility
...
     Then I choose 'No'
     Then I click 'Save and continue'
-    Then I should be on the 'policy_disregards' page showing 'schemes or charities'
-    Then I select "None of these"
-    And the page is accessible
-    Then I click 'Save and continue'
+#    Then I should be on the 'policy_disregards' page showing 'schemes or charities'
+#    Then I select "None of these"
+#    And the page is accessible
+#    Then I click 'Save and continue'
```
Obsiye, line 194: left overr comments

colinbruce: Annoyingly they need to be uncommented tomorrow!
>These were written for as expected behaviour and will be correct
after the 9th January
Branching logic in rspec handles the logic so these should
be restored on the 8th January

app/views/providers/means_reports/show.html.erb
```diff
@@ -73,4 +73,10 @@
 
   <%= render 'shared/check_answers/assets', read_only: true %>
 
+  <% if @legal_aid_application.passported? && @cfe_result.capital_contribution_required? %>
```
Obsiye, line 76: Is this just for passported applications?

colinbruce: Yes, the non-passported caseworker review section is handled further up in a different sequence

app/models/cfe/remarks.rb
```diff
@@ -36,17 +38,29 @@ def populate_review_transactions
 
     def populate_review_reasons
       reasons = []
-      @hash.each do |_category, reason_hashes|
-        reasons += reason_hashes.keys
+      @hash.each do |category, reason_hashes|
+        reasons += case reason_hashes
+                   when Array
+                     reasons << category
```
Obsiye, line 44: does reasons << category  return a duplication? I would assume just returning category would suffice

because to me the code is adding to the reasons array with << and the case is returning that value to reasons +=
Thus reaons will be [category1, category1]

features/accessibility/provider.feature
colinbruce: Working on a fix for these... It's something that will probably still need some love after the update!

app/models/cfe/remarks.rb
colinbruce: I think there was only ever one category with the new remark type so, for now, it will only ever be `[category]`
But may need addressing in the future if more types are added 🤔

app/services/cfe/create_explicit_remarks_service.rb
```diff
@@ -0,0 +1,22 @@
...
+    def request_body
+      {
+        explicit_remarks: [
+          policy_disregards.as_json
+        ]
+      }.to_json
+    end
+
+    def process_response
+      @submission.explicit_remarks_created!
```
Obsiye, line 19: would it be worth having a guard clause.  
`@submission.explicit_remarks_created! unless @submission.explicit_remarks_created?`

only because they can go back and resubmit to CFE, if it fails.

colinbruce: We don't on any of the other submission services... maybe with adding a ticket to discuss on each of the steps?

app/models/cfe/remarks.rb
```diff
@@ -36,17 +38,29 @@ def populate_review_transactions
...
+                   else
+                     reason_hashes.keys
+                   end
       end
       reasons.flatten.uniq.sort
```
Obsiye, line 49: Just saw this which removes duplicates :doh: my bad.

Obsiye (approved):
Looks awesome!

PR #2122, opened 2020-12-24
AP-1903 Display correct Assessment Results Panel if disregards have been selected

stephenrichards:
## Display correct Assessment Results Panel if disregards have been selected

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1903)

- moved the result panels into the `app/views/shared/assessment_results` folder to eliminate duplication
- New class `ResultsPanelSelector` to determine which panel to display.  This is configuration based, and esier to understand than nested ifs
- Added new method `policy_disregards?` to `LegalAidApplication` to return true if a `policy_disregards` record exists and has any of the disregards set to true
- Added a method `#any?` to `PolicyDisregards` model to return true if any of the disregards are set to true
- Changed the `CapitalAssessmentResultsController` and `CapitalIncomeAssessmentResultsController` to use the new `ResultsPanelSelector` to decide what partial name to pass to the view

## Checklist


colinbruce (approved):
👍  LGTM

PR #2126, opened 2020-12-30
AP-2155: Ruby-3 update

colinbruce:
## What

In an attempt to update faster than we managed to 2.7 this is updates ruby to ~3.0.0~ 3.0.1

It was a placeholder to remind me/us to keep retrying as gems update and patches arrive
<details>
<summary>25 Jun 2021 - Rebased on master</summary>
Following a BigDecimal fix in Rails 6.1.4 the previous errors with decimal rounding seem to have been addressed
</details>
<details>
<summary>23 Mar 2021 - Success 🎉 </summary>
Re-based on master - re-ran bundle install
</details>
<details>
<summary>15 Feb 2021 - Test failed - more rounding issues</summary>
Re-based on master, removed the prometheus-exporter update as this broke on 2.7 as well
webdack-uuid has been updated and that was causing some issues with uat deploys, so this is looking hopeful
</details>
<details>
<summary>5 Feb 2021 - Build failing, expand for details</summary>

Failure reason
--

The updated dockerfile does not seem to include `pip3` so, I assume, it will need to be manually installed in the Dockerfile 🤷‍♂️ 

This addresses
--

Prometheus exporter was quite out of date and needs webrick, as-of ruby 3, webrick 
is no longer packaged.  The updated version now adds webrick as a dependancy

*NOTE:* There is a warning that `prometheus_exporter will only bind to localhost by default as of v0.5` this may cause some issues on UAT (and above) if we see issues in the alerts channel, we may have to revert prometheus_exporter back to 0.4.17 and manually include webrick in the gemfile

This also includes @mpw5's fixes for capybara, geckoboard and CFE
</details>
<details>
<summary>30 Dec 2020 - Build failing because bundle install fails, expand for details</summary>
```ruby 
redis-namespace-1.8.0 requires ruby version ~> 2.4, which is incompatible with the current version, ruby 3.0.0p0
```
</details>

## Checklist


mpw5 (approved):
Looks good to me 🎉

PR #2144, opened 2021-01-07
Accept string or symbol for attribute keys in EntityAttributesGenerator

Obsiye:
# Don't merge this ticket unless the issue with attribute key (string or symbol) is resolved.

Attribute keys returned as string in testing.

- Fix this by allowing indifferent access - string or symbol key can return value in hash.

I made this as a draft PR because still needs debugging. Is this just a test issue. Is this something in master. When was this introduced? Has it always been like this? 

- Secondly, reorganise the rescue of exceptions into the call block to reveal which attribute / attribute_name is causing this issue.

-  Update the test to use :bank_acct and 'bank_acct' as either string or keys because some issue , as stated above, has meant keys are strings and not symbols : /

### Before implementing this:
- [ ] test  ActiveSupport::HashWithIndifferentAccess.new (also separate the [ :bank_acct , ' bank_acct'].sample into two tests) instead of using .sample because of a possibility of introducing flickering tests. 
- [ ] Test the EntityAttributesGenerator fully
- [ ] Find the reason for why attributes have keys that are strings now instead of symbols like :bank_acct . 
- [ ] document the issue somewhere with debugging

## Checklist

### DIFF
--- a/app/services/ccms/payload_generators/entity_attributes_generator.rb
+++ b/app/services/ccms/payload_generators/entity_attributes_generator.rb
@@ -18,7 +18,8 @@ module CCMS
         @entity_name = entity_name
         @requestor = requestor
         @options = options
-        @attr_config = ccms_attribute_keys[@entity_name]
+        ccms_attributes_hash = ActiveSupport::HashWithIndifferentAccess.new(ccms_attribute_keys)
+        @attr_config = ccms_attributes_hash[@entity_name]
       end
 
       def call
@@ -32,6 +33,9 @@ module CCMS
             @xml.__send__('ns0:ResponseValue', response_value)
             @xml.__send__('ns0:UserDefinedInd', config[:user_defined])
           end
+        rescue StandardError => e
+          Raven.capture_message("EntityAttributesGenerator #{e.class}: #{e.message} with config.inspect values of: #{config.inspect} and attribute_name: #{attribute_name}")
+          raise
         end
       end
 
@@ -68,9 +72,6 @@ module CCMS
         else
           raise CCMSError, "Submission #{submission.id} - Unknown response type in attributes config yaml file: #{config[:response_type]}"
         end
-      rescue StandardError => e
-        Raven.capture_message("EntityAttributesGenerator #{e.class}: #{e.message} with config.inspect values of: #{config.inspect}")
-        raise
       end
 
       def extract_raw_value(config)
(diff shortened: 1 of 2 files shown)


spec/services/ccms/payload_generators/entitiy_attributes_generator_spec.rb
```diff
@@ -7,11 +7,27 @@ module PayloadGenerators
...
       let(:submission) { double CCMS::Submission, id: '34343434', legal_aid_application: legal_aid_application }
       let(:legal_aid_application) { double LegalAidApplication }
-      let(:entity_name) { :bank_acct }
+      let(:entity_name) { [:bank_acct, 'bank_acct'].sample }
       let(:yaml_keys) { YAML.load_file(Rails.root.join('spec/fixtures/files/ccms_keys/standard_ccms_keys.yml')) }
       let(:xml) { double Nokogiri::XML::Builder }
       let(:options) { {} }
 
+      describe '#call' do
+        context 'non_passported attributes' do
```
Obsiye, line 16: - [ ] Update this to no longer be non_passported but to standard ccms keys (passported?)

app/services/ccms/payload_generators/entity_attributes_generator.rb
```diff
@@ -32,6 +33,9 @@ def call
             @xml.__send__('ns0:ResponseValue', response_value)
             @xml.__send__('ns0:UserDefinedInd', config[:user_defined])
           end
+        rescue StandardError => e
+          Raven.capture_message("EntityAttributesGenerator #{e.class}: #{e.message} with config.inspect values of: #{config.inspect} and attribute_name: #{attribute_name}")
```
Obsiye, line 37: - [ ] Remove this if it's still not neccessary and revert changes to tests.

app/services/ccms/payload_generators/entity_attributes_generator.rb
```diff
@@ -18,7 +18,8 @@ def initialize(requestor, xml, entity_name, options)
         @entity_name = entity_name
         @requestor = requestor
         @options = options
-        @attr_config = ccms_attribute_keys[@entity_name]
+        ccms_attributes_hash = ActiveSupport::HashWithIndifferentAccess.new(ccms_attribute_keys)
+        @attr_config = ccms_attributes_hash[@entity_name]
```
Obsiye, line 22: - [ ] Find out why in ActiveSupport::HashWithIndifferentAccess.new helped resolve an issue in testing where it returned nil
- [ ] What is the cause of ccms_attribute_keys having keys that are strings and failing tests.

spec/services/ccms/payload_generators/entitiy_attributes_generator_spec.rb
Obsiye: - [ ] Test EntityAttributesGenerator.call more thoroughly and not just for exceptions.

Obsiye (commented):
To do reminders

PR #2146, opened 2021-01-08
Fix: policy disregards tests

colinbruce:
## What

The policy disregards date switch-over missed a few things this PR addresses them:

* cucumber tests, it reverts some placeholder code that checked the date before executing
* cucumber tests, update the final features that, now the policy-disregards path is standard, began to fail
* rspec tests, One branch of the other_assets flow was mishandled and, following the date switch, was not covered causing the coverage to drop by 0.5% 

## Checklist

- [x] The PR description should say what you changed and why ~, with a link to the JIRA story.~


spec/requests/providers/other_assets_controller_spec.rb
```diff
@@ -204,7 +204,22 @@
...
+              context 'when the calculation date is prior to the policy disregards date' do
+                let(:application) do
+                  create :legal_aid_application,
+                         :with_positive_benefit_check_result,
+                         used_delegated_functions: true,
+                         used_delegated_functions_on: Date.new(2021, 1, 5)
+                end
+                let(:policy_disregards) { false }
+
+                it 'redirects to policy disregards' do
```
mpw5, line 217: Should this be 'redirects to check your answers'?

mpw5 (approved):
Looks good. One question about a spec description

spec/requests/providers/other_assets_controller_spec.rb
colinbruce: I don't think so, checking answers is 165-188
This is when they select nothing on their first runthrough and have a delegated function date before the cut-over
(mocked with `receive(:capture_policy_disregards?).and_return(false)`

mpw5: Ah, OK. I just noticed that the test checks for redirection to redirect_to(providers_legal_aid_application_check_passported_answers_path(application)), so it seemed like there was a bit of a mismatch between the description and the assertion.

PR #2149, opened 2021-01-08
AP-1917: Identify non live feedback

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1917)

Created a new version of the feedback template on Notify, that adds a `non_live_env` suffix to the subject line
Added a helper method to the feedback_mailer that checks for uat, staging, or localhost in the current host URL and sends ' -non live' to the template
If none of those are found, it will send an empty string ('') as notify does not handle nil gracefully!

## Checklist

### DIFF
--- a/app/mailers/feedback_mailer.rb
+++ b/app/mailers/feedback_mailer.rb
@@ -25,7 +25,8 @@ class FeedbackMailer < BaseApplyMailer
       originating_page: safe_nil(feedback.originating_page),
       provider_email: provider_email_phrase(feedback),
       application_reference: @legal_aid_application&.application_ref || '',
-      application_status: application_status || ''
+      application_status: application_status || '',
+      non_live_env: non_live_environment?
     )
   end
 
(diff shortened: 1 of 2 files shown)


app/mailers/feedback_mailer.rb
```diff
@@ -51,6 +52,10 @@ def yes_or_no(feedback)
     feedback['done_all_needed'] == true ? 'Yes' : 'No'
   end
 
+  def non_live_environment?
+    %w[staging uat localhost].any? { |env| Rails.configuration.x.application.host.include?(env) } ? '- non-live' : ''
```
stephenrichards, line 56: Conventionally, method names ending with a question mark return true or false

stephenrichards (changes requested):

app/mailers/feedback_mailer.rb
colinbruce: Thanks, nice catch.  It was boolean originally, then I realised that notify requires text!

stephenrichards (approved):

PR #2154, opened 2021-01-08
AP-1934 Show savings account as "none declared" on caseworker means report

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1934)

Under "which types of savings and investments does your client have?", if there is no Savings (or Current) Account uploaded, this will show 'None Declared'. If it is uploaded with a balance of 0, this will still show as £0.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -276,12 +276,12 @@ class LegalAidApplication < ApplicationRecord
 
   def online_savings_accounts_balance
     accounts = applicant.bank_accounts.reject { |c| c.account_type == 'TRANSACTION' }&.map(&:balance)
-    accounts.sum.to_s
+    accounts.present? ? accounts.sum.to_s : nil
   end
 
   def online_current_accounts_balance
     accounts = applicant.bank_accounts.reject { |c| c.account_type == 'SAVINGS' }&.map(&:balance)
-    accounts.sum.to_s
+    accounts.present? ? accounts.sum.to_s : nil
   end
 
   def other_assets?
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -280,7 +280,7 @@ class LegalAidApplication < ApplicationRecord
   end
 
   def online_current_accounts_balance
-    accounts = applicant.bank_accounts.reject { |c| c.account_type == 'SAVINGS' }&.map(&:balance)
+    accounts = applicant.bank_accounts.select { |c| c.account_type == 'TRANSACTION' }&.map(&:balance)
     accounts.present? ? accounts.sum.to_s : nil
   end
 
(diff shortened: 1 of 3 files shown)


config/locales/en/providers.yml
```diff
@@ -416,6 +416,7 @@ en:
           unknown_frequency: Frequency unknown
           residual_balance: Capital Contribution - Residual Balance Check
           policy_disregards: Discretionary disregards selected
+          multi_benefit: Multiple benefits in single bank transaction
```
mpw5, line 419: Is this needed? It doesn't seem to appear in the other files you've changed.

mpw5 (approved):
Looks good to me!

I've left one comment about the translation you've added. I'm not sure why it's needed, although possibly I'm missing something. It's not a biggie though so I've approved the PR anyway.

config/locales/en/providers.yml
eloisetait3: Ran into an error on localhost using Mock True Layer - required a translation so it's just covering that. Will tag you in the thread on apply dev :)

mpw5: Ah, I knew there would be a good reason 😄

app/models/legal_aid_application.rb
```diff
@@ -276,12 +276,12 @@ def offline_savings_value
...
     accounts = applicant.bank_accounts.reject { |c| c.account_type == 'TRANSACTION' }&.map(&:balance)
-    accounts.sum.to_s
+    accounts.present? ? accounts.sum.to_s : nil
   end
 
   def online_current_accounts_balance
     accounts = applicant.bank_accounts.reject { |c| c.account_type == 'SAVINGS' }&.map(&:balance)
-    accounts.sum.to_s
+    accounts.present? ? accounts.sum.to_s : nil
   end
```
stephenrichards, line 285: Are there definitely only two types of savings accounts, SAVINGS and TRANSACTIONS?  As things stand, if there were a third type, they would be included in both `online_savings_account_balance` and `online_current_account_balance`.

I think a better solution would be to have the `online_current_account_balance` use `#select` TRANSACTION to get all the current accounts, and leave the `online_savings_account_balance` to `#reject` TRANSACTION.  That way, you make sure you include all non-TRANSACTION account types in the savings account balance and there is no double counting.

stephenrichards (changes requested):

app/models/legal_aid_application.rb
colinbruce: According to the TrueLayer docs https://docs.truelayer.com/#retrieve-an-account yes, unless we expect applicants to hev BUSINESS_TRANSACTION or BUSINESS_SAVINGS accounts :)

colinbruce: Possibly could be adapted to 
```ruby
c.account_type.include?('SAVINGS')
```

colinbruce (approved):

PR #2163, opened 2021-01-12
AP-1959 BUGFIX do not set provider step to delete

stephenrichards:
## Do not set the provider_step to delete

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1959)

As part of the `Providers::ApplicationDependable` module, every controller set the `provider_step` field to the name of the controller.  This is to allow the application to be saved, and then picked up at the same place at a later date.

Unfortunately, this is also true for the DeleteController.  If the user accidentally hits the delete button, then backs out, the provider step remains set at delete.  This means that when they click on the application from the index page, it immediately asks them to confirm they want to delete the application (thank God!)

Solution: Don't set the provider step to `:delete`

## Checklist

### DIFF
--- a/app/controllers/concerns/providers/application_dependable.rb
+++ b/app/controllers/concerns/providers/application_dependable.rb
@@ -26,7 +26,7 @@ module Providers
         return if self.class.legal_aid_application_not_required?
         return process_invalid_application unless legal_aid_application.present?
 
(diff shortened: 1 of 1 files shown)


mpw5 (approved):
LGTM

willc-work (approved):

PR #2179, opened 2021-01-15
AP1925 multiple proceedings search

Obsiye:
## AP1925 multiple proceedings search

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1925)

- Add new multiple proceedings view/form behind feature flag. seperate the proceeding_types_form into two , one for the old implementation and the new one this will introduce.

- Update proceeding types controller and add a proceeding types service object which encapsulates the adding of proceeding types with a substantive scope limitation.

- Update the proceedings.js file for the multiple proceedings feature

- Update tests

## Checklist

### DIFF
--- a/app/controllers/providers/proceedings_types_controller.rb
+++ b/app/controllers/providers/proceedings_types_controller.rb
@@ -4,6 +4,7 @@ module Providers
 
     # GET /provider/applications/:legal_aid_application_id/proceedings_types
     def index
+      @multiple_proceedings = Setting.allow_multiple_proceedings?
       proceeding_types
     end
 
@@ -11,16 +12,18 @@ module Providers
     def create
       return continue_or_draft if draft_selected?
 
-      if legal_aid_application.proceeding_types.present?
+      if run_transaction
         go_forward
       else
         legal_aid_application.errors.add(:'proceeding-search-input', t('.search_and_select'))
+        @multiple_proceedings = Setting.allow_multiple_proceedings?
         proceeding_types
         render :index
       end
     end
 
     # PATCH /provider/applications/:legal_aid_application_id/proceedings_types/:id
+    # TODO: Could be removed after multiple proceedings go live.
     def update
       ActiveRecord::Base.transaction do
         legal_aid_application.reset_proceeding_types! # This will probably change when multiple proceeding types implemented!
@@ -31,6 +34,20 @@ module Providers
     end
 
     private
+
+    def run_transaction
+      proceeding_types_service.add(proceeding_type_id: form_params, scope_limitation: :substantive)
+    rescue ActionController::ParameterMissing
+      false
+    end
+
+    def proceeding_types_service
+      ProceedingTypesService.new(legal_aid_application)
+    end
+
+    def form_params
+      params.require(:id)
+    end
 
     def proceeding_type
       @proceeding_type = ProceedingType.find(params[:id])
--- a/app/controllers/providers/proceedings_types_controller.rb
+++ b/app/controllers/providers/proceedings_types_controller.rb
@@ -4,7 +4,7 @@ module Providers
 
     # GET /provider/applications/:legal_aid_application_id/proceedings_types
     def index
-      @multiple_proceedings = Setting.allow_multiple_proceedings?
+      @multiple_proceedings = allow_multiple_proceedings?
       proceeding_types
     end
 
@@ -16,7 +16,7 @@ module Providers
         go_forward
       else
         legal_aid_application.errors.add(:'proceeding-search-input', t('.search_and_select'))
-        @multiple_proceedings = Setting.allow_multiple_proceedings?
+        @multiple_proceedings = allow_multiple_proceedings?
         proceeding_types
         render :index
       end
@@ -25,11 +25,7 @@ module Providers
     # PATCH /provider/applications/:legal_aid_application_id/proceedings_types/:id
     # TODO: Could be removed after multiple proceedings go live.
     def update
-      ActiveRecord::Base.transaction do
-        legal_aid_application.reset_proceeding_types! # This will probably change when multiple proceeding types implemented!
-        legal_aid_application.proceeding_types << proceeding_type
-        AddScopeLimitationService.call(legal_aid_application, :substantive)
-      end
+      run_transaction
       go_forward
     end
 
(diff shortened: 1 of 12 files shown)


app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -4,23 +4,26 @@ class ProceedingsTypesController < ProviderBaseController
...
     # POST /provider/applications/:legal_aid_application_id/proceedings_types
     def create
       return continue_or_draft if draft_selected?
 
-      if legal_aid_application.proceeding_types.present?
+      if run_transaction
         go_forward
       else
         legal_aid_application.errors.add(:'proceeding-search-input', t('.search_and_select'))
+        @multiple_proceedings = Setting.allow_multiple_proceedings?
```
Alex-Swann, line 19: you could maybe stick lines 7 and 11 in their own private method and call them from there to avoid duplication. Or maybe not have these in the controller at all and just have `Setting.allow_multiple_proceedings?` in the relevant places in the template so they are easier to locate and remove when this feature flag isn't required anymore maybe.

app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -32,6 +35,20 @@ def update
 
     private
 
+    def run_transaction
+      proceeding_types_service.add(proceeding_type_id: form_params, scope_limitation: :substantive)
```
Alex-Swann, line 39: Hmm, looks like the controller stills only adds 1 proceeding type rather than multiple (but maybe that's still the intention for now and that this be developed further in a future ticket). Have sent you DM to check if this is so.

app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -4,23 +4,26 @@ class ProceedingsTypesController < ProviderBaseController
...
       else
         legal_aid_application.errors.add(:'proceeding-search-input', t('.search_and_select'))
+        @multiple_proceedings = Setting.allow_multiple_proceedings?
         proceeding_types
         render :index
       end
     end
 
     # PATCH /provider/applications/:legal_aid_application_id/proceedings_types/:id
+    # TODO: Could be removed after multiple proceedings go live.
```
Alex-Swann, line 26: I'm thinking it might be better to call run_transaction here and use the `Setting.allow_multiple_proceedings?` to dictate how to process the proceeding type so this isn't missed. Like:

```
    ActiveRecord::Base.transaction do
      @legal_aid_application.reset_proceeding_types! unless Setting.allow_multiple_proceedings?
      @legal_aid_application.proceeding_types << proceeding_type(proceeding_type_id)
      AddScopeLimitationService.call(@legal_aid_application, scope_limitation)
    end
    ```
in your service file. Then you could also remove the `def proceeding_type` private method below

File: app/services/proceeding_types_service.rb [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def add(proceeding_type_id: nil, scope_limitation: nil)
+    return false unless proceeding_type_id && scope_limitation
+
+    ActiveRecord::Base.transaction do
+      @legal_aid_application.proceeding_types << proceeding_type(proceeding_type_id)
+      AddScopeLimitationService.call(@legal_aid_application, scope_limitation)
+    end
+    true
```
Alex-Swann, line 13: I don't think it's necessary to return true here as line 9 returns an activerecord object anyway which is truthy

spec/services/proceeding_types_service_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    context 'correct params' do
+      let(:scope_limitation) { :substantive }
+      let(:params) do
+        {
+          proceeding_type_id: proceeding_type.id,
+          scope_limitation: scope_limitation
+        }
+      end
+
+      it 'adds a proceeding type' do
```
Alex-Swann, line 20: It might be good to have another test checking that when two/multiple proceeding types have been added they are both in `legal_aid_application.proceeding_types`. Maybe something along the lines of 
```
expect { subject }.to change(legal_aid_application.proceeding_types).by(2)
```
in the requests spec maybe by calling patch twice with 2 different ids maybe

app/controllers/providers/proceedings_types_controller.rb
Obsiye: I stuck with a private method. I think using Setting.allow_multiple_proceedings? in the view is less visible and introduces more logic in the view than neccessary. Just preference tbf.

app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -4,40 +4,53 @@ class ProceedingsTypesController < ProviderBaseController
...
+    def form_params
+      params.require(:id)
     end
 
     def proceeding_types
       @proceeding_types ||= ProceedingType.all
     end
+
+    def allow_multiple_proceedings?
+      Setting.allow_multiple_proceedings?
```
Alex-Swann, line 53: do `@multiple_proceedings = Setting.allow_multiple_proceedings?` then you don't need to assign twice above

Alex-Swann (approved):
LGTM nice work

app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -4,40 +4,53 @@ class ProceedingsTypesController < ProviderBaseController
 
     # GET /provider/applications/:legal_aid_application_id/proceedings_types
     def index
+      allow_multiple_proceedings?
       proceeding_types
```
stephenrichards, line 8: This has got a bit of a smell to me.  You're calling a method which returns a boolean, but doing nothing with it.  looking at the method, I see that. you're setting the `@multiple_proceedings` variable, but that is only used in the views as far as I can tell.

Can't we just call `Setting.allow_multiple_proceedings?` from the views?  It won't do any more database lookups, because rails returns cached values for identical SQL queries during the same request cycle.

If you don't want to do that, then I'd call the method something a bit more descriptive of what it's doing, like `set_multiple_proceedings_flag` (although rubocop will probably complain about using a method beginning `set_`)

stephenrichards (changes requested):

Alex-Swann (approved):

stephenrichards (approved):

PR #2180, opened 2021-01-15
AP-1966 Update date and time checks to consider time zones

Obsiye:
## AP-1966 Update date and time checks to consider time zones

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1966)

Use the rubocop-rails timezone https://www.rubydoc.info/gems/rubocop/0.57.2/RuboCop/Cop/Rails/TimeZone
and fix Time.now to Time.current or Time.zone.now which protects the application from BST/GMT changes

## Checklist


app/models/dashboard/single_object/applicant_email.rb
```diff
@@ -39,11 +39,11 @@ def dataset_definition
       end
 
       def log_start_message
-        log_message("***** ApplicantEmail starting job at #{Time.now} for dataset: #{dataset_name}")
+        log_message("***** ApplicantEmail starting job at #{Time.current} for dataset: #{dataset_name}")
```
Alex-Swann, line 42: This actually makes for a more readable Time than using Time.now which is nice

app/models/scheduled_mailing.rb
```diff
@@ -18,14 +18,14 @@ def deliver!
   end
 
   def cancel!
-    update!(cancelled_at: Time.now)
+    update!(cancelled_at: Time.current)
```
Alex-Swann, line 21: Just to check, because this is now used as part of process, bearing in mind
```
irb(main):020:0> Time.now
=> 2021-01-18 12:36:48.667092 +0000
irb(main):021:0> Time.current
=> Mon, 18 Jan 2021 12:36:50 GMT +00:00
```
the latter includes a more prettified date this shouldn't affect anything right as it's still a Time object correct? Asking in case it needs to be in a specified format but I'm guessing look at what was prior probably not

app/services/reports/mis/non_passported_applications_report.rb
```diff
@@ -1,7 +1,7 @@
 module Reports
```
Alex-Swann, line 1: Line 20, 
`Date.today.end_of_day`
should be
`Date.current.end_of_day`

bin/delete_ecr_images
```diff
@@ -13,7 +13,7 @@ images = JSON.parse(json_output)['imageDetails']
 images_to_delete = []
 images.each do |i|
   date_pushed = DateTime.strptime(i['imagePushedAt'].to_s, '%s')
```
Alex-Swann, line 15: should this be the following to ensure consistency with the new standard?
`DateTime.strptime(i['imagePushedAt'].to_s, '%s').in_time_zone`

spec/models/aggregated_cash_outgoings_spec.rb
```diff
@@ -366,7 +366,7 @@ def error_msg(name, month)
         let(:subject) { aco.update(corrected_valid_params) }
```
Alex-Swann, line 367: There's a few other references to Dates that haven't been updated to the new standard, i.e. line 50 `Date.new(2020, 12, 1).in_time_zone`? and 384, `Date.today.at_beginning_of_month` should be `Date.current.at_beginning_of_month`

Alex-Swann:
Seems there are still a few places where `Time.` and `Date.` are not using the better standard. Might be worth going through the application and making sure they are all addressed so everything's as consistent as it can be 🙂

app/models/scheduled_mailing.rb
Obsiye: postgres saves it as UTC so it should be ok.

Also, since we set timezone to europe/london then https://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html
will do the clever math of converting UTC to the relevant timezone

stephenrichards (changes requested):
There are lots of places where `Date.today` is still used, rather than `Date.current`.  Let's get those all changed as well.  I'm surprised the new rubocop extension doesn't bitch about that.

PR #2183, opened 2021-01-18
Reduce log pollution in the testing environment

Obsiye:
## Reduce log pollution in the testing environment

- Set the log_level to fatal which logs when something happens that causes the application to crash. This reduces log pollution and speeds up testing.

- Create an environment variable ENV['RAILS_ENABLE_TEST_LOG'] to allow default logging in the test environment e.g.
`RAILS_ENABLE_TEST_LOG=1 rspec ./spec`

- Update readme

resources: https://jtway.co/speed-up-your-rails-test-suite-by-6-in-1-line-13fedb869ec4

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -375,6 +375,12 @@ Staging and UAT environments.
   - [Staging](https://mojdt.slack.com/messages/GGWMW7M0F)
   - [Production](https://mojdt.slack.com/messages/GGWE9V9BP)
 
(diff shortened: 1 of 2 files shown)


mpw5 (approved):
Nice! I like this, seems to knock about 30 seconds of the running time of rspec 👍

PR #2184, opened 2021-01-18
AP-1967: Initial fix for mean summary CYA page for non passported jou…

Alex-Swann:
…rneys

## What

[AP-1967](https://dsdmoj.atlassian.net/browse/AP-1967)

Updated the means summary CYA page to ensure offline bank accounts is revealed for the passported journey only and to reveal the 'Do you have any offline savings' question so the user can change their answer. I've initially put in more hardcoded html for the headers but if there is a better way using the check_answer_link erb partials then please let me know, it was to ensure the styling was consistent. I wasn't splitting hairs on it too much as it might be this page is updated when the new formbuilder is implemented anyway.

Also, it appears the factory traits that generate an application with a passported or non_passported state do not update the benefit_check_result which ensures `legal_aid_application.non_passported?` is set to the right thing. For now, the simplest fix I could find was to include the `passported` and `non_passported` traits with the `:with_passported_state_machine` and `:with_non_passported_state_machine` traits respectively. This actually flagged a couple of tests which hadn't been setup properly which were only passing because the bug with `non_passported?`. Whether or not there is a better solution or this is fine for now, please let me know.

## Checklist

### DIFF
--- a/spec/factories/legal_aid_applications.rb
+++ b/spec/factories/legal_aid_applications.rb
@@ -32,6 +32,7 @@ FactoryBot.define do
         state_machine = FactoryBot.build(:non_passported_state_machine, legal_aid_application: application)
         application.update!(state_machine: state_machine)
       end
+      non_passported
     end
 
     trait :with_passported_state_machine do
@@ -39,6 +40,7 @@ FactoryBot.define do
         state_machine = FactoryBot.build(:passported_state_machine, legal_aid_application: application)
         application.update!(state_machine: state_machine)
       end
+      passported
     end
 
     trait :initiated do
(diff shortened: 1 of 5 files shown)


spec/factories/legal_aid_applications.rb
```diff
@@ -32,13 +32,15 @@
         state_machine = FactoryBot.build(:non_passported_state_machine, legal_aid_application: application)
         application.update!(state_machine: state_machine)
       end
+      non_passported
```
Alex-Swann, line 35: This is to set benefit_check_result to the right thing, however for some reason when the factory builds on application using `:non_passported_state_machine` or `:passported_state_machine` it doesn't update it, meaning `legal_aid_application.non_passported?` could be inaccurate.

mpw5 (approved):
LGTM. I think your approaches to the view and the factory are fine 👍

PR #2194, opened 2021-01-21
AP-1962 Remove overriden ccms_submission method

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1962)

The legal_aid_application model overrides the default ccms_submission method, causing it to create a new submission when it is called (if one doesn't already exist)

- remove the method 
- fix some tests which break as a result

## Checklist

### DIFF
--- a/app/models/concerns/base_state_machine.rb
+++ b/app/models/concerns/base_state_machine.rb
@@ -119,7 +119,9 @@ class BaseStateMachine < ApplicationRecord  # rubocop:disable Metrics/ClassLengt
 
     event :generated_reports do
       transitions from: :generating_reports, to: :submitting_assessment,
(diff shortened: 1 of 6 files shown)


Alex-Swann (approved):
Just checking `@legal_aid_application.ccms_submission.present?` doesn't need to be removed from the `app/views/admin/legal_aid_applications/submissions/show.html.erb` page right? Otherwise LGTM

stephenrichards (changes requested):
I don't know if this is an issue or not, but I'm a bit concerned that the `generated_reports` event in the State Machine calls ccms_submission and expects it to be created if it doesn't already exist.

stephenrichards (approved):

PR #2196, opened 2021-01-21
AP-1955 Reduce size of statement of case uploads

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1955)

Users are currently able to upload statement of case files up to 8mb in size. This can cause issues after base64 encoding - there is a 10mb limit on the AWS gateway, and encoded files can exceed that limit.

This amends the `StatementOfCase` model to reduce the permitted file size to 7mb, which is hopefully low enough to prevent the 10mb limit being hit.

## Checklist


willc-work (approved):
LGTM

PR #2200, opened 2021-01-22
BUGFIX - Don't send details for current account to CFE when nil

stephenrichards:
## DOn't send current account details to CFE when nil

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1993)

CFE is failing if on the non-passported journey the provider says yes to "Does your client have any savings account your client cannot access on-line".  

This is because in the non-passported journey, we only ask for savings account, not current account, so that field remains nil.  This value of nil was being sent to CFE, which rejected it as invalid.

The solution is not to send details of any field which is nil.

## Checklist


mpw5 (approved):
LGTM

PR #2213, opened 2021-01-27
AP-1974 Send alert emails to apply team when mail undeliverable

stephenrichards:
## Send alert mails to apply team if outgoing mail undeliverable

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1974)

If a provider mistypes an applicant's email address, the email can't be sent.  Currently, an error is posted to Sentry, but all too often this is missed.  This PR will send an email to the team email address `apply-for-legal-aid@digital.justice.gov.uk` with details of the failed delivery.  This is a first step - eventually the mails will go to the provider who will have the ability to update the email address of the applicant.

---

#### On this branch, the live GOVUK Notify key is used so that it can be tested in UAT.  This must be reverted before merging to master 

---

## Checklist


spec/mailers/undeliverable_email_alert_mailer_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe UndeliverableEmailAlertMailer, type: :mailer do
+  # let(:application) { create :legal_aid_application, :with_applicant }
+  # let(:email) { Faker::Internet.safe_email }
+  # let(:provider_name) { Faker::Name.name }
+  # let(:application_url) { 'test' }
```
mpw5, line 7: Can these be deleted?

spec/services/cfe/create_capitals_service_spec.rb
```diff
@@ -100,7 +100,7 @@ def savings_amount_with_nil_current_account
     end
 
     def expected_payload_hash # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-      {
+      payload = {
```
mpw5, line 103: These changes look like they've crept in from another PR

app/services/govuk_emails/email_monitor.rb
```diff
@@ -65,23 +68,21 @@ def send_email
...
-        "*GovUk email status:* #{email.status}",
-        email_args.to_s
-      ].join("\n")
+    def send_undeliverable_alert(error)
+      failure_reason = if error.is_a?(StandardError)
+                         error.class.to_s
+                       else
+                         error
+                       end
+      UndeliverableEmailAlertMailer.notify_apply_team(email_address, failure_reason, @mailer, @mail_method, @email_args)
```
mpw5, line 77: It looks like we're removing the sentry alert and just relying on the email. I'm just wondering if it's worth keeping the sentry alert too. Personally I'm more likely to see the slack notification for that alert than the email (I'm not sure if I even have access to that account 🤔)

mpw5 (changes requested):
A couple of minor comments/questions, plus it looks like thre are some unexpected CFE changes which I think need removing before this is merged.

spec/services/cfe/create_capitals_service_spec.rb
stephenrichards: No - I assign it to payload here because lower down I add elements in to the hash if they're not nil.  It's to fix a flickering test

app/services/govuk_emails/email_monitor.rb
stephenrichards: Fair point.  That email is monitored by Lizzie and Alice, but it can't help us being notified too.  I'll make it a specialised error so that it stands out a bit more.

spec/services/cfe/create_capitals_service_spec.rb
mpw5: Ok, that makes sense 👍

mpw5 (approved):
LGTM

PR #2214, opened 2021-01-27
Ap 1960 Assign lead proceeding to multi proceeding cases

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1960)

Add a method to check if a proceeding type is the first to be assigned to an application, if it is, a new attribute `lead_proceeding`,  is set to TRUE. All subsequent proceedings on the same application have this value set to FALSE.

Why? Apply does not actually care about a lead proceeding, it is a CCMS requirement. As such the first proceeding added is as good as any. However when we make subsequent proceeding types dependant on the first one then the first proceeding entered becomes important to Apply so the assigning it as lead_proceeeding seems like the best choice.

The ticket specifies it should be a domestic abuse type, but as all our proceedings are Domestic Abuse there is no need to do extra work to determine this is the case. Also once we had non Domestic Abuse cases we would need to revisit this.

The rake task:
When the lead_proceeding attribute is created it is set as default false, this means that on all existing records the setting is FALSE. As these are all single proceeding type cases these should be set to TRUE. The rake task changes all existing proc_types to be lead proc types.

Will this break Staging and Master UAT? maybe if multi-proc cases exists in the DB and they have not yet been completed and submitted to CCMS. Prod will be fine as no multi-proc cases exist there yet.

## Checklist


Obsiye (dismissed):
LGTM

PR #2219, opened 2021-01-28
AP-1957 delegated functions date confirmation page

Obsiye:
## AP-1957 delegated functions date confirmation page

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1957)

- Add a new page to confirm the delegated functions date

- Duplicate the used delegated functions form with some changes for the new page. This is because the used delegated functions form is already complicated enough and reusing it will make it more so, thus it seems better to have duplication.

- Add requests tests and form tests

- Update the  feature tests

## Checklist


colinbruce (approved):
LGTM 👍

PR #2221, opened 2021-01-28
AP-1956 Calculate balances for every transaction in CSV file

stephenrichards:
## Calculate balances for every transaction in CSV file

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1956)

This PR fixes two issues:
 - the balances as reported by the means report does not match the balance on the latest transaction of the CSV (that is because the means report balance was taken from the balance today, whereas the latest transaction may be several days ago)
 - Not all CSVs have a running balance (that's because not all banks (e.g. Monzo) return a running balance with each transaction)

### What I did
- Changed `TrueLayer::BankDataImportService` to collect all transactions from the start of the calculation period up to now, thus ensuring that the running balance on the last transaction (if present) matches the current balance received by a separate call to TrueLayer
- Created a `Banking:: BankTransactionBalanceCalculator` service to go through all the bank transactions and calculate the running balance from the current balance we've received.  We do this for all transactions, regardless of whether or not they have a running balance already because that removes issues we get when transactions on the same day are ordered slightly differently and the running balances don't match.
- Created a `Banking:: BankTransactionsTrimmer` service to delete any bank transactions we've collected after the end of the calculation period
- Call the above two services before `StateBenefitAnalyserService` in the `BankTransactionsAnalyserJob` which gets run at the end of the applicant journey
- Removed the date limitations on `LegalAidApplication#bank_transactions` as that's no longer needed due to the `Banking:: BankTransactionsTrimmer` service
- Changed the methods `#online_savings_accounts_balance` and `#online_current_accounts_balance` methods on `LegalAidApplication` to use the running balance on the last transaction of each account (i.e. at the end of the calculation period) rather than the balance on the account itself, which is today's balance
- Moved `StateBenefitAnalyserService` into the `Banking` module namespace.
- Deleted the services which sorted the transactions so that they matched the running balances on each transaction - this is no longer needed.

## Checklist


app/models/legal_aid_application.rb
```diff
@@ -150,12 +150,14 @@ def capture_policy_disregards?
...
+    # set_transaction_period
     return applicant.bank_transactions if Setting.mock_true_layer_data?
 
-    from = transaction_period_start_on.beginning_of_day
-    to = transaction_period_finish_on.end_of_day
-    applicant.bank_transactions.where(happened_at: from..to)
+    #
+    # from = transaction_period_start_on.beginning_of_day
+    # to = transaction_period_finish_on.end_of_day
+    # applicant.bank_transactions.where(happened_at: from..to)
```
Obsiye, line 159: comments that can be removed.

Obsiye (approved):
LGTM, just some leftover comments

app/services/banking/bank_transactions_trimmer.rb
```diff
@@ -0,0 +1,18 @@
...
+    def self.call(legal_aid_application)
+      new(legal_aid_application).call
+    end
+
+    def initialize(legal_aid_application)
+      @legal_aid_application = legal_aid_application
+    end
+
+    def call
+      deletion_period = (@legal_aid_application.transaction_period_finish_on..1.year.from_now)
```
Obsiye, line 12: Just to confirm, transaction_period_finish_on can be the used delegated functions date or todays date?

And that here, you destroy anything past it.

Since the used delegated functions date can't be more than 12 months ago, the  `..1.year.from_now`  ensures that all transactions are deleted after that.

Obsiye: Should it be `1.year.from_now` ... or `Date.current` (today) ?

stephenrichards: We destroy anything past the end of the calculation period, which is either the day delegated functions were used, or midnight last night.  That is because we now collect ALL transactions from the beginning of the calculation period to NOW, in order that we can get the latest balance and that ties in with the running balance on the last transaction if present, and we can calculate it if not.

But after we've done that, we don't need the bank transactions after the end of the calculation period to go into the CSV file, which is why we delete them.

PR #2223, opened 2021-01-29
BUGFIX Use DateTime instead of Time class for strptime method

Obsiye:
## BUGFIX Use DateTime instead of Time class for strptime method

- Use DateTime instead of Time class for strptime method. This is because it raises a bug in circleci:

```bash
./bin/delete_ecr_images:15:in `block in <main>': undefined method `strptime' for Time:Class (NoMethodError)
```

This is because we require 'date' in the delete_ecr_images file. Instead of requiring 'time' and using Time objects etc. I changed the code to use DateTime. This uses .in_time_zone and DateTime.current to meet TimeZone standardisations requirements.

## Checklist

### DIFF
--- a/bin/delete_ecr_images
+++ b/bin/delete_ecr_images
@@ -12,7 +12,7 @@ images = JSON.parse(json_output)['imageDetails']
 
 images_to_delete = []
 images.each do |i|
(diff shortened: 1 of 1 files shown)


mpw5 (approved):
LGTM.

Alex-Swann (approved):

PR #2237, opened 2021-02-02
AP-1974 Report undeliverable emails

stephenrichards:
## Report undeliverable emails

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1974)

Previously, if an email was undeliverable (usually because the provider mis-typed the applicant's email address), a message was captured to Sentry, but this was often missed.

This PR is the first stage in enabling the provider to handle this themselves.  For the time being, an email will be sent to the. apply team email address, as well as being reported to Sentry.  In addition, we keep a record of all the emails sent, and what their status is in a new table `sent_emails` in order to better debug what has happened. 

The code which previously took care of resending mails that GOVUK Notify were reporting as temporarily failed has been removed.  These mails are now reported as undeliverable, as in testing it was found that this status was reported for emails with addresses on non-existent domains.

## Note
**This branch uses the live GOVUK Notify API key for testing.  Revert this before merging with master**

## Checklist


colinbruce (changes requested):
I'm a bit worried that the email_monitor class is getting a bit bloated. 

Could the create_sent_email and associated methods be extracted to a new class?

colinbruce (approved):

PR #2240, opened 2021-02-02
Add a hold_uat manual approval to circleci

Obsiye:
## Add a hold_uat manual approval to circleci

Add a hold_uat manual approval to circleci

This will mean deployments to uat are made when needed as opposed to automatically and thus reducing our kubernetes pods/deployments. This will also reduce ingress limit reached issues e.g.
```
Quata-exceeded 
Namespace xyz-xyz-uat is using 91% of its pods quota
```

Consequently, Dependabot will not be deployed to UAT until  it's neccessary and other git branches.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -334,9 +334,13 @@ workflows:
       - build_and_push:
           requires:
           - lint_checks
+      - hold_uat:
+          type: approval
+          requires:
+            - build_and_push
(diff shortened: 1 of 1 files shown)


colinbruce:
Can we have a discussion around this? We removed the hold for UAT deliberately because we had some dependabot release where the tests passed but the build and deploy failed :/

PR #2250, opened 2021-02-05
AP-1970: Parse CFE V3 response and add relevant tests for it

Alex-Swann:
## What

[AP-1970](https://dsdmoj.atlassian.net/browse/AP-1970)

To complete the cash transactions work this PR parses the CFE V3 response when admin settings are set to include cash transactions. V3 CFE result factories and traits have been setup, separately to V2, with a view for the latter to be removed once this work is all live.

## Checklist

### DIFF
--- a/spec/factories/legal_aid_applications.rb
+++ b/spec/factories/legal_aid_applications.rb
@@ -627,6 +627,13 @@ FactoryBot.define do
       end
     end
 
+    trait :with_cfe_v3_result do
+      after :create do |application|
+        cfe_submission = create :cfe_submission, legal_aid_application: application
+        create :cfe_v3_result, submission: cfe_submission
+      end
+    end
+
     trait :with_means_report do
       with_cfe_v2_result
       after :create do |application|
--- a/app/models/cfe/v3/result.rb
+++ b/app/models/cfe/v3/result.rb
@@ -1,0 +1,242 @@
+module CFE
+  module V3
+    class Result < CFE::BaseResult # rubocop:disable Metrics/ClassLength
+      def assessment_result
+        return nil if result_hash[:assessment].nil?
+
+        result_hash[:assessment][:assessment_result]
+      end
+
+      def capital_assessment_result
+        capital[:assessment_result]
+      end
+
+      def capital_contribution_required?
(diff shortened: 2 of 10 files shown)


spec/factories/legal_aid_applications.rb
```diff
@@ -627,6 +627,13 @@
       end
     end
 
+    trait :with_cfe_v3_result do
```
mpw5, line 630: There are a few places where tests use the `with_cfe_v2_result` trait. I wonder if it's worth duplicating those tests and making sure they pass using the new `with_cfe_v3_result` trait too.

For example `spec/requests/providers/means_reports_spec.rb`, `spec/services/reports/means_report_creator_spec.rb` and several CCMS specs. 

Maybe it's overkill though.

mpw5 (commented):
LGTM. I've left a couple of comments:

 - there's a schema change which doesn't look like it's related to the rest of this PR 
 - an idle thought about expanding specs where we currently use the `with_cfe_v2_result` to also test the new `with_cfe_v3_result`. I'm not sure it's really necessary though, so feel free to ignore it.

db/schema.rb
```diff
@@ -434,7 +434,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "scanner_working"
-    t.index ["uploader_type", "uploader_id"], name: "index_malware_scan_results_on_uploader_type_and_uploader_id"
+    t.index ["uploader_type", "uploader_id"], name: "index_malware_scan_results_on_uploader"
```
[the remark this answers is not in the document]
Alex-Swann: @mpw5 it shouldn't be. But it keeps sneaking in there since rebasing and resetting my DB for some reason

spec/factories/legal_aid_applications.rb
Alex-Swann: Yeah I noticed those too, but tbh I was thinking we should leave those and replace the traits for those when we make V3 live. But maybe you are right, in case there are any breakages maybe worth sorting now and then we know we won't be blocked by it when live.

mpw5 (approved):
LGTM 🎉

PR #2252, opened 2021-02-05
AP-1965 Removed hard-coded data from BankAccount model

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-1965)

Methods in bank account model now filled. Added a factory trait and testing.

## Checklist

### DIFF
--- a/app/models/bank_account.rb
+++ b/app/models/bank_account.rb
@@ -26,15 +26,17 @@ class BankAccount < ApplicationRecord
 
   # rubocop:disable Naming/PredicateName
   def has_tax_credits?
-    true # TODO: CCMS placeholder - to be fixed by AP-1965
+    tax_credits = bank_transactions.pluck(:meta_data).map { |transaction| transaction[:code] }
+    tax_credits.include?('TC') || tax_credits.include?('WTC') || tax_credits.include?('CTC')
   end
 
   def has_wages?
-    true # TODO: CCMS placeholder - to be fixed by AP-1965
+    false
   end
 
   def has_benefits?
-    true # TODO: CCMS placeholder - to be fixed by AP-1965
+    benefits = bank_transactions.joins(:transaction_type).where(transaction_type: { name: 'benefits' })
+    benefits.present?
   end
   # rubocop:enable Naming/PredicateName
 end
--- a/app/models/bank_account.rb
+++ b/app/models/bank_account.rb
@@ -26,8 +26,8 @@ class BankAccount < ApplicationRecord
 
   # rubocop:disable Naming/PredicateName
   def has_tax_credits?
-    tax_credits = bank_transactions.pluck(:meta_data).map { |transaction| transaction[:code] }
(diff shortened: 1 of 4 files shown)


app/models/bank_account.rb
```diff
@@ -26,15 +26,17 @@ def bank_and_account_name
 
   # rubocop:disable Naming/PredicateName
   def has_tax_credits?
-    true # TODO: CCMS placeholder - to be fixed by AP-1965
+    tax_credits = bank_transactions.pluck(:meta_data).map { |transaction| transaction[:code] }
+    tax_credits.include?('TC') || tax_credits.include?('WTC') || tax_credits.include?('CTC')
```
mpw5, line 30: I think we could use a better variable name here - we don't necessarily know that the `meta_data` contains tax credits. 

Something like this might be better:

```suggestion
    meta_data_codes = bank_transactions.pluck(:meta_data).map { |transaction| transaction[:code] }
    meta_data_codes.include?('TC') || meta_data_codes.include?('WTC') || meta_data_codes.include?('CTC')
```

app/models/bank_account.rb
```diff
@@ -26,15 +26,17 @@ def bank_and_account_name
...
 
   def has_wages?
-    true # TODO: CCMS placeholder - to be fixed by AP-1965
+    false
   end
 
   def has_benefits?
-    true # TODO: CCMS placeholder - to be fixed by AP-1965
+    benefits = bank_transactions.joins(:transaction_type).where(transaction_type: { name: 'benefits' })
+    benefits.present?
```
mpw5, line 39: You don't really need the `benefits` variable here. This should work just as well and is a little neater:

```suggestion
    bank_transactions.joins(:transaction_type).where(transaction_type: { name: 'benefits' }).present?
```

spec/models/bank_account_spec.rb
```diff
@@ -33,4 +33,25 @@
       expect(bank_account.display_name).to eq 'Test Bank Acct 123456789'
     end
   end
+
+  describe '#benefits ' do
+    it 'returns true if benefits present' do
```
mpw5, line 38: Is it worth adding a test for `'returns false if benefits not present'`, to be consistent with the `tax_credits` spec below?

mpw5 (approved):
Good work 😁 I've left a couple of comments about small stylistic things, not blockers though.

app/models/bank_account.rb
eloisetait3: Good point - changed

spec/models/bank_account_spec.rb
eloisetait3: Added

eloisetait3:
Did not need to touch the Gemfile.lock to fix bundler, in the end. Just a quick alteration in the bin/bundle helper file

mpw5 (approved):
Great work 💯

PR #2267, opened 2021-02-10
AP-1968: Update styling and read_only values for the means report

Alex-Swann:
## What

[AP-1968](https://dsdmoj.atlassian.net/browse/AP-1968)

Refactored the align right styling as even though it was being set it appears it wasn't being used when an html file was set to read_only for the reports generation. Also wickedpdf sets out it's own formatting when creating a pdf which supersedes class name styling so by setting text-precision this assists with the rendering process to ensure there is whitespace between question/answers in the means (and other) reports. 

## Checklist


RoseSAK (approved):
LGTM!

PR #2275, opened 2021-02-12
AP-2020 Update other capital ccms attributes

Obsiye:
## AP-2020 Update other capital ccms attributes

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2020)

- Hardcode GB_INPUT_B_17WP2_7A to false 
- Add new attribute GB_INPUT_B_17WP2_8A with hardcode value of false
- remove ccms attribute OTHCAPITAL_INPUT_C_17WP2_14A

- Update tests

left to do:
- [ ] find meaning of GB_INPUT_B_17WP2_8A and add to attribute block config

## Checklist

### DIFF
--- a/config/ccms/attribute_block_configs/non_passported.yml
+++ b/config/ccms/attribute_block_configs/non_passported.yml
@@ -1597,19 +1597,18 @@ other_capital:
     response_type: boolean
     user_defined: true
     generate_block?: true
-  OTHCAPITAL_INPUT_C_17WP2_14A:
-    value: '0'
-    br100_meaning: 'Other capital: The value'
-    user_defined: true
-    response_type: currency
-    generate_block?: true
   OTHCAPITAL_INPUT_T_17WP2_13A:
     value: "."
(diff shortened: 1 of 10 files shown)


Alex-Swann (approved):
LGTM

config/ccms/attribute_block_configs/non_passported.yml
```diff
@@ -1597,12 +1597,6 @@ other_capital:
     response_type: boolean
     user_defined: true
     generate_block?: true
-  OTHCAPITAL_INPUT_C_17WP2_14A:
-    value: '0'
-    br100_meaning: 'Other capital: The value'
-    user_defined: true
-    response_type: currency
-    generate_block?: true
```
mpw5, line 1605: It might be better to set `generate_block?: false` instead of deleting this completely. That way the attribute won't be included in the payload, but if we want to enable it later we don't have to recreate this, we can just re-enable it.

mpw5 (approved):
One minor comment, not a blocker though.

PR #2283, opened 2021-02-17
AP-2045: Update ManualReviewDeterminer to flag submissions as needing…

Alex-Swann:
… a review even if there are restrictions

## What

[AP-2045](https://dsdmoj.atlassian.net/browse/AP-2045)

Removed a check on restrictions when flagging if a passported application needs a manual review

## Checklist

### DIFF
--- a/app/services/ccms/manual_review_determiner.rb
+++ b/app/services/ccms/manual_review_determiner.rb
@@ -26,11 +26,17 @@ module CCMS
     end
 
     def call
-      return true if Setting.manually_review_all_cases? && non_passported?
+      needs_review?
+    end
 
-      return true if capital_contribution_required? && has_restrictions?
+    private
 
-      false
+    def needs_review?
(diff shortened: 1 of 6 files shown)


app/services/ccms/manual_review_determiner.rb
```diff
@@ -26,11 +26,17 @@ def initialize(legal_aid_application)
...
 
-      return true if capital_contribution_required? && has_restrictions?
+    def needs_review?
+      manually_review_all_non_passported? || capital_contribution_required?
+    end
 
-      false
+    def manually_review_all_non_passported?
+      Setting.manually_review_all_cases? && non_passported?
     end
```
stephenrichards, line 40: What happens if there is no capital contribution required, but there are restrictions?  it should return true.

stephenrichards (approved):

stephenrichards (changes requested):
See comment regarding restrictions

stephenrichards (approved):

PR #2290, opened 2021-02-18
ap-2043 add quote marks for proper separation

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2043)

added a space and single quote marks to separate out the error name and class, erblint was failing when just a space was added to fix this, which is probably how it got introduced in the first place.

## Checklist


mpw5 (approved):
LGTM

PR #2295, opened 2021-02-19
AP-2041 No Undeliverable email alerts in non-prod environments

stephenrichards:
##  No Undeliverable email alerts in non-prod environments

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2041)

- Added Rails configuration variable alert_undeliverable_emails to control whether or not UndeliverableEmail alerts are sent to Sentry and the apply team email inbox.  This is set to false in development and test and UAT, and true in production, but can be overridden for testing purposes.
- Added the host env environment to the subject line of the Undeliverable email alert to make it obvious when testing emails were being sent to that box.
- Changed Undeliverable Alert email template so as no longer to accept all mail params - they are pretty meaningless, and the addressee, if available,  is now separated out on its own line.
- 
## Checklist


config/environments/production.rb
```diff
@@ -96,4 +96,7 @@
   config.webpacker.check_yarn_integrity = false
 
   config.x.policy_disregards_start_date = Date.parse(ENV['POLICY_DISREGARDS_START_DATE'] || '2021-1-8')
+
+  # says whether or not to send alerts to Sentry and support email for undeliverable emails
+  config.x.alert_undeliverable_emails = HostEnv.production?
```
colinbruce, line 101: Can this not just be `true`?

spec/services/govuk_emails/email_monitor_spec.rb
```diff
@@ -239,43 +241,73 @@
...
 
     context "govuk_notify can't find email" do
       before do
-        allow_any_instance_of(Notifications::Client)
-          .to receive(:get_notification)
-          .and_raise(Notifications::Client::NotFoundError, OpenStruct.new(code: 404, body: ''))
+        allow(Rails.configuration.x).to receive(:alert_undeliverable_emails).and_return(true)
+        allow(GovukEmails::Email).to receive(:new).and_raise(Notifications::Client::NotFoundError, OpenStruct.new(code: 404, body: ''))
       end
+      #   allow_any_instance_of(Notifications::Client)
```
colinbruce, line 291: Can this comment block be removed?

colinbruce (changes requested):
A couple of minor fixes needed and the HostEnv use also seems to cause Docker build to fail!

config/environments/production.rb
stephenrichards: If it's true, then mails from UAT and Staging will be sent as well, as they run in the Rails production environment.  This way, just those running in live production will be sent.

spec/services/govuk_emails/email_monitor_spec.rb
stephenrichards: yep

config/environments/production.rb
```diff
@@ -98,5 +98,5 @@
   config.x.policy_disregards_start_date = Date.parse(ENV['POLICY_DISREGARDS_START_DATE'] || '2021-1-8')
 
   # says whether or not to send alerts to Sentry and support email for undeliverable emails
-  config.x.alert_undeliverable_emails = HostEnv.production?
+  config.x.alert_undeliverable_emails = true
```
Alex-Swann, line 101: you can probably remove this line now as what is in `app/services/govuk_emails/email_monitor.rb` only checks the values of this for development or test

PR #2300, opened 2021-02-19
Ccms report generator

Alex-Swann:
## What

Using L&D time to generate a CCMS report with xml payload details based on options in the Admin reports UI.
Feel free to pull this down and test, though tricky without real submissions to CCMS.

In this PR I've extended factories to accommodate successfully submitted cases to CCMS in the CCMS Submission and CCMS Submission History tables. With that, I've then got a custom report generator that uses a UI to capture what a user would like to find in terms of Apply/CCMS cases, whether they are passported or not, date ranges, capital assessment result and also to extract XML attribute values into a CSV file.

Screenshot 2021-03-17 at 19 23 04

## Checklist

### DIFF
--- a/app/services/reports/reports_types_creator.rb
+++ b/app/services/reports/reports_types_creator.rb
@@ -1,0 +1,170 @@
+module Reports
+  class ReportsTypesCreator # rubocop:disable Metrics/ClassLength
+    include ActiveModel::Model
+
+    def self.call(params)
+      new(params).call
+    end
+
+    attr_reader :application_type,
+                :submitted_to_ccms,
+                :capital_assessment_result,
+                :records_to,
+                :records_from,
+                :payload_attrs
+
+    validate :validate_required
+
+    def initialize(params)
+      @application_type = params[:application_type]
+      @submitted_to_ccms = params[:submitted_to_ccms]
+      @capital_assessment_result = params[:capital_assessment_result]
+      @payload_attrs = params[:payload_attrs]
+      @records_from = process_date(params, :records_from) || Date.parse('2019-12-01')
+      @records_to = process_date(params, :records_to) || Date.current
+    end
+
+    def call
+      search_applications if valid?
+      self
+    end
+
+    def generate_csv
+      CSV.generate do |csv|
+        if @results.present?
+          csv << %w[application_ref case_ccms_reference] + default_opts[:payload_attrs]
+          @results.each do |row|
+            csv << row.map { |v| "\"#{v}\"" }
+          end
+        end
+      end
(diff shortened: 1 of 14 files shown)


app/controllers/admin/admin_base_controller.rb
```diff
@@ -25,5 +25,10 @@ def set_cache_buster
       response.headers['Pragma'] = 'no-cache'
       response.headers['Expires'] = 'Fri, 01 Jan 1990 00:00:00 GMT'
     end
+
+    def convert_date_params(params)
```
mpw5, line 29: This code exists in this class and `ProviderBaseController`. Would it be worth moving it to the `ApplicationController` so that we're only defining it once?

app/services/reports/reports_types_creator.rb
```diff
@@ -0,0 +1,178 @@
+module Reports
+  class ReportsTypesCreator # rubocop:disable Metrics/ClassLength
```
mpw5, line 2: This is a big class 😬  There's probably too much going on it.

Would it be possible to split some of it out into another class to handle searching (ie moving the `search_applications` method and everything it calls from this class into the new one) which would just leave this one to actually create the report.

I haven't tried it so I'm not sure how easy it would be to do, but maybe worth considering.

mpw5 (approved):
Looks good. I've left a couple of comments about possible refactors, but I don't think either of them are problematic enough to block it.

app/services/reports/reports_types_creator.rb
Alex-Swann: That was possibly something I was going to do later. Just didn't want to spend lots of time doing all that if a) we break it up later if design, UR have some input, b) it didn't get merged because people didn't like it lol.

app/controllers/admin/admin_base_controller.rb
Alex-Swann: 👍

app/services/reports/reports_types_creator.rb
Alex-Swann: I could also point your attention to `app/models/legal_aid_application.rb` in the meantime 😬 😂

mpw5: Fair point. Don't spend any time on it now. Let's get some views on the new functionality and then we can worry about refactoring.

mpw5: > I could also point your attention to `app/models/legal_aid_application.rb` in the meantime as a form of distraction 😬 😂

Good point, well made

PR #2302, opened 2021-02-22
AP-2011 email refactor

stephenrichards:
## Refactor mail system

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2011)

The old system for sending mail suffered from several drawbacks:
- There was one class that was responsible for both sending and monitoring emails
- The audit information for each email was incomplete (no legal aid application id, no addressee), and the architecture made it very difficult to add that data to an audit table
- Each mailer specified a delivery job, which is an undocumented interface
- The job to deliver/monitor a mail was just added to the queue, so it was very difficult to see what was in the queue, or indeed to remove something from the queue if it was being sent repeatedly because of an error.

This PR rectifies those issues and implements a new lifecycle for emails:
- Emails are scheduled for delivery (either immediately or at some future date) by writing a record in the `scheduled_mailings` table.  Class methods `send_now!` and `send_later!` on `ScheduledMailing` are provided to facilitate this and ensure that the correct state is recorded, and appropriate delivery jobs and monitor jobs are scheduled.
- `ScheduledMailingsDeliveryJob` runs periodically (or when scheduled by one of the methods above) to pick up all mails waiting to be delivered, and pass them to `GovukEmails::DeliveryMan` to deliver and record the GOVUK Notify message id in the table.
- `EmailMonitorJob` runs periodically to find all find all `ScheduledMailing` records relating to mails that have been sent to GOVUK Notify, but not yet confirmed as delivered, and calls `GovukEmails::Monitor` to query their status and record it in the database, sending alert messages to sentry and the team email if it could not be delivered (for production only).

The `sent_emails` table and associated model have been deleted as the job is now done by the `scheduled_mailings` table.

## Checklist


app/jobs/email_monitor_job.rb
```diff
@@ -0,0 +1,17 @@
+class EmailMonitorJob < ApplicationJob
+  queue_as :default
+
+  DEFAULT_DELAY = 2.minutes.freeze
+  END_OF_TIME = Time.zone.local(9999, 12, 31, 23, 59, 59)
```
willc-work, line 5: looks like the method using this has moved to app/services/job_queue.rb

app/jobs/scheduled_mailings_delivery_job.rb
```diff
@@ -1,7 +1,17 @@
 class ScheduledMailingsDeliveryJob < ApplicationJob
   queue_as :default
 
+  DEFAULT_DELAY = 2.minutes.freeze
```
Alex-Swann, line 4: Can't this be moved to a constants or config file for reuse both here and in `app/jobs/email_monitor_job.rb` ?

app/models/scheduled_mailing.rb
```diff
@@ -1,34 +1,59 @@
 class ScheduledMailing < ApplicationRecord
+  # records an email for immediate or later delivery. Use .send_now! or .send_later! methods to schedule mails.
+
   serialize :arguments, Array
 
-  belongs_to :legal_aid_application
+  VALID_STATUSES = [
```
Alex-Swann, line 6:
```
%w[
 waiting
 processing
 (etc...)
]
```
would be better, surprised the linter hasn't complained

app/models/scheduled_mailing.rb
```diff
@@ -1,34 +1,59 @@
...
+    'waiting', # record has been written, but mail not yet sent by ScheduledMailingsDeliveryJob
```
Alex-Swann, line 7: Might be better to remove all these comments and stuck in the readme.

app/models/scheduled_mailing.rb
```diff
@@ -1,34 +1,59 @@
...
 
-  scope :not_sent, -> { where(sent_at: nil) }
-  scope :not_cancelled, -> { where(cancelled_at: nil) }
-  scope :past_due, -> { where('scheduled_at < ?', Time.current) }
+  MONITORED_STATUSES = %w[processing created sending].freeze
 
-  def self.due_now
-    not_sent.not_cancelled.past_due
-  end
+  FAILURE_STATUSES = %w[technical-failure permanent-failure temporary-failure].freeze
```
Alex-Swann, line 20: Do all these statuses have to be strings or could we make them all syms and replace hyphens with underscores so then they work?

Alex-Swann: Oh I see, it's because the statuses all go into the DB and removes the issue of type conversion. Righto.

Alex-Swann (approved):
There's a couple of comments in regards to END_OF_TIME and DEFAULT_DELAY from both Will and I about moving it to one place. But it might be you want them separated in case you wanted to set them to something else separately in future. There is also a suggestion about moving the comments to the Readme but I can leave that up to you. Otherwise LGTM. You think sending an email would be easy wouldn't you.

app/jobs/scheduled_mailings_delivery_job.rb
stephenrichards: I think we want them separate in case we ever want different delays on the two jobs.

Alex-Swann: cool, thought that might be the case

app/models/scheduled_mailing.rb
stephenrichards: The trouble with having them in the README, as opposed to in the code is that they get out of date.  The code is updated with new statuses, but the readme isn't.

PR #2310, opened 2021-02-25
Ap 2039 assigned scope

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2039)

Created a new join table assigned_scope_limitations to handle relationship between legal_aid_application, proceeding_type and scope_limitation

Each proceeding now has its now set of scope limitations (previously the scope limitations were on the application)

Removing a proceeding will also remove both substantive and delegated_functions scope limitations for that proceeding

If delegated_functions are used, a service creates a delegated_function scope_limitation for each proceeding on the application. 

If delegated_functions are removed then all of the associated scope_limitations are removed but the substantive scope limitation will remain linked to the proceeding

## Checklist


app/services/add_assigned_scope_limitation_service.rb
```diff
@@ -0,0 +1,28 @@
+class AddAssignedScopeLimitationService
+  def initialize(legal_aid_application, proceeding_type_id, scope_type)
+    @legal_aid_application = legal_aid_application
```
Alex-Swann, line 3: this isn't being used anywhere is it? There's no attr_reader either so don't see how another outside of this would use it either.

app/models/legal_aid_application.rb
```diff
@@ -325,6 +328,11 @@ def clear_scopes!
     reset_delegated_functions
   end
 
+  def delete_assigned_scope!(id)
+    scopes_to_clear = ApplicationProceedingTypesScopeLimitation.where(application_proceeding_type_id: id)
+    scopes_to_clear.each(&:delete)
```
Alex-Swann, line 333: nice

app/controllers/providers/used_delegated_functions_controller.rb
```diff
@@ -17,10 +17,26 @@ def update
     def save_continue_or_draft_and_update_scope_limitations
       return false unless save_continue_or_draft(@form)
```
Alex-Swann, line 18: you could probably remove false here as without it it would be nil (still falsey)

Alex-Swann (changes requested):
There are some very small tweaks, but this is very good. Not easy at all and I think I can see where it's going as the code certainly looks clear. Be great to get an explanation about the whole process and how it maps together during Dev time or something.

app/services/add_assigned_scope_limitation_service.rb
willc-work: When you say this, do you mean the line `@legal_aid_application = legal_aid_application`? If so we use it in the next line along with proceeding_type_id to get the specific proceeding. Let me know if I have misunderstood.

app/models/application_proceeding_type.rb
```diff
@@ -3,11 +3,26 @@ class ApplicationProceedingType < ApplicationRecord
...
 
+  def add_default_substantive_scope_limitation
+    assigned_scope_limitations << proceeding_type.default_substantive_scope_limitation
+  end
+
+  def add_default_delegated_functions_scope_limitation
+    assigned_scope_limitations << proceeding_type.default_delegated_functions_scope_limitation
+  end
+
+  def remove_default_delegated_functions_scope_limitation
```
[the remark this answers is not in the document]
willc-work: exactly.👍🏻

app/controllers/providers/used_delegated_functions_controller.rb
```diff
@@ -17,10 +17,26 @@ def update
...
 
+      remove_delegated_scope_limitations unless @form.model.used_delegated_functions?
       AddScopeLimitationService.call(legal_aid_application, :delegated) if @form.model.used_delegated_functions?
```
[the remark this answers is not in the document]
willc-work: ah yes, i was placing comments for those, looks like i missed one

app/services/add_assigned_scope_limitation_service.rb
Alex-Swann: I mean, couldn't you just ditch line 3 and on line 4 do
```
  @application_proceeding_type = legal_aid_application.application_proceeding_types.find_by(proceeding_type_id: proceeding_type_id)
```
without the need to make it an instance variable?

app/services/add_assigned_scope_limitation_service.rb
```diff
@@ -0,0 +1,28 @@
...
+    @application_proceeding_type = @legal_aid_application.application_proceeding_types.find_by(proceeding_type_id: proceeding_type_id)
```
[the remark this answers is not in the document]
stephenrichards: @Alex-Swann I'm not a fan of always making instance variables into attributes readable from outside the class.  There's not really much in it, but it does slightly pollute the API of the class, and the @ is a handy indicator that it is an instance variable.

db/migrate/20210218165806_create_join_table_application_proceeding_type_scope_limitation.rb
```diff
@@ -0,0 +1,10 @@
+class CreateJoinTableApplicationProceedingTypeScopeLimitation < ActiveRecord::Migration[6.1]
+  def change
+    create_join_table :application_proceeding_types, :scope_limitations, column_options: { type: :uuid } do |t|
+      t.primary_key :id, :uuid
+      t.timestamps
+      t.index %i[application_proceeding_type_id scope_limitation_id], name: 'index_application_proceeding_scope_limitation'
+      t.index %i[scope_limitation_id application_proceeding_type_id], name: 'index_scope_limitation_application_proceeding'
```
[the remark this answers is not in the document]
stephenrichards: Yep, I agree.  I think we need to add something into the readme to explain this.

app/controllers/providers/used_delegated_functions_controller.rb
willc-work: I think I prefer it be left in for clarity

app/services/add_assigned_scope_limitation_service.rb
willc-work: I misunderstood, yes good idea.

app/services/add_assigned_scope_limitation_service.rb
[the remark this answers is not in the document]
Alex-Swann: yes you are completely right. I was initially thinking about what we are doing elsewhere but then those attrs are also being read outside the class. Ignore me @willc-work

app/controllers/providers/used_delegated_functions_controller.rb
Alex-Swann: 👍

Alex-Swann (approved):
well done 🎉

PR #2324, opened 2021-03-02
AP-2064: Move ccms attrs to the land entity in the ccms payload to av…

Alex-Swann:
…oid ccms errors

## What

[AP-2064](https://dsdmoj.atlassian.net/browse/AP-2064)

Move land attrs (LAND_INPUT_C_5WP2_13A & LAND_INPUT_B_5WP2_26A)  to land entity in the CCMS payload.

Also updated payload helper to use generic system and rerun to ensure example payload reflects what we have in example xml.

## Checklist

### DIFF
--- a/ccms_integration/example_payloads/NonPassportedFullMonty.xml
+++ b/ccms_integration/example_payloads/NonPassportedFullMonty.xml
@@ -839,7 +839,7 @@
                 </ns3:Entity>
                 <ns3:Entity>
                   <ns3:SequenceNumber>13</ns3:SequenceNumber>
-                  <ns3:EntityName>CLICACLICAPITALPITAL</ns3:EntityName>
+                  <ns3:EntityName>CLICAPITAL</ns3:EntityName>
                   <ns3:Instances>
                     <ns3:InstanceLabel>the client's capital bond1</ns3:InstanceLabel>
                     <ns3:Attributes>
--- a/ccms_integration/helpers/ccms_payload_analyser.rb
+++ b/ccms_integration/helpers/ccms_payload_analyser.rb
@@ -2,18 +2,18 @@ require 'nokogiri'
 require 'csv'
 
 #
-# This class is useful for printing out all the attributes in the merits and measn ssections of a ccms payload file
+# This class is useful for printing out all the attributes in the merits and means sections of a ccms payload file
 #
 #
 class CCMSPayloadAnalyser
   def initialize
-    @doc = File.open('/Users/stephenrichards/moj/apply/ccms_integration/example_payloads/NonPassportedFullMonty.xml') { |f| Nokogiri::XML(f).remove_namespaces! }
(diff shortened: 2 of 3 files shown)


ccms_integration/example_payloads/NonPassportedFullMonty.csv
```diff
@@ -102,12 +102,12 @@ MeansAssesments,PROCEEDING,MATTER_TYPE,text,true
...
-MeansAssesments,CLICAPITAL,CLICAPITAL_INPUT_B_9WP2_8A,boolean,true
-MeansAssesments,CLICAPITAL,CLICAPITAL_INPUT_B_9WP2_11A,boolean,true
-MeansAssesments,CLICAPITAL,CLICAPITAL_INPUT_C_9WP2_7A,currency,true
-MeansAssesments,CLICAPITAL,CLICAPITAL_INPUT_B_9WP2_10A,boolean,true
+MeansAssesments,CLICACLICAPITALPITAL,CLICAPITAL_INPUT_T_9WP2_12A,text,true
+MeansAssesments,CLICACLICAPITALPITAL,CLICAPITAL_INPUT_B_9WP2_9A,boolean,true
+MeansAssesments,CLICACLICAPITALPITAL,CLICAPITAL_INPUT_B_9WP2_8A,boolean,true
+MeansAssesments,CLICACLICAPITALPITAL,CLICAPITAL_INPUT_B_9WP2_11A,boolean,true
+MeansAssesments,CLICACLICAPITALPITAL,CLICAPITAL_INPUT_C_9WP2_7A,currency,true
+MeansAssesments,CLICACLICAPITALPITAL,CLICAPITAL_INPUT_B_9WP2_10A,boolean,true
```
mpw5, line 110: Did you mean to change this? Looks like a find and replace gone wrong!

mpw5 (approved):
Looks good to me. One question about a change to one of the example payloads which looks as though it might be accidental. Not a blocker though

ccms_integration/example_payloads/NonPassportedFullMonty.csv
Alex-Swann: Yes I did. I reran the example payload creator to make sure it was up to date. This is generated from an example xml that has this attr name so they match. It may be at some point we need to review this as it’s not that synced up to the reality in the config. But thought to update it for now so at least it is consistent. You can see I’ve updated the helper too as it was referencing Stephen’s system

mpw5: Ah, I see. Makes sense. I think it's supposed to be `CLICAPITAL` in both places rather than `CLICACLICAPITALPITAL`, although I'm not sure how long it's been out of sync for. I guess it doesn't really matter as it's test data, and as long as the two files are consistent then it will work 🤷

Alex-Swann: Yes will do.

Alex-Swann: all fixed

PR #2339, opened 2021-03-04
AP-2035 remove cash payment feature flag etc

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2035)

Remove the `allow_cash_payment` Admin setting and all functionality, including CFE V2 requests.

As part of this I've removed the CFE V2 response model which has cascaded changes to a large number of unrelated factories and specs 😬. I think it's worth doing as there is no need to retain the V2 model in future and we'll probably have to make these changes at some point in future anyway, but sorry for the hefty PR...

## Checklist


stephenrichards (approved):
nicely done.  A lot more to this ticket than originally thought.

Alex-Swann (approved):
Ah Stephen beat me to it, good job

Alex-Swann (changes requested):
It looks like there are still references to V2 in the spec files, also I think there are duplicate tests. See `spec/requests/providers/means_reports_spec.rb`, there's a context assessment_v3_response but now you have changed the top of the file and using the factory with v3 you are essentially doing the same thing twice. There might be others doing that.

Alex-Swann (approved):

PR #2352, opened 2021-03-08
Fix delegated functions date format in cucumber

mpw5:
## What

Feature tests are failing because they expect the delegated function used on date to have a leading zero. This date is not displayed with a leading zero in the app. We've only just reached a date when this becomes a problems, so this hasn't been an issue up until now.

Thi fixes the expected date format to strip the zero.

## Checklist


Obsiye (approved):
LGTM!

PR #2355, opened 2021-03-08
AP-2095 remove respondent-details.js

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2095)

Removed file respondent-details.js. 
This file contained code for updated the labels and showing the revealed fields and errors of conditionally revealing radio buttons on the police notified details question.
This is no longer needed as all happens automatically since the page was updated with the new formbuilder. 
 
NOTE: 
The label updates in the JS were used in conjunction with the aria-expanded attribute to notify users of conditionally revealed content. The current design system recommendation is not to use the aria-expanded attribute, as it is not an allowed aria-attribute and therefore is often missed by screen readers anyway, and so the new formbuilder does not implement aria-expanded on conditional radio buttons. 
Notifying screen readers of revealed content in radio buttons is still an issue which is not fixed in the design system, but there is a planned fix for the end of March 2021: https://github.com/alphagov/govuk-frontend/issues/979 
 
## Checklist


Obsiye (approved):
LGTM

PR #2357, opened 2021-03-08
AP-2087 Require Caseworker review if DWP result has been overridden

stephenrichards:
## Require Caseworker review if DWP result has been overridden

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2087)

- Put the decision making for caseworker review in one place, `ManualReviewDeterminer` to be used both by Means Report Generator and CCMS payload 
- Update `ManualReviewDeterminer` to add in `:dwp_override` review reason if DWP result overridden
- Update `#applicant_receives_benefits?`, `#passported?` and `#non_passported?` to take DWP override into account
- Ensure only one DWPOverride record per 'LegalAidApplication`

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -229,6 +229,8 @@ class LegalAidApplication < ApplicationRecord
   end
 
   def applicant_receives_benefit?
+    return true if dwp_override
+
     benefit_check_result&.positive? || false
   end
 
--- a/app/services/ccms/manual_review_determiner.rb
+++ b/app/services/ccms/manual_review_determiner.rb
@@ -23,9 +23,10 @@ module CCMS
     end
 
     def manual_review_required?
-      return true if dwp_override
-
-      manually_review_all_non_passported? || capital_contribution_required? || has_restrictions?
+      dwp_override.present? ||
+        manually_review_all_non_passported? ||
+        capital_contribution_required? ||
+        has_restrictions?
     end
 
     def review_reasons
--- a/app/services/ccms/manual_review_determiner.rb
+++ b/app/services/ccms/manual_review_determiner.rb
@@ -10,29 +10,44 @@ module CCMS
     attr_reader :legal_aid_application
 
     delegate :cfe_result,
+             :dwp_override,
              :passported?,
              :non_passported?,
              :has_restrictions?, to: :legal_aid_application
 
     delegate :capital_contribution_required?, to: :cfe_result
 
-    def self.call(legal_aid_application)
-      new(legal_aid_application).call
-    end
-
     def initialize(legal_aid_application)
       @legal_aid_application = legal_aid_application
       raise 'Unable to determine whether Manual review is required before means assessment' if legal_aid_application.cfe_result.nil?
     end
 
-    def call
-      needs_review?
+    def manual_review_required?
+      return true if dwp_override
+
+      manually_review_all_non_passported? || capital_contribution_required? || has_restrictions?
+    end
+
+    def review_reasons
+      cfe_review_reasons + application_review_reasons
+    end
+
+    def review_categories_by_reason
+      cfe_result.remarks.review_categories_by_reason
     end
 
     private
 
-    def needs_review?
-      manually_review_all_non_passported? || capital_contribution_required? || has_restrictions?
+    def cfe_result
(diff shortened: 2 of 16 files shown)


app/models/legal_aid_application.rb
```diff
@@ -228,6 +229,8 @@ def add_benefit_check_result
   end
 
   def applicant_receives_benefit?
+    return true if dwp_override
+
     benefit_check_result&.positive? || false
```
Alex-Swann, line 234: Could this be refactored to
```
  def applicant_receives_benefit?
    dwp_override || benefit_check_result&.positive?
  end
```

app/models/dwp_override.rb
```diff
@@ -1,2 +1,5 @@
 class DWPOverride < ApplicationRecord
+  belongs_to :legal_aid_application
+
+  validates :passporting_benefit, :evidence_available, presence: true
```
Alex-Swann, line 4: These two attributes are added across two pages according to the design. Can we rely on separated validations in the model for each of these pages? Or leave just `validates :passporting_benefit` here and then update with evidence available on the 2nd step of the dwp override flow?

Alex-Swann, line 4: I think this can be removed. These are added over two separate pages according to the design, so we are planning to validate using the form model on each of those pages there instead.

app/services/ccms/manual_review_determiner.rb
```diff
@@ -10,29 +10,44 @@ class ManualReviewDeterminer
...
       @legal_aid_application = legal_aid_application
       raise 'Unable to determine whether Manual review is required before means assessment' if legal_aid_application.cfe_result.nil?
     end
 
-    def call
-      needs_review?
+    def manual_review_required?
+      return true if dwp_override
+
+      manually_review_all_non_passported? || capital_contribution_required? || has_restrictions?
```
Alex-Swann, line 28: as the list of review reasons is getting ever longer, maybe
```
def manual_review_required?
  reasons_to_review.any? true
end

def reasons_to_review
  [
    dwp_override,
    manually_review_all_non_passported?, 
    capital_contribution_required?, 
    has_restrictions?
  ]
end
```
up to you

Alex-Swann (changes requested):
Just a couple of styling suggestions and a comment about removing a validation check on the DWPOverride model

app/models/dwp_override.rb
stephenrichards: Agreed

app/models/legal_aid_application.rb
```diff
@@ -228,6 +229,8 @@ def add_benefit_check_result
   end
 
   def applicant_receives_benefit?
+    return true if dwp_override
+
```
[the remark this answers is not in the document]
stephenrichards: yep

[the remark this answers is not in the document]
stephenrichards: My immediate reaction was yes, but looking more closely, I think it's better as it is.  I am a believer in methods with a question mark only returning `true` or `false`.  Your suggestion would return nil if there was no `dwp_override` AND no `benefit_check_result`.

It could be 

   ```dwp_override || (benefit_check_result&.positive? || false)```

but I think that is harder to understand.

app/models/dwp_override.rb
stephenrichards: Agreed

app/services/ccms/manual_review_determiner.rb
stephenrichards: The only issue with that is it has to evaluate all four conditions even if dwp_override is true.  So perhaps I'll go with:

```def manual_review_required?
  dwp_override ||
    manually_review_all_non_passported? || 
    capital_contribution_required? ||
    has_restrictions?
end```

app/models/legal_aid_application.rb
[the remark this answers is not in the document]
Alex-Swann: fair enough 👍

app/services/ccms/manual_review_determiner.rb
Alex-Swann: yeah, that was the alternative 👍

Alex-Swann (approved):

PR #2370, opened 2021-03-10
AP 2084 DWP Confirm Benefits Received

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2084)

Page to confirm which benefit is received if client details are correct. 
Provider can only select 1 option.
If the provider selects one of the 5 state benefit options it should continue to new page AP-2085.
Provider receives an error if no option is selected and tries to continue with the journey. 
If the provider selects ‘None of the above’ continue to normal non-passported journey.

## Checklist

### DIFF
--- a/app/controllers/providers/check_provider_answers_controller.rb
+++ b/app/controllers/providers/check_provider_answers_controller.rb
@@ -15,7 +15,6 @@ module Providers
     end
 
     def continue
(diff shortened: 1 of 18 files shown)


Alex-Swann (approved):
Just fix linting otherwise approved 🎉

PR #2374, opened 2021-03-11
AP-2083: Add the confirm client details page for the DWP Override flow

Alex-Swann:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2083)

Add the confirm client details page and add some styling corrections

## Checklist

### DIFF
--- a/app/webpack/stylesheets/govuk-secondary-button.scss
+++ b/app/webpack/stylesheets/govuk-secondary-button.scss
@@ -1,5 +1,5 @@
 .govuk-button.govuk-secondary-button {
-  background-color: #f3f2f1;
+  background-color: govuk-colour("light-grey");
   color: #0b0c0c;
   box-shadow: 0 2px 0 #b5babe;
 }
--- a/app/controllers/providers/check_benefits_controller.rb
+++ b/app/controllers/providers/check_benefits_controller.rb
@@ -5,13 +5,13 @@ module Providers
     helper_method :should_use_ccms?
 
     def index
-      legal_aid_application.applicant_details_checked! unless legal_aid_application.applicant_details_checked?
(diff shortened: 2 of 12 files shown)


app/webpack/stylesheets/govuk-secondary-button.scss
```diff
@@ -1,5 +1,5 @@
 .govuk-button.govuk-secondary-button {
-  background-color: #f3f2f1;
+  background-color: #dee0e2;
```
mpw5, line 2: Just curious why you're changing this? The design system [documentation](https://design-system.service.gov.uk/components/button/secondary/index.html) uses `#f3f2f1` (or `govuk-colour("light-grey")` which might be more readable).

mpw5 (approved):
Small question about the button colour change. You'll have to do some rebasing after my changes are released but we've discussed that already so happy to approve 👍

app/webpack/stylesheets/govuk-secondary-button.scss
Alex-Swann: so in the prototype it is in fact this hexidecimal value I proposed to change. I remember on another service they found the default secondary colour button did not have enough contrast against the page background and found it less accessible. I assumed this might have also been the reason for the difference in the prototype. I will check with Stevey

Alex-Swann: Screenshot 2021-03-11 at 18 44 51
Screenshot 2021-03-11 at 18 44 47

Alex-Swann: prototype is wrong so have updated using your suggestion 🎉

PR #2375, opened 2021-03-11
BugFix - Capture exceptions in DeliveryMan when mailers raise an error

Obsiye:
## BugFix - Capture exceptions in DeliveryMan when mailers raise an error

Related to this sentry issue has been discarded because it can reoccur in staging and we're okay with that. Under Sentry staging, settings, inbound filters, discarded >
image

When the DeliveryMan calls `deliver_now!` on a mailer job but that mailer raises an error then we should capture the issue and send the exception to Sentry.

This will mean that the ScheduledMailingsDeliveryJob can continue to call the DeliveryMan for 'waiting' jobs / scheduled mailings.

Update tests

## Checklist

### DIFF
--- a/app/services/govuk_emails/delivery_man.rb
+++ b/app/services/govuk_emails/delivery_man.rb
@@ -19,6 +19,8 @@ module GovukEmails
       return unless scheduled_mail.waiting? # in case another job has picked it up
 
       eligible_for_delivery? ? deliver_now : cancel!
+    rescue StandardError => e
+      Sentry.capture_exception(e)
     end
 
     private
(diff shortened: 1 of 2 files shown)


app/services/govuk_emails/delivery_man.rb
```diff
@@ -19,6 +19,9 @@ def call
       return unless scheduled_mail.waiting? # in case another job has picked it up
 
       eligible_for_delivery? ? deliver_now : cancel!
+    rescue StandardError => e
+      Sentry.capture_exception(e)
+      true
```
Alex-Swann, line 24: does it need to return true?

Obsiye: true doesn't need it 👍🏾

Alex-Swann (approved):

PR #2386, opened 2021-03-15
AP-2046Z Application/Scope Limitation migration scripts

stephenrichards:
## Application/Scope Limitation migration scripts

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2046)

This PR contains a migration script to convert the old application_scope_limitation records into application_proceeding_types_scope_limitation records.  This is non-destructive (i.e. it doesn't delete the original records) and does not install any code which relies on the new records.  It will allow us to ensure that the migration has gone smoothly on the live database before deploying any code which relies on it.  It is also idempotent, so can be run multiple times with no adverse effects.

## Checklist

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,6 +15,7 @@ AllCops:
   - 'ccms_integration/*'
   - 'tmp/**/*'
   - 'log/**/*'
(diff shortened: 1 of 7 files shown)


Obsiye (approved):
Nice! LGTM!

PR #2389, opened 2021-03-15
Tech debt/db anonymiser

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2117)

Create a method to allow us to anonymise and download databases.  
## Next steps
- We know the time will extend from the current ~2 mins so this should be created and run by an asynchronous job so then ...
- upload to a folder in the existing s3 buckets for each environment
- Work out how to alert the requesting user that it has completed! 
- create a method of monitoring the rules so if a new table is added a rule is created to match it, even if it's empty
- update the secrets to use the RDS database_url  rather than having to construct it

## Checklist


db/anonymise/rules.rb
```diff
@@ -0,0 +1,144 @@
...
+  # application_scope_limitations
+  # ar_internal_metadata
+  # attachments
+  # bank_holidays
+  # benefit_check_results
+  # ccms_submission_documents
+  # ccms_submissions
+  # cfe_results
+  # cfe_submission_histories
+  # cfe_submissions
```
willc-work, line 22: should these 3 cfe entries be removed as we do anonymise some content on these tables

db/anonymise/rules.rb
```diff
@@ -0,0 +1,144 @@
...
+  },
+  cfe_submissions: {
+    cfe_result: nil
+  },
+  debugs: {
+
+  },
+  dependants: {
+    name: -> { Faker::Name.name },
+    date_of_birth: -> { Faker::Date.birthday }
```
willc-work, line 105: Dont we use the dependant DOB to calculate if they are a child? If so we may want to retain this DOB as is in the original database.

db/anonymise/rules.rb
```diff
@@ -0,0 +1,144 @@
...
+  },
+  dwp_overrides: {
+
+  },
+  feedbacks: {
+    improvement_suggestion: -> { Faker::Lorem.paragraph(sentence_count: 2) }
```
willc-work, line 111: we also have `email` in this table which should be anonymised

db/anonymise/rules.rb
```diff
@@ -0,0 +1,144 @@
...
+  providers: {
+    username: -> { Faker::Internet.username },
+    name: -> { Faker::Name.name },
+    email: -> { Faker::Internet.email }
+  },
+  respondents: {
+    full_name: -> { Faker::Name.name }
+  },
+  scheduled_mailings: {
+    # Should we just not export this?
```
willc-work, line 137: this table also has `addressee` which is always going to be an email address

willc-work (commented):
in the `offices` table we also have `code` which is probably a good idea to anonymise as these are linked to individual firms, even if they are not widely known I think it would be good to err on the side of caution.

db/anonymise/rules.rb
stephenrichards: 👍

willc-work (approved):
Looks good

README.md
```diff
@@ -430,6 +430,34 @@ This will then allow you to connect to the database, eg:
...
+
+It requires that you have kubectl authenticated and your context set to the current live context
+
+It will connect to the required kubernetes namespace and pod and run the `rake db:export` task this will generate a filename of `[environment].anon.sql`
+
+It will then copy the compressed restore file to the `tmp` folder in the project and de-compress it
+
+If a file from the same environment exists, it will prompt you to overwrite the local copy
+
+It will then an output a restore command to enable you to restore it to your local psql instance at your convenience
```
willc-work, line 450: >It will then an output 
It doesn't need that 'an'

colinbruce: My brain farted on adding the word `create` 🤦 
>It will then create an output...

PR #2390, opened 2021-03-15
Ap 2046y include type in unique index

stephenrichards:
## Include type in the unique index on application_proceeding_types_scope_limitations table

Running the AP-2046Z migration failed on UAT master because of a unique index violation.  This is because the substantive and delegated functions scope limitations for some proceeding types are the same, meaning that when you add delegated functions it tries to add an `AssignedDfScopeLimtiation` record with the same `application_proceeding_type_id` and `scope_limtiation_id` as an existing `AssignedSubstantiveScopeLimitation` record.  Making the index unique on `type`, `application_proceeding_type_id` and `scope_limitation_id` gives us what we need.

## Checklist


colinbruce (approved):
Looks okay 👍

PR #2408, opened 2021-03-22
Ap 2149 fix full text search

Alex-Swann:
## What

## Checklist

### DIFF
--- a/app/webpack/src/modules/proceedings.js
+++ b/app/webpack/src/modules/proceedings.js
@@ -1,25 +1,41 @@
 import axios from 'axios';
 import { hide, show, pluralize } from '../helpers';
+
 const doneTypingInterval = 500; // time in ms, 500ms to make search work fairly quickly but avoid too many DB requests
-let typingTimer; // timer identifier
-let ariaText;
+let typingTimer,
+  ariaText,
+  proceedingMatches = [],
+  noMatchCount = 0;
 
 async function searchResults (searchTerm) {
-  var currentUrl = window.location.href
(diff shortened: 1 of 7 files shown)


stephenrichards (approved):

app/webpack/src/modules/proceedings.js
```diff
@@ -1,6 +1,6 @@
 import axios from 'axios';
 import { hide, show, pluralize } from '../helpers';
-const doneTypingInterval = 500; // time in ms, 500ms to make search work fairly quickly but avoid too many DB requests
+const doneTypingInterval = 10; // time in ms, 10ms to make search work fairly quickly but avoid too many DB requests
```
RoseSAK, line 3: I think if you're going to make the interval so small you may as well remove it completely - I really doubt anyone is going to be typing quicker than one letter per 0.01 seconds. I think this will mean that the search will run after every letter rather than after waiting for the user to finish typing - this makes the results refresh frequently giving the sort of flashing effect below the search box. The interval needs to be long enough to realistically cover the period between typing two letters, otherwise there is no point having it - happy to discuss

RoseSAK (approved):

features/providers/civil_application_journey.feature
```diff
@@ -361,7 +361,7 @@ Feature: Civil application journeys
     Then I click 'Save and continue'
     Then I should be on a page showing 'Check your answers'
 
-  @javascript @vcr
+  @javascript @vcr @alex
```
colinbruce, line 364: 🤔

Alex-Swann: Whoops. And this is why we code review! 😂

PR #2409, opened 2021-03-22
AP-2111 Use transaction_period as basis for cash payment months

stephenrichards:
## Use transaction_period as basis for cash payment months

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2111)

CFE expects the cash payment dates to be the three months prior to the calculation date, which is either today or the delegated functions date if that has been used.

This PR changes the AggregatedCashIncome model to use the calculation date, rather than today's date.

## Checklist

### DIFF
--- a/app/webpack/src/modules/proceedings.js
+++ b/app/webpack/src/modules/proceedings.js
@@ -9,6 +9,7 @@ let typingTimer,
   previousSearchTerm = null,
   containSimilarWords = false;
 
+
 async function searchResults (searchTerm) {
   const currentUrl = window.location.href;
   const url = `/v1/proceeding_types?search_term=${searchTerm}&sourceUrl=${currentUrl}`;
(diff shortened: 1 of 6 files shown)


app/webpack/src/modules/proceedings.js
```diff
@@ -1,6 +1,6 @@
 import axios from 'axios';
 import { hide, show, pluralize } from '../helpers';
-const doneTypingInterval = 500; // time in ms, 500ms to make search work fairly quickly but avoid too many DB requests
+const doneTypingInterval = 100; // time in ms, 500ms to make search work fairly quickly but avoid too many DB requests
```
Alex-Swann, line 3: You don't need to change this. All fixed in my PR @stephenrichards

Alex-Swann (approved):
Yep, all makes sense. Just a comment to remove that time interval as sorted in my PR.

PR #2417, opened 2021-03-24
Identify renamed tables in migration_hook

mpw5:
## What

The `migration_hook` script used by the 'Check anonymise rules' GitHub action does not identify cases where a table has been renamed. It only alerts if there is a difference in the number of tables between the rules file and the schema.

This change amends the script to identify any difference in table names by finding the difference between the two arrays and failing with an appropriate message.

## Checklist


colinbruce (approved):
Nice :+1:

PR #2426, opened 2021-03-25
AP-2137 namespace merits assessment

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2137)

Rename and namespace the existing MeritsAssessment model as ProceedingMeritsTask::ChancesOfSuccess.  

References to merits_assessments plural are chances_of_successes.

## Checklist


app/webpack/src/success_prospects.js
```diff
@@ -0,0 +1,17 @@
+$(function() {
```
Alex-Swann, line 1: I think this needs to be rewritten in vanilla JS not using jQuery otherwise Rose will have to redo it.

app/webpack/src/success_prospects.js
```diff
@@ -0,0 +1,17 @@
...
+  $('#success-prospect-text-area').hide();
+  let optionName = $("input[type=radio][name='chances_of_success[success_prospect]']:checked").val();
+  if (typeof(optionName) != "undefined" && optionName != 'likely') {
```
Alex-Swann, line 4: Use `!==` here. It's a best practice JS standard as opposed to Ruby.

app/webpack/src/success_prospects.js
```diff
@@ -0,0 +1,17 @@
...
+    showSuccessProspectTextArea(optionName)
```
Alex-Swann, line 5: `;` at the end

app/webpack/src/success_prospects.js
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  $('.success_prospect_option').click(function() {
```
Alex-Swann, line 8: replace `function() {` with ES6 synax `() => {`

Alex-Swann (changes requested):
LGTM. Just some corrections needed in the client side JS.

app/services/flow/flows/provider_merits.rb
```diff
@@ -2,8 +2,8 @@ module Flow
   module Flows
     class ProviderMerits < FlowSteps
       STEPS = {
-        start_merits_assessments: {
```
mpw5, line 5: Renaming this step will cause problems for any in-progress applications that have a `provider_step` of `start_merits_assessments`. If a provider tries to progress them after this has been released it will error (I know this because it happened after I renamed the `respondent` step to `opponent` 😬). There are currently 16 applications in production at the `start_merits_assessments` step.

I think we need a small rake file that finds all the applications at that step and updates them to the new step. We can run it manually after your branch has been deployed. There are some examples of similar tasks in `lib/tasks/`.

I think this one would need to do something like
```
LegalAidApplication.where(provider_step: 'start_merits_assessments').each do |app|
  app.provider_step = 'start_chances_of_successes'
  app.save
end
```

possibly with a rollback method too, in case we need to revert it for some reason.

app/webpack/src/success_prospects.js
RoseSAK: ^good point although actually this file has been deleted in AP-1821 which was merged last Tuesday, if you rebase off master then there is no need to update this :)

RoseSAK: @eloisetait3 just tagging you in case you don't get notifications for single comments

Alex-Swann: That's what I thought. Cool.

eloisetait3: @RoseSAK Just rebased on master but the file is still there 😕 Should I just manually delete the file?

app/services/flow/flows/provider_merits.rb
eloisetait3: @mpw5 this makes sense, hadn't thought of that at all. Thanks for the info too. I'm not sure I know how to create a rollback task. I'll get one of the devs not on leave to pair on that!

mpw5: @eloisetait3 I think it would be similar to the `migrate` task above, just going in the opposite direction, eg:

```
task rollback: :environment do
  LegalAidApplication.where(provider_step: 'start_chances_of_successes').each do |app|
    app.provider_step = 'start_merits_assessments'
    app.save
  end
end
```

Sorry for requesting changes and then disappearing off on leave!

Alex-Swann (approved):
I’ve approved as I know I’m away from tomorrow and don’t want to block you. Remove that file in a commit and address Mark’s comments 👍

lib/tasks/migrate_merits_assessment.rake
```diff
@@ -0,0 +1,9 @@
+namespace :merits_assessment_flow_steps do
+  desc 'Rename start_merits_assessments flow step to start_chances_of_successes'
+  task migrate: :environment do
+    LegalAidApplication.where(provider_step: 'start_merits_assessments').each do |app|
+      app.provider_step = 'start_chances_of_successes'
+      app.save
+    end
+  end
+end
```
eloisetait3, line 9: @mpw5 Is this and the rollback rake file all that's needed for now?

Assuming nothing else needed except running it manually after merging

PR #2429, opened 2021-03-26
Add BinaryChoiceForm controller and update additional accounts contro…

Alex-Swann:
…ller

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2109)

Create a base model for yes/no radio button choice pages so that all pages that use this have a consistent way of doing it. Also so a form object is sent to the frontend so generic errors are displayed properly with the new formbuilder.

## Checklist


app/views/providers/confirm_dwp_non_passported_applications/show.html.erb
```diff
@@ -1,25 +1,33 @@
-<% error_class = @error ? 'alternative-error-style' : '' %>
+<% error_class = @form.errors.any? ? 'alternative-error-style' : '' %>
```
RoseSAK, line 2: this could be done in a separate PR but I think a better way of doing this would be to just override the error styles in the scss file for the interruption panel class - since the error styling will always be different for these panels

RoseSAK (approved):
Looks great! 👍

app/views/providers/confirm_dwp_non_passported_applications/show.html.erb
Alex-Swann: I added that to fix the form errors which the additions I made. But yeah I could stick on ID or class or something to do it instead.

app/views/providers/confirm_dwp_non_passported_applications/show.html.erb
```diff
@@ -1,25 +1,33 @@
...
+                                                 OpenStruct.new(value: false, label: t('.no'))],
+                                                :value,
+                                                :label,
+                                                legend: {text: content_for(:page_title), tag: 'h1', size: 'l'} %>
       </div>
 
-      <% if @error %>
-        <%= link_to_accessible @error.values.first, "#dwp-results-correct-true-field", class: 'govuk-heading-s' %>
+      <% if @form.errors.any? %>
+        <%= link_to_accessible @form.errors.values.first, "#confirm_dwp_non_passported_application-true-field", class: 'govuk-heading-s' %>
```
willc-work, line 30: By changing 'dwp-results-correct-true-field' to 'confirm_dwp_non_passported_application-true-field' the link no longer takes the user to the relevant field. The GovUk design says that clicking the error message should move the focus to the relevant field, so in this case we should move focus to the Yes radio button.

Alex-Swann: Yeah not sure how this got here, pushing fix now

PR #2436, opened 2021-03-29
Tech debt:remove mimemagic

colinbruce:
## What

After the issues with MimeMagic, the Rails core team moved to an updated version of Marcel that changed it's mime data source and removed any reference to mimemagic

We used a specific call to mime magic to determine the Statement of Case upload, so I have replaced that with a call to the equivalent Marcel call and then removed the mimemagic gem entirely

## Checklist


mpw5 (approved):
LGTM 👍

PR #2444, opened 2021-04-01
AP-2188: prevent duplicate scope limitations

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2188)

Add feature and rspec tests to ensure duplicate scope limitations are not created on re-submission of the DF page

## Checklist


stephenrichards (changes requested):
I know you have a cuke test for this, but I think we could do with a test on ProceedingType itself.

stephenrichards (approved):

PR #2452, opened 2021-04-07
AP-2131 Apply Legal framework service call

Obsiye:
## AP-2131 Apply Legal framework service call

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2131)

- Add a legal framework service to call the [legal_framework_api](https://github.com/ministryofjustice/legal-framework-api) and handle the response.

- The response with merits tasks is serialized and stored into a new table called `LegalFramework::MeritsTaskList`

- Rename the index for :legal_framework_merits_task_lists references :legal_aid_application to index: { name: 'idx_lfa_merits_task_lists_on_legal_aid_application_id' } from 'index_legal_framework_task_lists_...' this is because it surpassed the max characters of 63.

## Checklist


db/migrate/20210407081304_create_legal_framework_merits_task_list.rb
```diff
@@ -0,0 +1,9 @@
+class CreateLegalFrameworkMeritsTaskList < ActiveRecord::Migration[6.1]
+  def change
+    create_table :legal_framework_merits_task_lists, id: :uuid do |t|
+      t.references :legal_aid_application, foreign_key: true, type: :uuid, index: { name: 'idx_lfa_merits_task_lists_on_legal_aid_application_id' }
+      t.string :serialized_data
+      t.timestamps
```
stephenrichards, line 6: Ideally, we would want to make the serialized data `text` rather than `string` because it could potentially be very big, especially when we start implementing other endpoints on this API.  On postgres in fact, there is no difference, but on other databases there are, so in the interests of being database agnostic ....

stephenrichards (changes requested):
Looks Good.  Have left a comment on one of the database migrations.  And now we know the staging and production URLs for the  Legal Framework API, you could add those into the secrets if they're not already there.

As I worked on the serialization bit with you, you should get someone else to review as well.

willc-work (approved):

stephenrichards (approved):

PR #2461, opened 2021-04-08
AP-2132 merits task list

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2132)

Created controller and view for merits task list
Call to Legal Framework API in the controller 
Updated flow to redirect to merits task list if multiple proceedings flag is switched on, but continue along current merits path if not 
Added rspec tests for MeritsTaskListsController and updated tests for SerializableMeritsTasksLIsts

## Checklist


stephenrichards (dismissed):

app/controllers/providers/merits_task_lists_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  class MeritsTaskListsController < ProviderBaseController
+    def show
+      @merits_tasks = merits_tasks
+    end
+
+    private
+
+    def merits_tasks
+      YAML.load(LegalFramework::MeritsTasksService.call(@legal_aid_application).serialized_data).tasks # rubocop:disable Security/YAMLLoad:
+    end
```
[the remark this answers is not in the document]
Obsiye: Ah that's my bad. ^ Told her to use yaml.load :grimacing:

stephenrichards (approved):

app/webpack/stylesheets/task-list.scss
```diff
@@ -4,72 +4,74 @@
...
+  }
 }
 
-.app-task-list__task-completed {
-  float: right;
+.app-task-list__task-name p {
+  margin-bottom: 0;
 }
+
+// The `app-task-list__task-completed` class was previously used on the task
```
willc-work, line 64: when you say` existing prototypes` do you mean existing deployments in UAT or is this related to the prototypes on heroku? If the latter, is it still required to be included?

config/locales/en/providers.yml
```diff
@@ -444,6 +444,30 @@ en:
...
+        states:
+          not_started: Not started
+          waiting_for_dependency: Cannot start yet
+          complete: Completed
+        urls:
+          latest_incident_details: date_client_told_incident
+          opponent_details: opponent
+          statement_of_case: statement_of_case
+          chances_of_success: chances_of_success_index
+          application_children: chances_of_success_index
```
willc-work, line 469: are these 3 keys supposed to have the same values?

willc-work (changes requested):
The changes to the `values-uat/staging/production.yaml` files need to be removed or rebased against master. The sentry keys that these files are using are the old ones which were changed today.

config/locales/en/providers.yml
RoseSAK: Yeah for now, they'll be updated when the other controllers and views for these pages are merged in but they don't exist yet on this branch

app/webpack/stylesheets/task-list.scss
RoseSAK: this scss is mostly just copied from the design system task list component (which is experimental - also some of it didn't work so I've updated it hence why it looks a bit different if you do compare to their scss) that comment was part of it. I guess they're just saying it's to avoid breaking existing task lists. Yeah I can probably remove this actually

PR #2476, opened 2021-04-15
AP-2191: Add delegated functions to the multiple proceedings journey

Alex-Swann:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2191)

For this PR, after some discussion with Will, I've added a name column to the proceeding types table so that each proceeding type not only has a meaning that is used as a label, but also a name that is attribute friendly for use in fields and the form model. I have then setup a new fork in the provider flow for add multiple delegated functions with dates. As per the link to the prototype in the ticket, all application proceeding types that have been previously selected are listed, and the user can select which proceeding type has an associated delegated function date.

The rest of the work validates these and the checkboxes with all date scenarios that had been implemented for single proceedings plus to accommodate for additional complexity. I've also updated some reference to used_delegation_functions? which is currently on the legal_aid_application model, with a view to have all references to this being pointed to the application proceeding types (APT) - please reference ([AP-2234](https://dsdmoj.atlassian.net/browse/AP-2234)). I have also updated this latter model, so any APT can indicate: 

1) if delegated functions have been used at all, 
2) which application proceeding type has the earliest delegated functions, 
3) what the earliest delegated functions date is

I have also updated the first CYA page to list all APTs as well as if they have a delegated function date. The user can select a change link and alter these. This PR has also been setup for task [AP-2225](https://dsdmoj.atlassian.net/browse/AP-2225) to list all APTs above this too and to ensure functionality still works on single proceedings.

*** Failing cucumber VCR tests. Just need to update these (pair with someone) but otherwise those are isolated, the rest can be reviewed in the meantime.

## Checklist


db/migrate/20210415094249_add_delegated_function_columns_to_proceeding_type.rb
```diff
@@ -0,0 +1,20 @@
+class AddDelegatedFunctionColumnsToProceedingType < ActiveRecord::Migration[6.1]
```
Alex-Swann, line 1: This file will be removed once AP-2172 is merged

features/providers/civil_application_journey.feature
```diff
@@ -79,7 +79,7 @@ Feature: Civil application journeys
     Then the results section is empty
     Then proceeding search field is empty
 
-  @javascript @vcr
+  @javascript @vcr @alex
```
Alex-Swann, line 82: this will be removed, just working out why vcrs aren't working

app/controllers/providers/used_multiple_delegated_functions_controller.rb
```diff
@@ -0,0 +1,92 @@
...
+      application_proceedings_by_name.map(&:application_proceeding_type)
+    end
+
+    def earliest_df
+      form.earliest_delegated_functions
+    end
+
+    def earliest_df_date
+      earliest_df&.used_delegated_functions_on
+    end
```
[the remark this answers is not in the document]
mpw5: I wonder if naming could be better on those two methods. Maybe avoid using `df` instead of `delegated_functions`, so `earliest_df_date` becomes `earliest_delegated_functions_date`(longer but clearer).

Also it's not completely clear what `earliest_df` actually returns - maybe something like `proceeding_with_earliest_delegated_functions` would be clearer.

app/forms/legal_aid_applications/used_multiple_delegated_functions_form.rb
```diff
@@ -0,0 +1,166 @@
...
+    def save(params)
+      update_proceeding_attributes(params)
+      update_populated_dates(params)
+
+      return false unless valid? || draft_nothing_selected?
+
+      save_proceeding_records
+    end
+
+    def earliest_delegated_functions
```
mpw5, line 50: Similar to above,  `proceeding_with_earliest_delegated_functions` might be a clearer name.

app/models/application_proceeding_type.rb
```diff
@@ -31,6 +31,28 @@ class ApplicationProceedingType < ApplicationRecord
...
+
+  def earliest_delegated_functions_date
+    proceedings.minimum('used_delegated_functions_on')
+  end
+
+  def earliest_delegated_functions_reported_date
+    earliest_delegated_functions.used_delegated_functions_reported_on
+  end
+
+  def earliest_delegated_functions
```
mpw5, line 49: Similar to above, `proceeding_with_earliest_delegated_functions` might be a clearer name.

app/views/shared/check_answers/_item.html.erb
```diff
@@ -13,7 +13,7 @@
   </dd>
   <% unless read_only %>
     <dd class="govuk-summary-list__actions">
-      <% unless read_only %>
+      <% unless read_only || url == :show_blank_action %>
```
mpw5, line 16: Is this `unless` needed given there's one on line 14? If so should they both have the same conditions? What happens if `read_only == false` and `url == :show_blank_action`? It looks like we would display the `govuk-summary-list` but not the `change` link - is that what's intended?

mpw5 (approved):
Looks good to me 🎉 

 I've left a few comments about stylistic things, but nothing to block progress. Take a look and feel free to make changes if you agree or ignore if not.

app/views/shared/check_answers/_item.html.erb
Alex-Swann: Yes it is to keep the spacing/styling, but you are right the 2nd read only is redundant.

app/controllers/providers/used_multiple_delegated_functions_controller.rb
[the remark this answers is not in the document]
Alex-Swann: 👍

mpw5 (approved):
👍

PR #2487, opened 2021-04-22
Ap 2230 date field bug

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2230)

## Checklist


app/forms/concerns/date_field_builder.rb
```diff
@@ -78,6 +78,10 @@ def form_date
     @form_date ||= Date.new(*date_attributes)
   end
 
+  def input_field_values
+    [''] + from_form
+  end
```
RoseSAK, line 83: what does this method do?

RoseSAK (approved):
LGTM 👍

app/forms/concerns/date_field_builder.rb
eloisetait3: It returns the original (invalid) input values in the right order if there is a validation error. This will probably need a deeper rails fix and refactor down the line so the forms aren't trying to retrieve 0th values (so ideally we don't need the empty array], but it fixes the issue for now.

PR #2523, opened 2021-05-05
AP-2202 link errors

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2202)

Fix links in error summary:
To add the error summary to every page automatically, and get it to link to the fields from the form, I updated the page template tag to use the error summary generated by the formbuilder, however this means the page template tag now needs to be rendered _within_ the form tag so that the form can be passed to it. 
As part of this ticket I also removed some of the forms from the /shared folder into the /provider folder as these forms are now only rendered in the provider journey.  

## Checklist


colinbruce (approved):
One possible extra form that could be deleted but otherwise looking good 👍

PR #2537, opened 2021-05-10
AP-2289 duplicate DF dates on application_proceeding_type

stephenrichards:
## Duplciate DF dates on Application Proceeding Type record

This is just in prepareation for the switch to multiple proceedings - we need all new applications to start recording their DF dates on the application proceeding type record AS WELL AS the legal aid application, so that the DF dates refactoring in AP-2234 will work. 

There is already a rake task to copy over the data for older applications.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2289)

## Checklist

### DIFF
--- a/app/controllers/providers/limitations_controller.rb
+++ b/app/controllers/providers/limitations_controller.rb
@@ -3,11 +3,26 @@ module Providers
     include PreDWPCheckVisible
 
     def show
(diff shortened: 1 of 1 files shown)


RoseSAK (approved):
👍

PR #2538, opened 2021-05-11
Ap 2233 check delegated functions page

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2233)

In the multi proc flow, this page only exists if the provider enters a DF over 1 month old. Page displays the date if over 1 month. If there are multiple dates, the radio buttons + error message for blank both pluralise.

## Checklist

### DIFF
--- a/app/views/providers/confirm_multiple_delegated_functions/show.html.erb
+++ b/app/views/providers/confirm_multiple_delegated_functions/show.html.erb
@@ -1,11 +1,12 @@
-<%= page_template page_title: t('.heading', date: @used_multiple_delegated_functions_on), template: :basic do %>
+<%= form_with(
+      model: @form,
+      url: providers_legal_aid_application_confirm_multiple_delegated_functions_path,
+      method: :patch,
+      local: true
+    ) do |form| %>
 
-  <%= form_with(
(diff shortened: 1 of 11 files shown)


app/views/providers/confirm_multiple_delegated_functions/show.html.erb
```diff
@@ -0,0 +1,51 @@
+<%= page_template page_title: t('.heading', date: @used_multiple_delegated_functions_on), template: :basic do %>
```
RoseSAK, line 1: the page template tag and the form_with tag will need to be the other way around - this is a recent change from about a week ago! basically the error summary is rendered by the page template tag, and so you need to be able to pass the form to the page template tag so that it links correctly. Currently the link in the error summary on this page doesn't work (it should move focus to the fieldset). You can look at other pages as an example or happy to pair tomorrow if you like

eloisetait3: Switched! Yes makes sense. Is the new commit all that was needed?

RoseSAK (approved):
LGTM!

PR #2548, opened 2021-05-14
Tech-debt/Flickering tests

colinbruce:
## What

* Ensure the InvolvedChildrenController name is escaped with by the html_compare method
* added additional debugging tests to the used_multiple_delegated_functions_form_spec if the new ones pass and the original ones fail then we can delete the original ones, or delete the new ones if they are impractical 🤷‍♂️ 

Since adding the debug step I cannot make the test flicker, so I've left it in place to see if it will fail longer term

## Checklist


Obsiye (approved):
LGTM

willc-work (approved):

PR #2554, opened 2021-05-18
AP-2241 LASPO page

RoseSAK:
## What

Added controller, view and rspec tests for the LASPO page
Added new column to legal_aid_application to record response
Updated the flow so this page is only triggered if Section 8 proceedings have been selected
Updated the CYA page and merits report to show the response to the LASPO question (only displays if the application has S8 proceedings)

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -44,6 +44,7 @@ class LegalAidApplication < ApplicationRecord
   has_many :scheduled_mailings, dependent: :destroy
   has_one :state_machine, class_name: 'BaseStateMachine', dependent: :destroy
   has_many :involved_children, class_name: 'ApplicationMeritsTask::InvolvedChild', dependent: :destroy
+  has_one :in_scope_of_laspo, dependent: :destroy
 
   before_save :set_open_banking_consent_choice_at
   before_create :create_app_ref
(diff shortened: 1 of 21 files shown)


app/models/legal_aid_application.rb
```diff
@@ -157,6 +157,10 @@ def application_proceedings_by_name
     types.sort_by(&:meaning)
   end
 
+  def section_8_proceedings?
+    proceeding_types.any? { |type| type.ccms_matter == 'Section 8 orders' }
```
colinbruce, line 161: I have something similar in my task list PR, how would you feel about using the ccms_matter_code, I felt that it had less scope for typos, e.g. 
``` ruby
type.ccms_matter.eql?('KSEC8')
```
String comparison isn't great (but all either of us could use) and `Section 8 orders` could be `section 8 order` or whatever sausage finger dev types in the future 😄

spec/factories/proceeding_types.rb
```diff
@@ -49,6 +49,22 @@
       involvement_type_applicant { true }
     end
 
+    trait :as_prohibited_steps_order do
+      code { 'PH001' }
+      ccms_code { 'SE003' }
+      meaning { 'Prohibited Steps Order-S8' }
```
colinbruce, line 55: :cry: Jim pointed out that I introduced typos on these seeds, it should be `Prohibited steps order`
 I'll pick up the amendments on mine when I rebase :+1:

colinbruce (approved):
One observation and a question, but otherwise 👍

app/models/legal_aid_application.rb
RoseSAK: Ah yes, this makes sense, I'll update it!

PR #2561, opened 2021-05-21
AP-2204 fix bug where df date_reported_on was not being stored for cases where df date was over a month old

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2204)

removed commented out code that was no longer needed
revised a test that was checking something incorrectly
Set reported_on_date value in the form instead of calling a method on the model
Updated multiple DF form to no longer leave the value nil if the date is over a month ago

## Checklist


Obsiye (approved):
LGTM

PR #2566, opened 2021-05-24
Ap 2242 gateway evidence upload page

eloisetait3:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2242)

Gateway Evidence upload page created to upload additional supporting evidence if required. 

## Checklist


app/controllers/providers/gateway_evidences_controller.rb
```diff
@@ -0,0 +1,107 @@
...
+      render :show
+    end
+
+    private
+
+    def save_or_continue
+      if form.files?
+        save_continue_or_draft(form)
+      else
+        legal_aid_application.reload
```
eloisetait3, line 30: Is this a good solution to stop the application saving an obsolete gateway evidence record if there are no files attached?

stephenrichards: I don't think you need to reload it here.  From what I can see, the `GatewayEvidence` record is only created when you've pressed the upload button for the first time. So I think, when you're here and `form.files?` returns false, then the `GatewayEvidence` record hasn't been built and attached to the `LegalAidApplication` and you could just return nil.  Or am I missing something?

app/forms/providers/gateway_evidence_form.rb
```diff
@@ -0,0 +1,147 @@
+module Providers
+  class GatewayEvidenceForm # rubocop:disable Metrics/ClassLength
+    include BaseForm
+
```
stephenrichards, line 4: A lot of this seems identical to the `StatementOfCaseForm`, in whcih case I think you could extract the common code out into a `BaseFileUploader` from, then derive the `StatementOfCaseForm` and this form from it.  It seems like more work now, but if ever we have to change it, it means a change in just one place.

app/services/flow/flows/provider_merits.rb
```diff
@@ -124,6 +124,10 @@ class ProviderMerits < FlowSteps # rubocop:disable Metrics/ClassLength
         },
         merits_task_lists: {
           path: ->(application) { urls.providers_legal_aid_application_merits_task_list_path(application) },
+          forward: :gateway_evidences
+        },
```
stephenrichards, line 128: From reading the ticket, gateway evidence is only required if multiple proceedings have been selected, so I think here we want to go straight on to check merits answers if there is only one proceeding.

stephenrichards (changes requested):
A good job on a complex task, but there are a couple of things I would suggest doing.  Am happy to discuss.

app/forms/providers/gateway_evidence_form.rb
eloisetait3: Yes great minds... I had written that as a ticket to do off the back of this. Thought I'd get this up and working with the right tests + functionality, and then look at refactoring the forms/uploader together later

stephenrichards (approved):

PR #2570, opened 2021-05-25
AP 2285 increase emergency cost limit

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2285)

Add a new form to an existing page to allow users to increase the limit
only appears if DF has been used
Add emergency costs section to CYA

## Checklist


spec/factories/legal_aid_applications.rb
```diff
@@ -320,6 +320,16 @@
       own_home { 'mortgage' }
     end
 
+    # trait :with_no_increased_cost_limit do
```
Obsiye, line 323: Are these purposefully commented out? I don't see why it needs to be commented out

features/providers/civil_application_journey.feature
```diff
@@ -153,6 +153,10 @@ Feature: Civil application journeys
     Then I should be on a page showing 'Harassment - injunction'
     Then I should be on a page showing 'Emergency certificate'
     Then I should be on a page showing 'Substantive certificate'
+    Then I should be on a page showing "Do you want to request a higher cost limit?"
```
Obsiye, line 156: Should this be testing if the Setting to allow multiple proceedings is disabled? 

because the emergency_cost_override.feature tests when it's enabled

Obsiye (changes requested):
Hi it looks good. I left some minor comments

spec/requests/providers/limitations_spec.rb
```diff
@@ -61,14 +72,71 @@
...
+      end
+
+      context 'with the incorrect params' do
+        let(:params) do
+          { legal_aid_application: { emergency_cost_override: true,
+                                     emergency_cost_requested: 'non numeric response',
+                                     emergency_cost_reasons: 'this is just a test' } }
+        end
+
+        it 'what is this doing?' do
```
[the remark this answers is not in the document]
willc-work: not meant to still be there, I will remove it

spec/factories/legal_aid_applications.rb
willc-work: I created them and then didn't use them. I will remove them, they can be recreated if required in the future.

features/providers/civil_application_journey.feature
willc-work: my reading was that this feature goes live with multiple proceedings. I will clarify with Dave/Simon because right now this feature will not be visible when multi proc is set to off

willc-work: Confirmed this is the correct behaviour, new feature will go live with MP

Obsiye (approved):
LGTM

PR #2617, opened 2021-06-15
AP-2331 Non-passported dependant child attributes in CCMS payload

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2331)

- Added code to generate the `CLIENT_RESIDING_PERSON` entity if there are any dependant children
- Added code to generate and test the following attributes for *each* child dependant within that entity for non-passported applications
  - `CLI_RES_PER_INPUT_B_12WP3_20A`
  - `CLI_RES_PER_INPUT_B_12WP3_21A`
  - `CLI_RES_PER_INPUT_B_12WP3_23A`
  - `CLI_RES_PER_INPUT_B_12WP3_24A`

## Checklist

### DIFF
--- a/app/models/dependant.rb
+++ b/app/models/dependant.rb
@@ -3,6 +3,8 @@ class Dependant < ApplicationRecord
 
   belongs_to :legal_aid_application
 
+  scope :child_relative, -> { where(relationship: 'child_relative') }
+
   DEFAULT_VALUES = {
     in_full_time_education: false,
     relationship: 'child_relative',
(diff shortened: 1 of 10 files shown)


app/models/dependant.rb
```diff
@@ -28,7 +30,15 @@ def over_fifteen?
   end
 
   def eighteen_or_less?
-    age < 19
+    age <= 19
```
Obsiye, line 33: should that be `age <= 18`

Obsiye: Also, wondering where this method is used?

Obsiye (approved):
LGTM. One comment around an 'unused method'

PR #2632, opened 2021-06-21
AP-2359: Reduce Dashboard::UpdaterJob calls to geckoboard

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2359)

Previously we would update geckoboard after each step of the CCMS `dance`
That could mean 20-30 calls in a few seconds and would occasionally trip
the geckoboard rate limits

This PR attempts to reduce that noise bu sending a ccms_submission_saved job
when a CCMS submission is initialised and then updating again when it concludes
either completed or failed.  It also updates the Applications call to send
whether the job completes successfully or not.  This should clear the
CCMS error widget when manually pushing a CCMS job though

## Checklist

### DIFF
--- a/app/services/dashboard_event_handler.rb
+++ b/app/services/dashboard_event_handler.rb
@@ -46,8 +46,8 @@ class DashboardEventHandler
   end
 
   def ccms_submission_saved
-    Dashboard::UpdaterJob.perform_later('Applications') if payload[:state] == 'failed'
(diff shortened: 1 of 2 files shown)


Obsiye (commented):
Awesome! LGTM! Is there a test for when the state is neither initialised, failed and completed ? e.g. 'document_ids_obtained' and thus it doesn't call any updater jobs

Obsiye (approved):
Awesome LGTM

PR #2636, opened 2021-06-22
AP-2372: BugFix for DF and DWP override routing issue

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2372)

Extended the substantive_applications_spec tests to ensure the correct routing when a provider overrides the DWP result
but does not have evidence of the benefit received
Updated the applicant_receives_benefit method of LegalAidApplication to only return true if the provider has selected yes to having evidence of the benefit

## Checklist


Obsiye (commented):
Lgtm

Obsiye (approved):
LGTM

PR #2638, opened 2021-06-22
Fix Rake Job for cleaning whitespaces 

Obsiye:
## Fix Rake Job for cleaning whitespaces 

Don't squish whitespaces when nil attributes. In particular the Applicant, which can exist at bare minimum with no details.

## Checklist


willc-work (approved):
Looks good. I saw the alert in staging and was a bit confused how it returned an error for applicant.first_name, wouldn't we always expect this to exist? I dont see an issue with using the safe operator though so I have approved it.

PR #2643, opened 2021-06-23
AP-2356: Always create task list

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2356)

As part of the MP work the task list was being created on arriving at the task list controller, this creates it after the capital assessment page.  

This means it will be created for all applications regardless of the MP flag state and allows applications in progress to route back to the task list seamlessly
mp-switch

Tasks completed while the flag is false are still marked as complete as the code to mark tasks as complete checks for the existence before flagging and is not behind the flag

## Checklist

### DIFF
--- a/app/controllers/providers/capital_assessment_results_controller.rb
+++ b/app/controllers/providers/capital_assessment_results_controller.rb
@@ -9,6 +9,7 @@ module Providers
     end
 
     def update
+      LegalFramework::MeritsTasksService.call(legal_aid_application)
       continue_or_draft
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/providers/capital_assessment_results_controller.rb
```diff
@@ -9,6 +9,7 @@ def show
     end
 
     def update
+      LegalFramework::MeritsTasksService.call(legal_aid_application)
```
Obsiye, line 12: All lgtm.

Should there be a TODO or a note somewhere to remove this when the MP flag is switched on? As I see the merits task list controller will then handle it.

willc-work (approved):
LGTM

app/controllers/providers/capital_assessment_results_controller.rb
colinbruce: I was tempted to leave both in... the MeritsTaskListController will create one if needed and this one somehow fails!

Obsiye (approved):

PR #2651, opened 2021-06-24
AP-2321 Show uploaded files names on check answers page

Obsiye:
## AP-2321 Show uploaded files names on check answers page

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2321)

- Add migration to create a new column original_filename for Attachment to store the files original names.

- Show the uploaded files names on the check answers page. This includes statement of case and gateway evidence.

- Create a helper to return file names concatenated with their file size. 
- Test this helper

**Also**:
```
Update html class '.govuk-summary-list--no-border' to class '.govuk-summary-list__row--no-border'

This is done for row by row basis. This is because it was not working when printing the check merits answers
page. The browser print page/popup would still show the border. Now with the __row--no-border it no longer shows as expected.
This is because the class was being added to a summary list row and thus required a row by row basis to remove the
border e.g.  'govuk-summary-list__row--no-border'
```

**EDIT:**

Fix sentry bug which could be the cause of UAT issues atm
```
ArgumentError Sidekiq/PdfConverterWorker
'gateway_evidence_1_pdf' is not a valid attachment_type
```
## Checklist

### DIFF
--- a/app/views/shared/check_answers/_item.html.erb
+++ b/app/views/shared/check_answers/_item.html.erb
@@ -5,11 +5,25 @@
 
 <div class="govuk-summary-list__row <%= no_border %> normal-word-break" id="app-check-your-answers__<%= name %>">
   <dt class="govuk-summary-list__key govuk-!-width-one-half">
-    <%= question %>
+    <h2 class="govuk-heading-m"><%= question %></h2>
   </dt>
   <!--  in-line styling due to issues around pdf generation in our production envs -->
   <dd class="govuk-summary-list__value"<% if read_only %> style="text-align: right;"<% end %>>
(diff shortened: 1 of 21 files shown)


app/views/shared/check_answers/_item.html.erb
```diff
@@ -9,7 +9,21 @@
   </dt>
   <!--  in-line styling due to issues around pdf generation in our production envs -->
```
RoseSAK, line 10: Just wondering if this same issue is going to affect the 'no_border' rendering properly above? As this used to probably be done in the same way as 'no_border' is now? Not sure though

RoseSAK (approved):

PR #2652, opened 2021-06-24
BugFix: Address flickering tests

colinbruce:
## What

There are some issues with BigDecimal, see their issues page and
this seems to be a new occurence.  An issue has been raised on
their repo https://github.com/ruby/bigdecimal/issues/194 but
this seems to work on the previously flickering values

It forces the value into a big decimal while setting the
precision manually based on the string length of the
random value generated, minus the decimal place separator

## Checklist


Obsiye (approved):
Awesome!

PR #2653, opened 2021-06-24
Ap 2367 implement cost override

willc-work:
## What
Remove the settings flag from the cost override feature so it is accessible to any user who use delegated functions
Update limitations/legacy_proceeding_types to show the new form and display errors correctly
Update check_provider_answers so the new section is shown if DF have been used and the chnage link works to allow users to change the information
Update merits page to show the new emergency cost override details only on merits reports and not on the merits_assessment page (it cannot be changed from here due to a state change)

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2367)

## Checklist

### DIFF
--- a/app/controllers/providers/limitations_controller.rb
+++ b/app/controllers/providers/limitations_controller.rb
@@ -9,7 +9,7 @@ module Providers
     end
 
     def update
-      if Setting.allow_multiple_proceedings? && @legal_aid_application.used_delegated_functions?
+      if @legal_aid_application.used_delegated_functions?
         clear_limit_and_reason
         @form = LegalAidApplications::EmergencyCostOverrideForm.new(form_params)
         render :show unless save_continue_or_draft(@form)
--- a/app/views/providers/limitations/_legacy_proceeding_types.html.erb
+++ b/app/views/providers/limitations/_legacy_proceeding_types.html.erb
@@ -1,3 +1,8 @@
+<%= form_with(model: @form,
+              url: providers_legal_aid_application_limitations_path(@legal_aid_application),
+              method: :patch,
+              local: true) do |form| %>
+
 <%= page_template(page_title: t('.h1-heading')) do %>
 
   <h2 class="govuk-heading-m">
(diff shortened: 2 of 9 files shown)


app/views/providers/limitations/_legacy_proceeding_types.html.erb
```diff
@@ -34,6 +39,27 @@
   <p><strong><%= t('.allowable_work') %></strong><%= application_proceeding_type.substantive_scope_limitation.description %></p>
   </p>
 
+  <p class="govuk-body">
+  <p><%= t('.default_limit_is') %> <strong><%= gds_number_to_currency(@legal_aid_application.default_delegated_functions_cost_limitation,  precision: 0) %></strong></p>
```
Obsiye, line 43: Does this need to be replicated onto the _multi_proceeding_types.html.erb?
There's a views/providers/limitations/show.html.erb that uses the Setting.allow_multiple_proceedings? and the` if @legal_aid_application.used_delegated_functions?` so maybe there's crossover/duplications there

app/controllers/providers/limitations_controller.rb
```diff
@@ -9,7 +9,7 @@ def show
     end
 
     def update
-      if Setting.allow_multiple_proceedings? && @legal_aid_application.used_delegated_functions?
+      if @legal_aid_application.used_delegated_functions?
```
Obsiye, line 12: Hi, the request test needs updating to ignore the whole MP flag etc.

features/providers/civil_application_journey.feature
```diff
@@ -296,6 +296,9 @@ Feature: Civil application journeys
     Then I click 'Save and continue'
     Then I should be on a page showing "Covered under an emergency certificate"
     Then I should be on a page showing "Covered under a substantive certificate"
+    When I choose 'Yes'
+    And I enter a emergency cost requested '5000'
+    And I enter legal aid application emergency cost reasons field 'This is why I require extra funding'
```
Obsiye, line 301: lgtm. Could you have an integration test for the updated merits section?

app/views/providers/limitations/_legacy_proceeding_types.html.erb
willc-work: there is a `multi_proceeding_types.html.erb `and a `multi_proceeding_types_with_df.html.erb` this code exists on the `_df `version. I think once we go like with MP we can then remove this `legacy_proceeding_types.html.erb`

app/controllers/providers/limitations_controller.rb
willc-work: I have added a test to the section where MP flag is set to false. So we test on both MP and Single proc as they generate different pages, again when MP is rolled out we can remove the legacy one and the associated tests.

features/providers/civil_application_journey.feature
willc-work: The merits section stuff is hidden behind an `<if read_only>` check. This means it is only visible on the caseworker merits report and not visible to the user on the `check_merits_answers` page

Obsiye (approved):
LGTM

Obsiye (approved):

PR #2661, opened 2021-06-25
BugFix: Flickering Multiple DF dates

colinbruce:
## What

Ensure the application proceeding types are always sorted
in the order expected by the each_with_index blocks

The params are sent to the controller in the sequence a->b
but sometimes are created as b->a this caused the sequence
of date - i.days to be in the wrong order

This aims to make the tests idempotent

## Checklist


willc-work (approved):
LGTM

PR #2671, opened 2021-06-29
BugFix File uploads / Attachment creation errors for multiple attachments

Obsiye:
## BugFix Attachment creation errors for PDFConvertor

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2379)

**Error in sentry:**

```
ArgumentError Sidekiq/PdfConverterWorker
'statement_of_case_3_pdf' is not a valid attachment_type
```
If say a statement of case was uploaded 3 times then the attachment_name will increment to statement_of_case_3
which is fine. However, this multi-file upload triggers the error below:

The problem is in the PDFConvertor worker placing the wrong value in attachment_type for the Attachment model.
Attachment#attachment_type is an enum and e.g. 'statement_of_case_3' is not matching with 'statement_of_case_pdf' in the model.

- The solution is to get the type value from the #attachment_type method instead of the #attachment_name

**### This fixes the bug but does not fix existing failed PDFConvertor workers in all environments.**

.
## Checklist

### DIFF
--- a/app/services/pdf_converter.rb
+++ b/app/services/pdf_converter.rb
@@ -19,7 +19,7 @@ class PdfConverter
       pdf_filename = "#{File.basename(@original_attachment.attachment_name, '.*')}.pdf"
       pdf_attachment = Attachment.create!(
         legal_aid_application_id: @original_attachment.legal_aid_application_id,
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
LGTM 👍

PR #2681, opened 2021-06-30
BugFix Allow for existing attachments with no original filenames

Obsiye:
## BugFix Allow for existing attachments with no original filenames

Some attachments in our current database do not have original_filename values in their Attachment records.
This caused a bug because the attachment helper used on the check merits answers page was expecting something.

Use the attachment_type as the default name for the file when no original_filename exists.

This should not be required going forward because all attachments will now have original_filenames attached.

Related ticket https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/2651
## Checklist

### DIFF
--- a/spec/helpers/attachments_helper_spec.rb
+++ b/spec/helpers/attachments_helper_spec.rb
@@ -20,5 +20,15 @@ RSpec.describe AttachmentsHelper, type: :helper do
         expect(attachments_with_size(attachments)).to eq []
       end
     end
+
+    context 'without original_filename' do
+      let(:attachment1) { create :attachment, original_filename: nil }
+      let(:attachment2) { create :attachment, original_filename: 'fake name' }
+      let(:attachments) { [attachment1, attachment2] }
+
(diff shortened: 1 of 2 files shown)


spec/helpers/attachments_helper_spec.rb
```diff
@@ -20,5 +20,15 @@
...
       end
     end
+
+    context 'without original_filename' do
+      let(:attachment1) { create :attachment, original_filename: nil }
+      let(:attachment2) { create :attachment, original_filename: 'fake name' }
+      let(:attachments) { [attachment1, attachment2] }
+
+      it 'returns attachment_type name in its place' do
+        expect(attachments_with_size(attachments)).to match ['statement_of_case (15.7 KB)', 'fake name (15.7 KB)']
```
willc-work, line 30: is there any chance these attachments could be created in a different order and therefore the test would fail? If so we could use the `match_array` option

willc-work (approved):

PR #2695, opened 2021-07-06
AP-2333 Dependant attributes

stephenrichards:
## Add remaining Dependant attributes to CLIENT_RESIDING_PERSON entity

[This ticket](https://dsdmoj.atlassian.net/browse/AP-2333)
 and
[this ticket](https://dsdmoj.atlassian.net/browse/AP-2352)

Originally, these two tickets were to add single attributes relating to dependants to the payload.  However, these attributes were for adults, and the CLIENT_RESIDING_PERSON entity was only being added for child dependants.  It turns out that there needs to be an instance block for each dependant, with some attributes for adults only, some for children only and some for both.  This rectifiers that situation and should be the last of the dependant attribtues.

## Checklist

### DIFF
--- a/spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_dependant_child_attributes_spec.rb
+++ b/spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_dependant_child_attributes_spec.rb
@@ -16,13 +16,14 @@ module CCMS
                  username: 4_953_649
         end
 
-        let!(:adult_dependant) { create :dependant, :over18, legal_aid_application: legal_aid_application, number: 3 }
+        let!(:adult_dependant) { create :dependant, :over18, legal_aid_application: legal_aid_application, number: 3, assets_value: 8_000.01 }
         let(:ccms_reference) { '300000054005' }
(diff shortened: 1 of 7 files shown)


spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_dependant_child_attributes_spec.rb
```diff
@@ -39,10 +40,134 @@ module Requestors # rubocop:disable Metrics/ModuleLength
...
+              context 'attribute CLI_RES_PER_INPUT_T_12WP3_17A - Person residing: relationship to client' do
+                before do
+                  legal_aid_application.dependants.map(&:destroy!)
+                  create :dependant, legal_aid_application: legal_aid_application, date_of_birth: dob, relationship: relationship
+                  legal_aid_application.reload
+                end
+
+                context 'adult_relative' do
+                  let(:relationship) { 'adult_relative' }
+                  let(:dob) { Time.current - 22.years }
```
willc-work, line 100: I think for DOB it would be better to use Date.today instead of Time.current e.g `Date.today - 22.years`

So we are checking  `9 July 2001` instead of` 2001-07-09 15:50:05 +0100`

willc-work (approved):
Just added a comment on Date v Time for DOB. Not a blocker though so I have approved.

PR #2700, opened 2021-07-08
Ap 2376 review cucumber tests

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2376)

Removed some steps from cucumbers tests and replaced with factory setups to avoid repetition and speed up the tests
Grouped tests that focus on back button functionality and have put them in a single feature file so they can be commented out and only turned on when there is a need to test back_button
Re-recorded some tests
Removed duplicated tests

## Checklist


features/providers/back_button.feature
```diff
@@ -0,0 +1,214 @@
+#  The test below checks the back button functionality across multiple pages
+#  Introduced as a check after issues with the the use of the back button were identified
```
stephenrichards, line 2: Should this file be commented out?

stephenrichards (changes requested):

features/providers/back_button.feature
willc-work: Yes. These are the tests that revolve around the use of the back button, they were written when we had issues with back button behaviour. We fixed that so with the ethos of feature tests only testing happy paths these can be commented out with the option to easily re-enable for testing in the future. I dont feel they need to be run every time but they have value as regression tests if we experience back button issues in the future.

stephenrichards: 👍

stephenrichards (approved):

PR #2704, opened 2021-07-09
AP-2112 Bank account balance when there are no trasactions

stephenrichards:
## Bank account balance when there are no transactions

Normally, the balance reported on the means report is the latest balance from the bank account transactions.  However, if there are no transactions in the calculation period, then take the balance from the BankAccout record.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -176,6 +176,10 @@ class LegalAidApplication < ApplicationRecord
     LegalAidApplications::CalculationDateService.call(self)
   end
 
+  def year_to_calculation_date
+    [calculation_date - 1.year + 1.day, calculation_date]
+  end
+
   def capture_policy_disregards?
     (calculation_date || Time.zone.today) >= POLICY_DISREGARDS_START_DATE
   end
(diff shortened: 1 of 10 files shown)


app/models/legal_aid_application.rb
```diff
@@ -176,6 +176,10 @@ def calculation_date
     LegalAidApplications::CalculationDateService.call(self)
   end
 
+  def year_to_calculation_date
+    [calculation_date - 1.year + 1.day, calculation_date]
```
Obsiye, line 180: Hi, just wondering why `1.year`, i thought it'd be 3 months of transactions? Also how comes the `1.day` addition?

Obsiye: ohh, is this related to "student loan" being annual? If so will that be the only type of irregular income (annual ones)?

Obsiye (approved):
LGTM! I left a question around the a `year_to_calculation_date`

app/models/legal_aid_application.rb
stephenrichards: yes - they declare the total amount they have received in the past year.  The means calculation works out what the monthly figure is, but on the certificate, we're just playing back to them what they have declared.  So the period is different (1 year as opposed to three months) .

PR #2724, opened 2021-07-20
Tech debt/rubocop fix

colinbruce:
## What

We had ignored the new `Layout/LineEndStringConcatenationIndentation` cop because of the number of false positives, these have been addressed in a subsequent update, so I've removed the override and fixed the remaining issue

## Checklist


kmahern (approved):
👍

willc-work (approved):
lgtm

PR #2726, opened 2021-07-20
AP-2412: CCMS submission on/off setting

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2412)

Allow admin users of Apply to turn off CCMS submissions, there is at least one known set of CCMS downtime and, to prevent lots of alerts, we need to stop submitting data until we get the green light to continue

Add a setting to the admin page to turn off the feature
Add a service class that monitors and enables us to set the state of applications to `paused`
Extend state machine with a submission-paused entry
Add a rake task, and class, that gathers any paused submissions and restarts them with a new state machine event

## Checklist


stephenrichards (approved):
This looks great.  The only thing I would do is use the new Setting to prevent submissions from development , rather than the `Rails.configuration.x.ccms_soa.submit_applications_to_ccms` - but we can always do that later.

PR #2736, opened 2021-07-22
Ap 2391 proceeding type search

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=233&projectKey=AP&modal=detail&selectedIssue=AP-2391)

Added 'cao' to additional_search_terms for Child arrangements order (contact) and Child arrangements order (residence)
Removed 'non molesation' from additional search_terms for Harassment - injunction
Removed the limit from the number of proceeding types returned, as there are only 12 (and adding the new cao additional search terms was making one of the specs fail when only 10 were returned)
Amended feature to search for 'injunction' (as search for 'non' is no longer returning Harrassment - injunction)
Added tests for terms that only exist in additional search terms ie 'cao'
Ran rails db:migrate

## Checklist


colinbruce (approved):
LGTM 👍

PR #2757, opened 2021-07-28
Fix: Change pause_submission event handling

colinbruce:
## What

This was originally implemented at haste due to the
short notice of CCMS shutdown.  It was not clear
when developing at pace that the first call to the
CCMS APIs were made in the ReportsCreatorWorker.

Therefore when we 'paused' teh submissions the reports
creator was failing to create the documents as there
was no way to fulfill the `ensure_case_ccms_reference_exists`
by requesting a `case_ccms_reference`

This PR moves the pause step forward to after check_merits
and removes the check on whether the step should run in the
controller

## Checklist

### DIFF
--- a/app/controllers/providers/check_merits_answers_controller.rb
+++ b/app/controllers/providers/check_merits_answers_controller.rb
@@ -8,7 +8,7 @@ module Providers
 
     def continue
       unless draft_selected?
-        legal_aid_application.generate_reports! if legal_aid_application.may_generate_reports?
+        legal_aid_application.generate_reports!
         legal_aid_application.merits_complete!
       end
       continue_or_draft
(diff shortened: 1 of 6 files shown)


app/controllers/providers/check_merits_answers_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
 
     def continue
       unless draft_selected?
-        legal_aid_application.generate_reports! if legal_aid_application.may_generate_reports?
+        legal_aid_application.generate_reports!
```
willc-work, line 11: If we are removing this, can we remove:
app/models/legal_aid_application.rb:96 ?

stephenrichards (approved):

PR #2759, opened 2021-07-28
Ap 2385 release merits task list

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2385)

remove MP flag setting from merits task list options.

Allows merits task list to be used before MP is released

Fix for the issues below:
When setting up the applications for the task list, it needs to be explicit about the proceeding type.

Previous issues were being caused by the factory bot assigning sequential numbers
```ruby
 sequence(:ccms_code) { |n| format('DA%<number>03d', number: n) }
```
Meaning that when recorded individually it would return DA001, but when run as part of the suite it could be DA008, DA0009, etc and the cassette would then be 'wrong' and the tests would fail

<details>
<summary>Issues:</summary>
currently 3 feature tests are failing:
features/accessibility/provider.feature:206 # Scenario: I complete the non-passported merits assessment and it is accessible
features/providers/civil_application_journey.feature:446 # Scenario: I am able to view the client completed means answers
features/providers/civil_application_journey.feature:505 # Scenario: Completes the merits application for applicant that does not receive passported benefits

Running civil_application_journey.feature:505 as a single test and it will pass, running it as part of the whole file means it will fail (and usually it is only this test that fails, 446 in this scenario is fine)
</details>
## Checklist

### DIFF
--- a/spec/factories/legal_aid_applications.rb
+++ b/spec/factories/legal_aid_applications.rb
@@ -298,32 +298,6 @@ FactoryBot.define do
       end
     end
 
-    trait :with_multiple_proceeding_types_no_section8 do
-      after(:create) do |application, evaluator|
-        if evaluator.proceeding_types.presence
-          application.proceeding_types = evaluator.proceeding_types
-        else
-          application.proceeding_types << create(:proceeding_type, :with_real_data)
-          application.proceeding_types << create(:proceeding_type, :as_occupation_order)
-        end
-        lead_apt = application.application_proceeding_types.find_by(lead_proceeding: true)
-        if lead_apt.nil?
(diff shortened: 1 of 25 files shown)


spec/factories/legal_aid_applications.rb
```diff
@@ -298,6 +298,32 @@
       end
     end
 
+    trait :with_multiple_proceeding_types_no_section8 do
```
kmahern, line 301: I can't see where we are using this trait?

kmahern (approved):
We have added a trait that it looks like we aren't using? Not sure if this is just for completeness/future use? Otherwise lgtm.

spec/factories/legal_aid_applications.rb
colinbruce: Good catch, I think Will added while debugging issues and then stopped using it.  
I'll remove it for now :+1:

PR #2772, opened 2021-08-03
AP-2439 Give namespaces meaningful names

stephenrichards:
### ~ON HOLD~ Temporary Fix applied for this issue: This currently has the Title for opponent and involved children set to `MR.` and `MASTER`.  This should not be merged until this issue is resolved 
---

### Fix for the above issue:  All cases(single proc and MP) will submit the opponent organisation details, that we have used as the default for all Apply applications. Once there is a permanent fix for the title issue in CCMS this change can be reverted to use the actual opponent details

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2439)

Relationship data in the payload was being ignored by CCMS.  This turned out to be because the attributes were in the wrong namespace.

Namespace naming in the files is a problem.  because they were originally taken from example PUI files, we followed the PUI convention of using `ns1`, `ns2`, `ns3` etc.  The problem is that PUI allocates these randomly.  So in one file, the `Header` namespace might be called `ns1` and the `Security Extension` namespace `ns2`, in another file they might be called `ns4` and `ns1` respectively.  This makes it easy to make get the namespace wrong when adding new attributes when reverse-engineering a PUI example file.

In this PR, we have standardised all he XML namespaces, so that the `Header` namespace is always named `hdr`, and the `Security Extension` namespace is always named `secext`.

This has the added advantage that the namespaces can now be specified once in the `CCMS::Requestors::BaseRequestor` class instead of having to be separately specified in each of the derived requestor classes, and adds context to the namespace names rather than just numbers.

## Checklist


app/services/ccms/requestors/base_requestor.rb
```diff
@@ -1,16 +1,40 @@
...
+
+      attr_reader :namespaces
+
+      def initialize
+        @namespaces = NAMESPACES
+      end
+
       class << self
-        attr_reader :wsdl, :namespaces
+        # attr_reader :wsdl, :namespaces
```
mpw5, line 28: This can be removed

app/services/ccms/requestors/base_requestor.rb
```diff
@@ -1,16 +1,40 @@
...
         def wsdl_from(filename)
           @wsdl = filename
         end
 
-        def uses_namespaces(namespaces)
-          @namespaces = namespaces.freeze
-        end
+        # def uses_namespaces(namespaces)
+        #   @namespaces = namespaces.freeze
+        # end
```
mpw5, line 37: These can be removed too.

app/services/ccms/requestors/case_add_requestor.rb
```diff
@@ -84,209 +73,216 @@ def request_xml
...
-                xml.__send__('ns2:FirstName', first_name)
+        xml.__send__('casebio:OtherParty') do
+          xml.__send__('casebio:OtherPartyID', "OPPONENT_#{child.generate_ccms_opponent_id}")
+          xml.__send__('casebio:SharedInd', false)
+          xml.__send__('casebio:OtherPartyDetail') do
+            xml.__send__('casebio:Person') do
+              xml.__send__('casebio:Name') do
+                xml.__send__('common:Title', 'MASTER')
+                xml.__send__('common:Surname', last_name)
+                xml.__send__('common:FirstName', first_name)
```
mpw5, line 187: Should these lines use the `casebio` namespace instead of `common`? In the original code they were all `ns2` which suggests they should all map to the same new namespace (`casebio`).

app/services/ccms/requestors/case_add_requestor.rb
```diff
@@ -362,54 +358,54 @@ def resolve_instance_label(config)
...
 
       def generate_vehicle_instance(xml, vehicle, config)
-        xml.__send__('ns0:Instances') do
-          xml.__send__('ns0:InstanceLabel', config[:instance_label])
-          xml.__send__('ns0:Attributes') { EntityAttributesGenerator.call(self, xml, config[:yaml_section], vehicle: vehicle) }
-          # xml.__send__('ns0:Attributes') { generate_attributes_for(xml, config[:yaml_section], vehicle: vehicle) }
+        xml.__send__('common:Instances') do
+          xml.__send__('common:InstanceLabel', config[:instance_label])
+          xml.__send__('common:Attributes') { EntityAttributesGenerator.call(self, xml, config[:yaml_section], vehicle: vehicle) }
+          # xml.__send__('common:Attributes') { generate_attributes_for(xml, config[:yaml_section], vehicle: vehicle) }
```
mpw5, line 393: Can this be removed?

mpw5 (approved):
There are a few comments that can be deleted, and I've asked one question about an inconsistent change of namespace in one file.

Other than that it looks good, so I'm approving 👍

app/services/ccms/requestors/case_add_requestor.rb
willc-work: I have changed this, it looks like it has been set as that since the first commit, but you are correct that it has always been `ns2/casebio `as per the others fields in this attribute. I will make sure that we get it checked on CCMS submission to see if there is any weird behaviour as the result of the change.

mpw5 (approved):
LGTM

PR #2776, opened 2021-08-05
AP-2429: Reduce time spent running unit tests

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2429)

### Stub out libreconv
As we only expect it to be called, not necessarily testing the outcome.
This reduces the time taken to run `rspec spec/services/pdf_converter_spec.rb`
from `Finished in 23.84 seconds` to `Finished in 1.63 seconds`

### Remove LibreOffice from the CircleCI

As we are stubbing it we may not need to install it ...
It has reduced the `install_packages_for_testing` step time from `55s` to `21s`

### Remove the i18n rspec tests

These could be run as a separate shell script via a github action?
It can be removed as it doesn't affect coverage

### Remove the ccms rspec tests

There are ~600 CCMS service tests and they take almost half the time to run the full suite.
As they rarely change, they have been tagged in rspec with `:ccms` and ignored by default

To run the CCMS tests locally, if you are changing the values you can run rspec with `INC_CCMS=true bundle exec rspec` and they will not be ignored and the coverage will be monitored

### Add scheduled, overnight, jobs

There a github action that runs:
* i18n checks _only_ at 7.05AM 

And a circle job that runs each night
* rspec (including ccms, excluding i18n) at 2:15AM

Both of them will output a message to the standard dev alerts channel when they fail

# Checklist

### DIFF
--- a/spec/models/dependant_spec.rb
+++ b/spec/models/dependant_spec.rb
@@ -74,7 +74,7 @@ RSpec.describe Dependant, type: :model do
         expect(dependant.over_fifteen?).to eq(false)
       end
 
-      it 'returns true' do
+      it 'returns false' do
         expect(dependant.sixteen_or_over?).to eq(false)
       end
     end
@@ -98,7 +98,7 @@ RSpec.describe Dependant, type: :model do
         expect(dependant.over_fifteen?).to eq(false)
       end
 
-      it 'returns true' do
+      it 'returns false' do
         expect(dependant.sixteen_or_over?).to eq(false)
       end
     end
--- a/spec/models/dependant_spec.rb
+++ b/spec/models/dependant_spec.rb
@@ -73,6 +73,10 @@ RSpec.describe Dependant, type: :model do
       it 'returns false' do
         expect(dependant.over_fifteen?).to eq(false)
       end
+
+      it 'returns true' do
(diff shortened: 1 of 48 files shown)


willc-work (approved):

mpw5:
Looks good to me. I think it would be worth getting the i18n tests runnable in some other fashion, either in github actions or even just as a rake task, especially if we ever do the Welsh translations. I don't think it should block this PR though 👍

stephenrichards (approved):

spec/models/dependant_spec.rb
```diff
@@ -73,6 +73,10 @@
       it 'returns false' do
         expect(dependant.over_fifteen?).to eq(false)
       end
+
+      it 'returns true' do
```
willc-work, line 77: is this supposed to be 'returns false'?

spec/models/dependant_spec.rb
```diff
@@ -89,6 +97,10 @@
...
```
willc-work, line 101: is this supposed to be 'returns false'?

willc-work (approved):
Looks great, just some queries on test names that might need amending for clarity.

PR #2784, opened 2021-08-09
AP-2447 Bugfix - handle Case already submitted response

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2447)

Consider a response with both`Case successfully created` and `Case successfully created.Case already exist, exiting.` to be successful.

## Checklist

### DIFF
--- a/spec/services/ccms/parsers/case_add_status_response_parser_spec.rb
+++ b/spec/services/ccms/parsers/case_add_status_response_parser_spec.rb
@@ -5,7 +5,6 @@ module CCMS
     RSpec.describe CaseAddStatusResponseParser do
       context 'successful response' do
         describe '#success?' do
-          let(:response_xml) { ccms_data_from_file 'case_add_status_response.xml' }
           let(:expected_tx_id) { '20190301030405123456' }
 
           context 'normal success response' do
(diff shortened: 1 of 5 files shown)


spec/services/ccms/parsers/case_add_status_response_parser_spec.rb
```diff
@@ -8,36 +8,64 @@ module Parsers
           let(:response_xml) { ccms_data_from_file 'case_add_status_response.xml' }
           let(:expected_tx_id) { '20190301030405123456' }
 
-          it 'extracts the status' do
-            parser = described_class.new(expected_tx_id, response_xml)
-            expect(parser.success?).to eq true
-          end
+          context 'normal success response' do
+            let(:response_xml) { ccms_data_from_file 'case_add_status_response.xml' }
```
colinbruce, line 12: Is this needed as well as line 8?

colinbruce (approved):
One observation on duplicated line, but otherwiseLGTM :+1:

spec/services/ccms/parsers/case_add_status_response_parser_spec.rb
stephenrichards: no you're right.  line 8 can go.  have removed it

PR #2785, opened 2021-08-09
AP-2416: Handle accented characters

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2416)

This was impossible to test in code, I wrote explicit Rspec and cucumber tests and they all passed, including `BROWSER=chrome cucumber`

Then I found [this unmerged PR](https://github.com/rails/rails/pull/41517) to fix it but in the meantime I used the fix from the original issue [here](https://github.com/rails/rails/issues/38080#issuecomment-589856401) and tested in browser

Because the headers do not seem to be generated when running tests the encode_header block needs `:nocov:` because the tests never hit it 😢 

## Checklist

### DIFF
--- a/app/controllers/concerns/providers/application_dependable.rb
+++ b/app/controllers/concerns/providers/application_dependable.rb
@@ -13,6 +13,7 @@ module Providers
     end
 
     included do
+      before_action :encode_upload_header
       before_action :set_legal_aid_application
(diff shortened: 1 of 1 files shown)


willc-work (approved):
LGTM

PR #2842, opened 2021-09-01
AP-2355 Use class inheritance for forms

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2355)

The main reason for this ticket is to extract the common code out of the two forms that upload files (`StatementOfCaseForm`, `GatewayEvidenceForm`) into the `BaseFileUploaderForm`.

 However making the two forms subclasses of `BaseFileUploaderForm` didn't play nicely with the fact that all forms included `BaseForm` as a module, rather than subclass it, so we have changed the `BaseForm` from a  module to a class, and made all the other forms derive from it rather than include it as a module.

One little wrinkle is that when there are errors on the form, the `ActiveModel::Naming` module  calls `#model_name` on the form ***and all its ancestors***.  Normally calling `#model_name` on a form will return whichever model is defined in the `.form_for` method, but for the superclasses `BaseFileUploaderForm` and `BaseForm`, this will be nil, which in turn will cause an exception.  As this information is never used by `ActiveModel::Naming`, we just return the superclass name in this case.

## Checklist


app/forms/providers/gateway_evidence_form.rb
```diff
@@ -1,33 +1,12 @@
 module Providers
-  class GatewayEvidenceForm # rubocop:disable Metrics/ClassLength
-    include BaseForm
+  class GatewayEvidenceForm < BaseFileUploaderForm
+    # include BaseForm
+    # include BaseFileUploaderForm
```
mpw5, line 4: I think these lines can be removed.

app/forms/statement_of_cases/statement_of_case_form.rb
```diff
@@ -1,33 +1,12 @@
 module StatementOfCases
-  class StatementOfCaseForm # rubocop:disable Metrics/ClassLength
-    include BaseForm
+  class StatementOfCaseForm < BaseFileUploaderForm
+    # include BaseForm
+    # include BaseFileUploaderForm
```
mpw5, line 4: I think these lines can be removed.

mpw5 (approved):
Looks good to me. There a few lines that have been commented out and can be removed, but not a blocker.

PR #2846, opened 2021-09-02
AP-2497 Alternative alerting mechanism

stephenrichards:
## Alternative alerting mechanism (to Sentry)

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2497)

This PR is a first stab at providing an alternative to Sentry for alerting about exceptions.

- New column on `settings` table to indicate whether or not to use Sentry for alerts
- New class `AlertManager` whose job it is to decide where to send the alert (Sentry or Slack depending on the value in `settings`), and to filter out alerts depending on the type of alert/message or the environment
- New class `SlackAlerter` to send alerts to Slack.  This class responds to the same methods as Sentry, i.e. `.capture exception` and `.capture_message`. Alerts messages and/or exceptions are sent to Slack using email for ease and speed of implementation.
- New mailer `ExceptionAlertMailer` to format the message and send to slack.  

Notes:  
- Slack alert emails are sent immediately, and not via a background job
- All filtering of messages should be done in `AlertManager` where a mechanism exists for this
- Currently, there is no way of capturing the URL of the request that caused the exception - I think this would require the URL to be passed to the `AlertManager#capture_expection` method as a parameter.
- I haven't added the new `settings` field to the Admin page - I expect it to be changed rarely, so we can do it manually in the database.

## Checklist


app/lib/slack_alerter.rb
```diff
@@ -0,0 +1,33 @@
+class SlackAlerter
+  SLACK_CHANNEL_EMAIL = 'apply-alerts-prod-aaaabbocbcpykkqticw7skurte@mojdt.slack.com'.freeze
```
colinbruce, line 2: Should this be public or regenerated and set in the deployment settings on a per-environment basis?
Wouldn't this mean that a single channel gets _all_ alerts?!

colinbruce (changes requested):
Some issues...

config/routes.rb
```diff
@@ -240,6 +240,12 @@
     end
   end
 
+  # dummy route to test errors in order to see how they are reported in Sentry/Slack
+  # namespace :test do
+  #   resource :generate_error, only: :show
+  # end
```
mpw5, line 246: Can these comments be removed?

spec/factories/addresses.rb
```diff
@@ -10,5 +10,9 @@
     trait :is_lookup_used do
       lookup_used { true }
     end
+
```
mpw5, line 13: This doesn't seem related to the rest of the PR?

spec/factories/applicants.rb
```diff
@@ -10,6 +10,10 @@
       addresses { build_list :address, 1 }
     end
 
+    trait :with_address_for_xml_fixture do
```
mpw5, line 13: This doesn't seem related to the rest of the PR?

spec/services/ccms/requestors/case_add_requestor_spec.rb
```diff
@@ -23,7 +23,7 @@ module Requestors
         end
         let(:applicant) do
           create :applicant,
-                 :with_address,
+                 :with_address_for_xml_fixture,
```
mpw5, line 26: This doesn't seem related to the rest of the PR?

app/lib/alert_manager.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      alert_klass.capture_message(message)
+    end
+
+    def alert_klass
+      Setting.setting.alert_via_sentry? ? Sentry : SlackAlerter
+    end
+
+    def sendable?(message_or_exception)
+      return false unless HostEnv.environment.in?(%i[production uat])
```
mpw5, line 20: Should this also include staging?

mpw5 (commented):
Looks good. I've left a few questions and comments. Happy to approve once they've been resolved 👍

spec/factories/addresses.rb
```diff
@@ -10,5 +10,9 @@
...
+    trait :with_address_for_xml_fixture do
+      postcode { 'GH08NY' }
+    end
```
mpw5, line 16: This doesn't seem related to the rest of the PR?

spec/factories/applicants.rb
stephenrichards: it is actually - it resolves a flickering test which other changes here made worse - I couldn't work out why. But this ensures that the applicant is created with the correct postcode for the xml fixture file.

spec/factories/addresses.rb
stephenrichards: it is actually - it resolves a flickering test which other changes here made worse - I couldn't work out why. But this ensures that the applicant is created with the correct postcode for the xml fixture file.

config/routes.rb
stephenrichards: removed

app/lib/alert_manager.rb
stephenrichards: yes

mpw5 (approved):
LGTM

PR #2848, opened 2021-09-02
AP-2448 Add translation for failed address lookup

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2448)

If a postcode lookup fails due to a connection failure (eg if wifi is turned off), the user should see the manual address entry screen with an appropriate message however instead  an error is presented to the user.

This commit adds the correct wording to the `config/locales/en/forms.yml` locale. It also corrects the wording which is displayed when an unsuccessful response is received from the OS API, as it is currently a bit misleading (new wordings have been agreed with Jim.)

## Checklist

### DIFF
--- a/config/locales/en/forms.yml
+++ b/config/locales/en/forms.yml
@@ -4,7 +4,8 @@ en:
     address:
       errors:
         no_results: We could not find any addresses for that postcode. Enter the address manually.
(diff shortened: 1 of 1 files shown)


kmahern (approved):
👍

RoseSAK (approved):
LGTM!

PR #2861, opened 2021-09-09
AP-2502 Add start_date and end_date to proceeding_types

kmahern:
Seed new proceeding_type records

Only index proceeding_types with start_date < Date.current

and end_date > Date.current

Add end_date to seed table.

Add proceeding_type_scope_lims  to cover the new proceedings.

### Important:
For UAT testing the dates have all been set back by 1 year. The final commit on this branch needs to be reverted before this is released to production.

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2502)

## Checklist


app/services/proceeding_type_scope_limitations_populator.rb
```diff
@@ -11,7 +11,9 @@ def call
     CSV.read(file_path, headers: true, header_converters: :symbol).each do |row|
       # TODO: Delete when the allow_multiple_proceedings? flag is
       # removed. At that point, all proceedings will be seeded again
-      next if ProceedingType.find_by(ccms_code: row[:proceeding_type_code]).nil?
+      next if ProceedingType.find_by(code: row[:apply_code]).nil?
+
+      # ccms_code
```
mpw5, line 16: Can this be removed?

spec/factories/proceeding_types.rb
```diff
@@ -16,6 +16,8 @@
     default_cost_limitation_substantive { 2 }
     involvement_type_applicant { false }
     additional_search_terms { nil }
+    start_date { '2018-08-07' }
+    end_date { '2021-09-12' }
```
mpw5, line 20: I'm wondering if this is going to cause problems with the tests when we run them after 12/09?

mpw5 (approved):
Couple of comments, one minor. I don't think the factory should be a problem but it might be worth double-checking what the implication of using that date will be in future.

spec/factories/proceeding_types.rb
willc-work: If I set this to use e.g Time.zone.now +/- 2.years, it causes a test to fail as the new proceeding type that is created is different to the one in the DB, so when we expect it to not create a duplicate row it actually does.

willc-work: If I set this to use e.g Time.zone.now +/- 2.years, it causes a test to fail as the new proceeding type that is created is different to the one in the DB, so when we expect it to not create a duplicate row it actually does. I will do some more checks to see how it affects things

mpw5: But if I backdate the `end_date` to `2021-09-09` then `spec/requests/proceeding_types/index_proceeding_types_spec.rb` fails. So as soon. as we pass 12/09 we're going to start seeing failures. 

If we can't change it in the factory then maybe we need to override it in that spec, change the setup to something like 
 ```     
create(:proceeding_type,
               :with_real_data,
               code: 'PH0002',
               ccms_code: 'PBM24',
               ccms_category_law: 'Other',
               ccms_matter: 'Public Law - Other',
               ccms_matter_code: 'APBLB',
               meaning: 'Not an application for a care order',
               description: 'Not to be represented on an application for a care order.',
               additional_search_terms: 'injunction',
               end_date: Time.zone.now + 2.years)

willc-work: thanks

PR #2873, opened 2021-09-13
AP-2502 Remove default cost limitations to a separate model

stephenrichards:
## Remove default cost limitations to a separate model

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2502)

Remove default cost limitations from the `ProceedingType` model to  `DefaultCostLimitation` model which can then have a start date to handle changes in default costs

## Checklist


db/schema.rb
```diff
@@ -743,6 +751,7 @@
     t.boolean "allow_welsh_translation", default: false, null: false
     t.boolean "allow_multiple_proceedings", default: false, null: false
     t.boolean "enable_ccms_submission", default: true, null: false
+    t.boolean "alert_via_sentry", default: false, null: false
```
mpw5, line 754: This looks like it might have crept in from another PR.

mpw5 (approved):
LGTM! Just one comment about a schema change which doesn't look related.

db/schema.rb
stephenrichards: That was part of the sentry alternative PR I did before I went away.  Shouldnt' be part of this PR

PR #2875, opened 2021-09-14
Tech-Debt: Remove Faraday warnings

colinbruce:
## What

There has been an update to faraday recently and it has made rspec very noisy.  This follows the guidance to update the handling of authorisation and remove the warnings

## Checklist


willc-work (approved):
looks good. I am assuming the change to ag_cash_income_spec is to fix the flaky test on there, hence its inclusion.

PR #2876, opened 2021-09-14
AP-2507- ccms failures 

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2507)
The lead_proceeding was not being set correctly for the tests. Previously when geckoboard was being notified it must also have been instantiating the application_proceeding_type and assigning the lead_proceeding, however when that notification was removed the lead_proceeding was not being saved.

Added a reload to the factory so that the lead_proceeding is saved after it is created.

## Checklist

### DIFF
--- a/app/services/true_layer/api_client.rb
+++ b/app/services/true_layer/api_client.rb
@@ -57,7 +57,7 @@ module TrueLayer
 
     def connection
       @connection ||= Faraday.new(url: TRUE_LAYER_URL) do |conn|
-        conn.authorization :Bearer, token
+        conn.request :authorization, 'Bearer', token
         conn.response :logger if Rails.configuration.x.logs_faraday_response
         conn.adapter Faraday.default_adapter
       end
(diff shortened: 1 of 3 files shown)


mpw5 (approved):
Good find!

app/services/true_layer/api_client.rb
```diff
@@ -57,7 +57,7 @@ def simple_error(error)
 
     def connection
       @connection ||= Faraday.new(url: TRUE_LAYER_URL) do |conn|
-        conn.authorization :Bearer, token
+        conn.request :authorization, 'Bearer', token
```
mpw5, line 60: You don't need this change as this is being fixed by Colin's PR.

willc-work: decided with Colin to put all the commits into this one PR

mpw5: 👍

PR #2877, opened 2021-09-14
update circle config rename master to main

willc-work:
## What

We have renamed master to main in github, this makes the same changes to circle_ci config. It is a cosmetic change and does not affect functionality

## Checklist


colinbruce (approved):
LGTM thanks for picking this up, I _think_ I did this on the rest of the repos but it always bugged me that I never did this one! 🙂

PR #2888, opened 2021-09-16
AP-2506: Improve handling of email errors

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2506)

When we send an email to an unallowed account using Notify on non-live environments it returns a BadRequestError.

These will never succeed, but previously we stored the ScheduledMail and kept retrying.  This now traps the error and swallows it so sentry will not get excessive alerts, it also cancels the ScheduledMail so that subsequent attempts will not occur

## Checklist


mpw5 (approved):
Nice!

PR #2905, opened 2021-09-28
Mike income featuretest - add auto grant test for non passported

mikekeen1:
## What

Add feature test to check auto grant doesn't turn on for non passported cases

## Checklist


features/providers/check_nonpassport_autogrant.feature
```diff
@@ -0,0 +1,163 @@
...
+    And I click 'Save and continue'
+    Then I should be on a page showing "Check your answers and submit application"
+    Then I click 'Submit and continue'
+    Then I should be on a page showing "Application complete"
+    Then I click 'View completed application'
+    Then the application must be manually reviewed in CCMS
+
+@javascript @vcr
+  Scenario: I am able to create a non passported application without Income or Cap Contribs and with no restrictions
+    Given the setting to manually review all cases is enabled
```
mpw5, line 85: This might be fine, but I'm not clear why this test requires the setting to be enabled when the one above doesn't.

If the test here is to check that a non-passported application with no contribs/restrictions doesn't auto-grant, isn't the outcome of that being affected by manually changing the setting?

Or is it that a non-passported application with no contribs/restrictions _does_ auto-grant and we're testing that we can force it to auto-grant by changing. the setting? In which case I think the scenario might just need rewording to make that clear.

mpw5 (commented):
@mikekeen1 I have one question about these tests (it's probably just my understanding of what you're trying to acheive, rather than a problem). Happy to chat about it.

mikekeen1:
Hi Mark, ah yes you're right. The manually review setting shouldn't be
included, as the test is simply to confirm non passported cases don't auto
grant. I'll remove it, thanks

features/providers/check_nonpassport_autogrant.feature
```diff
@@ -82,7 +82,6 @@ Feature: Checking ccms means does NOT auto grant for non passported applications
 
 @javascript @vcr
   Scenario: I am able to create a non passported application without Income or Cap Contribs and with no restrictions
-    Given the setting to manually review all cases is enabled
     And The means questions have been answered by the applicant
```
mpw5, line 85: @mikekeen1 I'm going to be really picky and suggest that as you've removed the `Given` line, you should change `And The means questions have been answered by the applicant` to `Given The means questions have been answered by the applicant, so we maintain the Given-When-Then format.

mikekeen1: Ok, 3rd time lucky :-)

mpw5 (approved):
All looks good to me. Thanks for making the changes.

PR #2924, opened 2021-10-05
AP-2533 file validation

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2533)

Added validation to file upload pages (statement of case and gateway evidence)
Use JS to check whether a file has been selected but not uploaded, and if so display an error message

TODO:
- [ ] Add cucumber tests for file upload pages

## Checklist

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -338,6 +338,7 @@ en:
           size_hint: The maximum file size is 7MB. You can attach more than one file.
           warning: Warning
           warning_text: You must provide a complete statement now. You will not be able to upload a statement later.
+          file_not_uploaded_error: Upload the chosen file
         uploaded_files:
           delete: Delete
           filename: Filename
--- a/app/views/providers/application_merits_task/statement_of_cases/show.html.erb
+++ b/app/views/providers/application_merits_task/statement_of_cases/show.html.erb
@@ -22,6 +22,9 @@
 
     <%= form.govuk_error_summary %>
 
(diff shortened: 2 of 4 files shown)


config/locales/en/providers.yml
```diff
@@ -338,6 +338,7 @@ en:
           size_hint: The maximum file size is 7MB. You can attach more than one file.
           warning: Warning
           warning_text: You must provide a complete statement now. You will not be able to upload a statement later.
+          file_not_uploaded_error: Upload the chosen file
```
willc-work, line 341: Might want to check this with Jim, but we have used 'selected' instead of 'chosen' elsewhere in the app. e.g. 

              file_empty: The selected file is empty
              file_too_big: The selected file must be smaller than %{size}MB
              file_virus: The selected file contains a virus

RoseSAK: I did check with Jim earlier and he said to go with 'chosen' but I can point this out to him - selected does seem more similar to other errors 🤔

RoseSAK: Jim agrees - have updated!

willc-work (approved):
~One query on the wording in the translation file.~

It might be worth considering creating a ticket for refactoring our error summary partials. The code in :
app/views/shared/forms/_error_summary_hidden.html.erb
is very similar to that used in:
app/views/shared/_error.html.erb
app/views/providers/legal_aid_applications/_error.html.erb
app/views/shared/forms/_error_summary.html.erb

Other than that, looks good.

PR #2986, opened 2021-10-28
AP-2599: Call HMRC Interface

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2599)

Add a method of submitting data to the HMRC Interface when a provider confirms their client is employed
This PR changes:-
* **HMRC::CreateResponsesService** - This is triggered from the ApplicantEmployedController and ensures that the application has had the `set_transaction_period` triggered and then creates two;
* **HMRC::Responses** - expressly link to `LegalAidApplications` and add a `url` field for calling the next step, they are then passed to;
* **HMRC::SubmissionWorker** - This Sidekiq job sends the data to the HMRC interface and records the `submission_id` and `url` in the hmrc_response, it then creates a, delayed by 5 seconds;
* **HMRC::ResultWorker** - this calls the hmrc_response.url and retrieves the result from the interface.  If the job is still running it re-tries until it gets a response of `created` or `failed` and then records the resulting JSON in the hmrc_response.response field
* It creates a number of shared classes for use in the new workers to prevent repetition warnings and complexity
* It also adds a section to the admin, application submission page so we can debug the responses from the interface 

## Checklist


spec/services/hmrc/interface/submission_service_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+
+        it 'raises an exception' do
+          expect {
+            described_class.call(hmrc_response)
+          }.to raise_error HMRC::SubmissionError, /Bad Request: URL:/
+        end
+      end
+
+      context 'connection error' do
+        it 'creates a LegalFramework::Submission error and writes a history record with a backtrace' do
```
mpw5, line 65: I think this is another C&P error 😁

spec/services/hmrc/interface/result_service_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+
+        it 'raises an exception' do
+          expect {
+            described_class.call(hmrc_response)
+          }.to raise_error HMRC::SubmissionError, /Bad Request: URL:/
+        end
+      end
+
+      context 'connection error' do
+        it 'creates a LegalFramework::Submission error and writes a history record with a backtrace' do
```
mpw5, line 145: I think this is a C&P error 😁

spec/services/hmrc/interface/result_service_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    context 'when the submission has failed' do
+      let(:expected_status) { 202 } # TODO: Response should be 200, also json response should equal - AP-2639 addresses
+      # {
+      #   data: [
+      #     { correlation_id: hmrc_response.submission_id },
+      #     { error: 'submitted client details could not be found in HMRC service' }
+      #   ]
+      # }
```
mpw5, line 93: Can this be removed?

mpw5 (commented):
Couple of v minor comments.

spec/services/hmrc/interface/result_service_spec.rb
colinbruce: It was kind of the continuation of the TODO ...also json response should equal
For when the updates get made to the interface

spec/services/hmrc/interface/result_service_spec.rb
colinbruce: How do you feel about this being named `SubmissionError`? As it became used in submission _and_ result, would `HMRC::InterfaceError` be better?

colinbruce: I deleted these and refactored the name anyway!

spec/services/hmrc/interface/result_service_spec.rb
mpw5: Sorry, didn't spot the comment on the previous line 🤦 That makes sense to me.

stephenrichards (approved):

PR #3013, opened 2021-11-02
AP-2642 Fix Scope Limitations bug in XML payload

stephenrichards:
## Fix Scope Limitations bug in XML payload

In certain cases, `ApplicationProceedingType` records had `assigned_scope_limitations` for both substantive and delegated functions, yet delegated functions aren't being used. As a result, both substantive and DF scope limitation details were being included in the XML payload sent to CCMS.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2642)

- Modified the Xml generation class (CaseAddRequestor) to not send the DF scope limitations if DF weren't used.
- Specs to test this

## Checklist


mpw5 (approved):
Looks good to me

PR #3021, opened 2021-11-03
AP-2535 fix table sort arrows

RoseSAK:
## What

add break which is hidden at certain breakpoints so the arrows wrap to a new line if there is not enough space

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2535)

## Checklist


mpw5 (approved):
LGTM

PR #3026, opened 2021-11-04
AP-2607 Extract applications to digest

stephenrichards:
## Extract applications to digest

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2607)

- Created `ApplicationDigest` model and table, with method to create or update digest from a `LegalAidApplication`
- Created `DigestExtractor` service to call ApplicationDigest for each `LegalAidApplcation` that has been updated since last run
- Created rake task `digest:extract` to run `DigestExtractor`
- Created cron job to call the `digest:extract` rake task

## Checklist


Gemfile
```diff
@@ -140,7 +140,7 @@ end
 group :test do
   gem 'axe-core-cucumber'
   gem 'capybara', '>= 3.32.2', '< 4.0'
-  gem 'cucumber', require: false
+  gem 'cucumber', '5.3.0', require: false
```
colinbruce, line 143: Is there a reason this PR needs bundler updates?

db/migrate/20211103112548_set_new_proceeding_fields_to_not_null.rb
```diff
@@ -0,0 +1,7 @@
+class SetNewProceedingFieldsToNotNull < ActiveRecord::Migration[6.1]
+  def change
+    change_column_null(:proceedings, :matter_type, false)
```
colinbruce, line 3: I assume this has to be separate to allow the data migration to run between?

colinbruce (commented):
A couple of observations/questions but looking good :+1:

Gemfile
stephenrichards: For some reason, this PR was bringing in cucumber 7.2.0, which doesn't work.  This change is to pin it at our existing version to prevent that from happening.

db/migrate/20211103112548_set_new_proceeding_fields_to_not_null.rb
stephenrichards: That is correct.

colinbruce (approved):
Still confused about why it needed bundle update at all, but otherwise, looks okay 👍

PR #2381, opened 2021-03-12
AP-2046-G Use ApplicationProceedingTypesScopeLimtiations model

stephenrichards:
## Use ApplicationProceedingTypesScopeLimtiations model

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2046)

This PR has been approved once, but then split into two parts: the actual migration of records from `application_scope_limitations` to `application_proceeding_types_scope_limitations`, and then using the new records as opposed to the old.
The migration has been done, and this PR is to use the new `application_proceeding_types_scope_limitations` records.

## Checklist

Before you ask people to review this PR:

- [ ] Tests and rubocop should be passing: `bundle exec rake`
- [ ] Github should not be reporting conflicts; you should have recently run `git rebase master`.
- [ ] There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- [ ] The PR description should say what you changed and why, with a link to the JIRA story.
- [ ] You should have looked at the diff against master and ensured that nothing unexpected is included in your changes.
- [ ] You should have checked that the commit messages say why the change was made.

PR #3027, opened 2021-11-04
FIX: Update tests

colinbruce:
## What

Changing the view to use lead_application_proceeding_type rather than
application_proceeding_type.first stops the issues with
scope limitations being assigned

## Checklist

### DIFF
--- a/app/views/shared/check_answers/_substantive_limitations.html.erb
+++ b/app/views/shared/check_answers/_substantive_limitations.html.erb
@@ -12,6 +12,6 @@
   <%= check_answer_no_link(
         name: :allowable_work,
         question: t('.work'),
-        answer: @legal_aid_application.application_proceeding_types.first.substantive_scope_limitation.description
+        answer: @legal_aid_application.lead_application_proceeding_type.substantive_scope_limitation.description
       ) %>
 </dl>


db/schema.rb
```diff
@@ -454,11 +454,14 @@
   end
 
   create_table "hmrc_responses", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
-    t.string "application_id"
```
mpw5, line 457: Are these schema changes meant to be here?

app/views/shared/check_answers/_substantive_limitations.html.erb
```diff
@@ -12,6 +12,6 @@
   <%= check_answer_no_link(
         name: :allowable_work,
         question: t('.work'),
-        answer: @legal_aid_application.application_proceeding_types.first.substantive_scope_limitation.description
+        answer: @legal_aid_application.lead_application_proceeding_type.substantive_scope_limitation.description
```
mpw5, line 15: I think this is probably fine - it seems slightly arbitrary to just use the first proceeding anyway, using the lead proceeding certainly doesn't seem to be a worse approach.

mpw5 (commented):
LGTM - apart from the rogue schema changes

app/views/shared/check_answers/_substantive_limitations.html.erb
stephenrichards: What SHOULD we be displaying in this partial - sounds like the scope limitations for the lead proceeding is right, as opposed to any old proceeding.  The `.first` probably dates back from when we only had one application proceeding type per application.

stephenrichards (approved):

app/views/shared/check_answers/_substantive_limitations.html.erb
colinbruce: It was added as part of PR #2381 so it possibly pre-dates the implementation of _lead_ _proceeding_types?

PR #3051, opened 2021-11-11
AP-2622 Remove unneeded SubmittedApplication code

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2622)

This is slightly different to what the ticket asked for. The lines of code that refer to `application_proceedings_types` aren't acually needed - they are passed to the `app/views/shared/check_answers/_merits.html.erb` partial however aren't ever used. I suspect they are a hangover from pre-multiple proceedings days.

This change simply deletes the unused code. The submitted application page appears as expected without it. It might be useful if someone could double-check my logic though 😁 

## Checklist


colinbruce (approved):
Seems fine code wise and your logic tracks... I'm sure the regression tests will be fine!
:+1:

PR #3058, opened 2021-11-12
add test for method only used in CCMS tests

willc-work:
## What

Bugfix - a method on delegated functions was only covered by tests in CCMS files. This meant that running rspec tests without CCMS tests would result in 99.99% coverage issue. This adds a test to cover the method in question and also a note to merge it with the method above when application_proceeding_type are removed.

## Checklist


colinbruce (approved):
👍

PR #3061, opened 2021-11-12
AP-2626 Create new proceeding removal service

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2626)

* Adds new `RemoveProceedingService` which replicates the functionality of the `RemoveProceedingTypeService` for `Proceedings` instead of `ProceedingTypes`.

* Renames `LeadProceedingAssignmentService` to `LeadApplicationProceedingTypeAssignmentService` to make it clear from the name which type of proceedings it involves.

* Creates `LeadProceedingAssignmentService` which replicates the functionality of the `LeadApplicationProceedingTypeAssignmentService` for `Proceedings` instead of `ProceedingTypes`.

* Amends `HasOtherProceedingsController` to call the new `RemoveProceedingService` and `LeadProceedingAssignmentService` classes in addition to the existing ones, to keep `ApplicationProceedingTypes` and `Proceedings` in sync.

* Amends `has_other_proceedings/show.html.erb` to use `Proceedings` instead of `ApplicationProceedingTypes`. To do this is necessary to pass the `ccms_code` of the proceeding to the `HasOtherProceedingsController` instead of the `code`, as this only exists on the `ApplicationProceedingType` model, not the `Proceeding` model.

* Removes the `after_delete` callback from the `ApplicationProceedingType` model and updates the `ProceedingSyncService` and various test config to account for this.

* Moves the `LeadApplicationProceedingTypeAssignmentService` and `LeadProceedingAssignmentService` into the `LegalFramework` namespace as they seem to make more sense there.

## Checklist


colinbruce (approved):
🤯 but LGTM :+1:

PR #3100, opened 2021-11-22
Fix live deployments for long branch names

mpw5:
## What

On branches with long names, deployments to the live cluster fail because the annotation we set in `uat_deploy_to_live` does not match the truncated ingress name expected by cloud platform.

This commit updates the `uat_deploy_to_live` so that it also truncates long names to 52 characters, which allows deployments to succeed.

## Checklist


kmahern (approved):
lgtm 👍

PR #3122, opened 2021-11-25
ap-2697 Add employment permissions

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2697)

Add a new permission and a method to check for employment permissions.

Currently employment is a child of non_passported permission. There is no specific reason why it is not `application.employment.*` and it is an easy change if this is preferred. My thinking is that employed would only be of use to users with non_passported permssion. However all users have non_passported permissions so it is not a strong argument for how I have implemented it.

I have tested using `provider.employment_permissions?` to change the flow and it works as expected.

  The ticket is light on details about the `done when` for this spike. We can either merge this PR in readiness for when it is needed, or wait until it is needed and it can be done as part of another ticket now we know that the existing permissions system we have can be used for employment.

## Checklist

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -45,6 +45,10 @@ class Provider < ApplicationRecord
     user_permissions.map(&:role).include?('application.non_passported.*')
   end
 
+  def employment_permissions?
+    user_permissions.map(&:role).include?('application.non_passported.employment.*')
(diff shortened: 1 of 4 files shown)


mpw5 (approved):
Nice!

PR #3126, opened 2021-11-26
bugfix: display correct response on means report

willc-work:
## What

The means report is showing £0 for the balance on accounts when they do not exist. This was noticed by a caseworker on a savings account, change made to both savings and current accounts to ensure we report these correctly on the means report. The safe operator (&) in the query returns a 0 as the savings accounts is an empty array

## Checklist


RoseSAK (approved):
LGTM!

PR #3127, opened 2021-11-26
AP-2565 Move HMRC call

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2565)

Moves the call to HMRC from the Is your client employed? page to the DWP results page.

The call only takes place if the provider confirms that the non-passported DWP benefit check is correct _and_ the employed journey feature flag is enabled _and_ the provider has the correct permissions _and_ the HMRC call setting is enabled.

Also adds hint text to the DWP results page, with a helper method to identify when to display this.

Note: requires confirmation of the wording of the hint text from Jim, but the rest of the code can be reviewed now.

## Checklist

### DIFF
--- a/app/controllers/providers/applicant_employed_controller.rb
+++ b/app/controllers/providers/applicant_employed_controller.rb
@@ -7,7 +7,6 @@ module Providers
 
     def create
       @form = Applicants::EmployedForm.new(form_params)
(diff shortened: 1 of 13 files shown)


colinbruce (approved):
LGTM :+1:

PR #3135, opened 2021-11-30
AP-2466 Create proceedings records from LFA data

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2466)

* Adds services to call LFA to retrieve data for a given proceeding.
* Updates `ProceedingsTypesController` to call LFA and create `Proceeding` records from returned data, leaving the existing approach in place for now so that we can continue to dual-run `Proceedings` and `ApplicationProceedingTypes`.
* Removes the `after_create` callback from the `ApplicationProceedingType` model.

## Checklist


app/controllers/providers/proceedings_types_controller.rb
```diff
@@ -25,7 +25,10 @@ def create
     private
 
     def run_transaction
-      # TODO: Should be fixed by future LFA ticket to change to Proceeding record
+      proceeding_to_add = proceeding_types.find { |proceeding| proceeding.ccms_code == form_params }
+      add_proceeding_service.call(ccms_code: proceeding_to_add.ccms_code)
```
stephenrichards, line 29: I thought the idea of this ticket was to create the Proceeding record directlly, and not rely on the ProceedingTypes

app/models/proceeding.rb
```diff
@@ -78,4 +82,9 @@ def application_proceeding_type
     proceeding_type = ProceedingType.find_by ccms_code: ccms_code
     legal_aid_application.application_proceeding_types.find_by(proceeding_type_id: proceeding_type.id)
   end
+
+  def highest_proceeding_case_id
+    rec = self.class.order(proceeding_case_id: :desc).first
+    rec.nil? || rec.proceeding_case_id.nil? ? FIRST_PROCEEDING_CASE_ID : rec.proceeding_case_id
```
stephenrichards, line 88: I think we need to add an index on proceeding_case_id for fast lookups - might as well make it a unique index to ensure we don't get dupes

stephenrichards (changes requested):

app/models/proceeding.rb
mpw5: Good point. I've pushed a commit to add an index.

stephenrichards (approved):

PR #3137, opened 2021-11-30
AP-2719 BUGFIX: Substantive scope limitations appearing on Emergency Applications

mpw5:
## What

Substantive scope limitations are incorrectly appearing on emergency applications in CCMS.

This is because the `generate_scope_limitations` method in the `CaseAddRequestor` class sets the `DelegatedFunctionsApply` attribute as `proceeding.used_delegated_functions?` for both the substantive and delegated functions scope limitations. This means that for a DF application, both scope limitations have a `DelegatedFunctionsApply` attribue of `true`.

This is incorrect - only the delegated functions scope limitation should have a `DelegatedFunctionsApply` attribue of `true`.

This PR changes that method so that the substantive scope limitation always has a `DelegatedFunctionsApply` attribute of `false`, and the delegated functions scope limitations always has a `DelegatedFunctionsApply` attribute of true.

## Checklist


RoseSAK (approved):
I'm pretty sure this looks right 👍

PR #3158, opened 2021-12-07
AP-2568 Parse HMRC response

stephenrichards:
## Parse HMRC response

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2568)

### Notes:
- after discussion with Pamela, it was agreed the Employment income summary page would show payment date rather than range as shown in the ticket.  The parsed response (`HMRC::EmploymentIncomeSummary`) reflects this.
- The parser currently only works when the applicant has one job - this is because we don't know how the data will be presented for applicants with more than one job.

This PR parses the HMRC and presents it in a way which is easy to display on the review income employment page.

- Created `HMRC::EmploymentIncomeSummary` to encapsulate all income from all employments
- Created `HMRC::Employment` to encapsulate each employment
- Created `HMRC::EmploymentPayment` to encapsulate each payment
- Added scope `use_case_one_for` on `HMRC::Response` to return the most recent use case `one` response for the specified `LegalAidApplication`
- Created `FactoryHelpers::HMRCResponse::UseCaseOne` to enable creation of HMRC responses in the factory

## Checklist

### DIFF
--- a/app/models/hmrc/employment.rb
+++ b/app/models/hmrc/employment.rb
@@ -1,0 +1,16 @@
+module HMRC
+  class Employment
+    attr_reader :payments
+
+    def initialize(hmrc_income_array)
+      @payments = hmrc_income_array.map { |payment_hash| EmploymentPayment.new(payment_hash) }
+      sort_payments_most_recent_first
+    end
+
+    private
+
+    def sort_payments_most_recent_first
+      @payments.sort! { |a, b| b.payment_date <=> a.payment_date }
+    end
+  end
+end
--- a/app/models/hmrc/employment_payment.rb
+++ b/app/models/hmrc/employment_payment.rb
@@ -1,0 +1,25 @@
+module HMRC
+  class EmploymentPayment
+    def initialize(payment_hash)
+      @payment_hash = payment_hash
+    end
+
+    def payment_date
+      Date.parse @payment_hash.fetch('paymentDate')
+    end
+
+    def gross_pay
+      @payment_hash.fetch('grossEarningsForNics').fetch('inPayPeriod1')
+    end
+
+    def tax
+      @payment_hash['taxDeductedOrRefunded'] || 0.0
+    end
+
+    def national_insurance
+      return 0.0 unless @payment_hash.key?('employeeNics')
+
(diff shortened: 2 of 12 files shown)


app/models/hmrc/response.rb
```diff
@@ -5,5 +5,9 @@ class Response < ApplicationRecord
     USE_CASES = %w[one two].freeze
     belongs_to :legal_aid_application, inverse_of: :hmrc_responses
     validates :use_case, presence: true, inclusion: { in: USE_CASES, message: 'Invalid use case' }
+
+    def self.use_case_one_for(laa_id)
```
colinbruce, line 9: Should there be a `use_case_two_for` as well?

app/models/hmrc/employment.rb
```diff
@@ -0,0 +1,16 @@
+module HMRC
+  class Employment
```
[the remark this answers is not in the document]
stephenrichards: Deleted

app/models/hmrc/employment_payment.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def tax
+      @payment_hash['taxDeductedOrRefunded'] || 0.0
+    end
+
+    def national_insurance
+      return 0.0 unless @payment_hash.key?('employeeNics')
+
+      @payment_hash.fetch('employeeNics').fetch('inPayPeriod1')
```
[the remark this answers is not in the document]
stephenrichards: I don't think so.  I want to return zero if there is no `employeeNics`, key, but if there is an `employeeNics` key but no `inPayPeriod1` key inside it, then I want to raise an error.

I think that's the right logic

app/models/hmrc/response.rb
stephenrichards: I think I get everything I need from use_case_one (date, gross pay, tax, national insurance), so I don't think we need it

colinbruce (approved):
:+1:

app/models/hmrc/response.rb
colinbruce: I had a vague recollection of DFab saying one filed was needed from UC2, but fair enough, we can always amend if needed :+1:

PR #3164, opened 2021-12-08
Fix flaky chances of success tests

mpw5:
## What

Tests in `spec/models/proceeding_merits_task/chances_of_success_spec.rb` are occasionally failing due to the way `application_proceeding_types` and `proceedings` are set up.

By creating the `proceeding` after the creation of the `application` and `application_proceeding_types`, and by creating different numbers of `application_proceeding_types` and `proceedings`, it is possible for the `proceeding_case_id` to be non-unique.

This changes the set up so that identical `application_proceeding_types` and `proceedings` are created simultaneously, replicating what happens in the real service. This should prevent the unique constraint being violated.

## Checklist


colinbruce (approved):
🏆

PR #3172, opened 2021-12-09
ap-2730: Create document_categories table

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2730)

Created migration to add new document_categories table, model and factory
Created document_category_populator to populate new table

## Checklist

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -103,9 +103,9 @@ ActiveRecord::Schema.define(version: 2021_12_09_083010) do
     t.string "unlock_token"
     t.datetime "locked_at"
     t.string "true_layer_secure_data_id"
-    t.boolean "employed"
     t.datetime "remember_created_at"
     t.string "remember_token"
+    t.boolean "employed"
     t.index ["confirmation_token"], name: "index_applicants_on_confirmation_token", unique: true
     t.index ["email"], name: "index_applicants_on_email"
     t.index ["unlock_token"], name: "index_applicants_on_unlock_token", unique: true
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -103,9 +103,9 @@
     t.string "unlock_token"
     t.datetime "locked_at"
     t.string "true_layer_secure_data_id"
+    t.boolean "employed"
```
mpw5, line 106: This change seems to creep in to migrations sometimes. I'm not entirely sure why but it shouldn't really be there. You can undo it by running:

```
rails db:rollback
git checkout origin/main db/schema.rb
rails db:schema:load
rails db:migrate
```
This should reset the schema and then reapply your migration but without the extra changes on lines 106 and 605.

Warning: `rails db:schema:load` will wipe your local db 😬

db/schema.rb
```diff
@@ -592,7 +602,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "scanner_working"
-    t.index ["uploader_type", "uploader_id"], name: "index_malware_scan_results_on_uploader"
+    t.index ["uploader_type", "uploader_id"], name: "index_malware_scan_results_on_uploader_type_and_uploader_id"
```
mpw5, line 605: As above.

mpw5 (commented):
Looks good but some unrelated schema changes have crept in. Happy to approve once they've been fixed. I've added some comments to help fix it but give me a shout if you want a hand.

mpw5 (approved):
LGTM!

db/schema.rb
kmahern: Thanks @mpw5, I also don't understand how the unrelated schema changes crept in (I definitely pulled the version from main before I started!) but your suggestion removed them anyway.

PR #3178, opened 2021-12-10
fix warning text icons

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2759)

In a couple of places the warning text icon has an 'i' instead of an exclamation mark

## Checklist


willc-work (approved):
Sergio who wrote this was Spanish:
From google:
in Spanish, the upside-down exclamation point (¡) is used at the beginning of all exclamatory sentences and clauses. For example: ¡Hola! (Hello!)

So I guess that answers why. 

Good spot.

PR #3185, opened 2021-12-13
AP-2362:improve report generation

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2362)

Update the report generation, previously we would see an occasional error when reports did not have a document attached (see the logging introduced by #2755).

This adds some logic to the means and merit report generators to extend the checks for existing records.
* If an attachment exists and its linked document is present, it will continue to log the state and exit the generator
* If an attachment exists and its linked document cannot be found, it destroys the existing attachment and recreates from scratch

It extends the logging (that can be checked on kibana) so we can monitor for extra occurences

## Checklist

### DIFF
--- a/app/services/reports/base_report_creator.rb
+++ b/app/services/reports/base_report_creator.rb
@@ -36,5 +36,27 @@ module Reports
         }
       }
     end
+
+    def delete_attachment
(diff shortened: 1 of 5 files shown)


willc-work (approved):
LGTM

PR #3195, opened 2021-12-15
AP-2766 update script to delete file

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2766)

After the rake task ends the script checks to see if it completed without any errors and if so it deletes the restore file.

## Checklist

### DIFF
--- a/scripts/restore_anonymised_db.sh
+++ b/scripts/restore_anonymised_db.sh
@@ -13,3 +13,11 @@ kubectl -n laa-apply-for-legalaid-uat cp ./tmp/production.anon.sql -c web laa-ap
 
 echo "Connect to the pod and run the rake task"
 kubectl -n laa-apply-for-legalaid-uat -c web exec "$POD" -- rake db:import_to_uat
+
+if [ $? -eq 0 ]; then
+   echo 'Restore complete'
+   echo "Delete the restore file from the pod"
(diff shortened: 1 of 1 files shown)


scripts/restore_anonymised_db.sh
```diff
@@ -13,3 +13,11 @@ kubectl -n laa-apply-for-legalaid-uat cp ./tmp/production.anon.sql -c web laa-ap
 
 echo "Connect to the pod and run the rake task"
 kubectl -n laa-apply-for-legalaid-uat -c web exec "$POD" -- rake db:import_to_uat
+
+if [ $? -eq 0 ]; then
+   echo 'Restore complete'
+   echo "Delete the restore file from the pod"
+   kubectl -n laa-apply-for-legalaid-uat -c web exec "$POD" -- sh -c 'rm tmp/anonymised_db.sql'
```
colinbruce, line 20: Do you need the second `-c` in front of the command?

willc-work: Yes it is required. It isn't the container flag that is used in kubectl, it is a flag passed to the shell to indicate that what follows is a command. At least that is my understanding but I cannot find any documentation to support it.

colinbruce (approved):
LGTM :+1:

PR #3197, opened 2021-12-15
AP-2722 Remove ApplicationProceedingType from UsedDelegatedFunctionsController

stephenrichards:
## Remove ApplicationProceedingType from UsedDelegatedFunctionsController

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2722)

The `UsedDelegatedFunctionsController` was set up to update both `Proceeding` records and `ApplicationProceedingType` records with DF date.  This PR removes the updates to the `ApplicationProceedingType`

## Checklist


kmahern:
@stephenrichards I had a question, https://dsdmoj.atlassian.net/browse/AP-2722 talks about removing the dependencies on proceeding_type in the remove_delegated_scope_limitations, add_delegated_scope_limitations and update_delegated_scope_limitations methods in UsedDelegatedFunctionsController. I know when I looked at this controller previously I thought we couldn't yet do that (because proceeding_type is required by AddAssignedScopeLimitationService which is called in multiple places) - do you think that is still the case?

stephenrichards:
Hi @kmahern - I'll look at this again, but I think that now the scope limitations data is held on the `Proceeding` record, and all the Check your answer pages, merits report and CCMS payload all use the `Proceeding` model for their data source, we don't really care any more what is on the application proceeding types scope limtiations records.  Having said that, I thought I'd removed a call to one of the services that takes care of all that, but I don't see that in the changes on this PR, so am going to take another close look.

Thanks for the heads-up

spec/forms/legal_aid_applications/used_multiple_delegated_functions_form_spec.rb
```diff
@@ -50,8 +44,6 @@
         end
 
         it 'updates the application types with no reported on date' do
```
colinbruce, line 46: Should this description be changed to `updates the proccedings with no reported on date`?

colinbruce (approved):
One minor comment about a test description but, otherwise, LGTM 👍

PR #3205, opened 2021-12-16
AP-2778 digest exporter cron

stephenrichards:
## Fix issues with digest exporter cron

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2778)

The digest exporter cron job was failing every night to to running out of memory, and taking 45 minutes to run before failing.  Several steps have been taken to solve this issue:
- increase memory on kubernetes cronjob definition
- Read the ids of the digest records and process, rather than read all 9,000 into memory at the same time
- format the output into an array and write to the spreadsheet once, rather than make an API call to the spreadsheet for each line (this has reduced the running time from 45 minutes to < 3 minutes
- Add logging to that progress can be monitored
- Add Extracted at time to the header line of the worksheet

## Checklist

### DIFF
--- a/app/services/digest_exporter.rb
+++ b/app/services/digest_exporter.rb
@@ -4,38 +4,73 @@ class DigestExporter
   end
 
   def initialize
+    log_message 'DigestExporter starting'
     secret_file = StringIO.new(google_secret.to_json)
     @session = GoogleDrive::Session.from_service_account_key(secret_file)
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM 👍 
It would be good to try and monitor this and see if the extra memory is needed on the pod if we are finishing in ~3 minutes instead of ~45! :)

PR #3244, opened 2021-12-29
Fix flickering code coverage bug on Circle CI

stephenrichards:
## Fix flickering code coverage bug on Circle CI

Coverage was below 100% on circle ci with certain seeds (This could be reliably
reproduced on Circle CI with seed 51620, but not locally).  The cause was found to be
that the two files app/models/cfe/v1/result.rb and app/models/cfe/v2/result.rb, both of which have
a :nocov: comment fore and aft were being reported as not tested.

This is fixed by adding those to files to the SimpleCov filter in spec_helper.rb

Also, the code to track this down has been added to the repo at lib/circle_ci_covereage_analyzer.rb

## Checklist

### DIFF
--- a/lib/circle_ci_coverage_analyzer.rb
+++ b/lib/circle_ci_coverage_analyzer.rb
@@ -1,6 +1,5 @@
 require 'json'
 
-
 # This class is useful for debugging why coverage is not 100% on Circle-CI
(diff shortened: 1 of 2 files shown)


stephenrichards:
Superceded by a PR which re-instates the tests

PR #3246, opened 2021-12-30
Re-instate testing for CFE V1 and V2 results

stephenrichards:
## Re-instate testing for CFE V1 and V2 results

Occasionally, unit tests on Circle-CI would fail due to test coverage being less than 100%.

This could be reliably reproduced on Circle CI by using the seed 51620.

It turns out that with certain seeds, the `:nocov:` at the beginning and end of `app/models/cfe/v1/results.rb` and 
`app/models/cfe/v1/results.rb` were not being honoured, and they were being counted as not tested.

This PR re-instates the tests (and factories, and mock result objects) for those two files.

## Checklist

### DIFF
--- a/app/models/cfe/base_result.rb
+++ b/app/models/cfe/base_result.rb
@@ -68,6 +68,10 @@ module CFE
     #                                                              #
     ################################################################
 
+    def additional_properties
+      property[:additional_properties]
+    end
+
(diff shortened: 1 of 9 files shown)


colinbruce (approved):
LGTM :+1:

PR #3262, opened 2022-01-05
Fix: CCMS tests were failing

colinbruce:
## What

After AP-2721 (#3227) removed the application_proceeding_type from chances_of_success
these all started failing because the factory could not add a non-existent field

## Checklist

### DIFF
--- a/spec/services/ccms/requestors/case_add_requestor_spec.rb
+++ b/spec/services/ccms/requestors/case_add_requestor_spec.rb
@@ -30,9 +30,8 @@ module CCMS
                  address: address
         end
         let(:proceeding) { legal_aid_application.proceedings.detect { |p| p.ccms_code == 'DA001' } }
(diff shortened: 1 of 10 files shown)


spec/services/ccms/requestors/case_add_requestor_spec.rb
```diff
@@ -30,9 +30,9 @@ module Requestors
                  address: address
         end
         let(:proceeding) { legal_aid_application.proceedings.detect { |p| p.ccms_code == 'DA001' } }
-        let(:application_proceeding_type) { create :application_proceeding_type, legal_aid_application: legal_aid_application }
+        # let(:application_proceeding_type) { create :application_proceeding_type, legal_aid_application: legal_aid_application }
```
kmahern, line 33: Can the commented out code be removed?

kmahern (approved):
One minor comment, otherwise lgtm.

PR #3270, opened 2022-01-05
AP-2725 Do not create ApplicationProceedingTypes

stephenrichards:
## Do not create ApplicationProceedingTypes

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2725)

This PR removes the creation of the ApplicationProceedingType records from the ProceedingTypesController and the HasOtherProceedingsController.  Also removes various service classes to do with creation/deletion of ApplicationProceedingType records, and marking them as lead proceeding.

## Checklist


spec/requests/providers/proceedings_types_spec.rb
```diff
@@ -140,8 +133,6 @@
         let(:add_proceeding_service) { double(LegalFramework::AddProceedingService, call: false) }
 
         before do
-          allow(LegalFramework::ProceedingTypesService).to receive(:new).with(legal_aid_application).and_return(proceeding_type_service)
```
colinbruce, line 143: Can the LegalFramework::ProceedingTypesService be removed too?
Is it in a separate Ticket or is it still needed?

colinbruce (approved):
LGTM 👍 
One question/observation but no reason not to merge

spec/requests/providers/proceedings_types_spec.rb
stephenrichards: It can be yes, but we can get rid of that in the general tidy up, which I think is all there is left to do on this apart from dropping the tables

PR #3283, opened 2022-01-07
hardcode the deploy name for staging Apply

willc-work:
## What
deploy to staging and prod uses an env_var in circle to name the deployment, this hardcodes the value as 
laa-apply-for-legal-aid
This also resolves an issue with the EKS migration where we have a mismatch between the actual and expected ingress names

A future PR will be required to make this change on production

## Checklist


colinbruce (approved):
:trophy:

PR #3285, opened 2022-01-07
AP-2737 Refactor ObtainDocumentIdService to use DocumentCategory

stephenrichards:
## Refactor ObtainDocumentIdService to use DocumentCategory

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2737)

ObtainDocumentIdService used a hard-coded list of attachment types to exclude from uploading to CCMS.  This PR refactors that code to use the `submit_to_ccms` field on the DocumentCategory record to decide whether or not to upload it to CCMS.  This is because this list will change over time, and the control of what to upload and what to omit is all in one place.

- Refactored `valid_document_categories` scope on `DocumentCategory` to a class method `.displayable_document_category_names` - it returns an array of names, not an array of records, and so isn't a scope
- Added class method `.submittable_category_names` to DocumentCategory
- Refactored `ObtainDocumentIdService` to use `DocumentCategory.submittable_category_names` rather than a hard-coded list

## Checklist

### DIFF
--- a/app/services/ccms/submitters/obtain_document_id_service.rb
+++ b/app/services/ccms/submitters/obtain_document_id_service.rb
@@ -1,7 +1,7 @@
 module CCMS
   module Submitters
     class ObtainDocumentIdService < BaseSubmissionService
-      NON_PDF_VERSION_ATTACHMENTS = %w[statement_of_case gateway_evidence].freeze
+      # NON_PDF_VERSION_ATTACHMENTS = %w[statement_of_case gateway_evidence].freeze
 
       def call
         return unless populate_documents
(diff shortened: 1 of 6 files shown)


app/services/ccms/submitters/obtain_document_id_service.rb
```diff
@@ -1,7 +1,7 @@
 module CCMS
   module Submitters
     class ObtainDocumentIdService < BaseSubmissionService
-      NON_PDF_VERSION_ATTACHMENTS = %w[statement_of_case gateway_evidence].freeze
+      # NON_PDF_VERSION_ATTACHMENTS = %w[statement_of_case gateway_evidence].freeze
```
mpw5, line 4: Can this be deleted now?

mpw5 (approved):
There's one commented-out line which I think can be deleted now, but apart from that this looks fine to me.

app/services/ccms/submitters/obtain_document_id_service.rb
stephenrichards: Done

PR #3309, opened 2022-01-11
AP-2802 Standardise Matter Type names on Proceeding rec

stephenrichards:
## Standardise Matter Type names

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2802)

Proceeding records are now created from the response from LFA, and the name of the matter type differs slightly from those on the Proceeding Type records, so we had some Proceeding records with `Domestic abuse` or `Section 8 proceedings`,  and others with `Domestic Abuse` and `Children - section 8`.  This was causing problems in the application digest export to the dashboard.

This PR sets all matter type names on the Proceeding records to the LFA standard, and sets the digest extracted setting back to the beginning of Apply-time so that all digest records are re-generated.

## Checklist

### DIFF
--- a/db/data/20220111135820_standardise_matter_types.rb
+++ b/db/data/20220111135820_standardise_matter_types.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+class StandardiseMatterTypes < ActiveRecord::Migration[6.1]
+  def up
+    execute "UPDATE proceedings set matter_type = 'Section 8 orders' where ccms_matter_code = 'KSEC8'"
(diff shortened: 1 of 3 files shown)


colinbruce:
Have you checked the new values with Dave? 
I have a vague memory that the names have specific legal values that we shouldn’t change/interpret and I thought it should be Domestic Abuse with both capitalised but I could be wrong

colinbruce (changes requested):
See comment

willc-work (commented):
As Colin mentioned , might be good to confirm with Dave for these values

PR #3349, opened 2022-01-18
AP-2630 Remove temporary method from Proceeding

stephenrichards:
## Remove temporary method from Proceeding

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2630)

During the migration from holding all Legal Framework data in Apply to using the LFA system to supply it, we were writing both old style `ApplicationProceedingType` and `ApplicationProceedingTypeScopeLimitation` records as well as the newer `Proceeding` records.  There was a temporary method on `Proceeding` to create a `Proceeding`  from an existing `ApplicationProceedingType` record.  This is no longer needed and removed.

This also necessitated some clean up of tests and other areas of the code, notably:
- Removal of the `ProceedingSyncService` and the `after_update` callback on `ApplicationProceedingType` that called it
- Change the dashboard widget to use `Proceeding` rather than `ApplicationProceedingType`
- Removal of the `#application_proceeding_types_by_name` method from `LegalAidApplication` (`#proceeding_by_name` is now used instead)
- Change all the setup steps for the feature tests which used `:with_proceeding_types` trait to use `:with_proceedings` instead
- Remove all the switching-on and -off of the `ApplicationProceedingType` callbacks in the feature tests

## Checklist


willc-work (approved):
LGTM

PR #3354, opened 2022-01-19
Implement data shortcut

colinbruce:
## What

This allows users to call the service_url/data to be redirected to the
google data studio.  It allows us to update it in the future if the
url changes and means end-users don't have to remember long google uuid
containing urls

## Checklist


mpw5 (approved):
LGTM

PR #3368, opened 2022-01-21
AP-2814 Tidy up application_proceeding_types in traits

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2814)

The ticket refers to removing the `with_proceeding_types` trait from feature test setup, however this has been done as part of https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/3349. Instead this pull request identifies a number of other related traits which can also be removed:

* `with_lead_proceeding_type`
* `with_delegated_functions`
* `with_substantive_scope_limitation`
* `with_proceeding_type_and_scope_limitations`

These are deleted and relevant feature tests and specs updated to make sure they still pass.

It also deletes the `spec/factory_helpers/application_proceeding_type_helper` which is no longer used.

## Checklist


spec/services/ccms/submitters/obtain_document_id_service_spec.rb
```diff
@@ -76,9 +76,9 @@ module Submitters # rubocop:disable Metrics/ModuleLength
           end
 
           it 'only populates the documents array with submittable documents' do
-            expect(legal_aid_application.attachments.map(&:attachment_type)).to eq all_attachment_types
+            expect(legal_aid_application.attachments.map(&:attachment_type).sort).to eq all_attachment_types.sort
             subject.call
-            expect(submission.submission_documents.map(&:document_type)).to eq submittable_document_types
+            expect(submission.submission_documents.map(&:document_type).sort).to eq submittable_document_types.sort
```
colinbruce, line 81: An alternative would be 
```ruby 
expect(submission.submission_documents.map(&:document_type)).to match_array submittable_document_types
```
That ensures the values in the array match in a non-deterministic sequence

colinbruce (approved):
LGTM :+1:
One, non-blocking, observation

spec/services/ccms/submitters/obtain_document_id_service_spec.rb
mpw5: Thanks. I knew there would be a method to do it properly 😁 I'll push an update.

PR #3369, opened 2022-01-21
AP-2827 Capture exceptions during Report Generation

stephenrichards:
## Capture exceptions during Report Generation

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2827)

Several changes to ensure that we are alerted when an exception occurs during generation of reports overnight:
- Full test coverage of ReportsUploader job
- backgrounding of report generation rather than in the cronjob ephemeral pod
- Capturing exceptions and explicitly sending them to Sentry.  The exception is not re-raised because we don't particularly want the job to be retried lots of times.

## Checklist


mpw5 (approved):
LGTM

PR #3378, opened 2022-01-24
ap-2575: Add employment income to cfe results page

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2575)

Add ‘Employment income' section to the 'Income calculation’ accordion on the CFE results page and populate with results data from CFE.
Fix a couple of failure points on mock_interface_response_service.rb to allow for testing.

## Checklist

### DIFF
--- a/spec/requests/providers/capital_income_assessment_results_spec.rb
+++ b/spec/requests/providers/capital_income_assessment_results_spec.rb
@@ -278,6 +278,61 @@ RSpec.describe Providers::CapitalIncomeAssessmentResultsController, type: :reque
         expect { subject }.to raise_error(/Unknown capital_assessment_result/)
       end
     end
+
+    context 'enable_employed_journey is true' do
+      let(:before_tasks) do
+        Setting.setting.update!(enable_employed_journey: true)
+        allow_any_instance_of(Provider).to receive(:employment_permissions?).and_return(true)
+        login_provider
+        subject
+      end
+
+      context 'applicant has employment(s)' do
+        let(:cfe_result) { create :cfe_v4_result, :with_employments }
+        it 'displays the employment income' do
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.other_income.title'))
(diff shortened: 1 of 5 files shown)


spec/requests/providers/capital_income_assessment_results_spec.rb
```diff
@@ -278,6 +278,61 @@
...
+      context 'applicant has employment(s)' do
+        let(:cfe_result) { create :cfe_v4_result, :with_employments }
+        it 'displays the employment income' do
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.other_income.title'))
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.employment_income.title'))
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.employment_income.monthly_income_before_tax'))
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.employment_income.benefits_in_kind'))
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.employment_income.tax'))
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.employment_income.national_insurance'))
+          expect(unescaped_response_body).to include(I18n.t('providers.capital_income_assessment_results.employment_income.fixed_employment_deduction'))
```
mpw5, line 299: This tests that the text is being displayed, which is good, but is it also worth testing that the correct values are displayed too? ie have a test like 
```
expect(unescaped_response_body).to include(cfe_result.employment_income_gross_income)
```
for each of the types of employment income.

It's probably overkill, so this is fine to be released without the extra tests if you don't want to add them.

mpw5 (approved):
LGTM. One question about tests, but not a blocker.

spec/requests/providers/capital_income_assessment_results_spec.rb
kmahern: Thanks @mpw5. I agree that we should test that the correct values are displayed. I have added some tests based on your suggestion - are you happy with these?

mpw5 (approved):
LGTM

PR #3382, opened 2022-01-24
Fix mock HMRC data generation

mpw5:
## What

The `HMRC::MockInterfaceResponseService` does not set the `submission_id` on the `HMRC::Response` record. This causes the process to fail as the `submission_id` is used to subsequently update the `Response` with the correct json payload.

This changes the `MockInterfaceResponseService` so that it generates a `submission_id` uuid, and updates the `Response record` with it, in its initializer.

## Checklist

### DIFF
--- a/app/services/hmrc/mock_interface_response_service.rb
+++ b/app/services/hmrc/mock_interface_response_service.rb
@@ -20,7 +20,7 @@ module HMRC
     end
 
     def call
-      @hmrc_response.update!(response: return_json.to_json, submission_id: @submission_id)
+      @hmrc_response.update!(response: return_json, submission_id: @submission_id)
     end
 
     private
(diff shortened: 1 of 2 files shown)


app/services/hmrc/mock_interface_response_service.rb
```diff
@@ -16,6 +16,7 @@ def self.call(*args)
     def initialize(hmrc_response)
       @hmrc_response = hmrc_response
       @application = @hmrc_response.legal_aid_application
+      @hmrc_response.submission_id = SecureRandom.uuid
```
colinbruce, line 19: Does this save automatically?
If not, setting line 20 to 
```ruby 
@submission_id = @hmrc_response.submission_id
```

and then changing line 24 to 
```ruby
@hmrc_response.update!(response: return_json.to_json, submission_id: @submission_id)
```
Which matches the action of the HMRC::SubmissionWorker

colinbruce (approved):
LGTM :+1:

PR #3400, opened 2022-01-27
AP-2735 new evidence upload page

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2735)

Add new page into the flow which is only accessible behind the feature flag
Create new table to ultimately replace gateway evidence
Add a drop down select element to update attachment_type
Remove delete functionality from the view due to conflict with multiple forms on a page
Add commnets around validation which need to be addressed on https://dsdmoj.atlassian.net/browse/AP-2739

## Checklist


spec/requests/providers/uploaded_evidence_collection_spec.rb
```diff
@@ -0,0 +1,357 @@
...
+      #   legal_aid_application.reload
+      #   expect(legal_aid_application.uploaded_evidence_collection.original_attachments.first).to be_present
+      # end
+
+      # it 'stores the original filename' do
+      #   subject
+      #   legal_aid_application.reload
+      #   attachment = uploaded_evidence_collection.original_attachments.first
+      #   expect(attachment.original_filename).to eq 'hello_world.pdf'
+      # end
```
RoseSAK, line 63: should these comments be removed?

config/locales/en/providers.yml
```diff
@@ -440,6 +440,32 @@ en:
         feedback_suffix: " to help us improve this service."
         view_completed_application: View completed application
         back_to_your_applications: Back to your applications
+    uploaded_evidence_collections:
+      upload_evidence:
+        list_text: 'Use this page to upload:'
+        benefit_evidence: evidence that your client receives Universal Credit
+        employment_evidence: evidence of your client's employment
```
RoseSAK, line 447: this should only be shown if the employed journey feature flag is enabled

app/models/legal_aid_application.rb
```diff
@@ -292,6 +293,10 @@ def benefit_check_result_needs_updating?
     applicant_updated_after_benefit_check_result_updated?
   end
 
+  def employment_evidence_required?
+    (applicant_employed? && !hmrc_employment_income?) || (hmrc_employment_income? && applicant_not_employed?)
+  end
```
RoseSAK, line 298: this should return false always unless the employed journey feature flag is on

RoseSAK (approved):
LGTM 👌

spec/requests/providers/uploaded_evidence_collection_spec.rb
willc-work: they are tests that we should have on the page but dont work at the moment because of validation. As there is a separate ticket for validation I commented them out so they could be addressed in the other ticket.

config/locales/en/providers.yml
willc-work: fixed below

PR #3427, opened 2022-02-03
Ap 2469 migrate to lfa

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2469)

Remove all models, specs, factories, populators etc that are now handled by LFA
Update test setups and remove all references to aforementioned models
I have also had to remove some old migrations e.g. where a migration added a field to a model that this PR removes.
I am happy to discuss alternatives to this approach if anyone has any suggestions.

## Checklist


db/migrate/20220204114732_drop_proceeding_type_tables.rb
```diff
@@ -0,0 +1,13 @@
+class DropProceedingTypeTables < ActiveRecord::Migration[6.1]
+  def up
+    drop_table :application_proceeding_types, force: :cascade
+    drop_table :proceeding_types, force: :cascade
+    drop_table :scope_limitations, force: :cascade
+    drop_table :proceeding_type_scope_limitations, force: :cascade
+    drop_table :application_proceeding_types_scope_limitations, force: :cascade
```
mpw5, line 7: Possibly some other tables that could be dropped here too:

`application_proceeding_types_linked_children`
`default_cost_limitations`
`service_levels`

mpw5 (approved):
Great work 🍾 

I think there might be a couple of extra tables that could be dropped, but I haven't checked 100% that it won't break anything. I think it would be fine not to add them now if it looks like it might be problematic. I'm sure we'll find other bits that need tidying up in the near future anyway.

db/migrate/20220204114732_drop_proceeding_type_tables.rb
willc-work: dropped them

db/schema.rb
```diff
@@ -371,9 +338,6 @@
     t.index ["proceeding_id"], name: "index_chances_of_successes_on_proceeding_id"
   end
 
-  create_table "data_migrations", primary_key: "version", id: :string, force: :cascade do |t|
```
colinbruce, line 374: Isn't this table needed for `rake db:migrate:with_data`?
If not, you'll need to delete it from the anonymise rules file

willc-work: Thanks Colin, I think that has crept in from rebasing. I will fix this.

PR #3442, opened 2022-02-07
AP-2793 update govuk frontend and formbuilder

RoseSAK:
## Update to most recent versions of govuk-frontend and govuk_design_system_formbuilder

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2793)

Main changes:
* Add JS to skip link
* Update error message HTML
* Update hint HTML
* Fix fallback logo 
* Fixed the proceedings search JS to look within the hint div instead of span
* Update the accordion headings (this completes [AP-2641](https://dsdmoj.atlassian.net/browse/AP-2641)

List of changes [here](https://docs.google.com/spreadsheets/d/1zVVG_QelvQ6Io6SPbv3SgosG7H1UiLhbWdmKHbSGWh0/edit#gid=0)

## Checklist


db/schema.rb
```diff
@@ -607,7 +607,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.boolean "scanner_working"
-    t.index ["uploader_type", "uploader_id"], name: "index_malware_scan_results_on_uploader_type_and_uploader_id"
+    t.index ["uploader_type", "uploader_id"], name: "index_malware_scan_results_on_uploader"
```
colinbruce, line 610: I think this migration change has crept in before and should be reverted?

colinbruce (approved):
🏆 
There's one random schema change that seems to have crept in, but otherwise :+1:

db/schema.rb
RoseSAK: Ah yes thanks for spotting, will revert it now!

PR #3493, opened 2022-02-15
ap-2885: add path to date_client_told_incidents path step

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-2885)

Flow::FlowError: :path of step :date_client_told_incidents is not defined" was being raised in Sentry. After investigations, this is raised when the user clicks Save and come back later on the 'Latest incident details' page. When they click on applicant name to continue the application they are then unable to progress the application, as they are taken to the applicant details page rather than the latest incident details page (resulting in a further 'enter_applicant_details' cannot transition from 'provider_entering_merits' sentry error)

I noticed that the same issue was affecting the chances_of_success page so I have also added a path to this step.

- Added path to date_client_told_incidents step in provider_merits flow
- Added path to chances_of_success step in provider_merits flow

## Checklist

### DIFF
--- a/app/services/flow/flows/provider_merits.rb
+++ b/app/services/flow/flows/provider_merits.rb
@@ -49,6 +49,7 @@ module Flow
           }
         },
         date_client_told_incidents: {
+          path: ->(application) { urls.providers_legal_aid_application_date_client_told_incident_path(application) },
           forward: :opponents,
           check_answers: :check_merits_answers
         },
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
🏆 
Awesome work identifying this ⭐

PR #3521, opened 2022-02-24
AP-2877 Employment data alignment

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2877)

When Tax or National Insurance values on the Employment Income Review screen have more then five significant digits they exceed the column width and wrap incorrectly.

1. Update `app/views/providers/employment_incomes/show.html.erb` so that the widths of the table header columns match the widths of the table body columns. This improves the amount of space available for the data.

2. Add the `govuk-table__cell--numeric` class to Income and Deductions values to bring them into line with other monetary values in the service, and also with the prototype design.

## Checklist


willc-work (approved):
LGTM

PR #3524, opened 2022-02-24
AP-2905 Persist HMRC data

stephenrichards:
Persist HMRC data as Employment and EmploymentPayment records

## What

## Checklist


spec/services/cfe/create_employments_service_spec.rb
```diff
@@ -24,19 +24,18 @@
       let(:expected_payload) { empty_payload }
 
       it 'sends empty array' do
+        # this tests passes if the service sends the same payload as is stubbed out above, i.e. expected_payload
```
colinbruce, line 27: But is this even a test?  There's no `expect` ?!
The instantiation of the class raises an error but it seems like a code smell not to have an actual expectation!
e.g. 
``` ruby
expect {described_class.call(submission) }.not_to raise_error StandardError
```

spec/services/cfe/create_employments_service_spec.rb
```diff
@@ -24,19 +24,18 @@
...
       let(:expected_payload) { full_payload }
-
-      before do
-        create :hmrc_response, :example1_usecase1, legal_aid_application: laa
-      end
+      let!(:employment) { create :employment, :example1_usecase1, legal_aid_application: laa }
 
       it 'sends array of payment data' do
+        # this tests passes if the service sends the same payload as is stubbed out above, i.e. expected_payload
         described_class.call(submission)
```
colinbruce, line 39: See comment above about tests 🤷‍♂️

colinbruce (changes requested):
Some slight oddness around the tests but otherwise looking good

spec/services/cfe/create_employments_service_spec.rb
stephenrichards: I agree it is confusing, and actually, this hasn't changed from what was there  before, except the comment to explain what was going on.  

we want to test that the correct payload is being generated.  This does that, but it's not at all clear.  I think the best way is for me to set up a method on the class which returns the payload, and then we can set an expecation on that.

colinbruce: I spotted Mark had done something similar in his PR [here](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/3532/files#diff-447b95f277fea6ea9808a9179048fd854f2fe78cac86f1565bba6f842f7c6dccR54)

colinbruce (approved):
LGTM :+1:

PR #3540, opened 2022-03-01
AP-2950 Include all manually entered employment income data for result panel/caseworker reviews

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2950)

* Adds a new method to the `LegalAidApplication` model to returns a boolean value indicating whether _any_ employment income data has been manually entered by the provider. This returns true if either `extra_employment_information` is true or `full_employment_details` contains data.
* Amends the `ManualReviewDetailer` service class to use that method to determine whether casworker review is required.
* Amends the `ResultsPanelSelector` service class to use that method to determine the correct result panel to display.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -471,6 +471,10 @@ class LegalAidApplication < ApplicationRecord
     out.transform_values { |attachment| attachment.map(&:original_filename) }
   end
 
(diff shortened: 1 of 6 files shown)


kmahern (approved):
lgtm 👍

PR #3555, opened 2022-03-07
ap-2947: Deactivate inactive provider accounts

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-2947)

This pr contains a new rake file with two tasks, to list and deactivate inactive providers.

## Checklist

### DIFF
--- a/lib/tasks/inactive_providers.rake
+++ b/lib/tasks/inactive_providers.rake
@@ -11,7 +11,7 @@ namespace :inactive_providers do
         updated_at
         portal_enabled
       ]
-      inactive_providers = Provider.where('current_sign_in_at < ? OR (current_sign_in_at IS NULL AND created_at < ?)', 6.months.ago.to_date, 6.months.ago.to_date)
+      inactive_providers = Provider.where('current_sign_in_at < :limit OR (current_sign_in_at IS NULL AND created_at < :limit)', limit: 6.months.ago.to_date)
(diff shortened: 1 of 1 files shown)


lib/tasks/inactive_providers.rake
```diff
@@ -0,0 +1,38 @@
...
+      csv << %w[
+        id
+        email
+        last_sign_in_at
+        current_sign_in_at
+        created_at
+        updated_at
+        portal_enabled
+      ]
+      inactive_providers = Provider.where('current_sign_in_at < ? OR (current_sign_in_at IS NULL AND created_at < ?)', 6.months.ago.to_date, 6.months.ago.to_date)
```
colinbruce, line 14: You could replace the ? with named variables to reduce duplication e.g. 
```ruby
Provider.where('current_sign_in_at < :limit OR (current_sign_in_at IS NULL AND created_at < :limit)', limit: 6.months.ago.to_date)
```

colinbruce (approved):
LGTM :+1:

One observation to reduce the duplication of the variables but it's not a blocker 👍

lib/tasks/inactive_providers.rake
kmahern: Thanks @colinbruce have implemented this suggestion 👍

PR #3574, opened 2022-03-09
AP-2974 fix validation errors not appearing the second time

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2974)

- Fixed the validation errors so that the correct error is displayed if there is a second error (on second upload)
- The errors are hidden if a valid file is then uploaded
- Fixed on both evidence upload and statement of case page
- Moved the code to hide error messages into a separate function, and called this function on the 'drop' event and on the 'Choose files' click event
- Added a new accepted document type that supports MS Word 2007 files 

## Checklist


mpw5 (approved):
LGTM 🏆

PR #3588, opened 2022-03-14
AP-2935 Add GOVUK Rubocop Gem

mpw5:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2935)

Replaces the generic rubocop gem with the GOVUK version to ensure we meet cross-gov code standards and have consistency between our applications.

* Adds all offences to a (very long) `.rubocop_todo.yml` file so that they can be addressed later.

* Where we have previously overriden cops in `.rubocop.yml`, I have left the configuration in if it targets specific file/s (as we know we have unresolvable issues with some classes) but removed it if it was a generic exclusion, as we should now be aiming to conform to the GOVUK standard.

## Checklist


.rubocop.yml
```diff
@@ -1,9 +1,18 @@
...
+    - config/default.yml
+    - config/rails.yml
+    - config/rspec.yml
+
+inherit_mode:
+  merge:
+    - Exclude
+
 AllCops:
   TargetRubyVersion: 2.7
```
colinbruce, line 17: Should we have updated this to 3.x? 🤔

colinbruce (approved):
Looks good (apart from something _I_ may have missed) :+1:

.rubocop.yml
mpw5: Good spot, I completely missed that. I'll remove it, if it's not present then rubocop falls back on `.ruby-version`. There doesn't seem to be any benefit in configuring it in multiple places 🤷

colinbruce: Me too - I've updated ruby multiple times and not spotted it 😬 
Thanks!

PR #3589, opened 2022-03-14
Upgrade to Rails 6.1.5

chris-groves:
-Edit rendering of inline 'read-only'
-Include 'autocomplete=off' in hidden field tests
-Edit syntax of 'error_id' 

## Checklist


app/views/shared/check_answers/_item.html.erb
```diff
@@ -8,7 +8,7 @@
     <%= question %>
   </dt>
   <!--  in-line styling due to issues around pdf generation in our production envs -->
-  <dd class="govuk-summary-list__value"<% if read_only %> style="text-align: right;"<% end %>>
+  <dd class="govuk-summary-list__value" style="<%= "text-align: right;" if read_only %>">
```
mpw5, line 11: Nice! I was looking at this earlier; it's a fix for the same problem which is causing https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/3485 to fail. I'll close that PR as you've fixed it here.

app/views/shared/partials/_revealing_checkbox.html.erb
```diff
@@ -9,7 +9,7 @@
     <% if error %>
       <% error_class='govuk-form-group--error'
          error_id = "#{model.model_name.to_s.underscore.dasherize}-#{name.dasherize}#{number}" %>
-      <p class='govuk-error-message' id="<% error_id %>"><%= error %></p>
+      <p class='govuk-error-message' id="error_id"><%= error %></p>
```
mpw5, line 12: I'm not sure this quite works. This will set the id to the string "error_id" rather than the contents of the `error_id` variable, which I think will cause issues if there's more than one error on the page. 

Would `id="<%= error_id %>"` work instead? [This](https://github.com/Shopify/better-html#syntax-restriction) seems to suggest it should be allowed.

mpw5 (changes requested):
Excellent work, but I think the fix to `app/views/shared/partials/_revealing_checkbox.html.erb` needs looking at a little more. Happy to approve once that's fixed.

mpw5 (approved):
LGTM

PR #3609, opened 2022-03-17
Fix flaky rspec array comparison

mpw5:
## What

In `spec/models/legal_aid_application_spec.rb` we test the output of `uploaded_evidence_by_category`, which is a hash of arrays. The elements of those arrays are not always populated in the same order, which causes a test to fail. 

This adds a `deep_match` method (shamelessly stolen from https://github.com/ministryofjustice/check-financial-eligibility/pull/826/commits/bd67f5f18032da34bc3a18dd834e77cbca09cccd) which allows for comparison of nested data structures and updates the test to use it.

## Checklist

### DIFF
--- a/spec/models/legal_aid_application_spec.rb
+++ b/spec/models/legal_aid_application_spec.rb
@@ -1310,7 +1310,7 @@ RSpec.describe LegalAidApplication, type: :model do
 
     context 'evidence has been uploaded' do
       it 'returns a hash of evidence filenames grouped by category' do
-        expect(laa.uploaded_evidence_by_category).to eq(uploaded_evidence_output)
+        deep_match(laa.uploaded_evidence_by_category, uploaded_evidence_output)
       end
     end
   end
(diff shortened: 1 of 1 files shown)


colinbruce (approved):
LGTM :+1:
Does the sequence of Gateway Evidence need to change?

mpw5:
> LGTM 👍 Does the sequence of Gateway Evidence need to change?

No, I changed it when I was testing and forgot to put it back 🤦

colinbruce:
> > LGTM 👍 Does the sequence of Gateway Evidence need to change?
> 
> No, I changed it when I was testing and forgot to put it back 🤦

I wasn't sure if it was 'Look it definitely works regardless of sequence' choice! 😄

mpw5:
> > > LGTM 👍 Does the sequence of Gateway Evidence need to change?
> > 
> > 
> > No, I changed it when I was testing and forgot to put it back 🤦
> 
> I wasn't sure if it was 'Look it definitely works regardless of sequence' choice! 😄

I guess I could have left it like that for that reason, but then it's just proving that a different hard-coded order works 🤷 Maybe I should have randomised the array just to make debugging more fun in future 😁

PR #3610, opened 2022-03-17
Fix GOVUK Rubocop Style/StringLiterals offences

chris-groves:
Fix GOVUK Rubocop Style/StringLiterals offences

"Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping. (https://rubystyle.guide#consistent-string-literals)"

## Checklist


mpw5 (approved):
LGTM. 

I've only checked a sample of changes given the size of the PR. Maybe in future we should look to keep the number of files changed to a minimum.

PR #3625, opened 2022-03-22
ap-2940: caseworker referrals

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2940)

- Added a check for policy_disregards? to manual_check_required? in manual_review_determiner
- Added restrictions and further_employment_information as caseworker manual review reasons on the means report

## Checklist

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -679,6 +679,8 @@ en:
           policy_disregards: Discretionary disregards selected
           multi_benefit: Multiple benefits in single bank transaction
           dwp_override: Provider disagrees with DWP result
+          restrictions: Restrictions on client's assets
+          further_employment_details: Further employment details
       show:
         heading: Means report
         caseworker-review-section-heading: Caseworker Review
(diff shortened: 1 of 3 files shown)


config/locales/en/providers.yml
```diff
@@ -679,6 +679,8 @@ en:
           policy_disregards: Discretionary disregards selected
           multi_benefit: Multiple benefits in single bank transaction
           dwp_override: Provider disagrees with DWP result
+          restrictions: Restrictions on client's assets
+          further_employment_details: Further employment details
```
jsugarman, line 683: For my benefit, what do we do about adding welsh translations?

kmahern: @jsugarman we have a rake task cy.rake that generates the dummy backwards welsh translations. However, I believe this only covers the citizens and not the providers journey (i.e there is no locales/cy/providers.yml file).

chris-groves (approved):

chris-groves (approved):

colinbruce (approved):

PR #3656, opened 2022-03-31
AP-3204: Improve admin CFE display

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2304)
* Improve handling of remarks - the array of GUIDs returned by CFE for remarks broke the logic of the hash loop this ensures that a hash is rendered the same as before, but anything else will be treated as a string and output simply
* Improve handling when a CFE Result is missing, this is an edge case but occurred while working on debugging the issue.  It should never occur in production but will ensure the page does not crash if it does

image

## Checklist

### DIFF
--- a/app/views/admin/legal_aid_applications/submissions/show.html.erb
+++ b/app/views/admin/legal_aid_applications/submissions/show.html.erb
@@ -1,7 +1,3 @@
-<%
-  empty_cfe_result = {result: 'nil'}
-%>
-
 <%= back_link() %>
 <div class='govuk-body'>
   <h1 class="govuk-heading-l">Legal Aid Application</h1>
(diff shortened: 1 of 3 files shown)


app/views/admin/legal_aid_applications/submissions/show.html.erb
```diff
@@ -65,7 +65,11 @@
           CFE result
         </summary>
         <section>
-          <%= format_hash(JSON.parse(cfe_submission.result&.result || empty_cfe_result)) %>
```
jsugarman, line 68: can the `empty_cfe_result` result be deleted from line 2 of this file. it does not appear to be used anywhere now

app/views/admin/legal_aid_applications/submissions/show.html.erb
```diff
@@ -65,7 +65,11 @@
...
+          <% if cfe_submission.result&.result.nil? %>
+            <p class="govuk-body">No result available</p>
+          <% else %>
+            <%= format_hash(JSON.parse(cfe_submission.result&.result)) %>
+          <% end %>
```
jsugarman, line 72: ****not a blocker** ** is `nil?` check needed?

```suggestion
          <% if cfe_submission.result&.result %>
            <%= format_hash(JSON.parse(cfe_submission.result&.result)) %>
          <% else %>
            <p class="govuk-body">No result available</p>
          <% end %>
```

jsugarman (approved):
👍 

the `empty_cfe_result` var is no longer used now by the looks of it

colinbruce:
Sorry @jsugarman - I realised the branch had the wrong name and had to close and open a new one... all of your points are addressed on #3670 I'll get that re-reviewed 👍

PR #3670, opened 2022-04-04
AP-3024: Improve admin CFE display

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3024)
* Improve handling of remarks - the array of GUIDs returned by CFE for remarks broke the logic of the hash loop this ensures that a hash is rendered the same as before, but anything else will be treated as a string and output simply
* Improve handling when a CFE Result is missing, this is an edge case but occurred while working on debugging the issue.  It should never occur in production but will ensure the page does not crash if it does

image

## Checklist


colinbruce:
original comments on #3656

kmahern (approved):
lgtm 👍

PR #3678, opened 2022-04-05
AP-2783: Rails 7 update

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2783)

This is a placeholder - once it is deployed I will add the rails 7 commits and see if
we can replicate the issues we saw in production

## Checklist


colinbruce:
rebased after circle env-var changes

config/application.rb
```diff
@@ -1,28 +1,25 @@
...
-require "active_storage/engine"
-require "action_controller/railtie"
-require "action_mailer/railtie"
-require "action_mailbox/engine"
-require "action_text/engine"
-require "action_view/railtie"
-require "action_cable/engine"
-# explicitly not loading 'sprockets/railtie'
-require "rails/test_unit/railtie"
+require "rails/all"
```
jsugarman, line 4: **not a blocker**, but the indvidual component give us more fine grained control i would have thought  - for example are we using action_cable and action_mailbox?

jsugarman (approved):
👍 LGTM

reminder to remove the last TEMP commit

config/application.rb
colinbruce: Generally speaking most Rails apps have `rails/all`, see [cccd](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/config/application.rb#L3)

We only swicthed to fine-grained to remove sprockets when we switched to (or tried to) WebPacker
Sub gems still required sprockets eventually 😢 

This just reverts to the standard rails import

PR #3698, opened 2022-04-11
AP-3053 Make dates in HMEC Mock payloads dynamic

stephenrichards:
The mock payloads returned by `HMRC::MockInterfaceResponseService` contained static dates, which would
soon be useless as they would not fall in the calculation period, unless you entered very old
delegated function dates

This PR makes them dynamic, by changing the `.json` files to `.json.erb` and including expressions that
generate dates according to the calculation date on the `LegalAidApplication` (or today's date if there is
no calculation date)

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3053)

## Checklist


jsugarman (approved):
👍

PR #3728, opened 2022-04-20
AP-2853: Remove legacy ingress values

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2853)

Updated the uat_deploy script to remove the `LEGACY_RELEASE_HOST` value and leave the `ingress.hosts` value as an array to avoid refactoring the `ingress.yaml` file and leave us able to add further hosts in the future if needed

## Checklist

### DIFF
--- a/bin/uat_deploy
+++ b/bin/uat_deploy
@@ -17,7 +17,7 @@ deploy() {
                 --set deploy.host="$RELEASE_HOST" \
                 --set image.repository="$IMG_REPO" \
                 --set image.tag="${CIRCLE_SHA1}" \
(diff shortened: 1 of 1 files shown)


jsugarman (commented):
👍 is the same logic applicable across the repos for UAT?

jsugarman (approved):

colinbruce:
> 👍 is the same logic applicable across the repos for UAT?

Yes, but the other repos will also require the values files for staging and prod changed too as they don't have service urls

PR #3738, opened 2022-04-22
AP-2894 Refactor UploadedEvidenceCollectionsController

stephenrichards:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2894)

This controller was very large with lots of complicated logic to handle displaying uploaded files,
the mandatory categorisations, deleting files, validation, etc.

This change refactors that complication out into a series of services.

For the update action, the controller calls UploadedEvidence::Base service, passing in itself
as a parameter, and the Base service determines which child service should be called depending on which
button was pressed.  For the other two actions, :show and :list, the child service is called directly
from the controller.

All the controllers inherit from Base so as to have access to common methods.

## Checklist


RoseSAK (approved):
LGTM 👍 my only comment would be to make sure Mike tests all the error message combinations, as there are a lot of different error messages on that page, just to ensure they are all still working as expected

willc-work (approved):

PR #3739, opened 2022-04-22
Tech-Debt: Try and reduce deploy-uat times

colinbruce:
## What

This is an attempt to see if we can reduce the times for deploy jobs

## Checklist


jsugarman:
does not yet look like it is having a significant impact but i still think it is a good "strategy" to use. CCCD has a 1:30 to 2:30 deployment for production for a similar sized image. rarely going over 2 minutes though. 

There is also an advantage in switching over to the new pods all at once in the case of migrations, for example, where you don't want old codebases running against the changed db

colinbruce:
@jsugarman check out the next deploy time! The Helm --wait command was slowing down and waiting for the entire process to complete 😕 
I'm not sure about running this on staging or prod yet, what do ou think?

jsugarman:
> I'm not sure about running this on staging or prod yet, what do you think?

I reckon its worth a shot and can remove if no obvious benefit in the coming week

colinbruce:
Agreed - I have restored the `helm --wait`

jsugarman (approved):
👍

PR #3740, opened 2022-04-22
Update ruby version to 3.1.2

colinbruce:
## What

Update ruby to the latest version

## Checklist


.circleci/config.yml
```diff
@@ -153,21 +153,29 @@ jobs:
         name: Run erblint
         command: bin/rails erblint
     - run:
-        name: Check if new apply-ci image is needed, and build if so
+        name: Check if new apply docker images is needed, and build if so
```
jsugarman, line 156: not-a-blocker
```suggestion
        name: Check if new apply docker images are needed, and build if so
```

jsugarman (approved):
looks good 👍 

That command looks like it could do with being extracted to a  script and called from the circleci step.
 e.g. `.circle/build.sh` or similar [cccd example](https://github.com/ministryofjustice/Claim-for-Crown-Court-Defence/blob/master/.circleci/build.sh) 

could even be split into two scripts (build_executor, build_base) and called in separate steps or same step.

PR #3751, opened 2022-04-25
[AP-3039] use net_housing_costs instead of gross_housing_costs

jsugarman:
## What
Replace monthly equivalent housing costs with net_housing_costs

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3039)

The monthly outgoing equivalents use gross_housing_costs
which does not take thresholds and other factors into account.

## Checklist

### DIFF
--- a/app/models/cfe/v4/result.rb
+++ b/app/models/cfe/v4/result.rb
@@ -249,6 +249,10 @@ module CFE
       #                                                              #
       ################################################################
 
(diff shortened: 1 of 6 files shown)


jsugarman:
see simpler change https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/3760

PR #3758, opened 2022-04-26
AP-2953 increase resources for admin_report crontab

stephenrichards:
## Increase resources for admin_report crontab

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2953)

this is the second attempt to increase the resource sizes of this cronjob so that it
runs to completion.

The timing of the jobs has also been increased so that it runs several times throughout the
day enabling us to see whether this has been effective without waiting overnight.  If successful,
it will be set back to one overnight run again.

## Checklist


kmahern (approved):
👍

PR #3760, opened 2022-04-26
[AP-3039] Minimal fix to use `net_housing_costs` instead of `rent_or_mortgage`

jsugarman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3039)

`net_housing_costs` deals with caps and other housing cost
edge cases, as well as - i believe from conversions with
other devs - being a month equivalent.

However this feels icky as it breaks the basic pattern of
having the values for moe in the `month_equivalent` response
from CFE.

**Should this be handled in CFE instead and apply remain
unchanged?**

## Checklist

### DIFF
--- a/app/models/cfe/v4/result.rb
+++ b/app/models/cfe/v4/result.rb
@@ -250,7 +250,7 @@ module CFE
       ################################################################
 
       def moe_housing
-        monthly_outgoing_equivalents[:rent_or_mortgage].abs
+        disposable_income_summary[:net_housing_costs].abs
       end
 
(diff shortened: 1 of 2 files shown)


kmahern (approved):
👍 lgtm from a code perspective - might be worth having a chat with @stephenrichards sbout whether the CFE response should be changed instead?

jsugarman:
closed in-lieu of https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/3751

PR #3773, opened 2022-04-28
[AP-3068] Do not alert on use_case "two" responses

jsugarman:
## What
Quieten the sentry output from HMRC responses.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3068)

Every query to HMRC interface will result in two
response. a use_case one and a use_case two. The
use_case two are not currently valid or usable so
are always considered unpersistable for employment
purposes and so we do not need to alert on them.

## Checklist


colinbruce (approved):
LGTM :+1:

PR #3774, opened 2022-04-28
[AP-3062] Do no validate individual against applicant

jsugarman:
## What
Do not validate for matching individual/applicant

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3062)

Since we have requested the HMRC details for the
applicant any record found should adhere to their
matching criteria. We therefore should not need
to confirm this in their response.

Discussed (CB/JS) and decided to remove for now pending 
further information from business (via DF). 
This follows on from actual responses being returned where the 
surname only matches was entered wrongly by provider but
matched correctly from HMRC based on their criteria (see HMRC
developer hub for more)

## Checklist

### DIFF
--- a/app/services/hmrc/parsed_response/persistor.rb
+++ b/app/services/hmrc/parsed_response/persistor.rb
@@ -22,7 +22,7 @@ module HMRC
       attr_reader :hmrc_response
 
       def persistable?
-        Validator.call(hmrc_response, applicant: @application.applicant)
+        Validator.call(hmrc_response)
       end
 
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM :+1:

stephenrichards (commented):
Given what we know about HMRC, I think not validating that they are returning the data for the person we asked for is unsafe.  At the very minimum, we should be checking that the NINO and DOB match - IMHO

PR #3775, opened 2022-04-28
[AP-3066] Hardcode section 8 level of service value

jsugarman:
## What
[TBC] hardcode default level-of-service for section 8 matter types

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3066)

The default "level of service" for a section 8 matter
in CCMS is "Family help (higher)", with a value of "1", not
as the current hadrcoded level-of-service, "Full representation"
with a value of "3".

This is the simplest and most immediate approach to sending a correct
default value for the section 8 matter type. Longer term we are looking
to store this data for the 6 level-of-services in LFA.

## Checklist

### DIFF
--- a/app/models/proceeding.rb
+++ b/app/models/proceeding.rb
@@ -1,12 +1,16 @@
 class Proceeding < ApplicationRecord
   FIRST_PROCEEDING_CASE_ID = 55_000_000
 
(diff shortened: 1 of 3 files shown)


jsugarman:
we are going to be implementing a solution involving LFA

PR #3782, opened 2022-04-29
[AP-3062] Do not validate on first or last name

jsugarman:
## What
Do not validate on first or last name, only nino and dob

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3062)

Avoid mismatches resulting from unusual
names and spelling mistakes either by our
users or in HMRC records.

We want some level of validation so we can
be confident our request of HMRC results in
a response for the individual/applicant entered
in our system. We have decided as a team (CB/JS/SR)
that the unique nature of a nino coupled with a date
of birth should suffice to give us warning of inappropriate
responses from HMRC.

## Checklist

### DIFF
--- a/app/services/hmrc/parsed_response/validator.rb
+++ b/app/services/hmrc/parsed_response/validator.rb
@@ -61,7 +61,6 @@ module HMRC
       def validate_response_individual
         errors << error(:individual, "individual must match applicant") unless individual &&
           applicant &&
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
Looks as agreed :+1:

stephenrichards (approved):

PR #3789, opened 2022-05-03
AP-3065 Retain cron pods so we can examine logs

stephenrichards:
## Retain cron pods so we can examine logs

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3065)

Recen change to create CSV in tempfile and attach that hasn't worked - jobs are still failing silently.  This change is to retain the cron job pods so that we can look at the logs and see what went wrong.

## Checklist


jsugarman (approved):
👍

PR #3829, opened 2022-05-10
Test: Log CFE call duration

colinbruce:
## What

Output CFE call durations (steps and total) so that the values can be monitored via kibana

## Checklist

### DIFF
--- a/app/services/cfe/submission_manager.rb
+++ b/app/services/cfe/submission_manager.rb
@@ -1,5 +1,6 @@
 module CFE
   class SubmissionManager
+    include ::DurationLogger
     COMMON_SERVICES = [
       CreateAssessmentService,
       CreateApplicantService,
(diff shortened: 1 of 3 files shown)


app/services/cfe/submission_manager.rb
```diff
@@ -46,15 +46,24 @@ def submission
...
       return if submission.passported?
 
-      NON_PASSPORTED_SERVICES.each do |s|
-        s.call(submission)
+      NON_PASSPORTED_SERVICES.each do |service|
+        log_duration(service)
       end
     end
+
+    def log_duration(service)
```
jsugarman, line 62: might be worth making reusable and use a block

for example, not test by any means

```
def log_duration(&block)
  start_time
  yield block if block_given?
  Rails.logger.info("duration: #{end_time - end_time}...blah")
  end_time
end
```

```
NON_PASSPORTED_SERVICES.each do |service|
        log_duration do
          service.call(submission)
      end
end
```

jsugarman: **not a blocker**

jsugarman (approved):
👍 

a reusable block call might be nice, as commented.

PR #3858, opened 2022-05-18
[AP-2640 ]Add feature test hooks to mock hmrc calls

jsugarman:
## Note
This is merge into branch to fix failing cukes

## What
Enable Mock HMRC calls during employed journey feature tests

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2640)

Employed journey tests are dependant on the
hmrc mock/sandobox data. This relies on a
config setting being in place and the test
were failing on CI as the mock was not enabled.

This stubs that config so feature tests can be run against it.

Should we be doing this however? We should, perhaps,
be using vcr to record (or otherise stub) real staging
hmrc responses - so long as no sensitive info is exposed.

## Checklist

### DIFF
--- a/features/providers/check_multiple_employment.feature
+++ b/features/providers/check_multiple_employment.feature
@@ -1,5 +1,5 @@
 Feature: check_multiple_employment
-  @javascript @vcr
+  @javascript @vcr @hmrc_use_dev_mock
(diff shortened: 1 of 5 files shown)


mikekeen1 (approved):
Approved

PR #3859, opened 2022-05-18
AP-2363 Improve alerting around document conversion

chris-groves:
Raise Sentry alerts only when 10 attempts have failed to convert a document to PDF.

Reduce the amount of errors raised so the key issues can be focused on.

## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-2363)

## Checklist

### DIFF
--- a/app/workers/pdf_converter_worker.rb
+++ b/app/workers/pdf_converter_worker.rb
@@ -1,8 +1,19 @@
 class PdfConverterWorker
   include Sidekiq::Worker
   include Sidekiq::Status::Worker
+  attr_accessor :retry_count
+
+  ALERT_ON_RETRY_COUNT = 3
+  sidekiq_options retry: ALERT_ON_RETRY_COUNT
+  sidekiq_retries_exhausted do |msg, _ex|
+    Sentry.capture_message Sidekiq::ExhaustedFailureMessage.call(msg)
+  end
 
   def perform(attachment_id)
     PdfConverter.call(attachment_id)
+  rescue StandardError => e
+    raise if retry_count.eql? ALERT_ON_RETRY_COUNT
+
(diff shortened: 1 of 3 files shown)


app/workers/pdf_converter_worker.rb
```diff
@@ -1,8 +1,19 @@
...
+  MAX_RETRIES = 10
+  sidekiq_options retry: MAX_RETRIES
+  sidekiq_retries_exhausted do |msg, _ex|
+    Sentry.capture_message Sidekiq::ExhaustedFailureMessage.call(msg)
+  end
 
   def perform(attachment_id)
     PdfConverter.call(attachment_id)
+  rescue StandardError => e
+    raise if @retry_count.eql? MAX_RETRIES
```
jsugarman, line 15: **nit**, since you have an `attr_accessor` can you not use that:
```suggestion
    raise if retry_count.eql? MAX_RETRIES
```

app/workers/pdf_converter_worker.rb
```diff
@@ -1,8 +1,19 @@
...
+
+    raise SentryIgnoreThisSidekiqFailError, "Attempt to convert file to PDF failed on retry #{@retry_count.to_i} with error #{e.message}"
```
jsugarman, line 17: and here:
```suggestion
    raise SentryIgnoreThisSidekiqFailError, "Attempt to convert file to PDF failed on retry #{retry_count.to_i} with error #{e.message}"
```

jsugarman (commented):
couple of nits

one question is what is the max retry count. sidekiq default is 25 but have not checked? And should we perhaps limit this to fewer retries if so. The name `MAX_RETRIES` might be clearer as `ALERT_ON_RETRY_COUNT` as this is what its purpose seems to be?!

jsugarman (approved):
👍

PR #3862, opened 2022-05-19
Tech-Debt: Update Helm ingressClassName handling 

colinbruce:
## What

Back in January, we updated our ingress files to use the new, v1, ingress template.
Although it has not impacted our system the `kubernetes.io/ingress.class` needed to be renamed to `spec.ingressClassName` 

This PR updates all environments to use the new convention, including adding a `default` value to UAT instead of relying on the service assuming a missing value was `default`

## Checklist


jsugarman (approved):
LGTM 👍

PR #3896, opened 2022-05-26
AP-3130: Update ModSec handling for Statement of Case page

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3130)

* Update the handling of ModSecurity on the Statement of Case page
* Resequence the ModSec rules to prevent clashes when we upgrade our ingress

## Checklist


jsugarman (approved):
👍 Have taken a look locally and cannot say I understand rule changes but, if it works 🤷

PR #3934, opened 2022-06-09
AP-3119 remove column and related content

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3119)
removed the column with the links
removed unused translations
updated the tests

## Checklist

### DIFF
--- a/app/views/providers/start/index.html.erb
+++ b/app/views/providers/start/index.html.erb
@@ -1,5 +1,5 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+  <div class="govuk-grid-column-full">
     <h1 class="govuk-heading-xl"><%= t('.apply_for_legal_aid_heading') %></h1>
 
     <div class="govuk-inset-text">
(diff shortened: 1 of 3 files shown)


app/views/providers/start/index.html.erb
```diff
@@ -22,21 +22,4 @@
                 class: 'govuk-button govuk-button--start',
                 id: 'start' %>
   </div>
-
-  <div class="govuk-grid-column-one-third">
```
colinbruce, line 26: This may be a question for @RoseSAK - if we remove a column one-third, should the other column-two-thirds be updated/renamed?

colinbruce (commented):
Looks good, but I have a question for Rose

app/views/providers/start/index.html.erb
willc-work: I wasnt sure about that but as removing the 1/3 didnt break the formatting for the rest if the page I left it as is, but I am happy to defer to @RoseSAK

RoseSAK: Heya, I think since the one-third column is removed, the remaining column should be changed to 'govuk-grid-column-full' rather than left as two-thirds, if that's alright, just to be consistent

RoseSAK: @willc-work

RoseSAK (approved):

PR #3949, opened 2022-06-13
[AP-3103] Prevent 0 byte file upload using dropzone

jsugarman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3103)

Zero byte files are getting through and causing
failures on CCMS submission since they are
raising the following error:

```
CCMS::CCMSError The following documents failed to upload
```

On investigation it appears that while the _fallback_ `BaseFileUploaderForm` class successfully rejects 
zero byte file uploads, when JS is DISABLED, a zero-byte file can get through when JS is ENABLED as 
dropzone bypasses that backend-but-form-only validation. To prevent this I have perform the same basic
validation using dropzone.

It might be more robust to prevent 0 byte file attachments via the attachment/activestorage
model itself rather than rely on the more "front-end(ish)" mechanism currently in place.

- [x] dropzone immediate warning when js enabled
- [x] backend fallback warning when js disabled
- [x] ensure request specs reflect actual behaviour

nb: currently `spec/requests/providers/application_merits_task/statement_of_case_spec.rb:296` is passing because, as a request spec, it tests tge non-javascript enabled mechanism. Feature tests have been added to exercise the javascript enabled/dropzone validated mechanism

## Checklist


RoseSAK (approved):
This all looks good to me 👍

PR #3959, opened 2022-06-15
Disable sentry alert when hmrc return no employments for applicant

kmahern:
## What

This is a pr to disable sentry alerting where HMRC returns that they have a record of the applicant, but no associated employments.

## Checklist

### DIFF
--- a/app/services/hmrc/parsed_response/validator.rb
+++ b/app/services/hmrc/parsed_response/validator.rb
@@ -103,16 +103,20 @@ module HMRC
         AlertManager.capture_message("HMRC Response is unacceptable (id: #{hmrc_response.id}) - #{errors.map(&:message).join(', ')}")
       end
 
-      def client_details_not_found
-        { "error" => "submitted client details could not be found in HMRC service" }
+      def client_details_not_found?
+        data&.include?({ "error" => "submitted client details could not be found in HMRC service" })
       end
 
-      def no_employments_found
+      def no_employments_found?
         errors == [error(:employments, "employments must be present")]
       end
 
+      def hmrc_not_found?
+        client_details_not_found? || no_employments_found?
+      end
+
       def errors_ignoreable?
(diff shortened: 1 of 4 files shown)


app/services/hmrc/parsed_response/validator.rb
```diff
@@ -107,8 +107,12 @@ def client_details_not_found
         { "error" => "submitted client details could not be found in HMRC service" }
       end
 
+      def no_employments_found
+        errors == [error(:employments, "employments must be present")]
+      end
+
       def errors_ignoreable?
-        errors.empty? || errors.map(&:attribute).include?(:use_case) || data&.include?(client_details_not_found)
+        errors.empty? || errors.map(&:attribute).include?(:use_case) || data&.include?(client_details_not_found) || no_employments_found
```
jsugarman, line 115: maybe a bit of method extraction for clarity as it is a bit on the cumbersome side now

```
def no_employments_found?
  errors == [error(:employments, "employments must be present")]
end

def use_case_invalid?
  errors.map(&:attribute).include?(:use_case)
end

def client_details_not_found?
  data&.include?(client_details_not_found)
end

....
def errors_ignoreable?
  errors.empty? || use_case_invalid? || client_details_not_found? || no_employments_found?
end
```


Also alternative method for employments as do we not want to handle the fact that there could be other errors  but that we don't want to alert if this is one of them? Or perhaps we are looking for situation when this is the only error

```
def no_employments_found?
  errors.map(&:message).include?("employments must be present")
end
```

kmahern: 👍 thanks @jsugarman I have refactored the errors_ignorable? method to hopefully make it more obvious or happy to pair if you think it still isn't clear. With regard to no_employments_found? I think we are looking for a situation where this is the only error (i.e. if there are other errors it should still alert) so have left this as is.

jsugarman (approved):
👍

PR #3966, opened 2022-06-16
Fix: Update the dashboard event handling for abandoning CCMS

colinbruce:
## What

This ensures that when we take the decision to abandon a CCMS submission the dashboard will update sooner. 

At the moment it waits until the next CCMS submission is kicked off before updating geckoboard or has to be manually triggered by a developer

## Checklist

### DIFF
--- a/spec/services/dashboard_event_handler_spec.rb
+++ b/spec/services/dashboard_event_handler_spec.rb
@@ -95,6 +95,19 @@ RSpec.describe DashboardEventHandler do
       end
     end
 
+    context "saved with_state abandoned" do
+      let(:state) { "abandoned" }
+
+      it "does not fire additional Application jobs" do
+        # one is fired from creating the LegalAidApplication required by the ccms_submission factory
+        expect { subject }.to have_enqueued_job(Dashboard::UpdaterJob).with("Applications").at_most(1).times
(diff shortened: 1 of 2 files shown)


spec/services/dashboard_event_handler_spec.rb
```diff
@@ -95,6 +95,19 @@
       end
     end
 
+    context "saved with_state abandoned" do
```
jsugarman, line 98: nit, can see this is used else where too but is it just a typo there and here `with_state` => `with state`

spec/services/dashboard_event_handler_spec.rb
```diff
@@ -95,6 +95,19 @@
...
+      let(:state) { "abandoned" }
+
+      it "does not fire additional Application jobs" do
+        # one is fired from creating the LegalAidApplication required by the ccms_submission factory
```
jsugarman, line 102: **not a blocker** but I am, after reading this, still unclear why this test and comment are similar to `initialised` but not failed/completed states.

willc-work (approved):
LGTM

jsugarman (approved):
✅ 👍 Just couple of questions/nits

PR #3975, opened 2022-06-20
AP-3240 Unemployed applicant with HMRC employment income

stephenrichards:
## Unemployed applicant with HMRC employment income

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3240)

In the above scenario, the provider should be told that's the case, shown the details and be given the opportunity to provide an explanation.  This has now been
fixed by:

 - making it follow the not_employed journey only if the applicant is not employed AND there are no eligible employment payments (i.e. within the transaction period).
 

Changes
- new method `#employment_payments` added to `LegalAidApplication` to return all employment_payments from all employments
- new method '#eligible_employment_payments` added to `LegalAidApplication` to return only those employment payments which fall within the transaction period
- `HMRC::StatusAnalyzer` amended to add new return status `:unexpected_employment_data` in this case
- The `ProviderCapital` flow updated to redirect to the correct page when `:unexpected_employment_data` returned from `HMRC::StatusAnalyser`

## Checklist

### DIFF
--- a/app/controllers/providers/client_completed_means_controller.rb
+++ b/app/controllers/providers/client_completed_means_controller.rb
@@ -12,7 +12,11 @@ module Providers
 
     def define_action_list
       @action_list = %w[sort_transactions dependants capital]
-      @action_list.prepend("review_employment") if @legal_aid_application.applicant.employed?
+      @action_list.prepend("review_employment") if employed_journey?
+    end
+
+    def employed_journey?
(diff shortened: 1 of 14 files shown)


colinbruce (changes requested):
I thought we'd agreed that WIP commit messages would be rebased, also the final commit misspells HMRC

colinbruce (approved):
There are a couple of typos, but otherwise LGTM :+1:

PR #4010, opened 2022-06-28
ap-3262: copy student finances controller to provider means namespace

kmahern:
## What

## Checklist


chris-groves:
@kmahern is it worth squashing some of the commits together?

app/controllers/citizens/student_finances_controller.rb
```diff
@@ -1,27 +1,38 @@
 module Citizens
   class StudentFinancesController < CitizenBaseController
     def show
-      @form = LegalAidApplications::StudentFinanceForm.new(model: legal_aid_application)
+      @form = ::StudentFinances::AnnualAmountForm.new(model: irregular_income)
```
chris-groves, line 4: Is this correct with nothing before the double colon? All the tests are passing but not seen this syntax before.

app/controllers/citizens/student_finances_controller.rb
```diff
@@ -1,27 +1,38 @@
...
-      @form = LegalAidApplications::StudentFinanceForm.new(form_params)
-
-      if @form.save
-        go_forward
+      @form = ::StudentFinances::AnnualAmountForm.new(form_params)
+      if student_finance
+        legal_aid_application.update!(student_finance:)
+        if student_finance == "false" || @form.save
+          return go_forward
+        end
```
chris-groves, line 13:
```suggestion
        legal_aid_application.update!(student_finance:) if student_finance
        return go_forward if student_finance == "false" || @form.save
```

spec/requests/citizens/student_finance_spec.rb
```diff
@@ -20,83 +20,92 @@
...
 
-      it "does not update the legal aid application record" do
-        expect(legal_aid_application.reload.student_finance).to be_nil
+      it "updates the legal aid application record" do
+        expect(legal_aid_application.reload.student_finance).to be false
       end
     end
 
-    context "when checking citizen answers" do
+    context "when tudent finance is nil" do
```
chris-groves, line 98:
```suggestion
    context "when student finance is nil" do
```

app/controllers/citizens/student_finances_controller.rb
kmahern: @chris-groves I don't think the suggested change will work, because the if statement logic needs the else clause - unfortunately the update action has become quite complex because of trying to combine the student_finances and annual_amounts update controller actions 🤔

kmahern:
👍 @chris-groves have squashed up the commits

chris-groves (approved):

PR #4038, opened 2022-07-05
Ap 3263 Move cash_outgoings to provider journey

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3263)
Copy controller to provider namespace
move views to shared
update show.html.erb files to point to the shared view
Add translations
Add tests
Added nocov to the provider version of the controller as the update path cannot be tested without the wider flow changes being done on another ticket.

## Checklist


jsugarman (approved):
Should we not be coping the `spec/requests/citizens/cash_outgoings_spec.rb` to `spec/requests/provider/means/cash_outgoings_spec.rb` and amending appropriately

maybe to be discussed but approving for now

spec/requests/providers/means/cash_outgoings_controller_spec.rb
```diff
@@ -0,0 +1,130 @@
+require "rails_helper"
+
+RSpec.describe Providers::Means::CashOutgoingsController, type: :request do
```
jsugarman, line 3:
```suggestion
RSpec.describe "Providers::Means::CashOutgoingsController", type: :request do
```

```suggestion
RSpec.describe "provider outgoings. request", type: :request do
```

willc-work: I have updated the file and renamed it to be consistent with the majority of other specs, however we do have instances where we use the model name in the spec and some where we have specs with 'controller' in the name. e.g. `full_employment_details_controller_spec.rb`. This appears to be due to govuk rubocop rules:
 RSpec/FilePath: Spec path should end with providers/means/cash_outgoings_controller*_spec.rb. (https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath)

PR #4045, opened 2022-07-07
AP-3206 change email address

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3206)

Added ability for the provider to go back and change the applicant email address once the email has already been sent. 

* Button to go back, change and resend email from the application submitted page
* Updated read only application page to resend email from there too
* Change to about_the_financial_assessments controller to allow email to be resent and state maintained
* Updated rspec tests
* Created feature test for changing email from various points

## Checklist


features/providers/change_email.feature
```diff
@@ -0,0 +1,74 @@
...
+    And I should be on a page showing "What is your client's employment status?"
+    And I select "None of the above"
+    When I click 'Save and continue'
+    Then I should be on a page with title "We need your client's bank statements from the last 3 months"
+    Then I choose 'Yes'
+    Then I click 'Save and continue'
+    Then I should be on a page showing 'What your client has to do'
+    Then I click link 'Continue'
+    Then I should be on a page showing "Enter your client's email address"
+    Then I fill 'email' with 'test@test.com'
```
colinbruce, line 16: I would change the first email e.g. `test@bad.com`
So when we change it at line 23 we expect to see the correct value at line 26 
The description says change email, but we don't actually change it! 😄

app/controllers/providers/about_the_financial_assessments_controller.rb
```diff
@@ -9,8 +9,10 @@ def show
     def update
       return continue_or_draft if draft_selected?
 
-      if ready_for_citizen_to_enter_financial_details?
-        legal_aid_application.await_applicant!
+      unless legal_aid_application.applicant_entering_means?
```
colinbruce, line 12: Can we have a chat about this? 
Multiple unless blocks seem a bit of a code smell!

colinbruce (changes requested):
Looking pretty good:+1:
A minor nitpick on the feature and a concern on the controller logic.
Lets catch up about the controller stuff 👍

features/providers/change_email.feature
RoseSAK: This is a good point! I changed it in the second test but not the first - have changed it now :)

app/controllers/providers/about_the_financial_assessments_controller.rb
RoseSAK: I forgot you would be off later in the day ...I have converted the nested unless block into a single statement which might be slightly better - but let's chat about it tomorrow :)

colinbruce (approved):
LGTM :+1:

PR #4046, opened 2022-07-07
AP-3269: remove Rails 7 deprecation warnings

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3269)

Remove legacy invocation of instance variables and replace
with a line in the view that creates an instance variable is
not passed in.

This allows the instance variable to be used in all the nested
partials that require them, a ticket will be raised to address
this by switching the nested partials to not require an instance
version of the required

## Checklist


kmahern (approved):
👍

PR #4047, opened 2022-07-07
add safe operator when checking for dwp_benefit type

willc-work:
## What

Bugfix
https://sentry.io/organizations/ministryofjustice/issues/3407948574/?project=5416054&referrer=slack

An error is being thrown when a user reaches uploaded evidence collection and they dont add mandatory evidence as the validation is looking for a benefit type, in this case the required evidence was for employment not benefits.

This change returns nil when a dwp_benefit override doesnt exist and subsequently the correct error is presented to the user

## Checklist


jsugarman (approved):
👍

PR #4048, opened 2022-07-08
AP-3295: Get and store client involvement type

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3295)

This PR enables calling LFA for the Client Involvement Types, storing the values in the database on a per-proceeding basis and then sending the stored values to CCMS

**Note:** The storing on the Proceeding type is a temporary solution until the designs are ready but allow us to show that a value (if chosen) _can_ be stored and submitted to CCMS

## Checklist


stephenrichards (approved):

colinbruce:
I've made some changes to the previous code - the DB no longer sets the filed s to prevent false values for `client_involvement_type_ccms_code` and `client_involvement_type_description` as these will be set after the proceeding is created.
The method of updating will still need to change on a future ticket when the design is finalised

kmahern (approved):
👍 lgtm

PR #4071, opened 2022-07-15
Ap 3265 bank statement result panel

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3265)
Add new result panel
Remove call to CFE for these cases

## Checklist

### DIFF
--- a/app/controllers/providers/means_summaries_controller.rb
+++ b/app/controllers/providers/means_summaries_controller.rb
@@ -6,9 +6,9 @@ module Providers
     end
 
     def update
-      if legal_aid_application.provider.bank_statement_upload_permissions?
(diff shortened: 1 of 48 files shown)


app/controllers/providers/means_summaries_controller.rb
```diff
@@ -6,6 +6,12 @@ def show
     end
 
     def update
+      if legal_aid_application.provider.bank_statement_upload_permissions?
```
jsugarman: can use `uploading_bank_statements?`

jsugarman (commented):
consider using the flow to redirect to a new blue panel page and controller. I have a branch i will push up

PR #4092, opened 2022-07-20
Ap 3338 create blank cfe result for bank_statement_upload journey

willc-work:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3338)

Add a new blank cfe result type
Create a cfe submission record indicating no cfe_calll was made
Crete a cfe_result record for use by subsequent pages
add tests for everything
Add new state to cfe state machine

## Checklist

### DIFF
--- a/app/controllers/providers/no_eligibility_assessments_controller.rb
+++ b/app/controllers/providers/no_eligibility_assessments_controller.rb
@@ -12,9 +12,7 @@ module Providers
     end
 
     def submission
-      # need to add a new aasm_state for cfe checks to indicate it was bypassed
       @submission ||= CFE::Submission.create!(legal_aid_application_id: legal_aid_application.id, aasm_state: "cfe_not_called")
-      # @submission ||= CFE::Submission.create!(legal_aid_application_id: legal_aid_application.id)
     end
 
     def write_cfe_result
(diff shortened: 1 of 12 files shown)


app/controllers/providers/no_eligibility_assessments_controller.rb
```diff
@@ -1,12 +1,27 @@
 module Providers
   class NoEligibilityAssessmentsController < ProviderBaseController
     def show
+      submission
+      write_cfe_result
       legal_aid_application.provider_enter_merits! unless legal_aid_application.provider_entering_merits?
```
stephenrichards, line 6: I don't think you need to call `#submission` here -you can just refer to `submission.id` instead of `@submission.id` in the `#write_cfe_result` method and it will do it there.

stephenrichards (changes requested):

stephenrichards:
One tiny stylistic change requested

app/controllers/providers/no_eligibility_assessments_controller.rb
willc-work: change made

RoseSAK (approved):
LGTM!

PR #4095, opened 2022-07-21
AP-3307 Handle CCMS returning Number Not Matched

stephenrichards:
## Handle CCMS returning Number Not Matched

During the "CCMS dance" to send data over to CCMS, we send name, dob, nino to ask CCMS if the applicant already exists on their side.  If it does, it returns a applicant ref which we use, if it doesn't we create a new applicant on ccms.

There has been a case where CCMS has returned a match, but it turns out that the match wasn't an exact match (different NINO, different first name), with a match level indicator of `Number Not Matched`

This PR examines the mactch level indicator, and if `Number Not Matched` returns a nil ccms_applicant reference, the same as if there had been no match at all.  This will cause a new applicant with the right details to be created.

## What

## Checklist


kmahern (approved):
lgtm (I think the failing feature test is just a flaky test) 👍

PR #4117, opened 2022-07-27
AP-3337 display error on field when file is not uploaded

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3337)

When the user does not upload a file on the bank statement upload page, they get an error. This error now displays on the dropzone as well as in the error summary, as expected.

## Checklist


willc-work (approved):
looks great. I see you have also fixed that issue with student finance response not being populated when you return to the page too 👍

PR #4139, opened 2022-08-02
[AP-3368] Remove cash payments once the category has been removed

jsugarman:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3368)

Remove Cash transactions for both income and outgoings when the corresponding payment category is removed

At the moment, if you have selected a payment category and entered cash payments for that category, if you return to the identify types of payments page and remove that category, the cash transactions are not removed and will still show up on the cash payments page and on the CYA page.

All cash transactions need to be removed when the payment category is removed.

## Checklist

### DIFF
--- a/app/controllers/providers/means/identify_types_of_incomes_controller.rb
+++ b/app/controllers/providers/means/identify_types_of_incomes_controller.rb
@@ -6,9 +6,20 @@ module Providers
       end
 
       def update
-        return continue_or_draft if none_selected? || transactions_added || draft_selected?
+        synchronize_credit_transaction_types_and_cash_transactions
 
-        remove_existing_transaction_types
+        if none_selected?
+          legal_aid_application.update!(no_credit_transaction_types_selected: true)
+
+          return continue_or_draft
+        elsif transaction_types_selected?
+          legal_aid_application.update!(no_credit_transaction_types_selected: false)
+
+          return continue_or_draft
+        end
+
+        return continue_or_draft if draft_selected?
+
         legal_aid_application.errors.add :transaction_type_ids, t(".none_selected")
         render :show
       end
@@ -23,27 +34,34 @@ module Providers
         @transaction_types ||= TransactionType.credits.find_with_children(legal_aid_application_params[:transaction_type_ids])
       end
 
+      def transaction_types_selected?
+        transaction_types.present?
+      end
+
       def none_selected?
-        return unless params[:legal_aid_application][:none_selected] == "true"
-
-        LegalAidApplication.transaction do
-          remove_existing_transaction_types
-          legal_aid_application.update!(no_credit_transaction_types_selected: true)
-        end
+        params[:legal_aid_application][:none_selected] == "true"
       end
 
-      def transactions_added
-        return if transaction_types.empty?
+      def synchronize_credit_transaction_types_and_cash_transactions
+        existing_credit_tt_ids = legal_aid_application.legal_aid_application_transaction_types.credits.map(&:transaction_type_id)
 
-        LegalAidApplication.transaction do
-          remove_existing_transaction_types
-          legal_aid_application.update!(no_credit_transaction_types_selected: false)
-          legal_aid_application.transaction_types << transaction_types
+        keep = transaction_types.each_with_object([]) do |form_tt, arr|
+          add_transaction_type(form_tt) if existing_credit_tt_ids.exclude?(form_tt.id)
+          arr.append(form_tt.id)
         end
+
+        destroy_all_credit_transaction_types_and_cash_transactions(except: keep)
       end
 
-      def remove_existing_transaction_types
-        legal_aid_application.legal_aid_application_transaction_types.credits.destroy_all
+      def add_transaction_type(transaction_type)
+        legal_aid_application.transaction_types << transaction_type
+      end
+
+      def destroy_all_credit_transaction_types_and_cash_transactions(except:)
+        LegalAidApplication.transaction do
+          legal_aid_application.legal_aid_application_transaction_types.credits.where.not(transaction_type_id: except).destroy_all
+          legal_aid_application.cash_transactions.credits.where.not(transaction_type_id: except).destroy_all
+        end
       end
     end
   end
--- a/app/controllers/providers/means/identify_types_of_incomes_controller.rb
+++ b/app/controllers/providers/means/identify_types_of_incomes_controller.rb
@@ -6,7 +6,7 @@ module Providers
       end
 
       def update
-        synchronize_credit_transaction_types_and_cash_transactions
+        synchronize_credit_transaction_types
 
         if none_selected?
           legal_aid_application.update!(no_credit_transaction_types_selected: true)
@@ -42,7 +42,7 @@ module Providers
         params[:legal_aid_application][:none_selected] == "true"
       end
 
-      def synchronize_credit_transaction_types_and_cash_transactions
+      def synchronize_credit_transaction_types
         existing_credit_tt_ids = legal_aid_application.legal_aid_application_transaction_types.credits.map(&:transaction_type_id)
 
         keep = transaction_types.each_with_object([]) do |form_tt, arr|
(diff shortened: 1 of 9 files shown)


spec/factories/legal_aid_applications.rb
```diff
@@ -816,6 +816,18 @@
       end
     end
 
+    # TODO: prefer in spec definition?!
```
jsugarman, line 819: remove

app/models/legal_aid_application_transaction_type.rb
```diff
@@ -4,4 +4,16 @@ class LegalAidApplicationTransactionType < ApplicationRecord
 
   scope :credits, -> { includes(:transaction_type).where(transaction_types: { operation: :credit }) }
   scope :debits, -> { includes(:transaction_type).where(transaction_types: { operation: :debit }) }
+
+  # TODO: after_destroy instead??
```
jsugarman, line 8: remove

jsugarman:
@RoseSAK there is a question related to removal of categorised transactions generally. _If you delete a transaction type/category are bank transactions from truelayer path categorised as that type also being deleted?_ This goes beyond the scope of this PR/Ticket but probably needs looking into as well, as impacts CFE results i imagine.

spec/requests/providers/means/identify_types_of_incomes_controller_spec.rb
```diff
@@ -118,45 +117,99 @@
...
+        laa.cash_transactions.create!(transaction_type_id: rent_or_mortgage_debit.id, amount: 301, month_number: 1, transaction_date: Time.zone.now.to_date)
+        laa.cash_transactions.create!(transaction_type_id: rent_or_mortgage_debit.id, amount: 302, month_number: 2, transaction_date: 1.month.ago)
+        laa.cash_transactions.create!(transaction_type_id: rent_or_mortgage_debit.id, amount: 303, month_number: 3, transaction_date: 2.months.ago)
+        laa
+      end
+
+      # xit "removes the deselected credit transaction types" do
+      #   expect { request }.to change(legal_aid_application.legal_aid_application_transaction_types.credits, :count).by(-1)
+      # end
+      # OR
```
jsugarman, line 187: remove

spec/requests/providers/means/identify_types_of_incomes_controller_spec.rb
```diff
@@ -118,45 +117,99 @@
...
+      end
+
+      it "does not remove existing debit transaction types" do
+        expect { request }.not_to change(legal_aid_application.legal_aid_application_transaction_types.debits, :count)
+      end
+
+      # xit "removes deselected credit cash transactions" do
+      #   expect { request }.to change(legal_aid_application.cash_transactions.credits, :count).by(-1)
+      # end
+      # OR
```
jsugarman, line 201: remove

app/controllers/providers/means/identify_types_of_incomes_controller.rb
```diff
@@ -6,9 +6,20 @@ def show
...
+          legal_aid_application.update!(no_credit_transaction_types_selected: true)
+
+          return continue_or_draft
+        elsif transaction_types_selected?
+          legal_aid_application.update!(no_credit_transaction_types_selected: false)
+
+          return continue_or_draft
+        end
+
+        return continue_or_draft if draft_selected?
```
cpjmcquillan, line 21: This method doesn't feel particularly readable to me, and it feels quite long —especially for a controller action.

I think ideally it would be nice if the controller action looked something like this

```ruby
module Providers
  module Means
    class IdentifyTypesOfIncomesController < ProviderBaseController
      def update
        if income_types.save
          return continue_or_draft
        else
          render :show, status: :unprocessable_entity
        end
      end
      
      def income_types
        SomeWellNamedClass.new(...)
      end
    end
  end
end
```

with the complex logic refactored to a separate class.

It would simplify the request specs too, since the class would be covered by unit tests.

If this sounds worthwhile to you, let's schedule some time to pair on it!

app/models/legal_aid_application_transaction_type.rb
```diff
@@ -4,4 +4,15 @@ class LegalAidApplicationTransactionType < ApplicationRecord
 
   scope :credits, -> { includes(:transaction_type).where(transaction_types: { operation: :credit }) }
   scope :debits, -> { includes(:transaction_type).where(transaction_types: { operation: :debit }) }
+
+  after_commit :cascade_delete_cash_transactions, on: :destroy
```
cpjmcquillan, line 8: I'd like to pair to better understand the relationships between legal aid applications, transactions, and transaction types.

It feels strange to define a custom `after_commit`, rather than relying on `dependent: :destroy` relationship options and `on_delete: :cascade` database constraints.

I wonder if there's some refactoring we could do to simplify the relationships, but it might not be straightforward.

spec/models/legal_aid_application_transaction_type_spec.rb
```diff
@@ -52,4 +52,49 @@
...
+      let(:action) { instance.destroy! }
+
+      it "does not call cascade_delete_cash_transactions" do
+        action
+        expect(instance).to have_received(:cascade_delete_cash_transactions)
+      end
+
+      it "cascades delete to cash transactions of that type" do
+        expect { action }.to change(legal_aid_application.cash_transactions, :count).by(-1)
+      end
```
cpjmcquillan, line 84: I wonder if we need both these tests. Perhaps we should focus on testing the behaviour, and delete the other test?

Something like this would probably be sufficient

```ruby
describe ".after_commit" do
  context "when destroying the transaction type" do
    it "destroys that type's cash transactions" do
      setup
      
      destroy_the_transaction_type

      assert_the_correct_transactions_are_destroyed
    end
  end 

  context "when updating the transaction type" do
    it "does not destroy that type's cash transactions" do
      setup
      
      destroy the transaction_type
      
      assert_the_transactions_are_not_destroyed
    end
  end
end
```

What do you think?

spec/requests/providers/means/identify_types_of_outgoings_controller_spec.rb
```diff
@@ -53,20 +53,24 @@
...
+      end
 
-    it "returns http success" do
-      request
-      expect(response).to have_http_status(:ok)
+      it "displays an error" do
+        request
+        expect(response.body).to match("govuk-error-summary")
+        expect(unescaped_response_body).to match(I18n.t("providers.means.identify_types_of_outgoings.update.none_selected"))
+        expect(unescaped_response_body).not_to include("translation missing")
```
cpjmcquillan, line 67: If we have `I18n` configured to raise errors on missing translations, do we need this expectation?

cpjmcquillan (commented):
Left a few comments.

If my suggestion doesn't sound worthwhile to you, let me know and I will review the rest of the PR and approve 👍

app/models/legal_aid_application_transaction_type.rb
jsugarman: there is no join table via which we can cascade the delete using a `dependant: :destroy` that I can see. Happy to discuss as this was not something I was happy with either.

spec/requests/providers/means/identify_types_of_outgoings_controller_spec.rb
jsugarman: Not something I changed, just moved i think, but happy to discuss. development will raise when you hit such a page but use of interpolation in some views and named scopes means I18n-missing gem won't pick up all missing translations

app/controllers/providers/means/identify_types_of_incomes_controller.rb
jsugarman: Lets discuss later but this was a refactor of existing, rather confusing, controller logic. It also changes the logic which was deleting all `transaction_types` and re-adding them using supplied params. This would not have been desirable with the cascade-delete logic.

RoseSAK (approved):
LGTM 👍

app/controllers/providers/means/identify_types_of_incomes_controller.rb
jsugarman: Agreed with @cpjmcquillan to make separate PR to refactor further.

PR #4147, opened 2022-08-04
AP-3363 Filter proceeding types

stephenrichards:
## Filter proceeding types

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3360)

LFA will soon start returning a list of all proceeding types in Domestic abuse and Section 8, but we don't want to show the full set of section 8 proceedings to providers that haven't been enabled for it.  This PR examines the permissions of the provder, and if. not enabled for full section 8, filters out those proceedings so that there are not returned as part of the proceeding type search.

- altered the call to `LegalFramework::ProceedingTypes::All` to take a new parameter `provider`
- altered `LegalFramework::ProceedingTypes::All` to filter out all proceedings marked `full_s8_only` as true unless the provider has full section 8 permisstions
- changed the spec for `LegalFramework::ProceedingTypes::All` to use a stubbed response rather than VCR
- re-recorded other VCR cassettes that called that LFA because the old ones didn't have the new `full_s8-only` attribute

## Checklist


db/schema.rb
```diff
@@ -20,7 +20,7 @@
     t.uuid "record_id", null: false
```
kmahern, line 20: I don't think there are any schema changes so do we need to commit this file as part of this pr? (we seem to keep cycling between adding and removing the precision: nils)

kmahern (approved):
@stephenrichards lgtm, I have made a comment about the schema.rb file but not a blocker.

PR #4183, opened 2022-08-16
Fix: Date display

colinbruce:
## What

The update to rails 7 removed the automatic conversion of dates to_s.  This meant that the default date format was not being applied and was falling back to YYYY-MM-DD

The hotfix applied means that we do not have to manually update each date with .to_fs any more

## Checklist

### DIFF
--- a/config/initializers/date_formats.rb
+++ b/config/initializers/date_formats.rb
@@ -1,21 +1,10 @@
-# The below hotfix is taken from https://stackoverflow.com/a/71328079
-# Overrides the rails 7 removal of to_s and allows our default
-# format to be applied rather than having to find each date
-# and manually add `.to_fs` to the end of it
+# The below hotfix overrides the rails 7 removal of to_s and
+# allows our default format to be applied rather than having
+# to find each date and manually add `.to_fs` to the end of it
 require "date"
 
 class Date
-  def to_s(format = :default)
-    if (formatter = DATE_FORMATS[format])
-      if formatter.respond_to?(:call)
-        formatter.call(self).to_s
-      else
-        strftime(formatter)
-      end
-    else
-      to_default_s
-    end
-  end
+  alias_method :to_s, :to_fs
 end
 
 # Note - if using `l(date)` in an erb, format is defined in locale at en.date.formats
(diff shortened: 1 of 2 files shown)


config/initializers/date_formats.rb
```diff
@@ -1,3 +1,23 @@
...
+    if (formatter = DATE_FORMATS[format])
+      if formatter.respond_to?(:call)
+        formatter.call(self).to_s
+      else
+        strftime(formatter)
+      end
+    else
+      to_default_s
+    end
+  end
```
cpjmcquillan, line 18: I understand it's a chunky piece of work to try and find all the dates and add `.to_fs` to them, but it doesn't feel great having to maintain this code snippet either.

The code on Stack Overflow looks like it was lifted wholesale from [`active_support/core_ext/date/conversions.rb` ](https://github.com/rails/rails/blob/d6973d50d01fe55e5c65368b92160708fb60a3da/activesupport/lib/active_support/core_ext/date/conversions.rb#L47-L57).

How do you feel about defining an alias_method between `:to_s` and `:to_fs` instead?

I haven't tried it to see if that will work, but if it does it might be a simpler (and more palatable) change.

What do you think?

cpjmcquillan (approved):
I added a comment for discussion, but ultimately I'm ok with whatever you decide.

config/initializers/date_formats.rb
colinbruce: Good idea :+1:
I'll add some tests to ensure dates are shown as expected - we weren't alerted as during the Rails 7 update 😞 
Then I'll try dropping this code and replacing with an `alias_method`

config/initializers/date_formats.rb
```diff
@@ -1,21 +1,10 @@
-# The below hotfix is taken from https://stackoverflow.com/a/71328079
-# Overrides the rails 7 removal of to_s and allows our default
-# format to be applied rather than having to find each date
-# and manually add `.to_fs` to the end of it
+# The below hotfix overrides the rails 7 removal of to_s and
+# allows our default format to be applied rather than having
+# to find each date and manually add `.to_fs` to the end of it
```
cpjmcquillan, line 3: Is it worth adding a link to the release notes? I don't mind either way, but it might be useful for anyone curious.

https://guides.rubyonrails.org/7_0_release_notes.html#active-support-deprecations

PR #4188, opened 2022-08-18
AP-3388: Add feature flag for next loop stage

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3388)

Add a feature flag for the next stage of the proceeding loop
There may be a third, `full-loop` stage so this will either become the full loop by default or, be an interstitial step between the mini and full loops

## Checklist

### DIFF
--- a/app/controllers/admin/settings_controller.rb
+++ b/app/controllers/admin/settings_controller.rb
@@ -23,7 +23,8 @@ module Admin
                                       :allow_welsh_translation,
                                       :enable_ccms_submission,
                                       :enable_employed_journey,
-                                      :enable_mini_loop)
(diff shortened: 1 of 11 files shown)


jsugarman (approved):
**NOT A BLOCKER** but just thinking naming wise if these these are to exist until January next year(?) should we be thinking of more descriptive names - enable_sec8_mini_loop, enable_sec8_loop for example

colinbruce:
Discussed with Joel, the mini-loop should be _very_ short-lived and there are plans to make the loop itself releasable before January, so 🤞 they should not be hanging around for too long

PR #4195, opened 2022-08-19
Move the credits and debits scope specs to correct class

jsugarman:
## What
Moved scope specs to correct model class spec and replaced with specs that 
excercise the owning classes scopes.

Both transaction_types and legal_aid_application_transaction_types
have `.credits` and `.debits` scopes. The transaction_types model's
scopes were being tested in the legal_aid_application_transaction_types
specs and there were no specs excercising the expected behaviour of
the transaction_types model's scopes. I have added the later and moved
and rejigged the former.

## Checklist


colinbruce (approved):
LGTM 👍

PR #4197, opened 2022-08-19
[AP-3423] - Add GovukComponents gem

cpjmcquillan:
Pretty much every view in the application leans primarily on GOV.UK Design
System components.

This introduces the [govuk-components] gem to help us manage this complexity.

Rather than relying on copy-paste to ensure we render the design system
components accurately, we can render these resuable components. They are well
tested, customisable, and should reflect the design system accurately.

[govuk-components]: https://github.com/DFE-Digital/govuk-components

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3423)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -99,6 +99,9 @@ gem "secure_headers"
 # DFE formbuilder
 gem "govuk_design_system_formbuilder", ">= 3.0.1"
 
+# DFE ViewComponent library
+gem "govuk-components"
+
 group :development, :test do
   gem "awesome_print", "~> 1.9.2"
   gem "byebug", platforms: %i[mri mingw x64_mingw]
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -49,19 +49,13 @@
     <a href="#main-content" class="govuk-skip-link" data-module="govuk-skip-link">Skip to main content</a>
 
     <%= render partial: 'layouts/header' %>
+
     <div class="govuk-width-container">
-      <div class="govuk-phase-banner no-print">
-        <p class="govuk-phase-banner__content">
-          <strong class="govuk-tag govuk-phase-banner__content__tag ">
-            <%= t 'layouts.application.header.phase' %>
-          </strong>
-          <span class="govuk-phase-banner__text">
-            <%= t 'layouts.application.header.phase_banner_text_1' %>
-            <%= link_to_accessible t('layouts.application.header.feedback'), new_feedback_path, class: 'govuk-link' %>
-            <%= t 'layouts.application.header.phase_banner_text_2' %>
-          </span>
-        </p>
-      </div>
+      <%= govuk_phase_banner(
+        tag: {text: t(".header.phase")},
+        text: t(".header.phase_banner_html", link: govuk_link_to(t(".header.feedback"), new_feedback_path)),
+        classes: "no-print"
+      ) %>
 
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-one-half show-empty">
--- a/config/locales/cy/layouts.yml
+++ b/config/locales/cy/layouts.yml
@@ -27,8 +27,7 @@ cy:
         gov_uk_link_title: egapemoh KU.VOG eht ot oG
         feedback: kcabdeef
         phase: ateb
-        phase_banner_text_1: ruoy – ecivres wen a si sihT
-        phase_banner_text_2: ".ti evorpmi ot su pleh lliw"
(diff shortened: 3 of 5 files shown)


config/locales/cy/layouts.yml
```diff
@@ -27,8 +27,7 @@ cy:
         gov_uk_link_title: egapemoh KU.VOG eht ot oG
         feedback: kcabdeef
         phase: ateb
-        phase_banner_text_1: ruoy – ecivres wen a si sihT
-        phase_banner_text_2: ".ti evorpmi ot su pleh lliw"
+        phase_banner_html: This is a new service – your %{link} will help us to improve it.
```
cpjmcquillan, line 30: I didn't see the value in reversing this.

app/views/layouts/application.html.erb
```diff
@@ -49,19 +49,13 @@
...
-          <span class="govuk-phase-banner__text">
-            <%= t 'layouts.application.header.phase_banner_text_1' %>
-            <%= link_to_accessible t('layouts.application.header.feedback'), new_feedback_path, class: 'govuk-link' %>
-            <%= t 'layouts.application.header.phase_banner_text_2' %>
-          </span>
-        </p>
-      </div>
+      <%= govuk_phase_banner(
+        tag: {text: t(".header.phase")},
+        text: t(".header.phase_banner_html", link: govuk_link_to(t(".header.feedback"), new_feedback_path)),
```
cpjmcquillan, line 56: The only tangible change here is an omitted `aria-label="feedback"`, which seemed pretty redundant since the link text itself `feedback`.

cpjmcquillan:
Let me know what you think about this.

The gem is actively maintained and used across several services in DfE-Digital.

It will take a while to migrate entirely, but it's something that can be done incrementally and alongside other work. I would be tempted to migrate one component at a time, but it might make more sense from a UAT perspective to migrate by template. 

Personally I think it's a no brainer in terms of shifting complexity and load outside the scope of our application. I'd expect it to reduce the maintenance load of the application and improve developer happiness, but I'm keen to hear other people's opinions too.

Gemfile
```diff
@@ -99,6 +99,9 @@ gem "secure_headers"
 # DFE formbuilder
 gem "govuk_design_system_formbuilder", ">= 3.0.1"
 
+# DFE ViewComponent library
+gem "govuk-components"
```
jsugarman, line 103: might be worth pessimistically versioning?!

app/views/layouts/application.html.erb
jsugarman: @RoseSAK should probably check this. Can the components not have arbitrary html attributes added to them in some way?

jsugarman (commented):
I suggested introducing this gem during a developer time meeting, [card here](https://trello.com/c/fmwwxoDL/114-shall-we-implement-govuk-components-or-roll-our-own), so am happy for this to go in. we should perhaps consider writing up tickets to move "bits" to using it now or wait to discuss in another developer time meeting, particularly with @RoseSAK  - who was not present during the initial discussion

cpjmcquillan:
> I suggested introducing this gem during a developer time meeting, [card here](https://trello.com/c/fmwwxoDL/114-shall-we-implement-govuk-components-or-roll-our-own)

Thanks. I'm not sure who manages access to that Trello board, but I've requested access.

> we should perhaps consider writing up tickets to move "bits" to using it now or wait to discuss in another developer time meeting, particularly with @RoseSAK - who was not present during the initial discussion

Agreed, the acceptance criteria of the task urges us to make a decision one way or the other:

> "Done when a decision has been taken whether to proceed (subsequent tasks added), or not (pull request closed)."

Gemfile
cpjmcquillan: I don't have a strong opinion either way. 

It's semantically versioned, and I'd expect us to be confident we aren't affected by breaking changes in a major upgrade of any gem, but I see no harm in adding it here. 👍

app/views/layouts/application.html.erb
RoseSAK: Agree this can be removed 👍

RoseSAK (approved):
I think this is a great idea, definitely worth doing. Agree with your suggestion that it should be done incrementally. I think one component at a time would be good, except perhaps components that appear in a large amount of pages, in which case perhaps we could split by journey/section. 

Would be good to create an epic in JIRA for this work, and perhaps create some initial tickets for components that we'd like to migrate first. Once this is done, happy for this PR to be merged and this ticket closed off. 

Would still be good to discuss this at dev time as well, it looks like a ticket has been added to the Trello already 👍

PR #4202, opened 2022-08-22
Remove custom rules from Rubocop config

cpjmcquillan:
Before alphagov/rubocop-govuk was adopted, we added some custom configuration
for some of the rules.

These rules have been disabled in alphagov/rubocop-govuk, so we can remove this
config.


jsugarman (approved):
👍 LGTM

PR #4203, opened 2022-08-22
Use bundler-audit to audit gems

cpjmcquillan:
This adds the `bundler-audit` gem to automatically check installed gems in
Gemfile.lock for security advisories, and prompt us to upgrade to patched
versions as soon as possible.

This uses GitHub actions to run the check as part of CI, and it is triggered by
default pull_request events (opened, synchronized, reopened) against the main
branch. This seemed to be the trigger used across the application, so I have
adopted the same approach.

### DIFF
--- a/.github/workflows/bundle-audit.yml
+++ b/.github/workflows/bundle-audit.yml
@@ -1,0 +1,15 @@
+name: Bundle audit
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  bundle-audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true
+      - run: bundle exec bundle-audit check --update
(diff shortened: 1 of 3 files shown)


jsugarman (commented):
Aside from the licensing issue that snyk raises I am happy to use this if other devs think it is worth it.

Do you know what it provides above what we already get from dependabot and snyk?

colinbruce:
+1 to Joel's question

cpjmcquillan:
I wasn't aware we used Snyk, and I'm not quite clear what it does (I've requested access but I'm not sure where that request goes).

Dependabot notifies and opens a PR if it can after a certain period of time, but dependabot upgrades aren't always timely and sometimes there aren't a top priority for a team.

`bundler-audit` is useful because it runs automatically as part of CI. Using that approach typically highlights vulnerabilities and necessitates fixes much sooner.

If we're comfortable using the Snyk SaaS to solve this problem, then we can close this PR.

cpjmcquillan:
I'm closing this because it's been up for a while now with little engagement.

Ultimately I think it's worthwhile, but it probably isn't _essential_ if we're using Snyk.

Plus, I don't have the energy to make the case that we can use gem despite Snyk complaining about licensing.

PR #4217, opened 2022-08-24
Fix flaky rack-attack spec

cpjmcquillan:
Before, if the offending spec took longer than 10 seconds to make the first two
requests, the third request would be successful and the test would fail.

This is something I'd seen a few times, most recently [in this run].

[in this run]: https://app.circleci.com/pipelines/github/ministryofjustice/laa-apply-for-legal-aid/17300/workflows/6b49be0f-8c15-4eb0-98c3-d94ec84deeee/jobs/90881

This freezes time so the test will always pass, no matter how slow our CI test
suite is.

### DIFF
--- a/spec/requests/feedbacks_spec.rb
+++ b/spec/requests/feedbacks_spec.rb
@@ -211,12 +211,16 @@ RSpec.describe "FeedbacksController", type: :request do
 
       describe "when three feedbacks are submitted in quick succession" do
         it "returns a 429 status for the final submission" do
+          freeze_time
+
           post feedback_index_path, params:, headers: { "HTTP_REFERER" => originating_page }
           expect(response).to have_http_status(:ok)
           expect(Feedback.count).to eq 1
+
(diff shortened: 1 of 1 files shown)


spec/requests/feedbacks_spec.rb
```diff
@@ -211,12 +211,18 @@
...
+
           post feedback_index_path, params:, headers: { "HTTP_REFERER" => originating_page }
           expect(response).to have_http_status(:ok)
           expect(Feedback.count).to eq 1
+
           post feedback_index_path, params:, headers: { "HTTP_REFERER" => originating_page }
           expect(response).to have_http_status(:ok)
           expect(Feedback.count).to eq 2
+
+          travel_to 11.seconds.from_now
```
cpjmcquillan, line 224: I don't like magic numbers, but I didn't think it was worth introducing a constant to the rack-attack initialiser for this test alone.

If you do, feel free to add it and merge this in my absence.

willc-work (approved):
LGTM

PR #4233, opened 2022-08-30
Remove `enable_employed_journey` column from settings

cpjmcquillan:
Before, in #4216, the `enable_employed_journey` setting was removed and the
associated column was ignored.

Now that change has been deployed successfully, we can safely remove the column
from the database table.

This removes the column from the settings table and removes the
`.ignored_columns` method defined in the Setting model.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-2637)


jsugarman (approved):
👍

PR #4234, opened 2022-08-30
Remove redundant blazer db_url config

cpjmcquillan:
Once upon a time in #933, we added the [blazer] gem.

Several years later, we removed the [blazer] gem in #2025.

We forgot to remove the `db_url` variable from the config though.

This removes the redundant config.

[blazer]: https://github.com/ankane/blazer


colinbruce (approved):
Nice find :+1:

PR #4251, opened 2022-09-06
Add enhanced bank upload journey permissions

cpjmcquillan:
In #4218 a new enhanced bank upload journey feature flag was added to the
Settings.

This adds a new role to permissions that will allow users of permitted firms to
use the enhanced bank upload journey.

### DIFF
--- a/db/seeds/permissions.rb
+++ b/db/seeds/permissions.rb
@@ -5,6 +5,7 @@ class PermissionsPopulator
     "application.non_passported.employment.*" => "Can create, edit, delete employment applications",
     "application.non_passported.bank_statement_upload.*" => "Can upload bank statements",
     "application.full_section_8.*" => "Can use full set of section 8 proceedings",
+    "application.non_passported.enhanced_bank_upload.*" => "Can use enhanced bank upload journey",
   }.freeze
 
   def self.run
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -57,6 +57,10 @@ class Provider < ApplicationRecord
     user_permissions.map(&:role).include?("application.full_section_8.*")
   end
 
+  def enhanced_bank_upload_permissions?
+    user_permissions.map(&:role).include?("application.non_passported.enhanced_bank_upload.*")
+  end
+
   def ccms_apply_role?
     return true if Rails.configuration.x.laa_portal.mock_saml == "true"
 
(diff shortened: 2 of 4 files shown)


cpjmcquillan:
For some permissions there are rake tasks that I suppose are intended to be ephemeral, although they still exist.

Should we:

- add a similar task, run it in non-production environments (and remember to delete it after)
- or, run a data migration in non-production environments
- or, just add the permissions in the console manually to each environment?

I'd be inclined to go with the last option, because it's fairly trivial and should be something we only do once.

However, I'm open to suggestions and keen to hear people's feedback.

db/seeds/permissions.rb
```diff
@@ -5,6 +5,7 @@ class PermissionsPopulator
     "application.non_passported.employment.*" => "Can create, edit, delete employment applications",
     "application.non_passported.bank_statement_upload.*" => "Can upload bank statements",
     "application.full_section_8.*" => "Can use full set of section 8 proceedings",
+    "application.non_passported.enhanced_bank_upload.*" => "Can use enhanced bank upload journey",
```
jsugarman, line 8: if we do implement such a role i would be tempted to make it a sub role

```suggestion
    "application.non_passported.bank_statement_upload.enhanced.part_1*" => "Can use enhanced bank upload journey",
```
or similar. 

My thinking is they relate to bank_statement_upload only and there are further enhancements to be consider in the future potentially

jsugarman: 🤔 or should we be using a flag instead?!

cpjmcquillan: We already have the enhanced bank upload flag, and the feature sits behind that.

The more I'm thinking about it, the more it feels like the feature flag will be sufficient on its own.

We probably need a product steer on this, but I'd imagine when we turn this feature on, we will want to turn it on for all providers at firms with existing bank upload permissions. 

@RoseSAK what do you think?

cpjmcquillan: If we do want to use firm permissions though, I agree that it makes sense for this role to be sub-role of `application.non_passported.bank_statement_upload`.

cpjmcquillan:
I'm going to close this, because I don't think it's necessary (given existing permissions and the feature flag).

If we need to re-evaluate we can revisit this.

PR #4258, opened 2022-09-07
[AP-3445] Providers can submit more accurate income data

cpjmcquillan:
Before, when providers with employment and bank upload permissions submitted
income data for their client, they could only specify which transactions their
client received.

This data is not granular enough, and makes it impossible to obtain a financial
eligibility assessment (CFE) result for the client.

As part of ongoing work to iterate on the bank upload journey, providers are now
able to tell us how much (amount), and how often (frequency) their client
receives these income payments.

Next steps will include:

- ensuring providers can select the cash transactions
- ensuring cash transactions are updated when regular payments are updated
- replicating this functionality for the outgoings page
- updating the frequencies to reflect the designs
- allowing providers to save and come back later

This user journey is restricted by feature flag at the flow and controller level.

## Video

https://user-images.githubusercontent.com/25187547/189706162-a35dbbde-3424-472c-af55-1514e2c89ae7.mp4

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3445)

### DIFF
--- a/app/controllers/providers/means/identify_types_of_incomes_controller.rb
+++ b/app/controllers/providers/means/identify_types_of_incomes_controller.rb
@@ -1,6 +1,8 @@
 module Providers
   module Means
     class IdentifyTypesOfIncomesController < ProviderBaseController
+      before_action :redirect_if_enhanced_bank_upload_enabled
+
       def show
         @none_selected = legal_aid_application.no_credit_transaction_types_selected?
       end
@@ -64,6 +66,12 @@ module Providers
             .credits
             .where.not(transaction_type_id: except)
             .destroy_all
+        end
+      end
+
+      def redirect_if_enhanced_bank_upload_enabled
+        if Setting.enhanced_bank_upload?
+          redirect_to providers_legal_aid_application_means_regular_incomes_path(legal_aid_application)
         end
       end
     end
--- a/app/forms/providers/means/regular_income_form.rb
+++ b/app/forms/providers/means/regular_income_form.rb
@@ -1,0 +1,165 @@
+module Providers
+  module Means
+    class RegularIncomeForm
+      include ActiveModel::Model
+
+      INCOME_TYPES = %w[
+        benefits
+        friends_or_family
+        maintenance_in
+        property_or_lodger
+        student_loan
+        pension
+      ].freeze
+
+      attr_reader :transaction_type_ids, :legal_aid_application
+
+      INCOME_TYPES.each do |income_type|
+        attr_accessor "#{income_type}_amount".to_sym,
+                      "#{income_type}_frequency".to_sym
+      end
+
+      validates :transaction_type_ids, presence: true, unless: :none_selected?
+      validate :all_regular_transactions_valid
+
+      def initialize(params = {})
+        @none_selected = none_selected.in?(params["transaction_type_ids"] || [])
+        @legal_aid_application = params.delete(:legal_aid_application)
+        @transaction_type_ids = params["transaction_type_ids"] || @legal_aid_application.transaction_type_ids
+
+        regular_transactions.each do |transaction|
+          transaction_type = transaction.transaction_type
+          public_send("#{transaction_type.name}_amount=", transaction.amount)
+          public_send("#{transaction_type.name}_frequency=", transaction.frequency)
+        end
+
+        super
+      end
+
+      def save
+        return false unless valid?
+
+        ApplicationRecord.transaction do
+          destroy_transactions!
+
+          regular_transactions.each(&:save!)
+
+          build_legal_aid_application_transaction_types
+          legal_aid_application.assign_attributes(
+            no_credit_transaction_types_selected: none_selected?,
+          )
+          legal_aid_application.save!
+        end
+
+        true
+      end
+
+      def transaction_type_ids=(ids)
+        @transaction_type_ids = if none_selected?
+                                  []
+                                else
+                                  ids.compact_blank
+                                end
+      end
+
+      def transaction_type_options
+        TransactionType.not_children.credits
+      end
+
+      def none_selected
+        "none"
+      end
+
+    private
+
+      def none_selected?
+        @none_selected
+      end
+
+      def transaction_types
+        transaction_type_options.where(id: transaction_type_ids)
+      end
+
+      def destroy_transactions!
+        destroy_legal_aid_application_transaction_types!
+        destroy_regular_income_transactions!
+        destroy_cash_income_transactions!
+      end
+
+      def destroy_legal_aid_application_transaction_types!
+        legal_aid_application
+          .legal_aid_application_transaction_types
+          .credits
+          .where.not(transaction_type_id: transaction_type_ids)
+          .destroy_all
+      end
+
+      def destroy_regular_income_transactions!
+        legal_aid_application
+          .regular_incomes
+          .without(existing_income_regular_transactions)
+          .destroy_all
+      end
+
+      def destroy_cash_income_transactions!
+        legal_aid_application
+          .cash_transactions
+          .credits
+          .where.not(transaction_type_id: transaction_type_ids)
+          .destroy_all
+      end
+
+      def existing_income_regular_transactions
+        legal_aid_application
+          .regular_incomes
+          .where(transaction_type_id: transaction_type_ids)
+      end
+
+      def build_legal_aid_application_transaction_types
+        transaction_type_ids.each do |transaction_type_id|
+          next if transaction_type_id.in?(legal_aid_application.transaction_type_ids)
+
+          legal_aid_application.legal_aid_application_transaction_types.build(
+            transaction_type_id:,
+          )
+        end
+      end
+
+      def regular_transactions
+        @regular_transactions ||= transaction_types.map do |transaction_type|
+          RegularTransaction.find_or_initialize_by(
+            legal_aid_application:,
+            transaction_type:,
+          )
+        end
+      end
+
+      def all_regular_transactions_valid
+        regular_transactions.each do |transaction|
+          transaction_type = transaction.transaction_type
+          transaction.amount = public_send("#{transaction_type.name}_amount")
+          transaction.frequency = public_send("#{transaction_type.name}_frequency")
+
+          next if transaction.valid?
+
+          transaction.errors.each do |error|
+            add_regular_transaction_error_to_form(
+              transaction.transaction_type.name,
+              error,
+            )
+          end
+        end
+      end
+
+      def add_regular_transaction_error_to_form(transaction_type, error)
+        attribute = if error.attribute.in?(%i[amount frequency])
+                      "#{transaction_type}_#{error.attribute}".to_sym
+                    else
+                      :base
+                    end
+
+        errors.add(attribute, error.type, **error.options)
+      end
+    end
+  end
+end
--- a/app/services/flow/flows/provider_capital.rb
+++ b/app/services/flow/flows/provider_capital.rb
@@ -12,7 +12,11 @@ module Flow
             when :hmrc_single_employment, :unexpected_employment_data
               :employment_incomes
             when :employed_journey_not_enabled, :provider_not_enabled_for_employed_journey, :applicant_not_employed
-              :identify_types_of_incomes
+              if Setting.enhanced_bank_upload?
+                :regular_incomes
+              else
+                :identify_types_of_incomes
+              end
             else
               raise "Unexpected hmrc status #{status.inspect}"
             end
@@ -20,6 +24,17 @@ module Flow
         },
         identify_types_of_incomes: {
           path: ->(application) { urls.providers_legal_aid_application_means_identify_types_of_income_path(application) },
+          forward: lambda do |application|
+            application.income_types? ? :cash_incomes : :student_finances
+          end,
+          check_answers: lambda do |application|
+            return :cash_incomes if application.income_types?
+
+            application.uploading_bank_statements? ? :means_summaries : :income_summary
+          end,
+        },
+        regular_incomes: {
+          path: ->(application) { urls.providers_legal_aid_application_means_regular_incomes_path(application) },
           forward: lambda do |application|
             application.income_types? ? :cash_incomes : :student_finances
           end,
--- a/config/locales/en/helpers.yml
+++ b/config/locales/en/helpers.yml
@@ -11,6 +11,8 @@ en:
         postcode: This must be a UK postcode, for example SW1A 2AA.
       applicant:
         national_insurance_number: For example, 'QQ 12 34 56 C'.
+      providers_means_regular_income_form:
+        transaction_type_ids: Select all that apply
       student_finance: This includes any type of student loan, grant or bursary.
     label:
       applicant:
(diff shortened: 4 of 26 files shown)


app/models/regular_payment.rb
```diff
@@ -0,0 +1,17 @@
+class RegularPayment < ApplicationRecord
+  FREQUENCIES = %w[weekly monthly].freeze
```
cpjmcquillan, line 2: These frequencies need updating once they are confirmed by design.

db/schema.rb
```diff
@@ -689,8 +689,8 @@
     t.string "category_of_law", null: false
     t.string "category_law_code", null: false
     t.string "ccms_matter_code"
-    t.string "client_involvement_type_ccms_code", null: false
-    t.string "client_involvement_type_description", null: false
+    t.string "client_involvement_type_ccms_code"
+    t.string "client_involvement_type_description"
```
cpjmcquillan, line 693: It looks like these fields are nullable, but I didn't remove the constraint.

Has anyone run in to this before?

spec/forms/providers/means/identify_income_form_spec.rb
```diff
@@ -0,0 +1,299 @@
...
+        benefits = create(:transaction_type, :benefits)
+        pension = create(:transaction_type, :pension)
+        params = {
+          "legal_aid_application_id" => "test-id",
+          "income_types" => ["", benefits.name, pension.name],
+          "benefits_amount" => "",
+          "benefits_frequency" => "weekly",
+          "pension_amount" => "100",
+          "pension_frequency" => "monthly",
+        }
```
cpjmcquillan, line 39: I constructed params like this to replicate how they might be submitted from the form in a real browser.

app/models/regular_payment.rb
cpjmcquillan: Covered by https://dsdmoj.atlassian.net/browse/AP-3459

db/schema.rb
```diff
@@ -20,7 +20,7 @@
     t.uuid "record_id", null: false
     t.string "record_type", null: false
     t.uuid "blob_id", null: false
-    t.datetime "created_at", null: false
+    t.datetime "created_at", precision: nil, null: false
```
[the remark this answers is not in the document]
cpjmcquillan: I'll revert these changes, but I thought the discussion could be useful.

db/migrate/20220902150227_create_regular_payments.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRegularPayments < ActiveRecord::Migration[7.0]
+  def change
+    create_table :regular_payments, id: :uuid do |t|
```
cpjmcquillan, line 3: `regular_payments` or `regular_transactions`?

In isolation, I think `RegularPayment` makes more sense and models the domain accurately and clearly.

However, given we already have `TransactionType`, maybe `RegularTransaction` would be more helpful in the context of the application?

What do other people think?

jsugarman: We have had a discussion on this already as a team when looking at the first basic ticket and decided to use the regular_transactions `RegularTransaction` naming. That naming has also been used in the [related CFE PR](https://github.com/ministryofjustice/check-financial-eligibility/pull/1076)

app/forms/providers/means/identify_income_form.rb
```diff
@@ -0,0 +1,100 @@
+module Providers
+  module Means
+    class IdentifyIncomeForm
```
jsugarman, line 3: **NOT a blocker** Just from a naming point of view, and following on from the controller renaming comment,  it might be clearer to name it IdentifyRegularIncomesForm. Have followed a similar naming in the CFE branch

jsugarman (commented):
looks good, 

have added some comments about naming of the model, controller and form and use of a constant to store the names of the credits, the last being the only real blocker/tbd

app/views/providers/means/identify_incomes/show.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= form_with model: @form, url: providers_legal_aid_application_means_identify_incomes_path(@legal_aid_application), method: :patch do |f| %>
+  <%= page_template page_title: page_title, template: :basic, form: f do %>
+    <%= f.govuk_check_boxes_fieldset :income_types, legend: {size: "xl"} do %>
```
RoseSAK, line 3:
```suggestion
    <%= f.govuk_check_boxes_fieldset :income_types, legend: {size: "xl", tag: "h1"} do %>
```

RoseSAK: Would suggest adding this so the legend renders within a h1 tag - otherwise the page is missing a h1

RoseSAK (commented):
One final comment, I noticed that there error message that appears when nothing is selected is just the default rails error message: 

Screenshot 2022-09-13 at 16 53 01

This should probably be updated

app/forms/providers/means/identify_income_form.rb
cpjmcquillan: In the spirit of the other comment, I'd probably lean toward something like `RegularIncomeForm`.

Does that convey the intent more accurately from your perspective?

app/services/flow/flows/provider_capital.rb
```diff
@@ -12,7 +12,11 @@ class ProviderCapital < FlowSteps
             when :hmrc_single_employment, :unexpected_employment_data
               :employment_incomes
             when :employed_journey_not_enabled, :provider_not_enabled_for_employed_journey, :applicant_not_employed
-              :identify_types_of_incomes
+              if Setting.enhanced_bank_upload?
+                :identify_incomes
+              else
+                :identify_types_of_incomes
```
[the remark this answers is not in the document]
cpjmcquillan: I agree it isn't very clear at all —when I started the work I just followed the `:identify_types_of_income` that existed already.

I'll leave that flow as is, but I'll refactor the enhanced flow to use more transparent naming.

Probably something like `:regular_incomes`.

How does that sound?

config/locales/en/helpers.yml
```diff
@@ -58,3 +80,11 @@ en:
           <<: *true_false
         enhanced_bank_upload:
           <<: *true_false
+    legend:
+      providers_means_identify_income_form:
+        benefits_frequency: Select frequency
+        friends_or_family_frequency: Select frequency
+        maintenance_in_frequency: Select frequency
+        property_or_lodger_frequency: Select frequency
+        pension_frequency: Select frequency
```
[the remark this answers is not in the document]
cpjmcquillan: I added them so we can rely on the form builder's default behaviour to lookup these labels automatically.

app/controllers/providers/means/identify_types_of_incomes_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Providers
   module Means
     class IdentifyTypesOfIncomesController < ProviderBaseController
+      before_action :redirect_if_enhanced_bank_upload_enabled
```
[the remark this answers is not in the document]
cpjmcquillan: I'd prefer us to use feature flags as routing constraints, but I didn't have the time to implement it right now.

Given we don't have that (and we don't write unit tests for the flow steps), I think it's appropriate to restrict feature-flagged functionality at the controller level.

app/views/providers/means/identify_incomes/show.html.erb
cpjmcquillan: Good spot, thanks!

app/forms/providers/means/identify_income_form.rb
```diff
@@ -0,0 +1,100 @@
...
+      include ActiveModel::Model
+
+      INCOME_TYPES = %i[benefits friends_or_family maintenance_in property_or_lodger pension].freeze
```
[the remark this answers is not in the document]
cpjmcquillan: Yeah that makes sense, thanks!

app/services/flow/flows/provider_capital.rb
```diff
@@ -29,6 +33,17 @@ class ProviderCapital < FlowSteps
...
         },
+        identify_incomes: {
+          path: ->(application) { urls.providers_legal_aid_application_means_identify_incomes_path(application) },
+          forward: lambda do |application|
+            application.income_types? ? :cash_incomes : :student_finances
+          end,
+          check_answers: lambda do |application|
+            return :cash_incomes if application.income_types?
+
+            application.uploading_bank_statements? ? :means_summaries : :income_summary
```
RoseSAK, line 44: Just noticed this - we don't need this conditional as we only need to redirect to the  `:income_summary` page on the Truelayer route - and anyone coming from this page will be on the enhanced bank upload route. So we can just redirect straight to `:means_summaries`

cpjmcquillan: Presumably that's the case for the existing `:identify_types_of_income` flow step too?

If it is, let's open a separate PR to fix that!

app/forms/providers/means/identify_income_form.rb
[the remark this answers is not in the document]
cpjmcquillan: I've had a look, but...

I don't think this change is worthwhile, since it necessitates us creating all the transaction type records in all the relevant unit and request specs —and I think the cost of maintaining the list of income types we want to display in the form is low.

If this is something you feel passionate about, we can add a ticket to the backlog and think about it outside the context of this PR.

How does that sound to you?

jsugarman (changes requested):
Need to catchup about this as it is not working because the transaction types themselves are, from what i can tell, not being saved as part of the `RegularIncomeForm` (`IdentifyRegularIncomeForm`, really) so you cannot get back to see them nor add cash transactions. 
Screenshot 2022-09-15 at 16 38 57


I wonder if there should be a join table as well . It may simpler to follow the pattern of the `IdentifyTypesOfIncomesController` more closely

```
has_many :regular_transactions, through: :legal_aid_transaction_type_regular_transactions
```

app/services/flow/flows/provider_capital.rb
RoseSAK: Yes so since the introduction of this page then the check answers redirect can be updated for that flow step too, since the pages are now separate! 

Separate PR is potentially a good idea to update the `:identify_types_of_income` step as we might need to update/remove tests too 👌

config/locales/en/helpers.yml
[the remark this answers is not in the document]
RoseSAK: That makes sense 👍

app/controllers/providers/means/identify_types_of_incomes_controller.rb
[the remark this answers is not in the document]
RoseSAK: The flow step files don't have their own unit test files but we do write rspec request tests which cover redirects for each page based on different feature flags being switched on/off. But happy to go with this method for now if you think it's suitable

spec/forms/providers/means/regular_income_form_spec.rb
```diff
@@ -0,0 +1,289 @@
+require "rails_helper"
+
+# rubocop:disable Rails/SaveBang
+RSpec.describe Providers::Means::RegularIncomeForm do
+  describe "validations" do
+    context "when neither a income type or none are selected" do
+      it "is invalid" do
```
jsugarman, line 7: I don't feel comfortable with the complete change in style for our rspec tests. 

I think we should keep the existing pattern of using rspec `before/after/around` and `let` to extract out setup as and keep the individual example "thin", as much as possible pending a wider cross-team discussion. 

Personally, i find this style less clear and less open to DRYing out, if appropriate, than existing styles and feels like a backward step to me.

I have already added this [ticket to our trello board](https://trello.com/c/POzgiM2O/127-use-of-before-and-lets-in-specs) to start that off.

app/services/flow/flows/provider_capital.rb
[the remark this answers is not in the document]
jsugarman: `identify_regular_incomes` ?!

cpjmcquillan:
@jsugarman I'm not sure any of this necessitates the `Changes requested` badge-of-doom, nor should it be blocking the PR from being merged.

The intention of this PR is to unblock and create a foundation for the rest of the work. From my perspective, it is counter-productive and unhelpful to have ~100 file change PRs that sit around for a long time —since it makes reviewing them so much more difficult. 

The intention of this PR _is not_ to deliver a complete solution that is ready to be released to users (although it can be tested and reviewed by testers/product managers/designers). 

Ideally this PR would have been reviewed and released already.

---

Addressing the specific points you raised:

- > it is not working because the transaction types are not being saved

As mentioned in the commit message:

> Next steps will include:
> - ensuring providers can select the cash transactions

and this is covered in #4279. 

- > I wonder if there should be a join table as well

As you discussed in the [initial ticket](https://dsdmoj.atlassian.net/browse/AP-3426), the preferred solution was to implement a new `regular_transactions` table. I'm not sure what purpose it would serve to duplicate some of that data in an extra table? 

I feel like it would add confusion and make the relationship between a `LegalAidApplication` and its `regular_transactions` and `transaction_type` unnecessarily complicated. 

We would need to maintain a single data point across 2 tables (`regular_transactions` and `legal_aid_application_transaction_types`). I'm not sure what problem that would solve in the long term.

Currently, you could define a relationship between a `LegalAidApplication` and its `transaction_types` like this:

```ruby
class LegalAidApplication < ApplicationRecord
  has_many :regular_transactions
  has_many :transaction_types, through: :regular_transactions
end
```

Ultimately, as we migrate from the existing flow to the enhanced bank upload flow, we will likely define that relationship. We can't currently though, because that relationship is already defined through the `legal_aid_application_transaction_types` table. 

- > It may simpler to follow the pattern of the IdentifyTypesOfIncomesController more closely

I don't agree with this. In my opinion that controller has far too many responsibilities and I think we should be aiming to improve the state of the codebase as we go, not contribute to further controller bloat. It might be "simpler" in the sense we can copy-paste code to get a working solution, but at what maintenance cost?

---

I hope that clarified things for you, and I hope we're able to move this PR forwards so that we can unblock the next steps and continue to iterate!

If you have any other questions, let me know and perhaps we can pair to discuss them?

jsugarman:
Yes, let's discuss in catchup I booked.

> I'm not sure any of this necessitates the Changes requested badge-of-doom..
I personally do not see it as a badge of doom but an indicator that something needs to be discussed addressed prior to merging. For me those things are primiarily

- I think we should be saving transaction types as part of this page/form's functionality so we have that relationship between application and regular "income" transaction in place. 
- the change to rspec test styling that we have discussed before

app/forms/providers/means/regular_income_form.rb
```diff
@@ -0,0 +1,91 @@
...
+        @none_selected
+      end
+
+      def regular_transactions
+        @regular_transactions ||= TransactionType.where(name: income_types).map do |transaction_type|
+          regular_transaction(transaction_type)
+        end
+      end
+
+      def regular_transaction(transaction_type)
```
jsugarman, line 63: add a join table record for that transaction_type if it does not exist already at this point perhaps?!

cpjmcquillan:
Closing this in favour of a new PR because the scope of the branch has changed significantly and much of the conversation here is no longer relevant.

PR #4279, opened 2022-09-13
Providers can specify client's cash income

cpjmcquillan:
Before, the cash income form relied on a legal aid application's transaction types to determine which options to render in the form.

The enhanced bank upload journey does not create
`LegalAidApplicationTransactionType` records, but instead creates `RegularPayment` records. This meant that providers using the enhanced bank upload journey were not presented with any options on the cash income page, despite selecting regular payments in the previous step.

This adds a `cash_income_options` method to the `AggregatedCashIncome` model, and updates the form to render inputs based on those options.

The method returns the correct transaction types depending on whether the enhanced bank upload journey is enabled or not.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3467)

### DIFF
--- a/app/models/aggregated_cash_income.rb
+++ b/app/models/aggregated_cash_income.rb
@@ -25,4 +25,12 @@ class AggregatedCashIncome < BaseAggregatedCashTransaction
 
     save_cash_transaction_records
   end
+
+  def cash_income_options
+    if Setting.enhanced_bank_upload?
+      legal_aid_application.regular_transactions.credits.map(&:transaction_type)
+    else
+      legal_aid_application.transaction_types.not_children.credits
+    end
+  end
 end
--- a/app/views/shared/forms/_cash_income_form.html.erb
+++ b/app/views/shared/forms/_cash_income_form.html.erb
@@ -11,7 +11,7 @@
       <p><%= t('.info') %></p>
       <div id="select-all-that-apply-hint" class="govuk-hint"><%= t('generic.select_all_that_apply') %></div>
       <div class="deselect-group" data-deselect-ctrl="#aggregated-cash-income-none-selected-true-field">
-        <% @legal_aid_application.transaction_types.not_children.credits.each do |transaction_type| %>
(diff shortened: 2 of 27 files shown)


app/models/aggregated_cash_income.rb
```diff
@@ -25,4 +25,12 @@ def update(params)
 
     save_cash_transaction_records
   end
+
+  def cash_income_options
+    if Setting.enhanced_bank_upload?
+      legal_aid_application.regular_transactions.credits.map(&:transaction_type)
+    else
+      legal_aid_application.transaction_types.not_children.credits
```
RoseSAK, line 33: Why does this line not work with regular payments already? I thought that transaction types were created and added to the application alongside regular payments when these are added on the previous page. I am still trying to understand that part of the code though so this might be an incorrect understanding?

app/views/shared/forms/_cash_income_form.html.erb
```diff
@@ -11,7 +11,7 @@
       <p><%= t('.info') %></p>
       <div id="select-all-that-apply-hint" class="govuk-hint"><%= t('generic.select_all_that_apply') %></div>
       <div class="deselect-group" data-deselect-ctrl="#aggregated-cash-income-none-selected-true-field">
-        <% @legal_aid_application.transaction_types.not_children.credits.each do |transaction_type| %>
+        <% @aggregated_cash_income.cash_income_options.each do |transaction_type| %>
```
RoseSAK, line 14: This part does look cleaner

RoseSAK (changes requested):

app/models/aggregated_cash_income.rb
cpjmcquillan: A `LegalAidApplication` only `has_many :transaction_types, through: :legal_aid_application_transaction_types` (i.e the join table).

`RegularTransaction` effectively models the same relationship between a `LegalAidApplication` and a `TransactionType`, except the relationship contains more information and thereby models the domain more literally.

Does that make sense?

PR #4302, opened 2022-09-16
AP-3460 enhanced bank upload regular outgoings

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3460)

- Add route, controller, form, view and content for regular outgoings page, so providers can select outgoing categories and add the amount and frequency of payment for each category
- Add rspec tests for form and controller
- Update existing rspec tests and feature tests

## Checklist

### DIFF
--- a/app/forms/providers/means/regular_outgoings_form.rb
+++ b/app/forms/providers/means/regular_outgoings_form.rb
@@ -1,0 +1,161 @@
+module Providers
+  module Means
+    class RegularOutgoingsForm
+      include ActiveModel::Model
+
+      OUTGOING_TYPES = %w[rent_or_mortgage child_care maintenance_out legal_aid].freeze
+
+      attr_reader :transaction_type_ids, :legal_aid_application
+
+      OUTGOING_TYPES.each do |outgoing_type|
+        attr_accessor "#{outgoing_type}_amount".to_sym, "#{outgoing_type}_frequency".to_sym
+      end
+
+      validates :transaction_type_ids, presence: true, unless: :none_selected?
+      validate :all_regular_transactions_valid
+
+      def initialize(params = {})
+        @none_selected = none_selected.in?(params["transaction_type_ids"] || [])
+        @legal_aid_application = params.delete(:legal_aid_application)
+        @transaction_type_ids = params["transaction_type_ids"] || @legal_aid_application.transaction_type_ids
+
+        assign_regular_transaction_attributes
+
+        super
+      end
+
+      def save
+        return false unless valid?
+
+        ApplicationRecord.transaction do
+          destroy_transactions!
+
+          build_legal_aid_application_transaction_types
+          legal_aid_application.assign_attributes(
+            no_debit_transaction_types_selected: none_selected?,
+          )
+          legal_aid_application.save!
+
+          regular_transactions.each(&:save!)
+        end
+
+        true
+      end
+
+      def transaction_type_ids=(ids)
+        @transaction_type_ids = if none_selected?
+                                  []
+                                else
+                                  ids.compact_blank
+                                end
+      end
+
+      def transaction_type_options
+        TransactionType.not_children.debits
+      end
+
+      def none_selected
+        "none"
+      end
+
+    private
+
+      def assign_regular_transaction_attributes
+        regular_transactions.each do |transaction|
+          transaction_type = transaction.transaction_type
+          public_send("#{transaction_type.name}_amount=", transaction.amount)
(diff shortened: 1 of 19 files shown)


cpjmcquillan:
I'll give this a full review on Monday, but it looks pretty good so far —nice work!

At a first glance, I reckon the transaction forms are ripe for some refactoring.

If you'd like to pair on it, feel free to chuck some time in my calendar anytime next week.

app/forms/providers/means/regular_outgoings_form.rb
```diff
@@ -0,0 +1,161 @@
...
+          .legal_aid_application_transaction_types
+          .debits
+          .where.not(transaction_type_id: transaction_type_ids)
+          .destroy_all
+      end
+
+      def destroy_regular_outgoing_transactions!
+        legal_aid_application
+          .regular_outgoings
+          .without(existing_outgoing_regular_transactions)
```
jsugarman, line 96: I realize this copies the pattern that went in for regular incomes, so maybe @cpjmcquillan can comment, but 
is this the `Array#without` method? in which case it seem to be mixing an array method with active an record chain plus a separate query. would not the same as for cash_transactions work and be more efficient
```
.where.not(transaction_type_id: transaction_type_ids)
```

app/forms/providers/means/regular_outgoings_form.rb
```diff
@@ -0,0 +1,161 @@
...
+          .destroy_all
+      end
+
+      def destroy_cash_outgoing_transactions!
+        legal_aid_application
+          .cash_transactions
+          .debits
+          .where.not(transaction_type_id: transaction_type_ids)
+          .destroy_all
```
jsugarman, line 105: Again, I realize this copies the pattern that went in for regular incomes, so maybe @cpjmcquillan can comment, but I do not believe this is needed as the `destroy_legal_aid_application_transaction_types!` method destroys the join table record which will fire this [after_commit hook](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/models/legal_aid_application_transaction_type.rb#L8), which will delete the cash_transactions.

by extension, the same could be done for regular transactions, to take the dependent destroy logic out of the controller and into the model, where it feels to me to be more appropriate.

Anyway, neither of there are blockers and should not really be done here as the pattern is already in but it would be good could look at this as a refactor I think.

spec/forms/providers/means/regular_outgoings_form_spec.rb
```diff
@@ -0,0 +1,317 @@
...
+        form = described_class.new(params)
+
+        expect(form).to be_invalid
+        expect(form.errors).to be_added(:transaction_type_ids, :blank)
+      end
+    end
+
+    context "when at least one outgoing type is selected, but an amount is missing" do
+      it "is invalid" do
+        rent_or_mortgage = create(:transaction_type, :rent_or_mortgage)
```
jsugarman, line 19: again, don't feel comfortable and don't like this "not very rspec" and inconsistent to codebase way of testing but, ok as it follows the pattern of the previous one.

jsugarman (approved):
👍  Great Work @RoseSAK  - some comments regarding the patterns used here which are rather beyond the scope of this work as they are following preceding patterns.

app/forms/providers/means/regular_outgoings_form.rb
cpjmcquillan: It's an [alias for `.excluding` in ActiveRecord](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-excluding), so there is no difference between the SQL generated by either statement.

If `.without(existing_outgoing_regular_transactions)` is too verbose though, then let's change it.

app/forms/providers/means/regular_outgoings_form.rb
cpjmcquillan: This is intentional.

It only isn't needed if the `TransactionType` record associated to the `{Regular/Cash/Bank}Transaction` record has a respective `LegalAidApplicationTransactionType` record. 

This relationship isn't enforced at the database or the application layer, though —so it's possible for records to exist that aren't destroyed by the callback (because the callback never happens).

I think it's safer to be intentional about destroying the records until we refactor the relationship between a `LegalAidApplication` and it's various transactions (i.e they are joined through the join table).

If we do that, we wouldn't need the callback method anymore either, since we could rely on `dependent: :destroy` to do the dirty work for us.

cpjmcquillan (approved):

PR #4314, opened 2022-09-20
Providers can submit more accurate income data

cpjmcquillan:
Before, when providers with employment and bank upload permissions submitted income data for their client, they could only specify which transactions their client received.

This data is not granular enough, and makes it impossible to obtain a financial eligibility assessment (CFE) result for the client.

As part of ongoing work to iterate on the bank upload journey, providers are now able to tell us how much (amount), and how often (frequency) their client receives these income payments.

Next steps will include:
- updating the list of frequencies for each type of transaction
- replicating this functionality for the outgoings page

This user journey is restricted by feature flag at the controller level.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3445)

#### Video

https://user-images.githubusercontent.com/25187547/191240188-b69916fb-f99b-4e3f-a5fe-b95a45cb263d.mp4

### DIFF
--- a/app/forms/providers/means/regular_income_form.rb
+++ b/app/forms/providers/means/regular_income_form.rb
@@ -1,0 +1,169 @@
+module Providers
+  module Means
+    class RegularIncomeForm
+      include ActiveModel::Model
+
+      INCOME_TYPES = %w[
+        benefits
+        friends_or_family
+        maintenance_in
+        property_or_lodger
+        student_loan
+        pension
+      ].freeze
+
+      attr_reader :transaction_type_ids, :legal_aid_application
+
+      INCOME_TYPES.each do |income_type|
+        attr_accessor "#{income_type}_amount".to_sym,
+                      "#{income_type}_frequency".to_sym
+      end
+
+      validates :transaction_type_ids, presence: true, unless: :none_selected?
+      validate :all_regular_transactions_valid
+
+      def initialize(params = {})
+        @none_selected = none_selected.in?(params["transaction_type_ids"] || [])
+        @legal_aid_application = params.delete(:legal_aid_application)
+        @transaction_type_ids = params["transaction_type_ids"] || @legal_aid_application.transaction_type_ids
+
+        assign_regular_transaction_attributes
+
+        super
+      end
+
+      def save
+        return false unless valid?
+
+        ApplicationRecord.transaction do
+          destroy_transactions!
+
+          build_legal_aid_application_transaction_types
+          legal_aid_application.assign_attributes(
+            no_credit_transaction_types_selected: none_selected?,
+          )
+          legal_aid_application.save!
+
+          regular_transactions.each(&:save!)
+        end
+
+        true
+      end
+
+      def transaction_type_ids=(ids)
+        @transaction_type_ids = if none_selected?
+                                  []
+                                else
+                                  ids.compact_blank
+                                end
+      end
+
+      def transaction_type_options
+        TransactionType.not_children.credits
+      end
+
+      def none_selected
+        "none"
+      end
+
+    private
+
+      def assign_regular_transaction_attributes
+        regular_transactions.each do |transaction|
+          transaction_type = transaction.transaction_type
+          public_send("#{transaction_type.name}_amount=", transaction.amount)
+          public_send("#{transaction_type.name}_frequency=", transaction.frequency)
+        end
+      end
+
+      def none_selected?
+        @none_selected
+      end
+
+      def transaction_types
+        transaction_type_options.where(id: transaction_type_ids)
+      end
+
+      def destroy_transactions!
+        destroy_legal_aid_application_transaction_types!
+        destroy_regular_income_transactions!
+        destroy_cash_income_transactions!
+      end
+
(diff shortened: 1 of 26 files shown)


app/forms/providers/means/regular_income_form.rb
```diff
@@ -0,0 +1,169 @@
...
+      end
+
+      def none_selected?
+        @none_selected
+      end
+
+      def transaction_types
+        transaction_type_options.where(id: transaction_type_ids)
+      end
+
```
RoseSAK, line 86: My only thought is that perhaps these methods could be extracted into a helper class, as some of this functionality will be replicated on the outgoings page as well as on the current `identify_types_of_income/outgoing` and the cash transactions pages. We currently have the TransactionType helper - maybe this would be a suitable place? I'm not suggesting this needs be done in this PR though, just a suggestion for later on

RoseSAK (approved):
Looks good 💯

app/models/legal_aid_application.rb
```diff
@@ -50,6 +50,8 @@ class LegalAidApplication < ApplicationRecord
   has_many :involved_children, class_name: "ApplicationMeritsTask::InvolvedChild", dependent: :destroy
   has_many :hmrc_responses, class_name: "HMRC::Response", dependent: :destroy, inverse_of: :legal_aid_application
   has_many :employments, dependent: :destroy
+  has_many :regular_transactions, dependent: :destroy
+  has_many :regular_incomes, -> { credits }, class_name: "RegularTransaction"
```
jsugarman, line 54: nice 👍

db/migrate/20220902150227_create_regular_transactions.rb
```diff
@@ -0,0 +1,11 @@
+class CreateRegularTransactions < ActiveRecord::Migration[7.0]
+  def change
+    create_table :regular_transactions, id: :uuid do |t|
+      t.references :legal_aid_application, null: false, foreign_key: { on_delete: :cascade }, index: false, type: :uuid
```
jsugarman, line 4: ❓ I assume we do not have any callbacks on regular transactions that would be missed by a cascade delete here or would the `has_many :regular_transactions, dependant: :destroy` trigger those in any event?

spec/models/regular_transaction_spec.rb
```diff
@@ -0,0 +1,100 @@
+require "rails_helper"
+
+RSpec.describe RegularTransaction, type: :model do
+  describe "validations" do
```
jsugarman, line 4: **nit** more of a context than a describe to my mind but either way

```suggestion
  describe "#validate" do
```

```suggestion
  context "when validating" do
```

app/forms/providers/means/regular_income_form.rb
```diff
@@ -0,0 +1,169 @@
+module Providers
+  module Means
+    class RegularIncomeForm
+      include ActiveModel::Model
+
+      INCOME_TYPES = %w[
```
jsugarman, line 6: would using something like

```
def income_types
  TransactionType.not_children.credits.map(&:name)
   # OR
   transaction_type_options.map(&:name)
 end
```

... be neater and avoid duplicate sources of these types? might require testing changes though

jsugarman (approved):
LGTM 👍 

A few nits and questions. 

Not sure we should use a constant in the Form object for the different income types though. is there a reason for not using the DB values `TransactionType.not_children.credits`?

db/migrate/20220902150227_create_regular_transactions.rb
cpjmcquillan: From my perspective it didn't make sense for a regular transaction to exist if it isn't associated to a legal aid application in any circumstance.

If a `legal_aid_application` record is deleted from the database, any `regular_transaction` records that reference that `legal_aid_application_id` are also deleted.

This cascades deletions at a database level, but I wouldn't really expect us to be deleting records this way normally.

Typically, we'd be destroying records with ActiveRecord and the database constraint wouldn't come into play.

app/forms/providers/means/regular_income_form.rb
cpjmcquillan: We could, but I hadn't yet.

The reason I've been reluctant to is because I don't feel like it necessarily makes sense for the attribute accessors and permitted params to depend explicitly on the `TransactionType` records.

I'm not 100% convinced `TransactionType` should be `ActiveRecord` models in the conventional sense —they feel more like "static" records to me. 

When it comes to the form rendered in the view, we do iterate the `TransactionType.not_children.credits` records though.

Let's revisit this if we run into it again (i.e we add a new transaction type and we create the record, but forget to add attribute accessors or permit the new parameters).

How does that sound?

app/forms/providers/means/regular_income_form.rb
cpjmcquillan: We might do, but we might not.

I'd rather wait for the right abstractions to present themselves clearly, rather than try and predict what the abstractions might be to "DRY" things up —and potentially making things more difficult for ourselves.

I propose we leave it for now.

Does that sound ok to you?

PR #4319, opened 2022-09-21
Introduce a `Flow::Steps::Step` struct

cpjmcquillan:
Each descendant of the `Flow::Flows::FlowSteps` class contains 
a mega `STEPS` hash that spans hundreds of lines and each step 
often has associated complexity and conditional logic.

These objects manage complex user flow and are an integral part 
of the application's functionality, however they cannot be easily 
unit-tested. This means we rely on more expensive integration tests 
to test each specific edge case of flow logic.

This introduces a minimal `Flow::Steps::Step` struct, which can be 
used to manage this complexity and introduces an interface for 
encapsulating this important user flow logic that can be more easily 
unit-tested.

### DIFF
--- a/app/services/flow/flows/provider_capital.rb
+++ b/app/services/flow/flows/provider_capital.rb
@@ -33,17 +33,7 @@ module Flow
             application.uploading_bank_statements? ? :means_summaries : :income_summary
           end,
         },
-        regular_incomes: {
-          path: ->(application) { urls.providers_legal_aid_application_means_regular_incomes_path(application) },
-          forward: lambda do |application|
-            application.income_types? ? :cash_incomes : :student_finances
-          end,
-          check_answers: lambda do |application|
-            return :cash_incomes if application.income_types?
-
-            application.uploading_bank_statements? ? :means_summaries : :income_summary
-          end,
-        },
+        regular_incomes: Flow::Steps::RegularIncomeStep,
         cash_incomes: {
           path: ->(application) { urls.providers_legal_aid_application_means_cash_income_path(application) },
           forward: :student_finances,
--- a/app/services/flow/steps/regular_income_step.rb
+++ b/app/services/flow/steps/regular_income_step.rb
@@ -1,0 +1,9 @@
+module Flow
+  module Steps
+    RegularIncomeStep = Step.new(
+      ->(application) { urls.providers_legal_aid_application_means_regular_incomes_path(application) },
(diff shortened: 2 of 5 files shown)


cpjmcquillan:
I'm interested to hear what people think of this proposed change.

We don't necessarily need to do _exactly_ this, but I don't think the way we're currently using `Flow::STEPS` and associated concepts is particularly maintainable.

cpjmcquillan:
I'm also interested to hear peoples thoughts on the potential impact this could have on the way we test different user flows.

i.e is there any appetite to reduce the number of edge cases we test with integration tests if we have unit tests for _some_ flow step object?

cpjmcquillan:
[From Slack](https://mojdt.slack.com/archives/CFK4B3GUR/p1663849935392379?thread_ts=1663838145.015629&cid=CFK4B3GUR)

> colinbruce:
> I like this in principle, simpler, smaller tests of discrete code.
>My concerns are around the amount of work it will be to complete the task with the number of priorities at the moment! :grimacing:
> It would be worse to do it piecemeal and leave it half completed (thinks about the CCCD API :scream:)

cpjmcquillan:
@jsugarman @RoseSAK I'd like to get both your thoughts on this.

---

> It would be worse to do it piecemeal and leave it half completed

@colinbruce I would probably disagree with this statement.

I think the current approach to managing the flow steps is a 
significant maintenance burden, and I think we should be trying
to mitigate that where we can.

Even if that means people introduce objects as they go, I think it 
is a worthwhile endeavour.

In my opinion we should be striving to leave the code in a better
state than we found it whenever we carry out a piece of work. 

If we feel that ethos is at odds with the team's pursuit of feature 
development, then we should communicate the business case that 
maintenance is an essential part of feature development and will 
help us move faster in the long term.

---

My tldr; is:

Let's not let the perfect be the enemy of the good!

jsugarman:
I like it in principle. Lets discuss verbally at some point, like dev time - which we should probably do more regularly.

My views summarised:
1. the flow files need rejigging and renaming in any event IMO
2. the flow files can be broken up into smaller more logical groupings to reduce the length of each IMO
3. individually the flow files, aside from some with extreme length, are relatively easy to read IMO. Adding an extra layer of indirection may make it more difficult to picture the whole.
4. testing wise we need to enable unit specs on the flow files in any event as exercising them via request specs is not ideal IMO

PR #4324, opened 2022-09-22
AP-3495: Refactor scope limitations

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3495)

This creates a new ScopeLimitaton object and updates handlers to create new ScopeLimitation objects going forward.

It adds helper methods to read data from the proceedings table (current records) _or_ the ScopeLimitations table (new records) depending on what is available

## Checklist

### DIFF
--- a/app/models/proceeding.rb
+++ b/app/models/proceeding.rb
@@ -53,27 +53,27 @@ class Proceeding < ApplicationRecord
   end
 
   def substantive_scope_limitation_code
-    substantive_scope_limitations.first.code
+    self["substantive_scope_limitation_code"] || substantive_scope_limitations&.first&.code
   end
 
   def substantive_scope_limitation_meaning
-    substantive_scope_limitations.first.meaning
+    self["substantive_scope_limitation_meaning"] || substantive_scope_limitations&.first&.meaning
   end
 
   def substantive_scope_limitation_description
-    substantive_scope_limitations.first.description
+    self["substantive_scope_limitation_description"] || substantive_scope_limitations&.first&.description
   end
 
   def delegated_functions_scope_limitation_code
-    emergency_scope_limitations.first.code
+    self["delegated_functions_scope_limitation_code"] || emergency_scope_limitations&.first&.code
   end
 
   def delegated_functions_scope_limitation_meaning
-    emergency_scope_limitations.first.meaning
+    self["delegated_functions_scope_limitation_meaning"] || emergency_scope_limitations&.first&.meaning
   end
 
   def delegated_functions_scope_limitation_description
-    emergency_scope_limitations.first.description
+    self["delegated_functions_scope_limitation_description"] || emergency_scope_limitations&.first&.description
   end
 
   def highest_proceeding_case_id
--- a/app/models/proceeding.rb
+++ b/app/models/proceeding.rb
@@ -44,13 +44,37 @@ class Proceeding < ApplicationRecord
     ccms_matter_code == "MINJN"
   end
 
-  # def default_level_of_service_level
-  #   "3"
-  # end
-  #
-  # def default_level_of_service_name
-  #   "Full Representation"
-  # end
+  def substantive_scope_limitations
+    scope_limitations.where(scope_type: :substantive)
+  end
+
+  def emergency_scope_limitations
+    scope_limitations.where(scope_type: :emergency)
+  end
+
+  def substantive_scope_limitation_code
+    substantive_scope_limitations.first.code
+  end
+
+  def substantive_scope_limitation_meaning
+    substantive_scope_limitations.first.meaning
+  end
+
+  def substantive_scope_limitation_description
+    substantive_scope_limitations.first.description
+  end
+
+  def delegated_functions_scope_limitation_code
+    emergency_scope_limitations.first.code
+  end
+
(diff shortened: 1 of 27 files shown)


app/models/proceeding.rb
```diff
@@ -42,13 +44,37 @@ def domestic_abuse?
...
-  # def default_level_of_service_level
-  #   "3"
-  # end
-  #
-  # def default_level_of_service_name
-  #   "Full Representation"
-  # end
+  def substantive_scope_limitations
+    scope_limitations.where(scope_type: :substantive)
+  end
```
jsugarman, line 49: **just a suggestion:** 

if there is only 1 expected and or order does not matter at this point, then could we do a [find_by](https://api.rubyonrails.org/v7.0.3.1/classes/ActiveRecord/FinderMethods.html#method-i-find_by) here, thereby removing need for `.first` below. should probably remove the `s` too in that case 
```suggestion
  def substantive_scope_limitation
    scope_limitations.find_by(scope_type: :substantive)
  end
```

app/models/proceeding.rb
```diff
@@ -42,13 +44,37 @@ def domestic_abuse?
...
+
+  def emergency_scope_limitations
+    scope_limitations.where(scope_type: :emergency)
+  end
```
jsugarman, line 53: [similarly](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4324/files#r977574055)

spec/forms/proceedings/emergency_defaults_form_spec.rb
```diff
@@ -59,9 +58,17 @@
...
+        end
+
+        context "without calling the subject" do
+          let(:skip_subject) { true }
+
+          it "creates a scope_limitation object" do
+            expect { save_form }.to change(proceeding.scope_limitations, :count).by(1)
+            expect(proceeding.scope_limitations.where(scope_type: :emergency).first.code).to eq "CV117"
+            expect(proceeding.scope_limitations.where(scope_type: :emergency).first.meaning).to eq "Interim order inc. return date"
+            expect(proceeding.scope_limitations.where(scope_type: :emergency).first.description).to eq "Limited to all steps necessary to apply for an interim order; where application is made without notice to include representation on the return date."
```
jsugarman, line 70:
```suggestion
            expect(proceeding.scope_limitations.find_by(scope_type: :emergency))
              .to have_attributes(code: "CV117",
                                    meaning: "Interim order inc. return date",
                                    description: "Limited to all steps necessary to apply for an interim order; where application is made without notice to include representation on the return date.")
```
but either way

spec/forms/proceedings/substantive_defaults_form_spec.rb
```diff
@@ -28,9 +30,17 @@
...
+        end
+
+        context "without calling the subject" do
+          let(:skip_subject) { true }
+
+          it "creates a scope_limitation object" do
+            expect { save_form }.to change(proceeding.scope_limitations, :count).by(1)
+            expect(proceeding.scope_limitations.where(scope_type: :substantive).first.code).to eq "AA019"
+            expect(proceeding.scope_limitations.where(scope_type: :substantive).first.meaning).to eq "Injunction FLA-to final hearing"
+            expect(proceeding.scope_limitations.where(scope_type: :substantive).first.description).to eq "As to proceedings under Part IV Family Law Act 1996 limited to all steps up to and including obtaining and serving a final order and in the event of breach leading to the exercise of a power of arrest to representation on the consideration of the breach by the court (but excluding applying for a warrant of arrest, if not attached, and representation in contempt proceedings)."
```
jsugarman, line 42: [similarly](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4324/files#r977579513)

jsugarman (commented):
LGTM 👍 , just some ignorable suggestion.

Happy to approve but maybe I should leave to blue team members?

jsugarman (approved):
LGTM 👍, thanks for clarifications

app/models/proceeding.rb
colinbruce: As discussed with Joel, this is being implemented to allow multiples so leaving as-is for now.
Similar to the use of `.first` that is the next thing to be removed when we have the multiple handlers done

spec/models/proceeding_spec.rb
```diff
@@ -86,6 +86,12 @@
     end
   end
 
+  describe "#delegated_functions_scope_limitation_meaning" do # TEMP for coverage
```
kmahern, line 89: Is it possible to make this a TODO? (with info on when/how it can be removed)

app/models/scope_limitation.rb
```diff
@@ -0,0 +1,7 @@
+class ScopeLimitation < ApplicationRecord
```
kmahern, line 1: not a blocker - would we normally create a new model spec for a new model? Possibly not worth it for this though (at least not yet).

kmahern (approved):
a couple of minor comments but lgtm 🥇

app/models/scope_limitation.rb
colinbruce: It was on oversight on my part - thanks for spotting :+1:
I have added a very basic placeholder and we can expand as needed

PR #4325, opened 2022-09-22
Remove CCMS testing section from README

cpjmcquillan:
Before, tests tagged with `ccms` were only run if explictly included in the test command.

In #3466, this was changed so that `ccms` tests ran as part of the test suite by default.

This updates the README to reflect that change.


colinbruce (approved):
💥 Thanks for catching this :+1

PR #4326, opened 2022-09-22
Ensure correct flows use enhanced bank upload journey

cpjmcquillan:
Before, the `Providers::Means::IdentifyTypesofIncomesController` redirected all
requests to the new `Providers::Means::RegularIncomesController` if the enhanced
bank upload setting was enabled.

Initially this seemed reasonable, however it was based on some incorrect
assumptions.

When the enhanced bank upload setting is enabled, only applications that are
using the bank upload journey should use the new flow.

This change seeks to redress the incorrect flows and ensure that only
non-passported applications that are using the bank statement upload journey use
the new how much, how often (enhanced bank upload) flow.

### User scenarios that should lead to the enhanced bank upload journey

_Note: In each scenario the setting is enabled and the application is uploading bank_ 
_statements. If either of those conditions aren't met, the user will be directed to the_
_existing `:identify_types_of_income` route._

1. Provider uploads bank statements, applicant is not employed
2. Client has completed means, applicant is not employed
3. Provider enters employment incomes
4. Provider enters full employment details

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -507,15 +507,26 @@ class LegalAidApplication < ApplicationRecord
   end
 
   def uploading_bank_statements?
-    citizen_consent_given = provider_received_citizen_consent.nil? ? true : provider_received_citizen_consent?
-    provider.bank_statement_upload_permissions? && (!citizen_consent_given || attachments.bank_statement_evidence.exists?)
+    return false unless provider.bank_statement_upload_permissions?
+
+    !client_open_banking_consent? || attachments.bank_statement_evidence.exists?
   end
 
   def has_transaction_type?(transaction_type)
     legal_aid_application_transaction_types.map(&:transaction_type_id).include?(transaction_type.id)
   end
 
+  def using_enhanced_bank_upload?
+    Setting.enhanced_bank_upload? && uploading_bank_statements?
+  end
+
 private
+
+  def client_open_banking_consent?
+    return true if provider_received_citizen_consent.nil?
+
+    provider_received_citizen_consent?
+  end
 
   def bank_transactions_by_type(type)
     bank_transactions.__send__(type).order(happened_at: :desc).by_parent_transaction_type
--- a/spec/models/legal_aid_application_spec.rb
+++ b/spec/models/legal_aid_application_spec.rb
@@ -1575,6 +1575,68 @@ RSpec.describe LegalAidApplication, type: :model do
     end
   end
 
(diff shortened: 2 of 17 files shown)


spec/models/legal_aid_application_spec.rb
```diff
@@ -1421,156 +1421,218 @@
...
+          .to receive(:uploading_bank_statements?)
+          .and_return(false)
 
-      context "when application does not have any transaction types" do
-        let(:laa) { create :legal_aid_application, transaction_types: [] }
-        let(:transaction_type) { benefits }
+        result = legal_aid_application.using_enhanced_bank_upload?
 
-        it { is_expected.to be_falsey }
+        expect(result).to be false
```
cpjmcquillan, line 1635: These additions should be the only material changes to the file.

cpjmcquillan:
I noticed this when doing some other work re-structuring feature tests for the new journey.

To resolve the issue, I looked for every instance where `:identify_types_of_income` is returned as a flow step and replaced it with this conditional:

```ruby
if application.using_enhanced_bank_upload_journey?
  :regular_incomes
else
  :identify_types_of_incomes
end
```

Specifically, I'd like feedback on whether these changes to the flow make sense. 

Are there any other scenarios we need to account for, or are there any scenarios where this change doesn't make sense?

app/models/legal_aid_application.rb
```diff
@@ -507,16 +507,27 @@ def hmrc_response_use_case_one
   end
 
   def uploading_bank_statements?
-    citizen_consent_given = provider_received_citizen_consent.nil? ? true : provider_received_citizen_consent?
-    provider.bank_statement_upload_permissions? && (!citizen_consent_given || attachments.bank_statement_evidence.exists?)
+    return false unless provider.bank_statement_upload_permissions?
+
+    !client_open_banking_consent? || attachments.bank_statement_evidence.exists?
```
cpjmcquillan, line 512: I made this change and introduced the private method because I found the existing method difficult to comprehend, and I felt like it didn't reveal the complexity (i.e the different conditional branches) clearly enough.

If people don't agree that it makes the method more readable, I can remove the change from this PR.

jsugarman: it looks ok to me though my brain is a bit frazzled. But I believe unit tests cover this pretty well?!

jsugarman (approved):
LGTM, paired with Rose to double check whether problems in live and looks like all is good.
But looks like features need changing, and coverage issue resolving.

✅ 🎖️ merge as and when ready

PR #4329, opened 2022-09-23
[AP-3451] refactor submission manager

jsugarman:
Refactor submission manager

## What

[relates to story](https://dsdmoj.atlassian.net/browse/AP-3451)

Remove version 4 refrences as CFE version has been removed. Also reorganise and rename constants and their
calling to prepare for a third option of enhanced bank statement upload CFE submissions.

## Checklist


colinbruce (approved):
LGTM 👍

app/services/cfe/submission_manager.rb
```diff
@@ -58,22 +28,12 @@ def submission
...
-
-    def call_non_passported_services
-      return if submission.passported?
+  private
 
-      NON_PASSPORTED_SERVICES.each do |service|
-        make_logged_call_to service
-      end
+    def services
+      ServiceSet.call(submission)
```
cpjmcquillan, line 34: This is really semantics and not a big deal, but I wonder if it might be helpful if the `ServiceSet` responded to `.for` instead of `.call` —since `.call` feels kind of arbitrary to me.

```ruby
def services
  ServiceSet.for(submission: submission)
end
```

I appreciate we might prefer every "service object" to be invoked in the same way though.

spec/services/cfe/service_set_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+
+# rubocop:disable RSpec/VerifiedDoubles
```
cpjmcquillan, line 3: Rather than disabling this rule (what's the point of it if we disable it), could we define a class in the test?

Something like this might work?

```ruby
class TestObject
  def passported?; end

  def non_passported?; end
end

let(:object) { instance_double(TestObject, passported?: true, non_passported?: false) }
```

Although to be honest I think some of the RSpec rubocop rules are a arbitrary/a waste of time/annoying anyway, so I wouldn't mind if you just turned it off completely 🤣

spec/services/cfe/submission_manager_spec.rb
```diff
@@ -49,113 +49,122 @@ module CFE
...
-          let(:application) { create :legal_aid_application, :with_everything, :with_positive_benefit_check_result, :applicant_entering_means, vehicle: }
+        it "calls expected services" do
+          expect(CreateAssessmentService).to receive(:call).once
+          expect(CreateProceedingTypesService).to receive(:call).once
+          expect(CreateApplicantService).to receive(:call).once
+          expect(CreateCapitalsService).to receive(:call).once
+          expect(CreatePropertiesService).to receive(:call).once
+          expect(CreateVehiclesService).to receive(:call).once
+          expect(CreateExplicitRemarksService).to receive(:call).once
+          expect(ObtainAssessmentResultService).to receive(:call).once
```
cpjmcquillan, line 82: Not a big deal, but it might be worth defining the passported/non-passported services and iterating them.

```ruby
let(:passported_services) do
  [
    CreateAssessmentService,
    CreateProceedingTypesService,
    ...
  ]
end

it "calls expected services" do
  passported_services.each do |service|
    allow(service).to receive(:call).and_return(true)
  end
  
  submission_manager.call
  
  passported_services.each do |service|
    expect(service).to have_received(:call).exactly(1).time
  end
end
```

I don't mind too much either way, but it might make it a bit more readable (my lazy eyes glaze over a see of `Service`s 😆.

What do you think?

app/services/cfe/service_set.rb
```diff
@@ -0,0 +1,52 @@
...
+      CreatePropertiesService,
+      CreateExplicitRemarksService,
+      CreateDependantsService,
+      CreateOutgoingsService,
+      CreateStateBenefitsService,
+      CreateOtherIncomeService,
+      CreateIrregularIncomesService,
+      CreateEmploymentsService,
+      CreateCashTransactionsService,
+      ObtainAssessmentResultService,
```
cpjmcquillan, line 31: Is the order these are called important?

I was wondering if it would be worthwhile defining a common set of services still, but I don't mind us being explicit either.

cpjmcquillan (approved):
Nice :shipit:

app/services/cfe/service_set.rb
jsugarman: after the third set goes in we could possibly refactor this again yes. There is also something in progress to perform call all but the `CreateAssessmentService` asynchronously to speed things up. The common set would be 

```
      CreateAssessmentService,
      CreateProceedingTypesService,
      CreateApplicantService,
      CreateCapitalsService,
      CreateVehiclesService,
      CreatePropertiesService,
      CreateExplicitRemarksService,
```

jsugarman: the order is important, but only insofar as the `CreateAssessmentService` must go first i believe. Other may no different - @stephenrichards?

spec/services/cfe/service_set_spec.rb
jsugarman: It's only disabled for this spec.
I went through several iterations with rubocop complaining of various things at various points. This cop seemed the least applicable in the context of this spec. I did look at using an anonymous class but though the effort and complextity is not worth it (your example would raise another cop violation i believe). We do not need to verify a fake class IMO overall, so disabled it

spec/services/cfe/submission_manager_spec.rb
jsugarman: I did not change this except to ensure the `once`, but regardless I  find the above approach good in some respects but have had experience with that "iteration approach" making failures harder to resolve. 🤔

app/services/cfe/service_set.rb
jsugarman: 👍 i will need to lock the test down as the test is not order sensitive but CFE is

app/services/cfe/submission_manager.rb
jsugarman: I see the advantage, but prefer the consistent use of `.call` myself

app/services/cfe/service_set.rb
jsugarman: order sensitive now

spec/services/cfe/submission_manager_spec.rb
jsugarman: have tidied up using a similar approach to that suggested, rubocop rspec suggest going further once that is in:

```
 expect(passported_services).to all(have_received(:call).once
```

PR #4331, opened 2022-09-23
[AP-3451] retrieve cfe result using regular transactions

jsugarman:
## What
Retrieve cfe result using regular transactions

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3451)

Add regular transaction CFE service "set" to store the CFE
endpoints that will need to be hit and add a regular transactions
service to to convert the db records to the expected payload objects to be sent to CFE.

## Checklist

### DIFF
--- a/app/controllers/providers/means_summaries_controller.rb
+++ b/app/controllers/providers/means_summaries_controller.rb
@@ -6,7 +6,7 @@ module Providers
     end
 
     def update
-      unless draft_selected? || legal_aid_application.provider_entering_merits? || legal_aid_application.uploading_bank_statements?
+      if cfe_call_required?
         redirect_to(problem_index_path) && return unless check_financial_eligibility
 
         legal_aid_application.provider_enter_merits!
@@ -15,6 +15,17 @@ module Providers
     end
 
   private
+
+    # TODO: if show is always called before update and converts
+    # state to `checking_non_passported_means` then I do not see
+    # how legal_aid_application.provider_entering_merits? can ever be true!
+    #
+    def cfe_call_required?
+      return false if draft_selected? || legal_aid_application.provider_entering_merits?
+      return false if legal_aid_application.uploading_bank_statements? && !legal_aid_application.using_enhanced_bank_upload?
+
+      true
+    end
 
     def check_financial_eligibility
       CFE::SubmissionManager.call(legal_aid_application.id)
--- a/app/models/concerns/cfe_submission_state_machine.rb
+++ b/app/models/concerns/cfe_submission_state_machine.rb
@@ -18,6 +18,7 @@ module CFESubmissionStateMachine
       state :other_income_created
       state :irregular_income_created
       state :explicit_remarks_created
+      state :regular_transactions_created
       state :cash_transactions_created
       state :results_obtained
       state :employments_created
@@ -71,14 +72,20 @@ module CFESubmissionStateMachine
 
       event :irregular_income_created do
         transitions from: :other_income_created, to: :irregular_income_created, guard: :non_passported?
+        transitions from: :dependants_created, to: :irregular_income_created, guard: :non_passported?
       end
 
       event :employments_created do
         transitions from: :irregular_income_created, to: :employments_created, guard: :non_passported?
       end
 
+      event :regular_transactions_created do
+        transitions from: :employments_created, to: :regular_transactions_created, guard: :non_passported?
+      end
+
       event :cash_transactions_created do
         transitions from: :employments_created, to: :cash_transactions_created, guard: :non_passported?
+        transitions from: :regular_transactions_created, to: :cash_transactions_created, guard: :non_passported?
       end
 
       event :explicit_remarks_created do
@@ -104,6 +111,7 @@ module CFESubmissionStateMachine
         transitions from: :other_income_created, to: :failed
         transitions from: :irregular_income_created, to: :failed
         transitions from: :explicit_remarks_created, to: :failed
+        transitions from: :regular_transactions_created, to: :failed
         transitions from: :cash_transactions_created, to: :failed
         transitions from: :employments_created, to: :failed
         transitions from: :cfe_not_called, to: :failed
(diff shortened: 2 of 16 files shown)


spec/requests/providers/means_summaries_spec.rb
```diff
@@ -82,84 +84,123 @@
...
-          expect(response).to redirect_to flow_forward_path
-        end
-
         it "transitions to provider_entering_merits state" do
+          request
           expect(legal_aid_application.reload.provider_entering_merits?).to be true
         end
 
-        context "when provider does not have bank_statement_upload permissions" do
+        context "when provider has bank_statement_upload permissions" do
```
RoseSAK, line 114: I guess you could update this context to explicitly say that enhanced bank upload is not enabled? minor nitpick though

spec/requests/providers/means_summaries_spec.rb
```diff
@@ -82,84 +84,123 @@
...
+          it "calls CFE::SubmissionManager" do
+            request
+            expect(CFE::SubmissionManager).to have_received(:call).with(legal_aid_application.id)
+          end
         end
 
-        context "Form submitted using Save as draft button" do
-          let(:params) { { draft_button: "Save as draft" } }
+        context "when form submitted using Save as draft button" do
+          let(:params) { { draft_button: "irrelevant" } }
```
RoseSAK, line 167: Is this because it doesn't matter what the text is on the 'Save as draft' button?

RoseSAK (approved):
I'm unfamiliar with this part of the application but reading through it all looks good and makes sense to me 💯 
Have added one q and a minor nit comment
Good job, looks like a lot of work! 👍

app/services/cfe/service_set.rb
```diff
@@ -40,10 +56,26 @@ def initialize(object)
     end
 
     def call
+      sets[key]
+    end
```
cpjmcquillan, line 60: I'm not suggesting we make this change in this PR, but I'm not entirely comfortable with the interface here. 

To me at least, I feel like we could reveal more intent than we are currently. We're working in a complex domain already, so any place we can reduce cognitive load feels like a win!

This class is called in one place: `app/services/cfe/submission_manager.rb:34`
```ruby
def services
  ServiceSet.call(submission)
end
```

I wonder whether we could refactor this `ServiceSet` class slightly. 

```ruby
module CFE
  class ServiceSet
    def self.for(submission)
      new(submission).services
    end
    
    def services
      services_for_submission
    end
    
    private
    
    def services_for_submission
      if submission.passported?
        PASSPORTED_SERVICES
      elsif ...
        ...
      end
    end
  end
end
```

This would address a couple of things I found myself thinking when I read the code:
  - what does `object` actually refer to?
  - what does `call` actually do?
  
Ultimately this isn't a big deal, and perhaps we're wedded to the `ServiceObject/.call` pattern —but it's just some thoughts I had 😄

cpjmcquillan (approved):
LGTM

spec/requests/providers/means_summaries_spec.rb
jsugarman: yeah, this is [because only presence](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/2fde11d007f1ff351d165fa84cd69b1a8dfb6e6f/app/controllers/concerns/providers/draftable.rb#L47-L49) is checked for

app/models/concerns/cfe_submission_state_machine.rb
```diff
@@ -71,14 +72,20 @@ module CFESubmissionStateMachine
 
       event :irregular_income_created do
         transitions from: :other_income_created, to: :irregular_income_created, guard: :non_passported?
+        transitions from: :dependants_created, to: :irregular_income_created, guard: :non_passported?
```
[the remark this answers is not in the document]
jsugarman: Many, if not all, of these transitions need to be removed, particularly if we implement an asynchronous calling CFE approach - which is being looked at by Colin.

app/services/flow/flows/provider_capital.rb
```diff
@@ -200,7 +200,12 @@ class ProviderCapital < FlowSteps
         means_summaries: {
           path: ->(application) { urls.providers_legal_aid_application_means_summary_path(application) },
           forward: lambda do |application|
-            application.uploading_bank_statements? ? :no_eligibility_assessments : :capital_income_assessment_results
+            # TODO: this bank statment logic can be removed once enchanced_bank_statement is released and deemed stable
+            if application.uploading_bank_statements?
+              application.using_enhanced_bank_upload? ? :capital_income_assessment_results : :no_eligibility_assessments
+            else
+              :capital_income_assessment_results
+            end
```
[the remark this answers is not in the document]
jsugarman: not sure, i can see its more succinct and nicer for that reason but negative boolean logic is, for me at least, always a head scratcher. A passported application is `!application.uploading_bank_statements?` for example - not that that kind should ever reach this step.

app/services/cfe/service_set.rb
jsugarman: yeah, lets simplify for now

jsugarman: but the point of `object` is to highlight the decoupled nature of this class. This would work for an application object too

app/models/concerns/cfe_submission_state_machine.rb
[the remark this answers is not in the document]
jsugarman: I add an integration test in the `spec/services/cfe/submission_manager_spec.rb` that covers this entire state machine flow, which would fail without this change. the `spec/concerns/cfe_submission_state_machine_spec.rb` unit specs are missing a lot transition event tests

app/controllers/providers/means_summaries_controller.rb
```diff
@@ -16,6 +16,17 @@ def update
 
   private
 
+    # TODO: if show is always called before update and converts
+    # state to `checking_non_passported_means` then I do not see
+    # how legal_aid_application.provider_entering_merits? can ever be true!
```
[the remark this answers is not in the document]
jsugarman: we can try it in separate PR

PR #4338, opened 2022-09-26
Fix flaky cucumber upload tests

cpjmcquillan:
Several cucumber tests that rely on uploading files are flaky, and have been failing at a higher rate recently.

They seem to fail because validations that ensure supporting evidence has been uploaded have not been met. This is likely due to latency in the system.

This adds assertions that the files have been uploaded before attempting to navigate to the next step of the user journey in two of these tests.

The intention of this change is to ensure that the file has been uploaded, but it also ensures we are testing more realistic and appropriate user behaviour (i.e we care that the user knows the file has been uploaded too).

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3505)


colinbruce (approved):
LGTM :+1:

🤞

jsugarman (approved):
👍 LGTM

PR #4339, opened 2022-09-26
Use clamav in tests when explicitly tagged

cpjmcquillan:
Before, clamav was used for malware scanning in all tests that interacted with the `MalwareScanner` class (either directly or indirectly).

While it is important and useful to test the entire stack in integration tests, clamav is a source of repeated flakiness across the test suite due to the latency it introduces (especially in CI).

This updates RSpec specs and Cucumber tests to mock a successful clamav response
unless they are explicitly tagged with `clamav: true` or `@clamav`.

### DIFF
--- a/features/providers/bank_statement_upload.feature
+++ b/features/providers/bank_statement_upload.feature
@@ -1,6 +1,6 @@
+@javascript @clamav
 Feature: Bank statement upload
 
-  @javascript
   Scenario: Uploading a file
     Given csrf is enabled
     And I have completed a non-passported employed application and reached the open banking consent with bank statement upload enabled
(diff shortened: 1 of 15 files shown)


jsugarman (approved):
🎖️ Brill. 

You could, perhaps, create a helper module or similar,  `clamav_helper`, to centralise the logic for rspec and cucumber

```
module ClamavHelper
  def stub_clamav
    stdout = "/some/file/path.pdf: OK"
    stderr = ""
    status = instance_double(Process::Status, success?: true, exitstatus: 0)
    allow(Open3).to receive(:capture3).and_call_original
    allow(Open3)
        .to receive(:capture3)
        .with(/clamdscan/, any_args)
        .and_return([stdout, stderr, status])
  end
end
```

cpjmcquillan:
Merging this in place of #4348. 

If we see any flaky failures related to document uploads after this (that are tagged with `clamav`), we should adopt the approach taken in #4338.

PR #4349, opened 2022-09-27
AP-3482 add list of disregarded benefits to regular incomes page

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3482)

Add details section to the regular income page which, when expanded, shows a list of all disregarded benefits, so the providers know which benefits to exclude from the benefits amounts. 
Added cucumber test to test some of the content for this section. 

## Checklist


features/support/steps_helper.rb
```diff
@@ -42,6 +42,11 @@
   all(:link, link_name).last.click
 end
 
+Then("I open the section {string}") do |section|
+  section = page.find(".govuk-details__summary-text", text: section)
```
jsugarman, line 46: **v.minor nit**, do the capybara built-ins work. Just thinking that sometimes I have seen  `page.find`  fail as it may not "wait", where as built ins all have a built in "wait" timeout i believe

```
click_link('id-of-link')
click_link('Link Text')
```

jsugarman (approved):
👍 LGTM, just one question/suggestion

features/support/steps_helper.rb
cpjmcquillan: To add to this, if we really want to get serious about accessibility-first test-driven-development, we might want to look into using [this gem from the citizensadvice team](https://github.com/citizensadvice/capybara_accessible_selectors).

That would allow us to use accessible capybara selectors like [`within_disclosure`](https://github.com/citizensadvice/capybara_accessible_selectors#within_disclosurename-find_options-block) in our tests:

```ruby
within_disclosure "Disregarded benefits" do
  expect(page).to have_css("h2", text: "Other benefits")
end
```

cpjmcquillan (approved):
I added a comment about not using an instance variable in a partial. 

I think that's worth addressing, but we don't have `erblint` configured to complain about it _yet_ —so you could ignore me if you wanted!

features/support/steps_helper.rb
RoseSAK: This gem looks great, and I think also had the added advantage of making the capybara steps more readable 💯 I would definitely be on board with adding this, using it to write new steps and maybe over time rewriting current steps that we have already

app/views/providers/means/regular_incomes/_disregarded_benefits.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= govuk_details(summary_text: t('.details_title')) do %>
+  <p class="govuk-body"><%= t('.details_text') %></p>
+  <% @form.disregarded_benefit_categories.each do |category| %>
```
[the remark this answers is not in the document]
RoseSAK: Good point, will update this

PR #4354, opened 2022-09-28
Define backoffLimit for jobs, with default as fallback

jsugarman:
## What
Reduce numbers of failed job pods on UAT

**backoffLimit**

UAT environments are erroring when perform many of the jobs
due to database pods being recycled and not migrated/seeded.
The large numbers of errored job pods were persisting in the UAT
namespace eating up pods unnecessarily and could, in-extremis,
cause deployment failures due to pod limits?!

The [backoffLimit](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy) allows erroring pods to be limited to x, which is 6 
by default in kubernetes so has been set to 6 by default if value not set 
in `Values` file

Combined with a restartPolicy of "Never" this will only allow
2 pods to error (quite why it is 2 is unclear but i surmise it is
one initial attempt plus one retry) before it marks it as failed
and does not retry.

**failedJobsHistoryLimit**

_NOTE: Could remove this change as it has no impact on number of errored pods 
persisting for UAT._

Also, this defines a variable `failedJobsHistoryLimit` with default as fallback

This sets the failedJobsHistoryLimit via a helm `Value` entry
and applies this in each cronjob template with a default which matches
the current hardcoded value - the default is applied if no value
exists in the respective Values yaml file for the environment.

However, this does NOT impact the number of pods spun up when
errors occur. A setting of zero also seems to be ignored - you
can test this by "describing" any cronjob with a failedJobsHistoryLimit of
zero, `kubectl describe cronjob <name of cronjob>`, and checking the setting.

## Checklist


helm_deploy/apply-for-legal-aid/templates/cronjob-destroy-purgeable.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   schedule: '30 1 * * *'
   concurrencyPolicy: Forbid
   successfulJobsHistoryLimit: 0
-  failedJobsHistoryLimit: 1
+  failedJobsHistoryLimit: {{ .Values.jobs.failedJobsHistoryLimit | default 0 }}
```
cpjmcquillan, line 14:
```suggestion
  failedJobsHistoryLimit: {{ .Values.jobs.failedJobsHistoryLimit | default 1 }}
```

cpjmcquillan (approved):

PR #4355, opened 2022-09-28
AP-3459 regular payment frequencies

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3459)

Add correct frequency options for each transaction type on the regular income and outgoings pages

NB: this branch was rebased on AP-3460-enhanced-bank-upload-regular-outgoings so it might be worth waiting until that branch is merged to review, as at the moment this PR contains all the changes from that branch too

## Checklist

### DIFF
--- a/app/forms/providers/means/regular_income_form.rb
+++ b/app/forms/providers/means/regular_income_form.rb
@@ -76,10 +76,6 @@ module Providers
         "none"
       end
 
-      def frequencies(transaction_type)
-        transaction_type.name == "benefits" ? RegularTransaction::FREQUENCIES.without("monthly") : RegularTransaction::FREQUENCIES
-      end
-
     private
 
       def assign_regular_transaction_attributes
--- a/app/forms/providers/means/regular_income_form.rb
+++ b/app/forms/providers/means/regular_income_form.rb
@@ -76,6 +76,10 @@ module Providers
         "none"
       end
 
+      def frequencies(transaction_type)
+        transaction_type.name == "benefits" ? RegularTransaction::FREQUENCIES.without("monthly") : RegularTransaction::FREQUENCIES
+      end
+
     private
 
       def assign_regular_transaction_attributes
--- a/app/models/regular_transaction.rb
+++ b/app/models/regular_transaction.rb
@@ -1,5 +1,5 @@
 class RegularTransaction < ApplicationRecord
-  FREQUENCIES = %w[weekly monthly].freeze
+  FREQUENCIES = %w[every_week every_two_weeks every_four_weeks monthly total_in_last_three_months].freeze
 
   belongs_to :legal_aid_application
   belongs_to :transaction_type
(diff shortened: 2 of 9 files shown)


app/forms/providers/means/regular_income_form.rb
```diff
@@ -66,6 +66,10 @@ def none_selected
         "none"
       end
 
+      def frequencies(transaction_type)
+        transaction_type.name == "benefits" ? RegularTransaction::FREQUENCIES.without("monthly") : RegularTransaction::FREQUENCIES
+      end
```
cpjmcquillan, line 71: How do you feel about adding a class method on the `RegularTransaction` model instead and adding some unit tests?

Maybe something like this:

```ruby
class RegularTransaction
  def self.frequencies_for(transaction_type)
    if transaction_type.name == "benefits"
      FREQUENCIES.without("monthly")
    else
      FREQUENCIES
    end
  end
end
```

I feel like the context of regular transaction frequencies is broader than this form specifically, so this might help capture that. 

What do you think?

app/models/regular_transaction.rb
```diff
@@ -1,5 +1,5 @@
 class RegularTransaction < ApplicationRecord
-  FREQUENCIES = %w[weekly monthly].freeze
+  FREQUENCIES = %w[every_week every_two_weeks every_four_weeks monthly total_in_last_three_months].freeze
```
jsugarman, line 2: be good too use the expected values in CFE so they can be persisted and passed across to cfe as is. You can see the [current values here](https://github.com/ministryofjustice/check-financial-eligibility/blob/main/app/constants/cfe_constants.rb#L72)

app/forms/providers/means/regular_income_form.rb
RoseSAK: Yeah that makes sense, I've updated it to basically move this method to the RegularTransaction model

app/models/regular_transaction.rb
RoseSAK: Have updated it to use these values in the backend and render the right content in the frontend :)

spec/models/regular_transaction_spec.rb
```diff
@@ -97,4 +97,28 @@
...
+    end
+
+    context "when transaction type is not benefits" do
+      it "includes monthly in the frequencies" do
+        child_care = create(:transaction_type, :child_care)
+
+        frequencies = described_class.frequencies_for(child_care)
+
+        expect(frequencies.count).to eq(5)
+        expect(frequencies).to include("monthly")
```
RoseSAK, line 120: I followed the existing pattern for this file

spec/models/regular_transaction_spec.rb
```diff
@@ -97,4 +97,28 @@
...
+      end
+    end
+  end
```
jsugarman, line 123: My personal preference for rspec DSL tests would be, but no biggy!

```suggestion
  describe ".frequencies_for" do
    subject(:frequencies) { described_class.frequencies_for(transaction_type) }

    context "when transaction type is benefits" do
      let(:transaction_type) { create(:transaction_type, :benefits) }

      it "does not include monthly in the frequencies" do
        expect(frequencies).to include(%w[weekly two_weekly four_weekly three_monthly])
      end
    end

    context "when transaction type is not benefits" do
      let(:transaction_type) { create(:transaction_type, :child_care) }

      it "includes monthly in the frequencies" do
        expect(frequencies).to include(%w[weekly two_weekly four_weekly monthly three_monthly])
      end
    end
  end
```

jsugarman (approved):
LGTM 👍  but suggestion on the tests

PR #4366, opened 2022-09-30
Providers can submit housing benefit data

cpjmcquillan:
As part of the enhanced bank upload journey, providers need to tell us if their client is in receipt of housing benefit or not. If they do receive housing benefit, then they need to tell us how much, and how often.

If providers tell us that their client makes housing payments, then they are asked whether the client receives housing benefit.

If they do, the information is used to create a regular transaction that can be submitted to the CFE API to give an eligbility assessment result.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3462)

https://user-images.githubusercontent.com/25187547/193324168-e225cf09-4d18-485a-bf7e-2153415959ed.mp4

### DIFF
--- a/app/forms/providers/means/housing_benefit_form.rb
+++ b/app/forms/providers/means/housing_benefit_form.rb
@@ -1,0 +1,117 @@
+module Providers
+  module Means
+    class HousingBenefitForm
+      include ActiveModel::Model
+
+      attr_accessor :housing_benefit,
+                    :housing_benefit_amount,
+                    :housing_benefit_frequency
+
+      attr_reader :legal_aid_application
+
+      validates :housing_benefit, inclusion: { in: %w[true false] }
+
+      validates :housing_benefit_amount,
+                currency: { greater_than: 0 },
+                if: :applicant_in_receipt_of_housing_benefit?
+
+      validates :housing_benefit_frequency,
+                inclusion: { in: ->(form) { form.frequency_options } },
+                if: :applicant_in_receipt_of_housing_benefit?
+
+      def initialize(params = {})
+        @legal_aid_application = params.delete(:legal_aid_application)
+
+        if params[:housing_benefit].nil?
+          assign_housing_benefit_attributes
+        end
+
+        super
+      end
+
+      def save
+        return false unless valid?
+
+        ApplicationRecord.transaction do
+          legal_aid_application.applicant_in_receipt_of_housing_benefit = housing_benefit
+
+          if applicant_in_receipt_of_housing_benefit?
+            build_housing_benefit_transaction_type
+            build_housing_benefit_regular_transaction
+          else
+            destroy_housing_benefit_transaction_type!
+            destroy_housing_benefit_regular_transaction!
+          end
+
+          legal_aid_application.save!
+        end
+
+        true
+      end
+
+      def frequency_options
+        RegularTransaction.frequencies_for(housing_benefit_transaction_type)
+      end
+
+    private
+
+      def applicant_in_receipt_of_housing_benefit?
+        housing_benefit == "true"
+      end
+
+      def assign_housing_benefit_attributes
+        @housing_benefit = @legal_aid_application.applicant_in_receipt_of_housing_benefit
+        @housing_benefit_amount = housing_benefit_regular_transaction&.amount
+        @housing_benefit_frequency = housing_benefit_regular_transaction&.frequency
+      end
+
+      def build_housing_benefit_transaction_type
+        legal_aid_application
+          .legal_aid_application_transaction_types
+          .find_or_initialize_by(
+            transaction_type: housing_benefit_transaction_type,
+          )
+      end
+
+      def build_housing_benefit_regular_transaction
+        housing_benefit_transaction = legal_aid_application
+          .regular_transactions
+          .find_or_initialize_by(
+            transaction_type: housing_benefit_transaction_type,
+          )
+        housing_benefit_transaction.assign_attributes(
+          amount: housing_benefit_amount,
+          frequency: housing_benefit_frequency,
+        )
+        housing_benefit_transaction.save!
+      end
+
+      def destroy_housing_benefit_transaction_type!
+        legal_aid_application
+          .legal_aid_application_transaction_types
+          .includes(:transaction_type)
+          .where(transaction_type: { name: "housing_benefit" })
+          .destroy_all
+      end
+
+      def destroy_housing_benefit_regular_transaction!
+        legal_aid_application
+          .regular_transactions
+          .includes(:transaction_type)
+          .where(transaction_type: { name: "housing_benefit" })
+          .destroy_all
+      end
+
+      def housing_benefit_regular_transaction
+        @legal_aid_application
+          .regular_transactions
+          .includes(:transaction_type)
+          .find_by(transaction_type: { name: "housing_benefit" })
+      end
+
+      def housing_benefit_transaction_type
+        TransactionType.find_by!(name: "housing_benefit")
+      end
+    end
+  end
+end
--- a/app/forms/providers/means/regular_outgoings_form.rb
+++ b/app/forms/providers/means/regular_outgoings_form.rb
@@ -87,6 +87,10 @@ module Providers
         destroy_legal_aid_application_transaction_types!
         destroy_regular_outgoing_transactions!
         destroy_cash_outgoing_transactions!
+
+        unless housing_payments_selected?
+          destroy_housing_benefit_transactions!
+        end
       end
 
       def destroy_legal_aid_application_transaction_types!
@@ -109,6 +113,24 @@ module Providers
           .cash_transactions
           .debits
           .where.not(transaction_type_id: transaction_type_ids)
+          .destroy_all
+      end
+
+      def housing_payments_selected?
+        transaction_types.exists?(name: "rent_or_mortgage")
+      end
+
+      def destroy_housing_benefit_transactions!
+        legal_aid_application
+          .legal_aid_application_transaction_types
+          .includes(:transaction_type)
+          .where(transaction_type: { name: "housing_benefit" })
+          .destroy_all
+
+        legal_aid_application
+          .regular_transactions
+          .includes(:transaction_type)
+          .where(transaction_type: { name: "housing_benefit" })
           .destroy_all
       end
 
--- a/app/forms/providers/means/regular_outgoings_form.rb
+++ b/app/forms/providers/means/regular_outgoings_form.rb
@@ -3,12 +3,18 @@ module Providers
     class RegularOutgoingsForm
       include ActiveModel::Model
 
-      OUTGOING_TYPES = %w[rent_or_mortgage child_care maintenance_out legal_aid].freeze
+      OUTGOING_TYPES = %w[
+        rent_or_mortgage
+        child_care
+        maintenance_out
+        legal_aid
+      ].freeze
 
       attr_reader :transaction_type_ids, :legal_aid_application
 
       OUTGOING_TYPES.each do |outgoing_type|
-        attr_accessor "#{outgoing_type}_amount".to_sym, "#{outgoing_type}_frequency".to_sym
+        attr_accessor "#{outgoing_type}_amount".to_sym,
+                      "#{outgoing_type}_frequency".to_sym
       end
 
       validates :transaction_type_ids, presence: true, unless: :none_selected?
@@ -17,7 +23,8 @@ module Providers
       def initialize(params = {})
         @none_selected = none_selected.in?(params["transaction_type_ids"] || [])
         @legal_aid_application = params.delete(:legal_aid_application)
-        @transaction_type_ids = params["transaction_type_ids"] || @legal_aid_application.transaction_type_ids
+        @transaction_type_ids = params["transaction_type_ids"] ||
+          @legal_aid_application.transaction_types.debits.not_children.pluck(:id)
 
         assign_regular_transaction_attributes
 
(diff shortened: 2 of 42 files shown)


app/models/transaction_type.rb
```diff
@@ -36,7 +38,7 @@ class TransactionType < ApplicationRecord
 
   scope :active, -> { where(archived_at: nil) }
   scope :debits, -> { active.where(operation: :debit) }
-  scope :credits, -> { active.where(operation: :credit) }
+  scope :credits, -> { active.where(operation: :credit).where.not(name: DISREGARDED_BENEFITS) }
```
cpjmcquillan, line 41: I think we need to think about how housing benefit transactions map onto our existing data model.

It makes sense that housing benefit is a regular transaction for the purposes of obtaining a financial eligibility result from CFE, but I'm not sure it makes sense across the application.

We already have an `excluded_benefits` transaction type, which seems to correspond to the domain concept of "Disregarded benefits".

It doesn't seem like this was implemented with the intention of there being multiple varieties of `excluded_benefits` transaction types, but I'm not really sure what the original intention was anyway.

It looks like the concept was introduced in #1425, and iterated on in #1447 and #1575 —but it's hard to glean much context from the commit messages.

Given `housing_benefit` is a type of `excluded_benefit`, I'm not sure what else we need to consider for this to go live.

- We want to submit it to CFE as a regular transaction
- We don't want to display it in forms alongside other income types

But, 
- Do we want to display it as a category for bank transactions?
- Are there things we need to consider in terms of means reports etc?

I'm not sure who knows the answers to these questions, so I'm interested to hear if anyone can shed some light on any of this!

jsugarman: > Do we want to display it as a category for bank transactions?

I do not believe so. There is "supposed to be" automatic categorisation of bank transactions as a state benefit of type housing_benefit if the description of the bank transaction contains the works "DWP" and "HB", however, I do not believe it is working as intended following some commentary from StephenR a few weeks ago. But it may need wider discussion with the BA and others

jsugarman: > Are there things we need to consider in terms of means reports etc?

That will hopefully come out of [AP-3498](https://dsdmoj.atlassian.net/browse/AP-3498) and the [design](https://docs.google.com/document/d/1LbPEoEM3tlqRt3KeCmD7hNsRi9dvzLzgQEMnWIbOsjI/edit)

jsugarman: > We want to submit it to CFE as a regular transaction

Yes, that is already handled by Apply's `CFE::SubmissionManager` and is accounted for in CFE (excluded from income, but negated from gross housing costs if necessary)

app/forms/providers/means/housing_benefit_form.rb
```diff
@@ -0,0 +1,113 @@
+module Providers
+  module Means
+    class HousingBenefitForm
+      include ActiveModel::Model
+
+      attr_accessor :housing_benefit,
```
jsugarman, line 6: naming wise, would `housing_benefit_receipt` be clearer for this attribute

jsugarman (changes requested):
Good work, it looks ok

Aside from some nits the are some logic problems with the flow when I UAT'd this. 

I believe these stem from the outgoings page not deleting housing benefit regular transactions if the housing payments transaction type is deselected/removed/deleted. This could lead to the the housing benefit record being submitted to CFE and Ideally should be dealt with as part of this PR.

spec/factories/transaction_types.rb
```diff
@@ -12,16 +12,17 @@ def repeating_sequence_for_transaction_type(type, current)
 
 FactoryBot.define do
   factory :transaction_type do
-    name { TransactionType::NAMES.values.flatten.sample }
-    operation { TransactionType::NAMES.keys.sample }
```
cpjmcquillan, line 16: We want our tests to be deterministic, so I don't think we should be doing things like `.sample` in our factories (or using Faker!)

app/models/transaction_type.rb
```diff
@@ -18,6 +18,7 @@ class TransactionType < ApplicationRecord
       maintenance_out
       legal_aid
     ],
+    null: %i[housing_benefit],
```
cpjmcquillan, line 21: I thought about how to approach this for quite a while, and I think this approach is reasonable —even though I'm not entirely satisfied.

I'm not convinced `null` is the right "operation", and I'm concerned it might add confusion. 

Having said that, I do think it conveys some meaning. Housing benefit transactions are neither credits or debits in the conventional sense, and we need to distinguish them somehow.

RoseSAK: Makes sense to me I think 👍

app/views/providers/means/housing_benefits/show.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_with model: @form,
+              url: providers_legal_aid_application_means_housing_benefits_path(@legal_aid_application),
+              method: :patch do |f| %>
+  <%= page_template page_title: page_title, template: :basic, form: f do %>
+    <%= f.govuk_radio_buttons_fieldset :housing_benefit, legend: {tag: "h1", size: "xl"} do %>
+      <%= f.govuk_radio_button :housing_benefit, "true", link_errors: true do %>
+        <%= f.govuk_text_field :housing_benefit_amount, width: "one-quarter", prefix_text: "£" %>
+        <%= f.govuk_collection_radio_buttons :housing_benefit_frequency,
+                                             RegularTransaction::FREQUENCIES,
```
RoseSAK, line 9: Should/does this display all the regular transaction frequencies? On the page in Miro it only has every week, every two weeks and every four weeks as options

config/locales/en/activemodel.yml
```diff
@@ -491,6 +491,16 @@ en:
               file_virus: "%{file_name} contains a virus"
               no_file_chosen: Upload your client's bank statements
               system_down: There was a problem uploading your file - try again
+        providers/means/housing_benefit_form:
+          attributes:
+            housing_benefit:
+              inclusion: Select one of the options
```
RoseSAK, line 497: Not sure if this has been content checked but I think usually for yes/no radio button pages, we have the none selected error message as something more informative like 'Select if your client receives housing benefit' or 'Select yes if your client receives housing benefits' - maybe one to check with Chris?

RoseSAK (approved):
This looks good to me! Sorry it took me ages to review

I have left a couple of comments but they are pretty minor 

💯 🙌

app/views/providers/means/housing_benefits/show.html.erb
RoseSAK: I have just realised this branch doesn't include the updated frequencies as that branch hasn't been merged to main yet. Once these are combined might just need to add a helper method to restrict the frequencies to those mentioned on the miro board (assuming these are accurate)

app/models/transaction_type.rb
jsugarman: It feels to me like it is fundamentally a credit and it will need to be sent to CFE as a `credit` operation as it currently stands.

jsugarman (commented):
As it currently stands the use of "null" as an operation "type" results in an error during CFE submission. 

Example error.
```
  "Unprocessable entity: URL: http://localhost:5050/assessments/371ae0af-c888-4fd4-afea-cffb8ef38b9d/regular_transactions, details: The property '#/regular_transactions/2/operation' value \"null\" did not match one of the following values: credit, debit in schema file://public/schemas/regular_transactions.json",
```

Lets discuss as it feels like the "null" approach could be replaced by and existing pattern that uses parent_id, plus use of scopes, for example

```
 #<TransactionType:0x00000001168ba328
  id: "056941ae-558c-4ef8-adbf-d1468bde7284",
  name: "housing_benefit",
  operation: "credit",
  created_at: Fri, 07 Oct 2022 14:09:03.962368000 BST +01:00,
  updated_at: Fri, 07 Oct 2022 14:09:03.962368000 BST +01:00,
  sort_order: 2000,
  archived_at: nil,
  other_income: false,
  parent_id: <uuid-for-benefits-transaction-type>]
```
And then you can use existing scope which is without `excluded_benefits` and `housing_benefit`, and which you are already using the regular_income_form (and regular_outgoing_form) by the looks of it

```
TransactionType.credits.not_children
```

cpjmcquillan:
> As it currently stands the use of "null" as an operation "type" results in an error during CFE submission.

Yeah, I thought this would probably happen.

How concrete are the operations in CFE? Are there any drawbacks to us adding an operation type? 

Does CFE consider housing benefit regular transactions the same as the other credit transactions? (i.e are they all summed in the same way?)

---

> Lets discuss as it feels like the "null" approach could be replaced by an existing pattern that uses parent_id

Sure. I saw that pattern, but I really couldn't decipher what the intent of it was from the code or commit history. If we do use that approach, it's probably worthwhile improving the readability and adding some context and clarity.

Would we consider `housing_benefits` a child of `benefits` or `excluded_benefits`? 

From my perspective it sounds like `housing_benefits` are a type of "disregarded benefit" from a business perspective, and `excluded_benefits` are synonymous with "disregarded benefits" in our application. 

It would be get that relationship cleared up, so we can map the domain to the code more clearly.

---

I've invited you to a pairing/mobbing session on Monday afternoon, but I'll pick this up again Monday morning and see where I get with the `parent/child` approach.

I suspect it might be worthwhile leaning on Dave's BA knowledge too (at least for me), because the domain still feels a bit murky.

jsugarman:
As mentioned in our tuple session the Housing benefits were coming up in the identify_types_of_income_controller page, as below, when I UAT'ed the branch deployment. However the discussion on the way forward will almost certainly eliminate that in any event, so just mentioning here for clarity. 🤔 perhaps we should extend any applicable tests to check that housing benefit is excluded.

Screenshot 2022-10-10 at 13 17 27

app/models/transaction_type.rb
```diff
@@ -45,55 +50,39 @@ def self.for_income_type?(transaction_type_name)
...
     parent_id.present?
   end
 
-  def parent?
-    TransactionType.where(parent_id: id).any?
-  end
-
-  def parent
-    TransactionType.find_by(id: parent_id)
-  end
```
cpjmcquillan, line 90: These methods aren't used anywhere else, so I've removed them in the interests of keeping the public API small.

app/models/transaction_type.rb
```diff
@@ -45,55 +50,39 @@ def self.for_income_type?(transaction_type_name)
...
   end
 
-  def self.any_type_of(name)
-    top_level_id = TransactionType.find_by(name:)&.id
-    return [] if top_level_id.nil?
-
-    find_with_children(top_level_id)
+  def label_name(journey: :citizens)
+    I18n.t("transaction_types.names.#{journey}.#{name}")
   end
```
cpjmcquillan, line 63: Similarly, these methods were no longer used, so they have been removed.

spec/forms/providers/means/regular_outgoings_form_spec.rb
```diff
@@ -122,6 +122,51 @@
...
+          transaction_type: maintenance_out,
+          amount: 250,
+          frequency: "weekly",
+        )
+
+        form = described_class.new(legal_aid_application:)
+
+        expect(form.transaction_type_ids).to contain_exactly(maintenance_out.id)
+        expect(form.maintenance_out_amount).to eq(250)
+        expect(form.maintenance_out_frequency).to eq("weekly")
```
jsugarman, line 165: **not a blocker**, either way, but prefer single expectations if possible

```suggestion
        expect(form).to have_attributes(transaction_type_ids: [maintenance_out.id],
                                        maintenance_out_amount: 250,
                                        maintenance_out_frequency: "weekly)
```

app/forms/providers/means/housing_benefit_form.rb
```diff
@@ -0,0 +1,117 @@
...
+          amount: housing_benefit_amount,
+          frequency: housing_benefit_frequency,
+        )
+        housing_benefit_transaction.save!
+      end
+
+      def destroy_housing_benefit_transaction_type!
+        legal_aid_application
+          .legal_aid_application_transaction_types
+          .includes(:transaction_type)
```
jsugarman, line 92: **nit** I don't believe we need this `include` if we are just trying to destroy the join table record

app/forms/providers/means/housing_benefit_form.rb
```diff
@@ -0,0 +1,117 @@
...
+          .where(transaction_type: { name: "housing_benefit" })
+          .destroy_all
+      end
+
+      def destroy_housing_benefit_regular_transaction!
+        legal_aid_application
+          .regular_transactions
+          .includes(:transaction_type)
```
jsugarman, line 100: **nit** I don't believe we need this `include` if we are just trying to destroy the join table record

app/forms/providers/means/regular_outgoings_form.rb
```diff
@@ -112,6 +116,24 @@ def destroy_cash_outgoing_transactions!
...
+      def destroy_housing_benefit_transactions!
+        legal_aid_application
+          .legal_aid_application_transaction_types
+          .includes(:transaction_type)
+          .where(transaction_type: { name: "housing_benefit" })
+          .destroy_all
+
+        legal_aid_application
+          .regular_transactions
+          .includes(:transaction_type)
```
jsugarman, line 132: **nit** I don't believe we need this `include` if we are just trying to destroy the join table record

spec/factories/transaction_types.rb
jsugarman: Generally agree that non-detereministic test data is a pain - probably worthy of a wider discussion as limited non-deterministic test data, over time, can identify bugs that a deterministic will not. overall 👍

RoseSAK (approved):
All looks good to me 👍 🎉

jsugarman (approved):
LGTM 👍

AM going to UAT it this PM as well

jsugarman:
UAT of this branch checking the "Check your answers" for truelayer journey reveals the same problem so should be fixed here. My [PR here should fix](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4428)

app/forms/providers/means/housing_benefit_form.rb
cpjmcquillan: If we want to filter by `transaction_type: { name: "housing_benefit" }` we do

PR #4386, opened 2022-10-04
Refresh our SimpleCov config

cpjmcquillan:
This makes some small (but hopefully useful) changes to our SimpleCov
configuration.

1. Use SimpleCov for Cucumber feature tests

In the absence of traditional Rails system (feature) tests, we use Cucumber to
write feature level integration tests. Until now, though, our coverage reports
have not taken into account the code covered by these important tests. This
seemed like an oversight, so it has been added.

2. Simplify coverage configuration

Now that SimpleCov was being invoked in several places, it made sense to
centralise configuration in a `.simplecov` file as [recommended by the
documentation].

This simplification means we can inline the work to process coverage in the task
itself.

[recommended by the documentation]: https://github.com/simplecov-ruby/simplecov#using-simplecov-for-centralized-config

3. Add branch coverage

Branch coverage is useful because it can highlight untested conditional logic in
the codebase. Currently there is no minimum level (we should do some work to
address this before setting it), however no drop in coverage is allowed.


spec/simplecov_helper.rb
```diff
@@ -1,20 +0,0 @@
...
-module SimpleCovHelper
-  def self.process_coverage
-    SimpleCov.start do
-      add_filter "config/initializers/"
-      add_filter "spec/"
-      add_filter "services/migration_helpers/"
-      add_filter "config/environments/"
-
-      minimum_coverage(line: 100)
-      merge_timeout 3600
```
cpjmcquillan, line 13: This line didn't make it into the new config, because it wasn't actually doing anything beforehand.

`SimpleCov.collate` ignores timestamps by default.

.simplecov
```diff
@@ -0,0 +1,13 @@
+SimpleCov.start "rails" do
+  add_filter "config/"
```
cpjmcquillan, line 2: I updated this to filter everything In `config/`

.simplecov
```diff
@@ -0,0 +1,13 @@
...
+  add_filter "spec/"
+  add_filter "services/migration_helpers/"
+  add_filter "lib/tasks"
```
cpjmcquillan, line 5: I added this filter because they weren't covered. I'm not sure why it wasn't picked up beforehand.

spec/spec_helper.rb
```diff
@@ -21,23 +21,6 @@
...
-    add_filter "config/environments/"
-    minimum_coverage 100 unless ENV["CIRCLE_JOB"]
-  end
-
-  unless ENV["CIRCLE_JOB"]
-    SimpleCov.at_exit do
-      say("<%= color('Code coverage below 100%', RED) %>") if SimpleCov.result.coverage_statistics[:line].percent < SimpleCov.minimum_coverage[:line]
-      SimpleCov.result.format!
-    end
-  end
```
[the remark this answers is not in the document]
cpjmcquillan: I tried to add it back, but this line was raising `NoMethodError` after Cucumber tests

```ruby
say("<%= color('Code coverage below 100%', RED) %>")
```

I've left in the result formatting part though.

jsugarman (approved):
I agree with all three points and this is very useful IMO.

.simplecov
jsugarman: we do have some specs for `lib/tasks/helpers/` which are getting excluded by this but not sure why it was not complaining about other unspeced  `lib/tasks/helpers/` though so 🤷‍♂️

cpjmcquillan:
Added a task to the backlog to think about branch coverage.

https://dsdmoj.atlassian.net/browse/AP-3542

.simplecov
cpjmcquillan: Ah ok, let's remove this filter then and add tests for what is missing!

cpjmcquillan:
Added a task to the backlog to address the `lib/tasks` helpers missing coverage.

https://dsdmoj.atlassian.net/browse/AP-3548

PR #4388, opened 2022-10-04
Use strong_migrations gem

cpjmcquillan:
As the application continues to scale, we will need to think more 
about how we do safe migrations.

This gem reduces some of that cognitive load by catching unsafe 
migrations in development and giving developers insight on how 
to mitigate problematic migrations.

Default configuration has been applied, except the targeted 
database version has been set to 10.

### DIFF
--- a/config/initializers/strong_migrations.rb
+++ b/config/initializers/strong_migrations.rb
@@ -1,0 +1,28 @@
+# Mark existing migrations as safe
+# rubocop:disable Style/NumericLiterals
+StrongMigrations.start_after = 20221017093914
+# rubocop:enable Style/NumericLiterals
+
+# Set timeouts for migrations
+# If you use PgBouncer in transaction mode, delete these lines and set timeouts on the database user
+StrongMigrations.lock_timeout = 10.seconds
+StrongMigrations.statement_timeout = 1.hour
+
+# Analyze tables after indexes are added
+# Outdated statistics can sometimes hurt performance
+StrongMigrations.auto_analyze = true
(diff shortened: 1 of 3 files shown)


config/initializers/strong_migrations.rb
```diff
@@ -0,0 +1,28 @@
...
+StrongMigrations.lock_timeout = 10.seconds
+StrongMigrations.statement_timeout = 1.hour
+
+# Analyze tables after indexes are added
+# Outdated statistics can sometimes hurt performance
+StrongMigrations.auto_analyze = true
+
+# Set the version of the production database
+# so the right checks are run in development
+StrongMigrations.target_version = 10
```
colinbruce, line 17: Production (and staging) DB is actually 11 
The UAT postgres image is currently 10 (bitnami limitation we discussed the other day)
Not sure how this affects the gem? 🤷‍♂️

colinbruce (approved):
Overall looks pretty good 
There is a slight overhead of needing to review the DSL when you infringe and get pulled up, 
 but improving the stability of the system is always a plus :+1:

One comment about the DB version but I'm not sure how that will affect it in use?

config/initializers/strong_migrations.rb
cpjmcquillan: Ok this should be updated to 11 then.

This configuration setting basically determines the Postgres version that `strong_migrations` checks the migrations against.

So even if your development database is version 14, it will highlight issues for the target version.

PR #4389, opened 2022-10-04
Eager load the application in CI

cpjmcquillan:
This ensures the application is eager loaded when running tests in CI.

This is the default behaviour since Rails 7+, and can help catch certain errors before deployment.

It also ensures all application code is checked by simplecov when doing performaing coverage checks.


colinbruce (approved):
LGTM :+1:

PR #4390, opened 2022-10-04
Configure erblint to use Rubocop

cpjmcquillan:
This lints the Ruby code in our .erb templates with Rubocop  
—helping us to write consistently styled Ruby code in our views.


.erb-lint.yml
```diff
@@ -1,8 +1,20 @@
...
+    enabled: true
+    rubocop_config:
+      inherit_from:
+        - .rubocop.yml
+      Layout/InitialIndentation:
+        Enabled: false
+      Layout/TrailingEmptyLines:
+        Enabled: false
+      Lint/UselessAssignment:
+        Enabled: false
```
cpjmcquillan, line 20: `erblint` evaluates Ruby code in isolation, so these rules don't make any sense.

cpjmcquillan:
Given the large number of failures, I propose we incrementally include more templates to the Rubocop checks.

This should help us to fix and UAT things in a more manageable way.

colinbruce:
> Given the large number of failures, I propose we incrementally include more templates to the Rubocop checks.
> 
> This should help us to fix and UAT things in a more manageable way.

I like the idea of automated linters, but I have some concerns about ignoring most views by folder... It won't prevent us adding more infractions, see the 1200 line rubocop_todo 😱 

How would you propose reducing the number of ignored folders? Jira ticket per folder for addressing during tech-debt time?

cpjmcquillan:
Thanks @colinbruce

> see the 1200 line rubocop_todo 😱

ouch 😬

> How would you propose reducing the number of ignored folders?

The main issue isn't fixing failures, those are mostly auto correctable or trivial, but it results in huge diffs that are difficult to review and UAT.

If we have an appetite for that risk, I'd say _just do it_.

**But**, we probably don't —which is why I've kicked us off folder by folder.

> Jira ticket per folder for addressing during tech-debt time?

If we do want to fix things incrementally, I think this would be a reasonable approach!

I'd be interested to hear some of the rest of the team's thoughts too.

cpjmcquillan:
I've created [an epic new Epic in JIRA](https://dsdmoj.atlassian.net/browse/AP-3620) to help us track this work, as well as tasks to address the outstanding folders.

colinbruce (approved):

PR #4392, opened 2022-10-04
Use prosopite to detect N+1 queries

cpjmcquillan:
This adds the prosopite gem to detect N+1 queries.

Default configuration has been applied.

This means N+1 queries raise errors in test environments and log in development environments.

Prosopite does not run in production.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -78,4 +78,9 @@ Rails.application.configure do
 
   # Switch to determine whether or not o collect HMRC data
   config.x.collect_hmrc_data = true
+
+  config.after_initialize do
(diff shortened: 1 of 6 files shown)


config/environments/test.rb
```diff
@@ -78,4 +78,9 @@
 
   # Switch to determine whether or not o collect HMRC data
   config.x.collect_hmrc_data = true
+
+  config.after_initialize do
+    Prosopite.rails_logger = true
+    Prosopite.raise = true
```
cpjmcquillan, line 84: I'd expect us to disable this initially if there are a significant number of failures, and devise a strategy to incrementally remove any N+1 queries.

PR #4399, opened 2022-10-06
Use default Docker version in CircleCI

cpjmcquillan:
This change was [recommended] by the operations team ahead of 
some changes to CircleCI.

[recommended]: https://mojdt.slack.com/archives/C02D2NEF9CJ/p1665061637638169

We were already using the default version in most places, however 
this removes the version flag from the two remaining places it was used.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,8 +133,7 @@ jobs:
     executor: linting-executor
     steps:
     - checkout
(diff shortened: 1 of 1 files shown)


colinbruce (approved):
LGTM 👍

naseberry (approved):

PR #4400, opened 2022-10-06
AP-3447 Check answers page and CFE results for enhanced bank upload

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3447)

Update the means summary page and financial eligibility results pages to include details of regular transactions and uploaded bank statements

## Checklist

### DIFF
--- a/app/helpers/transaction_type_helper.rb
+++ b/app/helpers/transaction_type_helper.rb
@@ -40,7 +40,7 @@ module TransactionTypeHelper
   end
 
   def regular_transaction_answer_by_type(legal_aid_application:, transaction_type:)
-    regular_transaction = legal_aid_application.regular_transactions.where(transaction_type:).first
+    regular_transaction = legal_aid_application.regular_transactions.find_by(transaction_type:)
     if regular_transaction
       [number_to_currency(regular_transaction.amount), t("transaction_types.frequencies.#{regular_transaction.frequency}")]
     else
--- a/app/helpers/transaction_type_helper.rb
+++ b/app/helpers/transaction_type_helper.rb
@@ -38,4 +38,13 @@ module TransactionTypeHelper
       t("generic.none")
     end
   end
+
+  def regular_transaction_answer_by_type(legal_aid_application:, transaction_type:)
+    regular_transaction = legal_aid_application.regular_transactions.where(transaction_type:).first
+    if regular_transaction
+      [number_to_currency(regular_transaction.amount), t("transaction_types.frequencies.#{regular_transaction.frequency}")]
+    else
+      t("generic.none")
+    end
+  end
 end
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -484,6 +484,10 @@ class LegalAidApplication < ApplicationRecord
     __send__("#{transaction_type}_transactions").amounts.fetch(category_id, 0)
   end
 
+  def regular_transaction_category(transaction_type)
+    regular_transactions.where(transaction_type:)
+  end
+
   def hmrc_employment_income?
     employments.any?
   end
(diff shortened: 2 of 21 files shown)


app/helpers/transaction_type_helper.rb
```diff
@@ -38,4 +38,13 @@ def answer_for_transaction_type(legal_aid_application:, transaction_type:)
       t("generic.none")
     end
   end
+
+  def regular_transaction_answer_by_type(legal_aid_application:, transaction_type:)
+    regular_transaction = legal_aid_application.regular_transactions.where(transaction_type:).first
```
jsugarman, line 43: For a single return you can use `find_by`
```suggestion
    regular_transaction = legal_aid_application.regular_transactions.find_by(transaction_type:)
```

spec/models/regular_transaction_spec.rb
```diff
@@ -97,4 +97,28 @@
...
+      it "includes monthly in the frequencies" do
+        child_care = create(:transaction_type, :child_care)
+
+        frequencies = described_class.frequencies_for(child_care)
+
+        expect(frequencies.count).to eq(5)
+        expect(frequencies).to include("monthly")
+      end
+    end
+  end
```
jsugarman, line 123: better to use rspec DSL, and to be explicit about the values rather than just counting IMO
```
describe ".frequencies_for" do
  subject(:frequencies_for) { described_class.frequencies_for(transaction_type) }

  context "when transaction type is benefits" do
    let(:transaction_type) { create(:transaction_type, :benefits) }

    it "includes all but monthly in the frequencies" do
      expect(frequencies_for).to match(%w[weekly two_weekly four_weekly three_monthly])
    end
  end

  context "when transaction type is not benefits" do
    let(:transaction_type) { create(:transaction_type, :child_care) }

    it "includes all frequencies" do
      expect(frequencies_for).to match(%w[weekly two_weekly four_weekly monthly three_monthly])
    end
  end
  ```

File: app/models/legal_aid_application.rb [hunk 3]
```diff
@@ -484,6 +484,10 @@ def transactions_total_by_type(transaction_type, category_id)
     __send__("#{transaction_type}_transactions").amounts.fetch(category_id, 0)
   end
 
+  def regular_transaction_category(transaction_type)
+    regular_transactions.where(transaction_type:)
+  end
```
jsugarman, line 489: is this method being used? can't see it called anywhere. If not I would remove. If it is, then a minor nit is that  it is returning a collection of results too, rather than a single category, as implied by the singular name - regular_transaction_category -> regular_transaction_categories

jsugarman (approved):
A few nits but 👍 . I have not UAT'ed it though

app/views/shared/check_answers/_income_summary.html.erb
```diff
@@ -11,13 +11,26 @@
     </div>
 </div>
 
-<dl class="govuk-summary-list govuk-!-margin-bottom-9" data-check-your-answers-section="<%= income_type.parameterize %>">
+<% if @legal_aid_application.using_enhanced_bank_upload? %>
+  <dl class="govuk-summary-list govuk-!-margin-bottom-9" data-check-your-answers-section="<%= income_type.parameterize %>">
```
jsugarman, line 15: perhaps the income_type can be passed to the partial and this would resolve the bug that the <dl> is defined outside the partial with the <dt> and <dd>.

app/helpers/transaction_type_helper.rb
cpjmcquillan: Should an application be able to have more than one regular transaction for a given category?

If the answer is no, then I think we should probably be validating that and adding a unique index to the database.

That doesn't need to be a part of this pull request though.

If the answer is yes, then we should be explicit about _which_ regular transaction we want to find.

What do you think?

jsugarman: sure, but the same applies to cash transactions, at least. Feel free to add a tech debt ticket or similar.

PR #4404, opened 2022-10-07
Refactor regular transaction forms

cpjmcquillan:
This introduces a new `RegularTransactionForm` model for persisting regular
transaction data for an application (and destroying dependent records).

Each of `RegularIncome`, `RegularOutgoings` and `HousingBenefit` forms make
use of this new abstract form.

The basic premise of this refactor is that each of the forms exhibit the same
characteristic behaviour:

1. They validate user input (i.e an option is selected, inputs are valid).

2. They facilitate validation and creation of regular transaction records that
   belong to a given legal aid application.

3. They persist some data on the legal aid application record itself.

4. They destroy the correct combination of associated records. This is currently
   complicated by several factors:
    - The relationship between LegalAidApplication <> TransactionType <>
      Transaction (regular/cash) is indirect.
    - Housing payments (rent_or_mortgage outgoings) are coupled to housing
      benefit. This means housing benefit transactions can only exist if there is
      a rent or mortgage transaction. This also impacts user flow.
    - Housing benefit is also an income (i.e credit) transaction (although it is
      a child transaction type), but it is not handled by the income form and
      income form changes should not affect housing benefit transactions.

5. They assign regular transaction data from the legal aid application if it
   exists (i.e so correct data is displayed when changing answers).

Each form subclass defines a `transaction_type_conditions` private method that
sets out the conditions with which associated transactions should be scoped.
This means the `RegularIncomeForm` is only interested in transactions where the
transaction type has a "credit" operation and "nil" parent_id, for example.

Each form subclass also defines a `legal_aid_application_attributes` private
method that outlines the attributes to be updated for a given form.

This change also highlights further areas for refactoring -in particular the
relationship between a legal aid application, transaction types, and types of
transaction (i.e regular, cash, bank).

Subsequent work will be done to explore the feasibility of simplifying those
relationships. Essentially all types of transaction should be joined to an
application through the existing `LegalAidApplicationTransactionType` join
table.

### DIFF
--- a/app/forms/providers/means/housing_benefit_form.rb
+++ b/app/forms/providers/means/housing_benefit_form.rb
@@ -1,116 +1,30 @@
 module Providers
   module Means
-    class HousingBenefitForm
-      include ActiveModel::Model
+    class HousingBenefitForm < RegularTransactionForm
+      attr_accessor :housing_benefit_amount, :housing_benefit_frequency
 
-      attr_accessor :housing_benefit,
-                    :housing_benefit_amount,
-                    :housing_benefit_frequency
+      delegate :applicant_in_receipt_of_housing_benefit, to: :legal_aid_application
 
-      attr_reader :legal_aid_application
-
-      validates :housing_benefit, inclusion: { in: %w[true false] }
-
-      validates :housing_benefit_amount,
-                currency: { greater_than: 0 },
-                if: :applicant_in_receipt_of_housing_benefit?
-
-      validates :housing_benefit_frequency,
-                inclusion: { in: ->(form) { form.frequency_options } },
-                if: :applicant_in_receipt_of_housing_benefit?
-
-      def initialize(params = {})
-        @legal_aid_application = params.delete(:legal_aid_application)
-
-        if params[:housing_benefit].nil?
-          assign_housing_benefit_attributes
-        end
-
-        super
-      end
-
-      def save
-        return false unless valid?
-
-        ApplicationRecord.transaction do
-          legal_aid_application.applicant_in_receipt_of_housing_benefit = housing_benefit
-
-          if applicant_in_receipt_of_housing_benefit?
-            build_housing_benefit_transaction_type
-            build_housing_benefit_regular_transaction
-          else
-            destroy_housing_benefit_transaction_type!
-            destroy_housing_benefit_regular_transaction!
-          end
-
-          legal_aid_application.save!
-        end
-
-        true
+      def housing_benefit_transaction_type
+        TransactionType.find_by(name: "housing_benefit")
       end
 
       def frequency_options
         RegularTransaction.frequencies_for(housing_benefit_transaction_type)
       end
 
+      def housing_benefit_selected?
+        transaction_type_ids.include?(housing_benefit_transaction_type.id)
+      end
+
     private
 
-      def applicant_in_receipt_of_housing_benefit?
-        housing_benefit == "true"
+      def transaction_type_conditions
+        { name: "housing_benefit" }
       end
 
-      def assign_housing_benefit_attributes
-        @housing_benefit = @legal_aid_application.applicant_in_receipt_of_housing_benefit
-        @housing_benefit_amount = housing_benefit_regular_transaction&.amount
-        @housing_benefit_frequency = housing_benefit_regular_transaction&.frequency
-      end
-
-      def build_housing_benefit_transaction_type
-        legal_aid_application
-          .legal_aid_application_transaction_types
-          .find_or_initialize_by(
-            transaction_type: housing_benefit_transaction_type,
-          )
-      end
-
-      def build_housing_benefit_regular_transaction
-        housing_benefit_transaction = legal_aid_application
-          .regular_transactions
-          .find_or_initialize_by(
-            transaction_type: housing_benefit_transaction_type,
-          )
-        housing_benefit_transaction.assign_attributes(
-          amount: housing_benefit_amount,
-          frequency: housing_benefit_frequency,
-        )
-        housing_benefit_transaction.save!
-      end
-
-      def destroy_housing_benefit_transaction_type!
-        legal_aid_application
-          .legal_aid_application_transaction_types
-          .includes(:transaction_type)
-          .where(transaction_type: { name: "housing_benefit" })
-          .destroy_all
-      end
-
-      def destroy_housing_benefit_regular_transaction!
-        legal_aid_application
-          .regular_transactions
-          .includes(:transaction_type)
-          .where(transaction_type: { name: "housing_benefit" })
-          .destroy_all
-      end
-
-      def housing_benefit_regular_transaction
-        legal_aid_application
-          .regular_transactions
-          .includes(:transaction_type)
-          .find_by(transaction_type: { name: "housing_benefit" })
-      end
-
-      def housing_benefit_transaction_type
-        TransactionType.find_by!(name: "housing_benefit")
+      def legal_aid_application_attributes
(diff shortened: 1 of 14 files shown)


app/forms/providers/means/regular_transaction_form.rb
```diff
@@ -0,0 +1,158 @@
...
+        end
+      end
+
+      def dependent_transaction_models
+        %i[
+          legal_aid_application_transaction_types
+          regular_transactions
+          cash_transactions
+        ]
+      end
```
cpjmcquillan, line 115: I think this highlights the awkwardness of the data modelling currently.

In an ideal world, we wouldn't need to explicitly destroy these records —it would be covered by `dependent: :destroy` declarations in ActiveRecord or `on_cascade: :delete` relationships defined in Postgres.

When I can, I'll spend a bit of time exploring how painful/less that change might be.

cpjmcquillan: Similarly, we might be able to compress some of the complexity around building/validating/persisting regular transaction records by defining nested attribute relationships —but I don't want to get ahead of myself just yet!

app/forms/providers/means/regular_outgoings_form.rb
```diff
@@ -1,190 +1,56 @@
...
       end
 
-      def transaction_types
-        transaction_type_options.where(id: transaction_type_ids)
+      def legal_aid_application_attributes
+        {
+          no_debit_transaction_types_selected: none_selected?,
+          applicant_in_receipt_of_housing_benefit: (
+            applicant_in_receipt_of_housing_benefit if housing_payments_selected?
+          )
```
cpjmcquillan, line 29: Unless housing payments are selected, this attribute should be reset to `nil`.

app/forms/providers/means/regular_outgoings_form.rb
```diff
@@ -1,190 +1,56 @@
...
+        { operation: "debit", parent_id: nil }
       end
 
-      def transaction_types
-        transaction_type_options.where(id: transaction_type_ids)
+      def legal_aid_application_attributes
+        {
+          no_debit_transaction_types_selected: none_selected?,
+          applicant_in_receipt_of_housing_benefit: (
+            applicant_in_receipt_of_housing_benefit if housing_payments_selected?
```
cpjmcquillan, line 28: Unless housing payments are selected, this attribute should be reset to nil.

jsugarman (commented):
Feels like a premature optimisation at this point with a fair bit of meta-programming to over come this, but will try and take a deeper look.

cpjmcquillan:
> Feels like a premature optimisation at this point with a fair bit of meta-programming to over come this

@jsugarman I'm not sure what you mean by this?

There are 3 different forms (income, outgoings, housing benefit) that depend on the same behaviour. We don't expect these requirements to change, so I'm not sure I agree this refactor is premature.

This refactor has several objectives:

1. Reduce unnecessary duplication between forms
2. Highlight areas of unnecessary complexity, so they can be addressed in future changes
3. Address existing issues that impact the user journey as detailed in [AP-3562](https://dsdmoj.atlassian.net/browse/AP-3562)

I think all those objectives have been met.

If we're unhappy with the use of inheritance to solve the problem, then perhaps we should explore an approach that uses composition instead. I'm happy to pair on that if you think it would be helpful, or I can add a commit that demonstrates that approach?

Equally if any aspect of the changes are unclear, please highlight them so I can make the code as readable and maintainable as possible.

jsugarman:
> There are 3 different forms

Jut noticed the 3 forms, i thought it was 2 being replaced. That makes sense to move out to a superclass. although i wonder in a mixin might be cleaner.

I see it has dealt with similar things to that on the [AP-3550 branch](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4428).

cpjmcquillan:
@jsugarman 

> Would you be averse to getting AP-3550 in first and rebasing this on that.

As things stand, I don't think #4428 is ready to be merged in.

I appreciate it can be tricky working on similar areas of code, but I think it makes more sense to base the check your answers work on this change instead. 

That way you can keep the feature test changes and check the behaviour is as expected without worrying about any flow issues (which are handled by this PR)

Do you want to schedule some time and we can pair on that?

I don't expect this branch to change at all at this point, except being kept up to date with main.

> feature test coverage is a bit lacking for the various journeys 

I agree. The different user flows are really complicated and nuanced, and it feels like a thankless (and expensive) task trying to cover every combination with feature tests. The sooner we can break up our flow objects and begin unit testing them, the sooner I think we'll be able to move more quickly and with more confidence when it comes to user flow.

app/models/legal_aid_application.rb
```diff
@@ -53,8 +53,6 @@ class LegalAidApplication < ApplicationRecord
   has_many :hmrc_responses, class_name: "HMRC::Response", dependent: :destroy, inverse_of: :legal_aid_application
   has_many :employments, dependent: :destroy
   has_many :regular_transactions, dependent: :destroy
-  has_many :regular_incomes, -> { credits }, class_name: "RegularTransaction"
-  has_many :regular_outgoings, -> { debits }, class_name: "RegularTransaction"
```
cpjmcquillan, line 57: These associations aren't used anywhere, so they can be safely removed.

cpjmcquillan:
I think there is an argument for preferring composition to an inheritance based approach, but it wasn't presenting itself clearly to me right away and I didn't want to sink too much time into this.

Having said that, I still think the refactor is worthwhile and helpful in its current form.

app/forms/providers/means/housing_benefit_form.rb
```diff
@@ -1,52 +1,12 @@
...
-
-      validates :housing_benefit, inclusion: { in: %w[true false] }
-
-      validates :housing_benefit_amount,
-                currency: { greater_than: 0 },
-                if: :applicant_in_receipt_of_housing_benefit?
-
-      validates :housing_benefit_frequency,
-                inclusion: { in: ->(form) { form.frequency_options } },
-                if: :applicant_in_receipt_of_housing_benefit?
```
jsugarman, line 20: Is the conditional `if` logic on these validations accounted for in the superclass or elsewhere?

jsugarman (commented):
👍 Can't see any obvious problems with the inheritance used here but had a question on the validations for housing benefit specifically and a suggestion for tests.

Just want to UAT to check for regressions that features don't cover this pm. Be good to get others on that too.

app/forms/providers/means/housing_benefit_form.rb
cpjmcquillan: It's covered by `validate :all_regular_transactions_valid`.

If "No" is selected, there are no `transaction_type_ids`, and no regular transactions to validate.

Does that make sense?

spec/forms/providers/means/regular_transaction_form_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+        stub_const "ConditionsNotImplementedForm", conditions_not_implemented_form
+
+        ConditionsNotImplementedForm.class_eval do
+        private
+
+          def legal_aid_application_attributes
+            {}
+          end
+        end
+      end
```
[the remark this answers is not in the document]
cpjmcquillan: Sure 👍

jsugarman (approved):
🎖️ - baring the error handling probs that Mike noticed

cpjmcquillan:
> 🎖️ - baring the error handling probs that Mike noticed

Those are resolved too. 

I was tempted to add super granular feature tests to cover it, but once I started adding them it felt like overkill.

PR #4405, opened 2022-10-07
[AP-3536] Add current and saving account totals to bank statement uploads

jsugarman:
## What
Add "Which bank accounts does your client have?" step to bank statement upload

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3536)

This seems to have been missed from existing/basic bank 
statement upload flow and should be included in both the 
basic and enhanced flow.

- [x] add section into flow for basic and enhanced banks statement upload
- [x] add section into means summaries CYA page
- [x] add section into means report

NOTE: In the context of non-passported bank statement
uploads the name `offline_accounts` is wrong as well
as the field saving_amounts#offline_current_accounts
and saving_amounts#offline_saving_accounts. renaming
or adding new columns to reflect the new logic (to just
collect a flat amount for all current accounts whether
off or online) would require rejigging various other
classes.

## Checklist

### DIFF
--- a/features/providers/enhanced_bank_upload/check_your_answers.feature
+++ b/features/providers/enhanced_bank_upload/check_your_answers.feature
@@ -27,6 +27,7 @@ Feature: Enhanced bank upload check your answers
 
     And I should see "Uploaded bank statements"
     And I should see 'Does your client receive student finance?'
+    And I should not see 'Does your client have any savings accounts they cannot access online?'
 
     When I click Check Your Answers Change link for "bank statements"
     And I upload an evidence file named "hello_world.pdf"
--- a/features/providers/enhanced_bank_upload/check_your_answers.feature
+++ b/features/providers/enhanced_bank_upload/check_your_answers.feature
@@ -6,8 +6,27 @@ Feature: Enhanced bank upload check your answers
     And I have completed a non-passported employed application with enhanced bank upload as far as the end of the means section
     Then I should be on the "means_summary" page showing "Check your answers"
 
+    And the following sections should exist:
+      | tag | section |
+      | h1  | Check your answers |
+      | h3  | Bank statements |
+      | h2  | Your client's income |
+      | h3  | Employment income |
+      | h3  | What payments does your client receive? |
+      | h3  | Student finance |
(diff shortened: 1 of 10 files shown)


features/providers/enhanced_bank_upload/check_your_answers.feature
```diff
@@ -6,8 +6,27 @@ Feature: Enhanced bank upload check your answers
...
+      | h3  | Vehicles |
+      | h2  | Which bank accounts does your client have? |
+      | h2  | Which savings or investments does your client have? |
+      | h2  | Which assets does your client have? |
+      | h2  | Restrictions on your client's assets |
+      | h2  | Payments from scheme or charities |
+
     And I should see "Uploaded bank statements"
-    And I should not see "Which bank accounts does your client have?"
+    And I should see 'Does your client receive student finance?'
```
cpjmcquillan, line 29: Are these 2 expectations needed?

I can see `Bank statements` and `Student finance` sections listed above, but I'm not sure what the difference is between those assertions and these two.

Is there something special about these 2 questions?

cpjmcquillan (approved):

features/providers/enhanced_bank_upload/check_your_answers.feature
jsugarman: I think they come out of the fact that we should ideally be testing all of the "Questions" in a similar way to how we do for the means report but that this was a simpler, non-exhaustive, earlier attempt to define the expected sections content (or absence thereof). The ` And I should see "Uploaded bank statements"` is a difference for this journey but the student finance question could arguably go as i believe it is applicable to all journeys, not just this.

I am going to add another expectation i missed from the basic bank upload flow though which should apply to the enhanced too.

```
And I should not see 'Does your client have any savings accounts they cannot access online?'
```

jsugarman: have added this [commit](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4405/commits/de9e8d46d8a25fc21b392462d1ba670b2a10e5e7) to sync them for now

PR #4413, opened 2022-10-11
Chore: remove webhint ignored dependencies

naseberry:
## What

Remove Webhint's ignored dependencies via dependabot.

Webhint and its dependencies were removed from the codebase, therefore, we no longer need to ignore these via dependabot

## Checklist


cpjmcquillan (approved):

colinbruce (approved):
:trophy

PR #4417, opened 2022-10-12
Fix flaky document upload spec

cpjmcquillan:
This feature test would fail intermittently if the "Save and continue" 
button was clicked before the document had finished uploading.

Similarly to #4375, this adds an assertion that the document has 
been uploaded in order to make the test more predictably reliable 
(and better reflect the user experience).


jsugarman (approved):
👍

PR #4422, opened 2022-10-13
ap-3504: add ineligible to means report caseworker review reasons

kmahern:
where cfe returns a result of ineligible

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3504)

## Checklist

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -1001,7 +1001,7 @@ en:
           multiple_employments: Multiple employments
           refunds: Tax or NI refunds
           uploaded_bank_statements: Bank statements uploaded
-          ineligible: Ineligible
+          ineligible: Applicant is ineligible
       show:
         client_details_heading: Client details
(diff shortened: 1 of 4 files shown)


config/locales/en/providers.yml
```diff
@@ -1001,6 +1001,7 @@ en:
           multiple_employments: Multiple employments
           refunds: Tax or NI refunds
           uploaded_bank_statements: Bank statements uploaded
+          ineligible: Ineligible
```
colinbruce, line 1004: Is this the value that is added to the means report for caseworker review?
I _think_ it is supposed to say `Applicant is ineligible` but you may need to chat with Dave to confirm

kmahern: 👍 Asked Dave and he said if he had the choice and it is easy to amend then 'Applicant is ineligible' is probably clearer so have amended the value

colinbruce (approved):
LGTM :+1:

PR #4431, opened 2022-10-17
Use `govuk_accordion` ViewComponent helper

cpjmcquillan:
Before, accordion components were rendered adhoc by replicating and tweaking HTML examples from the GOV.UK Design System.

This replaces accordion components with their govuk-components counterparts.

### DIFF
--- a/app/views/providers/capital_assessment_results/_result_details.html.erb
+++ b/app/views/providers/capital_assessment_results/_result_details.html.erb
@@ -1,23 +1,16 @@
 <div class="govuk-details govuk-!-margin-top-6" data-module="govuk-details">
-    <h1 class="govuk-heading-l">
-      <%= t('.client_eligibility_calculation') %>
-    </h1>
+  <h2 class="govuk-heading-l">
+    <%= t('.client_eligibility_calculation') %>
+  </h2>
 
-  <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-breakdown">
-    <div class="govuk-accordion__section">
-      <div class="govuk-accordion__section-header">
-        <h2 class="govuk-accordion__section-heading">
-        <span class="govuk-accordion__section-button" id="accordion-capital-calculation">
-          <%= t('.capital_calculation') %>
-        </span>
-        </h2>
-      </div>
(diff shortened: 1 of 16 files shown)


app/views/providers/capital_assessment_results/_result_details.html.erb
```diff
@@ -1,23 +1,16 @@
 <div class="govuk-details govuk-!-margin-top-6" data-module="govuk-details">
-    <h1 class="govuk-heading-l">
-      <%= t('.client_eligibility_calculation') %>
-    </h1>
+  <h2 class="govuk-heading-l">
```
cpjmcquillan, line 2: The parent view renders a `h1` tag already.

app/views/providers/capital_assessment_results/show.html.erb
```diff
@@ -8,9 +8,7 @@
     <%= render @result_partial %>
   </div>
 
-  <%= render 'result_details' %>
-
-  <br><br>
```
colinbruce, line 13: Someone obviously wanted a space after the result_details partial, is this being added anywhere?

app/views/providers/capital_income_assessment_results/show.html.erb
```diff
@@ -9,9 +9,7 @@
     <%= render @result_partial %>
   </div>
 
-  <%= render 'result_details' %>
-
-  <div class="govuk-!-padding-bottom-6"></div>
```
colinbruce, line 14: Ditto on spacing... does the govuk_accordion css add spacing below?

colinbruce (commented):
Generally looks fine but a couple of questions about removed spacing

app/views/providers/capital_assessment_results/show.html.erb
colinbruce: Not arguing that a double line break is 'good' spacing, but it may be worth checking with design if they are happy for it to be removed, or if it should be replaced with a `govuk-!-padding-bottom-6` on the accordion itself?
image

RoseSAK (approved):
All looks good to me 👍

PR #4432, opened 2022-10-17
Use `govuk_inset_text` ViewComponent helper

cpjmcquillan:
Before, inset text components were rendered adhoc by replicating and tweaking HTML examples from the GOV.UK Design System.

This replaces all usages of the inset text component with its `govuk-components` counterpart.


colinbruce (approved):
LGTM :+1:

PR #4437, opened 2022-10-18
Discussion: check DB anonymizer rules when commiting

colinbruce:
## What

This adds a pre-commit check to overcommit to run the DB Anonymiser check

If you have added a new table to the schema but not created a new rule for it, it will cause your commit to fail

It means you won't get pinged on github, but this will only work if you have activated overcommit, so I would suggest leaving the Github action as a backup

## Checklist

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -29,3 +29,6 @@ PreCommit:
     command: ['standard']
     requires_files: true
     include: 'app/webpack/**/*.js'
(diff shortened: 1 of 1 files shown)


cpjmcquillan (approved):
I don't see any issue with adding this, and I agree it's important to keep the GitHub action running too 👍

jsugarman (approved):
👍

PR #4439, opened 2022-10-18
Replace `.to_not` expectations with `.not_to`

cpjmcquillan:
As Hamlet almost once said:

> ".to_not or .not_to be"

This replaces the former with the later in our tests, in the interests of consistency and correctness.


jsugarman (approved):
indeed 👍

PR #4449, opened 2022-10-19
Prevent basic uploads navigating to truelayer `income_summary`

jsugarman:
Prevent basic uploads navigating to truelayer income_summary

## What
This is a live bug whereby if the provider does not add any
outgoings on the outgoings page but has provided incomes
on the earlier incomes page they will be take to the
income_summary page - this page is only for categorising
bank transactions and part of the truelayer path only.

ps: this would not effect the enhanced journey.

## Checklist

### DIFF
--- a/app/services/flow/flows/provider_capital.rb
+++ b/app/services/flow/flows/provider_capital.rb
@@ -57,7 +57,7 @@ module Flow
           forward: lambda do |application|
             if application.outgoing_types?
               :cash_outgoings
-            elsif application.income_types?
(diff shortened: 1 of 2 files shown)


RoseSAK (approved):
All LGTM, good catch 😅 🎖️

cpjmcquillan (approved):
Good spot 👌🏼

This highlights some of my concerns about how our current approach to flow can scale as the complexity of the application continues to increase.

I wonder if other people share these concerns and whether this something we should discuss as a team?

PR #4455, opened 2022-10-20
bump axious 1.1.3 and fix

jsugarman:
## What
bump axios from 1.1.2 to 1.1.3 and fix errors arising

Amend @babel/plugin-proposal-class-properties to not use loose mode
to remove warning and amend overcommit's command to run standard using
`yarn run` explicitly so uses yarn "context".

import 'axios' rather than require it. Previoulsy
this resulted in a CI lint failure as below:

```
FAIL spec/javascript/worker_waiter.spec.js
  ● worker is complete after 2 API calls › polls the correct endpoint twice

    TypeError: Cannot read properties of undefined (reading 'get')

      3 | async function checkWorkerStatus () {
      4 |   const workerId = document.querySelector('.worker-waiter').getAttribute('data-worker-id')
    > 5 |   const response = await axios.get(`/v1/workers/${workerId}`)
        |                                ^
      6 |   return response.data
      7 | }
      8 |

      at Object.get [as checkWorkerStatus] (app/webpack/src/worker_waiter.js:5:32)
      at Object.checkWorkerStatus (spec/javascript/worker_waiter.spec.js:36:18)

Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-property-in-object since the "loose" mode option was set to "true" for @babel/plugin-pr
oposal-class-properties.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this w
arning by explicitly adding
  ["@babel/plugin-proposal-private-property-in-object", { "loose": true }]
to the "plugins" section of your Babel config.
PASS spec/javascript/modules/proceedings.spec.js
```

## Checklist

### DIFF
--- a/.overcommit.yml
+++ b/.overcommit.yml
@@ -26,7 +26,7 @@ PreCommit:
     requires_files: true
   StandardJS:
     enabled: true
-    command: ['standard']
+    command: ['yarn', 'run', 'standard']
     requires_files: true
     include: 'app/webpack/**/*.js'
   DBAnonymizer:
--- a/app/webpack/src/worker_waiter.js
+++ b/app/webpack/src/worker_waiter.js
@@ -1,4 +1,4 @@
-const axios = require('axios').default
+import axios from 'axios'
 
 async function checkWorkerStatus () {
   const workerId = document.querySelector('.worker-waiter').getAttribute('data-worker-id')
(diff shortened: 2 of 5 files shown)


jsugarman:
this replaces https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4441

PR #4463, opened 2022-10-21
Resolve SonarCloud Bugs

naseberry:
## What

[SonarCloud - Apply - Bugs](https://dsdmoj.atlassian.net/browse/AP-3520)

Resolve SonarClouds reported bugs. This resolves some of the existing issues in order to reduce and resolve our quality gate thresholds.

## Checklist

### DIFF
--- a/app/views/citizens/legal_aid_applications/index.html.erb
+++ b/app/views/citizens/legal_aid_applications/index.html.erb
@@ -1,13 +1,13 @@
 <%= page_template page_title: t('.heading_1'), back_link: :none do %>
   <p class="govuk-body">
-    <b> <%= t('.name') %></b>
+    <strong> <%= t('.name') %></strong>
     <%= current_applicant.first_name %>
     <%= current_applicant.last_name %>
     <br>
-    <b><%= t('.case_reference') %></b>
+    <strong><%= t('.case_reference') %></strong>
     <%= current_applicant.legal_aid_application.application_ref %>
     <br>
-    <b><%= t('.your_solicitor') %></b>
+    <strong><%= t('.your_solicitor') %></strong>
     <%= current_applicant.legal_aid_application.provider.firm.name %>
   </p>
 
--- a/app/views/layouts/pdf.html.erb
+++ b/app/views/layouts/pdf.html.erb
@@ -2,6 +2,7 @@
 <html lang="en" class="govuk-template">
   <head>
     <meta charset="utf-8">
+    <title><%= t("layouts.application.header.title") %></title>
     <%= apply_webpacker_stylesheet_pack_tag('styles') %>
     <%= render 'layouts/govuk_base64_fonts' %>
   </head>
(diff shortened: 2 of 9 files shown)


app/views/layouts/pdf.html.erb
```diff
@@ -2,6 +2,7 @@
 <html lang="en" class="govuk-template">
   <head>
     <meta charset="utf-8">
+    <title><%= t("layouts.application.header.title") %></title>
```
colinbruce, line 5: I'd like to UAT this and ensure it isn't doing something weird with the report generator! :+1:

app/views/saml_idp/idp/saml_post.html.erb
```diff
@@ -1,8 +1,9 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset='utf-8'>
     <meta http-equiv='X-UA-Compatible' content='IE=edge,chrome=1'>
+    <title><%= t("layouts.application.header.title") %></title>
```
colinbruce, line 6: Ditto portal login

colinbruce (approved):
Looks good, but a couple of UAT tests identified :+1:

jsugarman (approved):
looks good, baring Colin's UAT comments on pdf generation.

app/views/citizens/legal_aid_applications/index.html.erb
```diff
@@ -1,13 +1,13 @@
 <%= page_template page_title: t('.heading_1'), back_link: :none do %>
   <p class="govuk-body">
-    <b> <%= t('.name') %></b>
+    <strong> <%= t('.name') %></strong>
```
cpjmcquillan, line 3: I have a couple questions about the use of `<strong>` tags.

I'm intrigued what the justification is from SonarCloud for these being "Code smells"?

The [`<b>` tag is a valid HTML element with its own purpose](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/b).

Do we want to communicate to the user that all of these elements are important or urgent?

```html
This is some text that is <strong>seriously important!</strong>
```

Or maybe we want to draw attention to elements without indicating that they are especially important?

```html
This is <b>some text that isn't important</b>, but I want to draw attention to it.
```

Maybe we just want to style some text with a bold weight?

```html
This is <span class="govuk-!-font-weight-bold">some text I want to look bold</span>
```

I think it's important we are conscious about the the tags we use, and we should be writing semantic HTML wherever we can. 

I'm worried SonarCloud just says "\<strong> tag good" and we follow it blindly without proper consideration.

What do you think?

app/views/saml_idp/idp/saml_post.html.erb
naseberry: > Ditto portal login

Great. I decided to use the service name in these instances as we weren't yielding any page title values to the layout page.

colinbruce:
UAT'd the PDF and it seems okay :+1:
Not sure we can do much with Portal login until it's on staging.  
@naseberry shout when you are deploying and I can UAT the staging login before you deploy to prod :+1:

PR #4464, opened 2022-10-21
AP-3529: Handle latest incident details task removal

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3529)

Update the handling of latest incidents, now the question is optional

Ensure the partial that renders it skips if the underlying incident object has not been created

DO NOT MERGE YET - THIS CONTAINS A UAT LINK TO THE LFA BRANCH WITH THE QUESTION DISPLAY LOGIC

## Checklist


app/views/shared/check_answers/_merits.html.erb
```diff
@@ -1,30 +1,31 @@
...
-        question: t('.latest-incident-heading'),
-        read_only: read_only
-      ) %>
-
-  <dl class="govuk-summary-list govuk-!-margin-bottom-2">
-    <%= check_answer_no_link(
-          name: :notification_of_latest_incident,
-          question: t('.items.notification_of_latest_incident'),
-          answer: incident&.told_on,
+<% if @legal_aid_application&.latest_incident %>
```
cpjmcquillan, line 3: Can `@legal_aid_application` really be `nil`?

I tend to avoid using the safe navigator unless it is absolutely necessary.

In this instance, I think I would definitely want to know if `@legal_aid_application` is `nil`, because it suggests to be that something isn't quite right.

What do you think?

cpjmcquillan (approved):

app/views/shared/check_answers/_merits.html.erb
colinbruce: That's a safe navigation operator on latest_incident.  
There is an edge case as part of the section 8 changes where we will no longer be creating a latest_incident record for the legal_aid_application being shown here.

colinbruce: Would switching it to 
```suggestion
<% if @legal_aid_application.latest_incident.present? %>
```
be clearer?

cpjmcquillan: Hmm, maybe I'm missing something.

I evaluate `@legal_aid_application&.latest_incident` as `@legal_aid_application && @legal_aid_application.latest_incident`, so I just wanted to check whether that conditional check made sense and was necessary.

I can future me trying to understand why that check was there, and what the implications of it are.

The latter seems clearer to me, but ultimately I don't mind.

colinbruce: Thanks for clarifying - good spot, thanks for helping me see the issue.
The first pass at the code had the safe navigator further down the object list and got pulled back up a level, but you are right, the legal_aid_application will never be nil, we should only be checking for the nil or presence of the latest_incident value.
Removed the safe_navigation operator :+1:

PR #4465, opened 2022-10-21
Use `<h2>` tag for nested fieldset headings

cpjmcquillan:
Before, these nested fieldsets in the regular transaction forms were using a `<p>` tag.

This was an accessibility issue, so the fieldset headings have been updated to use `<h2>` tags.


jsugarman (approved):
looks fine to me but have not checked rendering impact. Might be worth getting @naseberry to vet as an accessibility aficionado

naseberry (approved):

PR #4467, opened 2022-10-21
Enable RSpec/ExampleLength cop

jsugarman:
## What
Enable RSpec/ExampleLength

Do we want to enable this cop to

1. prevent further high block lengths
2. enable us to whittle some down

Have configured to allow certain structures over multiple
lines and increased default of 5 lines to 10 lines. This
reduces violations from 384 to 73.

We can exclude for now and work on them via the usual `rubocop_todo.yml`
but wanted to show them failing to start with.

https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecexamplelength

## Checklist

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,6 +25,11 @@ AllCops:
   - 'tmp/**/*'
   - 'log/**/*'
 
+RSpec/ExampleLength:
+  Enabled: true
+  CountAsOne: ['array', 'hash', 'heredoc']
+  Max: 10
+
 Lint/AmbiguousBlockAssociation:
   Exclude:
   - 'spec/**/*'
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -303,6 +303,48 @@ RSpec/ContextWording:
     - 'spec/services/submit_provider_reminder_service_spec.rb'
     - 'spec/services/substantive_application_deadline_calculator_spec.rb'
 
+
+RSpec/ExampleLength:
+  Exclude:
+    - 'spec/forms/citizens/other_assets_form_spec.rb'
+    - 'spec/forms/providers/means/housing_benefit_form_spec.rb'
+    - 'spec/forms/providers/means/regular_income_form_spec.rb'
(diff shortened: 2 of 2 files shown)


cpjmcquillan:
What problem are we trying to solve by adding a rule like this?

I'm struggling to understand how this could be helpful, so I hope this discussion shines some light on what we hope to gain from this!

There are two main reasons I think this is something we should not be doing:

1. It will negatively impact how easy it is to understand our tests.
2. Lines of code as a metric is unhelpful and often counter-productive.

Let’s explore the first point.

Understandability is always preferable to brevity —and this forces us to potentially compromise _everything else_ in the pursuit of... example blocks less than 10 lines?

Generally speaking we try to make our code as readable and maintainable as possible.

That can mean lots of things, but in application code it usually means shorter methods, defined in smaller objects, with fewer responsibilities.

When we’re writing tests, though, that isn’t always the case.

The rule [dictates that](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecexamplelength)

> A long example is usually more difficult to understand

But I don’t think this holds true.

That premise applies the way we think about code to tests, which would be fine —except tests are fundamentally different.

For starters, I would argue the main objective of a test (besides correctness and repeatability) is to tell a story.

This means writing [tests as documentation](http://xunitpatterns.com/Goals%20of%20Test%20Automation.html#Tests%20as%20Documentation).

When we dive into a before unseen test, we need to familiarise ourselves quickly with our surroundings. 

If that test tells a clear story, it will be easier to understand. 

If the test cannot tell a story on its own, but necessitates us grokking the entire test file to grasp what it is doing, then it will be much harder to understand and make changes to.

That concept is diametrically opposed to the idea we should be writing as few lines of code as possible in our tests.

If we think about the second point, I don’t think lines of code is a helpful or useful metric at all in any context.

When we try to make our code take up as few lines as possible, it typically leads to us writing more complex and less readable code.

That doesn't seem like a worthwhile endeavour to me.

We don't enforce any other arbitrary stylistic cops, and I don’t think we should start now.

Cops like this are entirely subjective, and they are almost always a complete waste of time and energy.

PR #4468, opened 2022-10-21
Fix flaky proceeding helper spec

cpjmcquillan:
The tests for the `#scope_limits` helper method introduced recently were [flaky] 
because they expected scope limitations for a given proceeding to be returned in 
a particular order.

[flaky]: https://app.circleci.com/pipelines/github/ministryofjustice/laa-apply-for-legal-aid/18383/workflows/ce9fab1b-fdf1-46c5-855a-a0684d15e294/jobs/96235

This updates the tests to prefer the less specific `contain_exactly` helper over `eq`.

The tests have also been refactored with the aim of telling more of a story, since 
at first it wasn't clear to me what the tests were doing, or where the data was 
coming from.

The `#scope_limits` method itself has also been simplified.

### DIFF
--- a/app/helpers/proceeding_helper.rb
+++ b/app/helpers/proceeding_helper.rb
@@ -11,11 +11,10 @@ module ProceedingHelper
   end
 
   def scope_limits(proceeding, scope_type)
-    scope_limits = []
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
LGTM 👍

PR #4477, opened 2022-10-25
TEMP: Add an email for user research on main UAT

colinbruce:
## What

Set up a branch for user research

## Checklist

### DIFF
--- a/db/seeds/test_provider_populator.rb
+++ b/db/seeds/test_provider_populator.rb
@@ -14,6 +14,7 @@ class TestProviderPopulator
     "Groves & Co." => [989, %w[7F809X:23 4X809Z:24]],
     "Sugarman & daughters" => [810, %w[6F810X:25 6X810Z:26]],
     "Test Connor Co." => [118, %w[5F809X:27 5X809Z:28]],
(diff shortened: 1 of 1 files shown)


db/seeds/test_provider_populator.rb
```diff
@@ -14,6 +14,7 @@ class TestProviderPopulator
     "Groves & Co." => [989, %w[7F809X:23 4X809Z:24]],
     "Sugarman & daughters" => [810, %w[6F810X:25 6X810Z:26]],
     "Test Connor Co." => [118, %w[5F809X:27 5X809Z:28]],
+    "User Research Assoc." => [118, %w[4F808X:26 4X808Z:27]],
```
naseberry, line 17: Is it ok/safe for multiple firms to have the same number `118`?

naseberry:
Also for note, should we / can we remove firms and providers related to deactivated emails?

db/seeds/test_provider_populator.rb
colinbruce: 😱 probably not, although it has deployed 🤔

PR #4491, opened 2022-10-26
AP-3485: Add specific issue merits question

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3485)

Implement the Specific Issue, section 8 question.
Add a DB object to store the responses
Add a new form for validating the answer
Add the route, controller and view
Update the merits task flow and routing
Updated the merits partial to show the response on the CYA, review and merits_report pages

## Checklist

### DIFF
--- a/app/controllers/providers/proceeding_merits_task/specific_issue_controller.rb
+++ b/app/controllers/providers/proceeding_merits_task/specific_issue_controller.rb
@@ -1,0 +1,34 @@
+module Providers
+  module ProceedingMeritsTask
+    class SpecificIssueController < ProviderBaseController
+      def show
+        @form = SpecificIssueForm.new(model: specific_issue)
+      end
+
+      def update
+        @form = SpecificIssueForm.new(form_params.merge(proceeding_id: proceeding.id))
+        render :show unless update_task_save_continue_or_draft(proceeding.ccms_code.to_sym, :specific_issue)
+      end
+
+    private
+
+      def specific_issue
+        @specific_issue ||= proceeding.specific_issue || proceeding.build_specific_issue
+      end
+
+      def proceeding
+        @proceeding ||= Proceeding.find(params[:merits_task_list_id])
+      end
+
+      def legal_aid_application
+        @legal_aid_application ||= proceeding.legal_aid_application
+      end
+
+      def form_params
+        merge_with_model(specific_issue) do
+          params.require(:proceeding_merits_task_specific_issue).permit(:details, :confirmed)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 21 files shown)


colinbruce:
Rebased after rubocop changes

app/controllers/providers/proceeding_merits_task/specific_issue_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+        @specific_issue ||= proceeding.specific_issue || proceeding.build_specific_issue
+      end
+
+      def proceeding
+        @proceeding ||= Proceeding.find(params[:merits_task_list_id])
+      end
+
+      def legal_aid_application
+        @legal_aid_application ||= proceeding.legal_aid_application
+      end
```
cpjmcquillan, line 25: How do you feel about us scoping proceedings by legal aid application (that's assigned in the `Providers::ApplicationDependable` module included in `ProvidersBaseController`) instead?

```suggestion
      def proceeding
        @proceeding ||= legal_aid_application.proceedings.find(params[:merits_task_list_id]
      end
```

spec/requests/providers/proceeding_merits_task/specific_issue_controller_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+          context "with Continue button pressed" do
+            let(:submit_button) do
+              {
+                continue_button: "Continue",
+              }
+            end
+
+            it "updates the record" do
+              expect(proceeding_two.specific_issue.reload.details).to eq(details)
+              expect(proceeding_two.specific_issue.reload.confirmed).to eq(confirmed)
```
cpjmcquillan, line 64: We could assert this in a single expectation if you wanted, but I don't really mind.

```suggestion
              expect(proceeding_two.specific_issue.reload).to have_attributes(
                details:,
                confirmed:,
              )
```

cpjmcquillan (approved):
:shipit:

app/controllers/providers/proceeding_merits_task/specific_issue_controller.rb
colinbruce: This one won't work without changes... a while ago we had conversations about URLs getting 'too long' and could we 'shorten' them 🙄 
Because tasks on the task list were needed by proceeding they would have ended up as `/providers/applications/436c3998-606f-456b-84be-d9c3b742548e/merits_task_list/1e516e08-a991-48af-b3da-63ffa7655a6c/specific_issue`
So we truncated the url (when in the task list) to be `/providers/merits_task_list/1e516e08-a991-48af-b3da-63ffa7655a6c/chances_of_success`
Which means that partial doesn't have the legal_aid_application_id available 😅 
Hence (I assume) why we load the proceeding and then extrapolate the application from it :+1:

cpjmcquillan: URLs being too long is a new one!

Thanks for the context, I eagerly await the days of even shorter URLs 😆 

https://bit.ly/3Dj7lmU

PR #4513, opened 2022-11-01
Ap 3487 additional merits q3

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-3487)

## Checklist

### DIFF
--- a/app/controllers/providers/application_merits_task/nature_of_urgency_controller.rb
+++ b/app/controllers/providers/application_merits_task/nature_of_urgency_controller.rb
@@ -5,12 +5,22 @@ module Providers
         @form = ApplicationMeritsTask::NatureOfUrgencyForm.new(model: urgency)
       end
 
-      def update; end
+      def update
+        @form = ApplicationMeritsTask::NatureOfUrgencyForm.new(form_params)
+        render :show unless update_task_save_continue_or_draft(:application, :nature_of_urgency)
+      end
 
     private
 
       def urgency
-        legal_aid_application.urgency
+        legal_aid_application.urgency || legal_aid_application.build_urgency
+      end
+
+      def form_params
+        merged_params = merge_with_model(urgency) do
+          params.require(:application_merits_task_urgency).permit(:nature_of_urgency, :hearing_date_set, :hearing_date, :additional_information)
+        end
+        convert_date_params(merged_params)
       end
     end
   end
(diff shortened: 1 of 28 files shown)


colinbruce (approved):
LGTM :+1:

app/controllers/providers/application_merits_task/nature_of_urgency_controller.rb
```diff
@@ -0,0 +1,27 @@
+module Providers
+  module ApplicationMeritsTask
+    class NatureOfUrgencyController < ProviderBaseController
```
cpjmcquillan, line 3: I've mentioned this (i.e pluralisation) a couple times across various PRs.

I definitely think it's worthwhile defaulting to convention (e.g NatureOfUrgencies) unless there is a clear reason not to.

How do the rest of the team feel about this?

cpjmcquillan: I'd even suggest this could just be an Urgencies controller!

jsugarman: i had started with a plural controller name but changed to singular as the route and relations are all singular and convention only [favours pluralisation](https://guides.rubyonrails.org/action_controller_overview.html#controller-naming-convention) but does not adhere to it itself (i.e. ApplicationController).

kmahern: I have renamed the controller (+route etc.) urgencies_controller.rb. @colinbruce @jsugarman @cpjmcquillan are we all happy with the new naming?

jsugarman: I would have said its fine to make it plural if its already done as it is not something i think we need to sweat over too much, but `urgencies` for `nature_of_urgenc[ies|y]` i have to say I am not keen on

cpjmcquillan: I meant to delete my `urgencies` comment 😆 

I think `NatureOfUrgenciesController` makes most sense in the context of using a `NatureOfUrgency` form model to create `Urgency` records.

kmahern: 👍 ok, I have named the controller NatureOfUrgenciesController

PR #4518, opened 2022-11-02
[AP-3489] add "vary order" merits question 5

jsugarman:
## What
Implement the Vary order, section 8 question.

~~**DEPENDENCY:** requires a change to LFA, so this PR https://github.com/ministryofjustice/legal-framework-api/pull/787 will need merging before this can be merged~~ 

**REMOVE** last commit points UAT to a branch of LFA
[Link to story](https://dsdmoj.atlassian.net/browse/AP-3489)

 - [X] Add vary_orders table
 - [X] Add model and associations
 - [X] Add single line association tests
 - [x] Add new form for validating answer
 - [x] Add the route, controller and view
 - [x] Update the merits task flow and routing
 - [x] Updated the merits partial to show the response on the CYA, review and merits_report pages

## Checklist


spec/models/proceeding_spec.rb
```diff
@@ -32,6 +32,14 @@
                               :used_delegated_functions_reported_on)
   }
 
+  it { is_expected.to have_one(:attempts_to_settle).dependent(:destroy) }
+  it { is_expected.to have_one(:chances_of_success).dependent(:destroy) }
+  it { is_expected.to have_one(:vary_order).dependent(:destroy) }
+
+  it { is_expected.to have_many(:proceeding_linked_children).dependent(:destroy) }
+  it { is_expected.to have_many(:involved_children).through(:proceeding_linked_children).source(:involved_child) }
+  it { is_expected.to have_many(:scope_limitations).dependent(:destroy) }
```
jsugarman, line 41: happy to get rid of these if we do not think they are worth it.  I find them helpful in a TDD flow

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "rails_helper"
+
+RSpec.describe Providers::ProceedingMeritsTask::VaryOrderForm do
+  subject(:form) { described_class.new(params) }
+
+  let(:params) { { details: } }
+  let(:details) { Faker::Lorem.sentence }
```
cpjmcquillan, line 7: This has been mentioned before, but I think it's important we write deterministic tests.

If we introduce randomness, we invite flakiness —and flaky tests are useless tests.

If we truly are concerned about exploring the problem space, and ensuring our form can handle all manner of different inputs, then we should be specific and explicit —and write deterministic tests that cover those cases.

I suspect we aren't though, and that some arbitrary test data would suffice.

What do you think?

[This article by Martin Fowler](https://martinfowler.com/articles/nonDeterminism.html) puts it much better than I ever could, if people are interested.

app/services/flow/merits_loop.rb
```diff
@@ -40,6 +40,7 @@ def convert_task_to_flow_name(task)
         prohibited_steps: :prohibited_steps,
         laspo: :in_scope_of_laspos,
         nature_of_urgency: :nature_of_urgencies,
+        reason_for_new_application: :vary_order,
```
cpjmcquillan, line 43: It might just be me...

... but I think I prefer `reason_for_new_application` 😄

For me, it conveys more context and meaning than `vary_order` does, but that might be because I have limited domain knowledge.

Having said that, I would advocate for something like `NewApplicationReason` when it comes to model naming.

If the "order" aspect is an important detail, then I think `OrderVariation` works too.

I'd be interested to hear what you (and others) think!

app/views/providers/proceeding_merits_task/vary_order/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= form_with(
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder,
```
cpjmcquillan, line 2: This is our default form builder, so we shouldn't need to specify it.

spec/requests/providers/proceeding_merits_task/vary_order_controller_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+
+    context "when authenticated as a different provider" do
+      let(:login_provider) { login_as create(:provider) }
+
+      it_behaves_like "an authenticated provider from a different firm"
+    end
+  end
+
+  describe "PATCH /providers/applications/merits_task_list/:merits_task_list_id/vary_order" do
+    subject(:post_vary_order) do
```
cpjmcquillan, line 35: Even though I wish we were making a post request, we're not 😆 

Should we name this `patch_vary_order` to better reflect that?

cpjmcquillan (approved):
None of my comments are blocking, so feel free to take them or leave them.

I'm interested to hear other people's perspective on the naming of these records, so we can reach a shared understanding and encapsulate as much context as possible in the name.

Regarding the comments related to testing, I don't care too much about the tautological shoulda tests, but I do care quite a lot about us writing deterministic tests.

spec/models/proceeding_spec.rb
jsugarman: cool, will remove.

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
jsugarman: i agree, mostly, but in terms of free text i think non-deterministic can cover the random nature of inputs and is a pattern I see elsewhere and have used before. I have found non-deterministic tests which can be fail when it randomly puts characters that one would not expect and these can expose problems that might otherwise only be caught by "live" testing.

In this instance I am therefore not inclined to remove but do not feel particularly strongly about it

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      let(:details) { Faker::Lorem.sentence }
+
+      it { is_expected.to be_valid }
+    end
+
+    context "when details absent" do
+      let(:details) { nil }
+
+      it { is_expected.to be_invalid }
+      it { is_expected.not_to be_valid }
```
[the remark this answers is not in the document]
jsugarman: there is a slight difference actually. and wondered what people would think, but I will remove one

db/migrate/20221101141154_create_vary_order.rb
```diff
@@ -0,0 +1,10 @@
+class CreateVaryOrder < ActiveRecord::Migration[7.0]
+  def change
+    create_table :vary_orders, id: :uuid do |t|
```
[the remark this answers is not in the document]
jsugarman: legal naming - various proceeding types are applications to "vary the [original] order" as reason for a new application felt too generic to me. Myself and Colin had a discussion on this and settled on this naming.

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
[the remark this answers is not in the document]
cpjmcquillan: What's the difference?

I think [calling `.invalid?` is the same as calling `!valid?`](https://github.com/rails/rails/blob/8015c2c2cf5c8718449677570f372ceb01318a32/activemodel/lib/active_model/validations.rb#L373-L375).

app/services/flow/merits_loop.rb
jsugarman: this line should have ben removed as i have changed LFA to use vary_order too, for consistency

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
cpjmcquillan: My point is that a unit test that sometimes fails and sometimes passes is useless, because it can give us no confidence in our system.

What are we trying to test?

That the form can handle atypical characters? 
(e.g `I can input \";* strange and unexpected characters`)

Or that it is safe from XSS vulnerabilities? 
(e.g `Don't click this <script>alert("This is really bad!");</script>`)

Or that is can handle specific random combinations of words? 
(e.g `sausages and bacon breaks the form`)

As I said, if we are really interested in exploring the problem space and testing a range of inputs, then we should be explicit about them and write deterministic tests to cover them.

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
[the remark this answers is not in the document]
jsugarman: purely from the rspec output, but for rails validations it is fundamentally the same behaviour yes

```
    when details absent
      is expected to return true for `invalid?` or `invalids?`
      is expected not to be valid truthy
```

spec/forms/providers/proceeding_merits_task/vary_order_form_spec.rb
jsugarman: Clearly you feel strongly about it, but personally I don't think we need to sweat too much over such things.

I will remove

PR #4519, opened 2022-11-02
Simplify `LegalAidApplications::HasOtherProceedingsForm`

cpjmcquillan:
This prefers default Rails validations to custom validation, 
reduces the external API of the form model, and prefers 
default error message lookups to custom ones.


colinbruce (approved):
LGTM :+1:

PR #4520, opened 2022-11-02
Refactor routes

cpjmcquillan:
Introduce the `scope` helper with `module:` arguments to simplify
the application and proceeding level merits task list routes, as well
as the proceeding loop routes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -155,9 +155,6 @@ Rails.application.routes.draw do
       resource :address_selection, only: %i[show update]
       resource :check_benefit, only: %i[index update]
       resource :has_national_insurance_number, only: %i[show update]
-      resource :statement_of_case, only: %i[show update destroy], controller: "application_merits_task/statement_of_cases" do
-        get "/list", to: "application_merits_task/statement_of_cases#list"
-      end
       resources :check_benefits, only: [:index]
       resources :applicant_employed, only: %i[index create]
       resource :open_banking_consents, only: %i[show update], path: "does-client-use-online-banking"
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -248,15 +237,30 @@
...
 
     resources :merits_task_list, only: [] do
-      resource :attempts_to_settle, only: %i[show update], controller: "proceeding_merits_task/attempts_to_settle"
-      resource :linked_children, only: %i[show update], controller: "proceeding_merits_task/linked_children"
-      resource :prohibited_steps, only: %i[show update], controller: "proceeding_merits_task/prohibited_steps"
-      resources :chances_of_success, only: %i[index create], controller: "proceeding_merits_task/chances_of_success"
-      resource :success_prospects, only: %i[show update], controller: "proceeding_merits_task/success_prospects"
-      resource :specific_issue, only: %i[show update], controller: "proceeding_merits_task/specific_issue"
+      scope module: :proceeding_merits_task do
+        resource :attempts_to_settle, only: %i[show update], controller: "attempts_to_settle"
```
cpjmcquillan, line 257: In an ideal world we wouldn't need to specify the controller, but we're too far down the road of pluralising `attempts_to_settle` in some places but not others to row back now.

This is the simplest way to move forwards without significant refactor, and keeps things as they were.

cpjmcquillan: Actually, are we too far down that road?

Has this part of the user journey been released to any users yet?

If it hasn't, I would advocate for pluralising the controller and updating the `step` to reflect that.

An alternative would be defining the inflection as uncountable and updating the table name etc, but that would be a bit more involved.

Interested to hear what others think.

cc @colinbruce because I feel like you will know what's been released to who and when in this area!

config/routes.rb
```diff
@@ -248,15 +237,30 @@
...
+        resource :linked_children, only: %i[show update]
+        resource :prohibited_steps, only: %i[show update]
+        resources :chances_of_success, only: %i[index create]
+        resource :success_prospects, only: %i[show update]
+        resource :specific_issue, only: %i[show update], controller: "specific_issue"
```
cpjmcquillan, line 262: ⬆️

config/routes.rb
colinbruce: The Attempts to settle question has been live for some months now, this is my least favourite part of rails 😄 
Trying to work out their grammar logic. attempts_to_settles is not and never will be acceptable! 😁

colinbruce:
I like this idea in principle @cpjmcquillan but doing immediately is going to impact all of the in-progress question tickets.
How do you feel about parking it for a week until they are all done?
That way it will be the definitive work and not causing pain for other devs :+1:

cpjmcquillan:
> How do you feel about parking it for a week until they are all done?

Sure, that works.

I figured we'd want it in early or late to minimise disruption either way!

colinbruce (approved):
:+1:

PR #4523, opened 2022-11-02
AP-3493: Move LASPO question to merits task list

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3493)

Moved the in_scope_of_laspos controller, view, form, etc from the central namespace into application_merits_task
Updated the routing to ensure it is handled with the other merits tasks
Update associated tests to reflect the new flow.

This is not actually related to Section 8 so is not required to be behind any feature flags.  
When it is released it will go live immediately and this has been agreed with the PM.

## Checklist


kmahern (approved):
lgtm

spec/requests/providers/has_other_proceedings_controller_spec.rb
```diff
@@ -102,8 +102,8 @@
 
       let(:params) { { legal_aid_application: { has_other_proceeding: "false" } } }
 
-      it "redirects to the LASPO page" do
-        expect(response).to redirect_to(providers_legal_aid_application_in_scope_of_laspo_path(legal_aid_application))
+      it "no longer redirects to the LASPO page, but continues to delegated functions" do
```
cpjmcquillan, line 109: Should we care what the test used to test when we're reading it?

I'd be tempted to say

```ruby
it "redirects to the delegated functions page" do
```

and rely on the commit messages to tell the story of what changed and why.

What do you think?

cpjmcquillan (approved):

spec/requests/providers/has_other_proceedings_controller_spec.rb
colinbruce: I'm all about the history! 😆 
Fair point - will get updated when I remove the temp redirect commit 👍

PR #4525, opened 2022-11-03
Update default firm permissions

jsugarman:
Add employment permission to all firms by default

## What

Add employment permission to all firms inline
with production.

This will primarily mean we do not have to set
the permission on UAT environments.

_QUESTION: this script updates ALL firms permissions as part
seeding, but surely it should be just the test providers
we update here?!_

## Checklist

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -18,7 +18,6 @@ Seeder.monitor DocumentCategory
 Dir[Rails.root.join("db/seeds/*.rb")].each do |seed|
   load seed
 end
-Rake::Task["employ"].invoke if Rails.env.development?
 
 Rails.logger.info Seeder.report.join("\n")
 Rails.logger.info "Seeding completed"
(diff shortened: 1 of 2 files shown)


colinbruce:
This rake task was added for the inverse... we could auto-add it to all providers on UAT and prevent it running on Prod.
There is also [this ticket](https://dsdmoj.atlassian.net/browse/AP-3566) to remove permissions that are now universal

jsugarman:
> This rake task was added for the inverse... we could auto-add it to all providers on UAT and prevent it running on Prod. There is also [this ticket](https://dsdmoj.atlassian.net/browse/AP-3566) to remove permissions that are now universal

can we chat about this? not sure what you mean by the inverse.

jsugarman:
closing in favour of https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4699

PR #4528, opened 2022-11-03
Providers can categorise disregarded benefit transactions (again)

cpjmcquillan:
This behaviour was erroneously removed in #4366.

Once a client has provided access to their transactions using TrueLayer, providers need to be able to categorise their bank transactions.

One of these categories is "Disregarded benefits".

In #4366, the `TransactionType.find_with_children` method was removed (alongside its use in `Providers::Means::IdentifyTypesOfIncomesController`.

What we didn't know then (but we do now) is that the intention of this method was to allow child transaction types to be associated to an application when their parent transaction type was identified.

For instance:

  When "Benefits" are released, we want to create "Disregarded benefits"
  transaction types too.

This updates the affected controller to create transaction type records for all identified credit transaction types including their children (but not housing benefit).

A feature test has been added, since this behaviour was previously untested --hence the regression!


features/providers/non_passported_journey.feature
```diff
@@ -65,9 +66,28 @@ Feature: Non-passported applicant journeys
...
+      | tag | section |
+      | h2  | Housing benefits |
+
     Then I click the first link 'View statements and add transactions'
     Then I select the first checkbox
     And I click 'Save and continue'
+
+    # Then I click the '2nd' link 'View statements and add transactions'
+    # Then I select the first checkbox
+    # And I click 'Save and continue'
```
cpjmcquillan, line 85: I want to uncomment this, but I'm not sure how to set up the appropriate VCR cassette so that state benefit data is returned 🤔

features/step_definitions/generic_steps.rb
```diff
@@ -18,12 +18,12 @@
 
 Then("the following sections should exist:") do |table|
   table.hashes.each do |row|
-    expect(page).to have_selector(row[:tag], text: /\A#{Regexp.quote(row[:section])}\z/), "expected to find tag \"#{row[:tag]}\" with text: \"#{row[:section]}\""
+    expect(page).to have_selector(row[:tag], text: row[:section]), "expected to find tag \"#{row[:tag]}\" with text: \"#{row[:section]}\""
```
cpjmcquillan, line 21: I removed the `Regexp.quote` because it was confusing, and it didn't match on the income summary page.

Just matching on the provided string seems to work nicely for all the integration tests.

spec/requests/providers/means/identify_types_of_incomes_controller_spec.rb
```diff
@@ -2,8 +2,20 @@
 
 RSpec.describe Providers::Means::IdentifyTypesOfIncomesController do
   let(:legal_aid_application) { create(:legal_aid_application) }
-  let!(:outgoing_types) { create_list(:transaction_type, 3, :debit_with_standard_name) }
-  let!(:income_types) { create_list(:transaction_type, 3, :credit_with_standard_name) }
+  let!(:outgoing_types) do
```
jsugarman, line 5: nit - linebreak

spec/requests/providers/means/identify_types_of_incomes_controller_spec.rb
```diff
@@ -2,8 +2,20 @@
...
+    [
+      create(:transaction_type, :rent_or_mortgage),
+      create(:transaction_type, :maintenance_out),
+      create(:transaction_type, :child_care),
+    ]
+  end
+  let!(:income_types) do
```
jsugarman, line 12: nit - linebreak

spec/requests/providers/means/identify_types_of_incomes_controller_spec.rb
```diff
@@ -2,8 +2,20 @@
...
+    [
+      create(:transaction_type, :benefits),
+      create(:transaction_type, :pension),
+      create(:transaction_type, :maintenance_in),
+    ]
+  end
```
jsugarman, line 18: nit - linebreak

jsugarman: Personally I'd like to see a request spec here as well. bit belt and braces with the feature test but prefer to have to much

```
    context "when benefits selected" do
      let(:transaction_type_ids) { [benefits.id] }
      let(:benefits) { create(:transaction_type, :benefits) }

      before do
        create(:transaction_type, :excluded_benefits, parent_id: benefits.id)
      end

      it "adds excluded_benefits transaction type as well as benefits" do
        expect { request }.to change(LegalAidApplicationTransactionType, :count).by(2)
        expect(legal_aid_application.reload.transaction_types.pluck(:name)).to match_array(%w[benefits excluded_benefits])
      end
    end
```

app/models/transaction_type.rb
```diff
@@ -48,6 +48,10 @@ class TransactionType < ApplicationRecord
   scope :without_disregarded_benefits, -> { not_children }
   scope :without_housing_benefits, -> { where.not(name: "housing_benefit") }
 
+  def self.with_children(ids:)
+    where(id: ids).or(where(parent_id: ids).where.not(parent_id: nil))
+  end
```
jsugarman, line 53: 👍 preferable to the old method i'd say

features/providers/non_passported_journey.feature
jsugarman: could pair on tomorrow if you like, for vcr i generally ensure the local machine config is pointing at relevant staging remote apps then delete the exist cassette and rerun feature test.

jsugarman (commented):
looks good, 

- wary of removing the regex myself as i think it prevent false positives but no biggee (was it causing failures?)
- fairly sure this controller is only used for truelayer since EBSU was enabled  so we do not need to worry about disregarded benefits for bank statement upload flows 🤔 

And yes, UAT of truelayer and EBSU journeys as a minimum is necessary

app/models/transaction_type.rb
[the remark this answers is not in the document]
jsugarman: i guess it avoids an unnecessary DB hit too so 👍

features/providers/non_passported_journey.feature
cpjmcquillan: Sounds good, I'll try and schedule something. 

I don't like Cucumber or VCR so it's like two nemeses colliding 😆

features/step_definitions/generic_steps.rb
cpjmcquillan: Ok, that makes sense.

I'm happy to leave it if we can figure out why it wouldn't match `Benefits` on that page 🤔

jsugarman: @cpjmcquillan am happy to approve this PR if we can just revert this change that removes the regex. Can you put it in a separate branch to discuss so it does not block this PR which we need to get in ASAP. I am just concerned that changing this generic step used in many other features could introduce false positives in future if we remove.

features/providers/non_passported_journey.feature
```diff
@@ -65,9 +66,28 @@ Feature: Non-passported applicant journeys
...
     Then I should be on the 'income_summary' page showing "Sort your client's income into categories"
+    And the following sections should exist:
+      | tag | section |
+      | h2  | Benefits |
+      | h2  | Disregarded benefits |
+      | h2  | Financial help from friends or family |
+
+    And the following sections should not exist:
+      | tag | section |
+      | h2  | Housing benefits |
```
jsugarman, line 77: not a blocker but for this page I would use be tempted to use a more general

And I should not see 'Housing benefit'

features/step_definitions/generic_steps.rb
cpjmcquillan: Sure 👍

jsugarman (approved):
🎖️

PR #4529, opened 2022-11-03
AP-3488 Opponent application merits task list question

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3488)

- Added a new page for the opponents application question (controller, form, view, route)
- Added a new DB table opponents_applications
- This question appears on a proceeding level, for any proceeding which has the client involvement type as defendant

TODO: 
- Write tests

## Checklist

### DIFF
--- a/app/controllers/providers/proceeding_merits_task/opponents_application_controller.rb
+++ b/app/controllers/providers/proceeding_merits_task/opponents_application_controller.rb
@@ -26,7 +26,7 @@ module Providers
 
       def form_params
         merge_with_model(opponents_application) do
-          params.require(:proceeding_merits_task_opponents_application).permit(:opponents_application, :details)
+          params.require(:proceeding_merits_task_opponents_application).permit(:has_opponents_application, :reason_for_applying)
         end
       end
     end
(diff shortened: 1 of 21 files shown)


app/controllers/providers/proceeding_merits_task/opponents_application_controller.rb
```diff
@@ -0,0 +1,34 @@
+module Providers
+  module ProceedingMeritsTask
+    class OpponentsApplicationController < ProviderBaseController
```
cpjmcquillan, line 3: It might seem like a trifling issue, but I think we should be keeping our naming conventions as conventional as possible.

I _think_ that would mean an `OpponentsApplication` record backed by an `opponents_applications` table with an `OpponentsApplicationsController` by default.

Making sure we name things (and pluralise them) intentionally by convention should help us keep things simple (like our routes).

This is one of those things I normally need to double check myself, but I think it's worthwhile us trying to minimise these deviations from convention where we can!

jsugarman: i had started with a plural controller name but changed to singular as the route and relations are all singular and convention only [favours pluralisation](https://guides.rubyonrails.org/action_controller_overview.html#controller-naming-convention) but does not adhere to it itself (i.e. ApplicationController).

I would approve either way

jsugarman (approved):
👍 

The coverage issue is one and colin have dealt with. I am happy to pair on it!

PR #4532, opened 2022-11-07
Add application instance var to path call

jsugarman:
Add explicit instance vars to route path helper calls.

## What

Fixes potential failed merits report caused by various new merits question renderings

Errors are raised - see Sidekiq/ReportsCreatorWorker errors such as below

```
ActionView::Template::Error: No route matches {:action=>"show", :controller=>"providers/application_merits_task/client_denial_of_allegations", :locale=>:en}, missing required keys: [:legal_aid_applicat...
While route helpers can deduce missing ke
```

This is because while path helpers can deduce a missing require route key when rendered in the rails
standard MVC way (via ActionDispatch cleverness) the way the MeritsReportCreator (and MeansReportCreator) work is to call `Controller.renderer.render` directly, which cannot deduce the missing route keys (I assume).

Solution: always be explicit with your route params

see https://thoughtbot.com/blog/avoid-making-path-helpers-guess#searching-for-missing-route-values
for more.

## Checklist

### DIFF
--- a/app/views/shared/check_answers/_merits.html.erb
+++ b/app/views/shared/check_answers/_merits.html.erb
@@ -124,7 +124,7 @@
     <dl class="govuk-summary-list govuk-!-margin-bottom-2">
       <%= check_answer_link(
             name: :allegation_denies_all,
-            url: providers_legal_aid_application_client_denial_of_allegation_path,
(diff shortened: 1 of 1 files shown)


cpjmcquillan (approved):
💫

colinbruce (approved):

PR #4544, opened 2022-11-08
ap-3524: display multiple scope limitations on limitations and

kmahern:
review and print application pages

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3524)

## Checklist

### DIFF
--- a/app/helpers/proceeding_helper.rb
+++ b/app/helpers/proceeding_helper.rb
@@ -11,21 +11,22 @@ module ProceedingHelper
   end
 
   def scope_limits(proceeding, scope_type)
-    proceeding
-      .scope_limitations
-      .where(scope_type:)
-      .map { |scope_limitation| scope_limitation_meaning(scope_limitation) }
+    scope_limitations = proceeding.scope_limitations
+                                  .where(scope_type:)
+                                  .map { |scope_limitation| scope_limitation_details(scope_limitation) }
+                                  .join("<br><br>")
+    sanitize(scope_limitations)
   end
 
 private
 
-  def scope_limitation_meaning(scope_limitation)
-    scope_limitation_meaning = scope_limitation.meaning
+  def scope_limitation_details(scope_limitation)
+    scope_limitation_details = "#{scope_limitation.meaning}<br>#{scope_limitation.description}"
     if scope_limitation.hearing_date
-      scope_limitation_meaning += "<br>Date: #{scope_limitation.hearing_date}"
+      scope_limitation_details += "<br>Date: #{scope_limitation.hearing_date}"
     elsif scope_limitation.limitation_note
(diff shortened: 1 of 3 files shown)


app/helpers/proceeding_helper.rb
```diff
@@ -11,21 +11,22 @@ def position_in_array(proceeding)
...
+    scope_limitation_details = "#{scope_limitation.meaning}<br>#{scope_limitation.description}"
     if scope_limitation.hearing_date
-      scope_limitation_meaning += "<br>Date: #{scope_limitation.hearing_date}"
+      scope_limitation_details += "<br>Date: #{scope_limitation.hearing_date}"
     elsif scope_limitation.limitation_note
-      scope_limitation_meaning += "<br>Note: #{scope_limitation.limitation_note}"
+      scope_limitation_details += "<br>Note: #{scope_limitation.limitation_note}"
     end
-    sanitize(scope_limitation_meaning)
+    scope_limitation_details
```
cpjmcquillan, line 30: I'm intrigued by the use of the double line break to separate scope limits.

Could we use paragraphs instead? If the purpose of them is to add spacing between distinct bodies of text, [I don't think line breaks are recommended](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br).

Also, how do you feel about [using `safe_join` instead of joining, then using `sanitize`](https://api.rubyonrails.org/classes/ActionView/Helpers/OutputSafetyHelper.html#method-i-safe_join)?

I wonder if something like this could work, what do you think?

```ruby
def scope_limits(proceeding, scope_type)
  scope_limitations = proceeding.scope_limitations
                                .where(scope_type:)
                                .map { |scope_limitation| scope_limitation_details(scope_limitation) }
                                                       
  safe_join(scope_limitations)
end


def scope_limitation_details(scope_limitation)
  scope_limitation_details = [scope_limitation.meaning, scope_limitation.description]
  
  if scope_limitation.hearing_date
    scope_limitation_details << "Date: #{scope_limitation.hearing_date}"
  elsif scope_limitation.limitation_note
    scope_limitation_details << "Note: #{scope_limitation.limitation_note}"
  end

  tag.p(safe_join(scope_limitation_details, "<br>"), class: "govuk-!-margin-bottom-8")
end
```

cpjmcquillan (approved):

app/helpers/proceeding_helper.rb
kmahern: Thanks @cpjmcquillan 👍 for the use of safe_join, have amended proceeding_helper to use it. I went with the use of <br><br> rather than <p> tag because the screen shots had the "Work you can do" heading and the first scope limitation meaning on the same line on the limitations page, and I don't think this is possible with the <p> tag? (but might be missing something 🙂 )

PR #4558, opened 2022-11-10
Fix: Update overcommit rule for DB anonymisation

colinbruce:
## What

It was running (slowly) on each commit this changes it so that it only runs when files in the db/migrate folder are in the commit, this retains the speed of not tripping the github action with a speed increase on every other commit!

## Checklist


jsugarman (approved):
👍

PR #4566, opened 2022-11-14
Fix flaky bank statement spec

cpjmcquillan:
Before, this test was flaky because it would try to navigate to the next page while a document was being uploaded.

In #4375, that document upload spec expected to see "document.pdf UPLOADED" content before clicking "Save and continue". This ensured that the document had finished uploaded before navigating away.

This test appeared to work similarly, except it would fail sporadically.

On closer inspection, it turns out the "hello_world.pdf" file was already uploaded. This meant the test attempted to navigate away by clicking "Save and continue" straight away (rather than waiting for the upload to finish).

This updates the test to upload a different second file, and expect both files to be uploaded before navigating away from the page.

As in #4375, similar considerations should be made, but in the meantime this prevents the test from being flaky.

### DIFF
--- a/features/providers/check_your_answers.feature
+++ b/features/providers/check_your_answers.feature
@@ -321,11 +321,11 @@ Feature: Checking answers backwards and forwards
     And I should not see 'Does your client have any savings accounts they cannot access online?'
 
     When I click Check Your Answers Change link for 'bank statements'
-    And I upload an evidence file named 'hello_world.pdf'
     Then I should see "hello_world.pdf UPLOADED"
+    And I upload an evidence file named 'hello_world1.pdf'
+    Then I should see "hello_world1.pdf UPLOADED"
     And I click 'Save and continue'
     Then I should be on the 'means_summary' page showing 'Check your answers'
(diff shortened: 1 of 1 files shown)


jsugarman (approved):
👍

naseberry:
Good spot 👍 
Does this however highlight another issue? Should we be able to upload the same file multiple times?

cpjmcquillan:
> Does this however highlight another issue? Should we be able to upload the same file multiple times?

No, there isn't an issue with uploading a file multiple times.

Perhaps the commit message could have been clearer, but the specific issue here is confined to the feature test.

```
When a document has previously been uploaded,
then we upload the same document again,
and we expect to see "document UPLOADED"
```

Capybara will return straight away, because the expected content is already there.

In this instance, we want to assert that the second document has been uploaded, so we need to choose two files with different names.

Outside the confines of our test, that doesn't matter.

I think it does highlight an issue with the way we test these interactions (that I touch on in #4375), but that doesn't mean we should resolve the flakiness of this test in the first instance.

PR #4568, opened 2022-11-15
Fix student finance irregular income bug 

cpjmcquillan:
## Fix student finance irregular income bug
Before, student loan irregular income records were created every time the
`StudentFinances::AnnualAmountForm` was saved.

This meant when a provider specified that their client *did not* receive student
finance, an irregular income record with an amound of `nil` was persisted.

Since the boolean answer to this "Yes/No" question is stored on the legal aid
application already, there is no need to create an irregular income record that
captures the absence of data.

The existing implentation was confusing, and apparently unintentional. If we
consider the "Review and print your application" page, student loan data is
rendered in the "Income, payments, and assets" section if there is *any*
irregular income data. This implied that irregular income data should only exist
if the client is in receipt of student finance.

This change resolves that bug by ensuring that student loan irregular income
records are only persisted when the client is in receipt of student finance, and
they have specified an amount.

A separate bug relating to "orphaned" irregular income records has also been
fixed.

Before, if a provider selected "No" after previously selecting "Yes" and
specifying how much, the irregular income record was never updated.

This resolves that issue by destroying associated student loan irregular income
records when the client does not receive student finance.

## Communicate that there is no student finance in review

Before, when a provider reviewed or printed an application, they would see an
empty "Student loan" row in the income table if they had specified that the
client did not receive any student finance.

This was due to a bug involving erroneous irregular income records being created
when answering "No" to the student finance question.

Now that issue has been resolved, there is no indication to the provider whether
or not the applicant was in receipt of student finance or not when they review
or print the application.

This updates the review or print page to clearly communicate when the provider
has told us that the client does not receive student finance.

Now, a section has been added (similarly to the non-TrueLayer page) that
displays

```
Does your client receive student finance?             No
```

Existing behaviour (i.e a "Student loan" row in the income table) persists for
applications where the client is in receipt of student finance.

## Screenshots

<details>
<summary>Bank upload journey</summary>

BankUpload

</details>

<details>
<summary>TrueLayer journey (before)</summary>

TrueLayer_Before

</details>

<details>
<summary>TrueLayer journey (after bug fix)</summary>

TrueLayer_After_v1

</details>

<details>
<summary>TrueLayer journey (with question)</summary>

TrueLayer_After_v2

</details>


cpjmcquillan:
Given the existing implementation seemed unintentional (i.e there was a bug), I think there might be space for a discussion with Product about whether we want to keep the change outlined in 85673ab41e0c19efbcd23e11233a248afdc6748a.

From my perspective I think it is useful information and worthwhile adding, but it does seem out of place considering the structure of the rest of the data. Perhaps we should be displaying `£0` instead?

db/data/20221114151504_destroy_redundant_irregular_incomes.rb
```diff
@@ -0,0 +1,15 @@
...
+      legal_aid_applications.each do |legal_aid_application|
+        legal_aid_application.irregular_incomes.destroy_all
+      end
+    end
+  end
+
+  def down
+    nil
+  end
+end
```
cpjmcquillan, line 15: Generally I'm wary of using migrations that reference application code, but we've done it lots already.

I used `data_migrate` because it was there, but I don't mind whether we use this, a normal migration, or a temporary rake task.

What do we prefer?

cpjmcquillan: Also, are these run automatically on deploy?

jsugarman: yes they are run automatically on deploy

jsugarman: since this is purely a data cleanup i think it makes sense to have it here

jsugarman: currently only 257 applications with student_finance: false so should be quick/fine. 

However, Was wondering if these data migrations implicitly remove the transaction block, as i believe we do not need this, possibly adding our own per batch if needed?


```
disable_ddl_transaction!

def up
    LegalAidApplication.where(student_finance: false).find_in_batches do |legal_aid_applications|
      legal_aid_applications.each do |legal_aid_application|
        legal_aid_application.irregular_incomes.destroy_all
      end
    end
    sleep(0.1) # throttle, but no effect when < 1000
 end

```

features/providers/review_and_print.feature
```diff
@@ -128,13 +128,75 @@ Feature: Review and print your application
...
+      | Student loan |
+
+    And I should not see "Housing benefit total"
+
+    And I should see "Housing payments"
+    And I should see "Childcare payments"
+    And I should see "Maintenance payments to a former partner"
+    And I should see "Payments towards legal aid in a criminal case"
+
+  Scenario: For a non-passported truelayer bank transactions journey
```
jsugarman, line 147:
```suggestion
  Scenario: For a non-passported truelayer bank transactions journey without student finance
```

features/providers/review_and_print.feature
```diff
@@ -128,13 +128,75 @@ Feature: Review and print your application
...
+
+    And the "Income, regular payments and assets" review section should contain:
+      | question |
+      | Benefits total |
+      | Disregarded benefits total |
+      | Financial help from friends or family total |
+      | Maintenance payments from a former partner total |
+      | Income from a property or lodger total |
+      | Pension total |
+
```
jsugarman, line 197: This appears to be a repetition of above scenario, so wonder if it is worth it - might be clearer/simpler to just focus on the student loan aspect?!

spec/requests/providers/means/student_finances_controller_spec.rb
```diff
@@ -0,0 +1,217 @@
...
+
+    context "when the provider continues" do
+      let(:button_clicked) { continue_button }
+
+      context "when the provider selects `Yes`" do
+        let(:student_finance) { "true" }
+        let(:amount) { "5000" }
+
+        it "updates the application" do
+          login_as provider
```
jsugarman, line 86: personally i'd prefer a context "when authenticated", with before { login_as provider } to remove the setup from these examples

jsugarman (approved):
👍 Code wise, aside from [spec DRYing](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4568#discussion_r1022747316) it looks ok.

Could do with UATing due to the scope of the changes

PR #4576, opened 2022-11-16
Update Puma to v6.0

cpjmcquillan:
This is a [major update](https://github.com/puma/puma/releases/tag/v6.0.0) of arguably the most important part of our application 
(the webserver), so proceed with caution!

It looks like the only breaking change that impacted us was the one that broke
Capybara. This has since been [resolved](https://github.com/teamcapybara/capybara/pull/2530) and [released](https://github.com/teamcapybara/capybara/releases/tag/3.38.0).

CI is green, but theres no place quite like production. This change should be
sanity checked in UAT and Staging environments, and monitored in Production.


jsugarman (approved):
👍 have UAT'd a BSU flow and no probs

PR #4583, opened 2022-11-17
ap-3624: prevent duplicate scope limitations

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3624)

It is possible that the user has already entered non-default scope limitations when they arrive at the default scope limitation page (e.g. if they use the back button). The application was adding the default scope limitation in addition to any existing scope limitations. Instead, delete any existing scope limitations. This applies to both substantive and emergency scope limitations.

## Checklist


colinbruce (approved):
LGTM 👍

PR #4587, opened 2022-11-18
Update GOV.UK FormBuilder and Frontend

cpjmcquillan:
- Update GOVUK Formbuilder to v3.3.0
- Update GOVUK Frontend to v4.4.0
- Update `govuk-error-summary` usage

In [GOVUK Frontend 4.4.0](https://github.com/alphagov/govuk-frontend/releases/tag/v4.4.0), the HTML for the error summary was updated.

The error summary is typically rendered by the form builder (which has updated
the HTML), but in some instances we are rendering the error summary outside the
form builder.

In these scenarios, the HTML as been updated in line with the recommended
changes.

1. Remove `aria-labelledby="error-summary-title"` and `role="alert"` from the
parent element (`.govuk-error-summary`).
2. Add a div wrapper around the contents of `govuk-error-summary` with the
attribute `role="alert"`.
3. Remove `id="error-summary-title"` from the error summary `h2` element.


app/views/shared/_error.html.erb
```diff
@@ -1,18 +1,20 @@
...
+      <ul class="govuk-list govuk-error-summary__list">
+        <% if @error.is_a? String %>
+          <li>
+            <%= link_to_accessible @error, "#correct" %>
+          </li>
+        <% else %>
+          <li>
+            <%= link_to_accessible @error.values.first, "##{@error.keys.first}" %>
+          </li>
+        <% end %>
```
naseberry, line 16: As the only thing that changes, based on the conditon, is the link, could this be improved, for example:

```
<li>
  <% if @error.is_a? String %>
    <%= link_to_accessible @error, "#correct" %>
  <% else %>
    <%= link_to_accessible @error.values.first, "##{@error.keys.first}" %>
  <% end %>
</li>
```

naseberry:
The "Error summary" component is currently manually written in at least 7 files, as this is being used outside the scope of the FormBuilder gem, and the Frontend gem does not have this as a component, would it be ideal to either write our own custom helper OR open a PR to add this to the Frontend gem?

cpjmcquillan:
> The "Error summary" component is currently manually written in at least 7 files, as this is being used outside the scope of the FormBuilder gem, and the Frontend gem does not have this as a component, would it be ideal to either write our own custom helper OR open a PR to add this to the Frontend gem?
> 
> 
> 
> This would make it reusable and easy to maintain if other future changes need to be applied.

I would argue that we shouldn't be using the error summary outside the context of a form, so all usages should be using the form builder.

As far as I know, the component isn't intended to be used in any other context.

PR #4589, opened 2022-11-18
Update schema.rb to reflect production

cpjmcquillan:
1. Timestamp precision

This was originally addressed in 22fe6852d095d7d096d1de790af58413e0187bd0 but
the changes were since merged in a different change inadvertently.

Since Rails 7.0, datetime column precision is set to 6 by default.

This means `t.datetime "partied_at"` would generate a column in postgres of
`timestamp(6) without time zone` type.

Before Rails 7.0, datatime column precision was set to nil by default.

This meant `t.datetime "partied_at"` would have generated a column in postgres
of `timestamp without time zone` type.

It seems that we always used the default, which means any datetime columns added
to the database **before** Rails 7.0 have a type without any precision, but any
datetime columns added to the database **after** Rails 7.0 have a type with
precision (6).

Currently our schema suggests that all datetime columns in the database have
precision (6), but this isn't the case.

This updates the datetime columns in the schema to reflect that some columns
have a type without precision (nil) while others have a type with precision (6).

2. Nullable columns

This was originally addressed in ebdf9e8522ab65cbee8881ab0a2fed6131ebe452 but
the changes were since merged in a different change inadvertently.

These columns were added as nullable string columns in #4048, but for some
reason the schema did not reflect this until #4530.

Since then, #4529 undid the change in #4530, so this redoes the change again.

Before, the columns were nullable but specified as `null: false` in the schema.

Now, they are still nullable, and the schema reflects that.

3. Column ordering

Less significant, but discrepancies between production and the schema make
conflicts more likely and make it harder to ensure the correct changes are added
to the schema after running migrations.

4. Misnamed index

The index is named unconventionally in the schema. It has been updated to
reflect the naming in production.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -111,9 +111,9 @@ ActiveRecord::Schema[7.0].define(version: 2022_11_13_212453) do
     t.string "unlock_token"
     t.datetime "locked_at", precision: nil
     t.string "true_layer_secure_data_id"
+    t.boolean "employed"
     t.datetime "remember_created_at", precision: nil
     t.string "remember_token"
-    t.boolean "employed"
     t.boolean "self_employed", default: false
     t.boolean "armed_forces", default: false
     t.boolean "has_national_insurance_number"
@@ -558,9 +558,9 @@ ActiveRecord::Schema[7.0].define(version: 2022_11_13_212453) do
     t.boolean "no_cash_income"
     t.boolean "no_cash_outgoings"
     t.date "purgeable_on"
+    t.string "required_document_categories", default: [], null: false, array: true
     t.boolean "extra_employment_information"
     t.string "extra_employment_information_details"
-    t.string "required_document_categories", default: [], null: false, array: true
     t.string "full_employment_details"
     t.datetime "client_declaration_confirmed_at", precision: nil
     t.boolean "substantive_cost_override"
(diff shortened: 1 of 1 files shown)


colinbruce (approved):
Why do these keep re-setting?
Should we encourage all Devs to delete their local DB and rebuild, ensuring they are using PSQL 11?

cpjmcquillan:
> Why do these keep re-setting?

@colinbruce I'm not sure what you mean?

This isn't an issue with database versions as far as I know. Timestamp precision has been a thing in postgres for a long time.

This addresses an issue similar to the one addressed in #3992.

That change updated the schema to reflect that `precision: 6` was now the default in Rails when generating datetime columns.

Currently the schema implies that lots of datetime columns have the default Rails precision (i.e 6), when they really have the database default (i.e nil).

This change updates the schema to reflect that `precision: nil` refers to the database default when generating datetime columns.

Is that helpful?

colinbruce:
We have seen some of these schema settings come and go...
The sequence of columns, the nullable ccms_codes.  They get added, a dev runs db:migrate, they get removed and accidentally committed.
It wasn't necessarily aimed at this PR, but the schema file seems to have been on a roundabout generally for the last few months 🤷‍♂️

PR #4590, opened 2022-11-18
Update schema.rb to reflect nullable columns

cpjmcquillan:
These columns were added as nullable string columns in #4048, but for some reason the schema did not reflect this until #4530.

Since then, #4529 undid the change in #4530, so this redoes the change again.

Before, the columns were nullable but specified as `null: false` in the schema.

Now, they are still nullable, and the schema reflects that.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -710,8 +710,8 @@ ActiveRecord::Schema[7.0].define(version: 2022_11_10_161139) do
     t.string "category_of_law", null: false
     t.string "category_law_code", null: false
     t.string "ccms_matter_code"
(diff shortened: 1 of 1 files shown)


colinbruce:
No idea why this one keeps getting re-set! 😖

colinbruce (approved):

cpjmcquillan:
> No idea why this one keeps getting re-set! 😖

It will get added to schema if people's databases are out-of-sync. 

In this case, both columns are nullable on production, so they should be nullable on our local databases too.

I suspect the `null: false` was added by mistake in #4529.

Once we have these schema issues resolved, it should make life a lot easier for people when updating it after running migrations.

PR #4592, opened 2022-11-18
Update schema.rb to reflect column order

cpjmcquillan:
This column was probably added to the table at a similar time to other columns.

This meant the schema ended up out of sync, so this updates the schema to reflect reality.


colinbruce (approved):

cpjmcquillan:
Closing this because it isn't an issue for other team members, and the issue isn't reflected on production.

PR #4605, opened 2022-11-21
Address Rubocop erblint failures in app/views/shared

cpjmcquillan:
Before, all non-nested templates in `app/views/shared` were excluded from Rubocop linting.

All failures were resolved with the `--autocorrect` flag, and are all trivial

One caveat is that when assigning local variables in partials inside a multiline block, Rubocop thinks the first line is empty. This is a limitation of using Rubocop in erblint, since each line is parsed in isolation.

### DIFF
--- a/app/views/shared/partials/_notification_banner.html.erb
+++ b/app/views/shared/partials/_notification_banner.html.erb
@@ -1,6 +1,6 @@
-<% success_class = local_assigns[:success] ? 'govuk-notification-banner--success' : ''
-   header = local_assigns[:success] ? t('generic.success') : t('generic.important')
-   link_to = false unless local_assigns.has_key?(:link_to) %>
+<% success_class = local_assigns[:success] ? "govuk-notification-banner--success" : ""
+   header = local_assigns[:success] ? t("generic.success") : t("generic.important")
+   link_to = false unless local_assigns.key?(:link_to) %>
 <div class="govuk-notification-banner <%= success_class %>" role="alert" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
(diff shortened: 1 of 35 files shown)


app/views/shared/partials/_notification_banner.html.erb
```diff
@@ -1,6 +1,6 @@
-<% success_class = local_assigns[:success] ? 'govuk-notification-banner--success' : ''
-   header = local_assigns[:success] ? t('generic.success') : t('generic.important')
-   link_to = false unless local_assigns.has_key?(:link_to) %>
+<% success_class = local_assigns[:success] ? "govuk-notification-banner--success" : ""
```
colinbruce, line 1: Should this be split into 3 lines like the other examples?

app/views/shared/partials/_revealing_checkbox.html.erb
```diff
@@ -1,21 +1,21 @@
-<% field_name="check_box_#{name}"
+<% field_name = "check_box_#{name}"
```
colinbruce, line 1: Should these be split too?
```suggestion
<% field_name = "check_box_#{name}" %>
<% model.__send__("#{field_name}=", model.send(field_name).present?)  %>
<% error_class = nil  %>
```

colinbruce (approved):
A couple of observations, but otherwise :+1:

app/views/shared/partials/_notification_banner.html.erb
cpjmcquillan: They could be, but it doesn't really matter.

The reason they haven't been is because none of the lines were parsed as "empty", so there weren't any blank line "failures" that I needed to manually correct.

PR #4608, opened 2022-11-22
Fix: Update test REGEXs for tasks

colinbruce:
## What

When these were first written we only had a maximum of 8 tasks as we start increasing, more and more often we hit errors when the regex fails to match as it moves from 9 to 11.  

Some have been updated to fix these errors as they've occurred, this PR globally replaces all of them so the tasks will be matched whether they are dependency 1 or 10

## Checklist

### DIFF
--- a/spec/models/legal_framework/merits_task_list_spec.rb
+++ b/spec/models/legal_framework/merits_task_list_spec.rb
@@ -34,8 +34,8 @@ module LegalFramework
 
       it "updates the task list" do
         # match pattern => key: :value, newline, any amount of spaces, repeat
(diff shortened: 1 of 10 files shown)


cpjmcquillan (approved):
Personally I've found the current style of assertion quite difficult to comprehend.

How do you feel about defining a more descriptive method instead?

Something like 

```ruby
it "is complete" do
  # other setup
  expect(matter_opposed_task.state).to eq(:complete)
end

def matter_opposed_task
  application_tasks = merits_task_list.reload.task_list.tasks[:application]
  application_tasks.detect { |task| task.name == :why_matter_opposed }
end
```

colinbruce:
Nice idea... let me close this and try creating a custom matcher :+1:
Make it simpler for future devs, too

PR #4609, opened 2022-11-22
Ap 3497 multiple scope limitations ccms

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3497)

- Modified CCMS payload to send multiple emergency/substantive scope limitations for each proceeding where these exist.
- Specify "MULTUPLE" as requested scope where there exist multiple emergency or substantive scope limitations for a proceeding, or where delegated functions have been used.

## Checklist

### DIFF
--- a/app/services/ccms/attribute_value_generator.rb
+++ b/app/services/ccms/attribute_value_generator.rb
@@ -308,7 +308,12 @@ module CCMS
     end
 
     def multiple_scope_limitations?(proceeding)
-      proceeding.scope_limitations.where(scope_type: :emergency).count > 1 || proceeding.scope_limitations.where(scope_type: :substantive).count > 1
+      proceeding
+        .scope_limitations
+        .select(:scope_type)
+        .group(:scope_type)
+        .count
+        .any? { |_type, count| count > 1 }
     end
 
     def other_party_full_name(options)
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
Looks good code-wise, but will need in depth UAT for confidence with the CCMS end ! 👍

app/services/ccms/attribute_value_generator.rb
```diff
@@ -302,11 +302,15 @@ def benefit_check_passed?(_options)
...
       proceeding = options[:proceeding]
-      return "MULTIPLE" if proceeding.used_delegated_functions?
+      return "MULTIPLE" if proceeding.used_delegated_functions? || multiple_scope_limitations?(proceeding)
 
       proceeding.substantive_scope_limitation_code
     end
 
+    def multiple_scope_limitations?(proceeding)
+      proceeding.scope_limitations.where(scope_type: :emergency).count > 1 || proceeding.scope_limitations.where(scope_type: :substantive).count > 1
+    end
```
cpjmcquillan, line 312: If you only want to query the database once, I think you should be able to do something like this

```ruby
def multiple_scope_limitations?(proceeding)
  proceeding
    .scope_limitations
    .select(:scope_type)
    .group(:scope_type)
    .count
    .any? { |type, count| count > 1 }
end
```

kmahern: Thanks! have amended the query.

PR #4616, opened 2022-11-23
Fix: Update ScheduledMailingsDeliveryJob

colinbruce:
## What

Currently, the scheduled mailer runs each morning on each UAT instance.  If K8s has deleted the postgres pod sidekiq will try multiple times to connect and send mails.  This results in thousands of sentry errors
image
This job runs every 30 minutes, so when the DB has been recycled there are a _lot_ of errors

This PR attempts to allow the job to fail politely.  If it cannot find the DB it will send a single Sentry message and then allow the job to exit, this should mean that sidekiq won't keep retrying as the job has not 'failed'

We can monitor this by tracking the sentry stats for the UAT instance - I think there will still be many alerts but fewer than there are now

## Checklist

### DIFF
--- a/app/jobs/scheduled_mailings_delivery_job.rb
+++ b/app/jobs/scheduled_mailings_delivery_job.rb
@@ -9,6 +9,8 @@ class ScheduledMailingsDeliveryJob < ApplicationJob
     end
     reschedule unless JobQueue.enqueued?(ScheduledMailingsDeliveryJob)
     EmailMonitorJob.perform_later unless JobQueue.enqueued?(EmailMonitorJob)
+  rescue ActiveRecord::StatementInvalid
+    Sentry.capture_message("Database for #{ENV.fetch('APP_BRANCH', 'unknown branch')} not available to send scheduled mailings")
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
💥 Looks good

cpjmcquillan (commented):
I wonder whether it might be worthwhile being more specific and explicit with our retry strategy?

I haven't given it too much thought yet, nor looked in to our current configuration —but I thought I'd raise it as an idea!

colinbruce:
> I wonder whether it might be worthwhile being more specific and explicit with our retry strategy?
> 
> I haven't given it too much thought yet, nor looked in to our current configuration —but I thought I'd raise it as an idea!

It's a UAT issue. We want this job running regularly. When K8s UAT gets swamped with pods and runs out of memory, it tends to terminate postgres pods.  The controller then brings up a new pod which creates the DB but doesn't run migrate and seed. 
So, often, UAT DB's are present but not populated, then the regular cron job runs and fails because the DB is not ready

PR #4617, opened 2022-11-23
Use `govuk_notification_banner` component

cpjmcquillan:
Before, custom markup in a partial was used to render 
the [GOV.UK Notification Banner][component].

Rather than maintaining custom HTML, we can make use 
of the `govuk_notification_banner` ViewComponent instead.

This adopts that component and removes the now unused 
notification banner partial.

[component]: https://design-system.service.gov.uk/components/notification-banner/


kmahern (approved):
lgtm

PR #4620, opened 2022-11-23
Include Capybara matchers in request specs

cpjmcquillan:
In #4252 we added Capybara support for integration tests (i.e request specs).

This worked mostly well, except any failures would hang for a long time, and eventually threw an error in the `super_diff` gem.

I documented [the issue] and isolated it in an [example repository], where it was possible to ascertain that:

1. The issue only occured when using `super_diff`
2. The issue only occured in request specs
3. The issue only occured in RSpec (although `super_diff` is only compatible with RSpec

Still, the root cause was elusive.

Eventually I found time to do some more digging, and I noticed a discrepancy between the matcher in the monkey patched `handle_failure` method for a request spec and a system spec.

In the request spec, the matcher was an instance of `Capybara::RSpecMatchers::Matchers::HaveSelector`, but in system specs it was an instance of `RSpec::Matchers::BuiltIn::Has`.

Therein lies the rub.

If we explicitly include the Capybara matchers in request specs, our worries fade away into distant memories and failures behave as expected.

[the issue]: https://github.com/mcmire/super_diff/issues/161
[example repository]: https://github.com/cpjmcquillan/super_diff_capybara_request_specs


jsugarman (approved):
👍 

should we not explicitly remove any
```ruby
include Capybara::RSpecMatchers
```
from request specs here as well then

spec/requests/providers/bank_statements_spec.rb
spec/requests/providers/has_national_insurance_numbers_controller_spec.rb
spec/requests/providers/means_reports_spec.rb

from my quick search

PR #4622, opened 2022-11-24
Refactor age calculator spec

jsugarman:
## What

Refactor age calculator spec

[relates to spike](https://dsdmoj.atlassian.net/browse/AP-3629)

For clarity and to cover leap years

## Checklist

### DIFF
--- a/app/services/age_calculator.rb
+++ b/app/services/age_calculator.rb
@@ -1,6 +1,6 @@
 class AgeCalculator
-  def self.call(date_of_birth, date)
-    age = date.year - date_of_birth.year
-    date_of_birth > date.years_ago(age) ? age - 1 : age
+  def self.call(date_of_birth, as_of)
+    age = as_of.year - date_of_birth.year
+    date_of_birth > as_of.years_ago(age) ? age - 1 : age
   end
 end
--- a/spec/services/age_calculator_spec.rb
+++ b/spec/services/age_calculator_spec.rb
@@ -3,29 +3,51 @@ require "rails_helper"
 RSpec.describe AgeCalculator do
   subject { described_class.call(date_of_birth, date) }
 
-  let(:date) { Date.parse("15-08-2010") }
+  let(:date_of_birth) { Date.parse("15-08-2000") }
 
-  context "birthday is just before date" do
-    let(:date_of_birth) { Date.parse("14-08-2000") }
+  context "when date is 1 day after birthday" do
+    let(:date) { Date.parse("16-08-2010") }
 
-    it("gives the right age") { expect(subject).to eq(10) }
+    it { is_expected.to eq(10) }
   end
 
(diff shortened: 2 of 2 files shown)


spec/services/age_calculator_spec.rb
```diff
@@ -3,29 +3,51 @@
...
+      it { is_expected.to eq(11) }
+    end
+
+    context "with date of 28th February in a non-leap year" do
+      let(:date) { Date.parse("28-02-2010") }
+
+      it { is_expected.to eq(9) }
+    end
+
+    context "with date of 1st March February in a non-leap year" do
```
kmahern, line 47: Should this be "with date of 1st March in a non-leap year"?

kmahern (approved):
One small typo otherwise lgtm

app/services/age_calculator.rb
```diff
@@ -1,6 +1,6 @@
 class AgeCalculator
-  def self.call(date_of_birth, date)
-    age = date.year - date_of_birth.year
-    date_of_birth > date.years_ago(age) ? age - 1 : age
+  def self.call(date_of_birth, as_of)
+    age = as_of.year - date_of_birth.year
+    date_of_birth > as_of.years_ago(age) ? age - 1 : age
   end
 end
```
cpjmcquillan, line 6: I appreciate we use the so called "service object" pattern throughout the application, but I wish we could call it like this instead

```ruby
Age.calculate(date_of_birth, as_of)
```

cpjmcquillan: Also, would it be clearer if the method was more explicit?

```ruby
class Age
  def self.calculate(date_of_birth, as_of)
    years_since_birth = as_of.year - date_of_birth.year
    next_birthday = date_of_birth.years_since(years_since_birth)
    
    if as_of.before?(next_birthday)
      years_since_birth - 1
    else
      years_since_birth
    end
  end
end
```

cpjmcquillan (approved):
I added a suggestion that might help reveal the method's intention a bit more clearly.

What do you think?

Other than that, even renaming the variable to `as_of` helps massively, nice work 👍

spec/services/age_calculator_spec.rb
jsugarman: thanks, done

app/services/age_calculator.rb
jsugarman: I was more interested in refactoring the spec myself and limited myself to only the minor change in the method/class itself.

Feel free to PR one for discussion re: services versus "utility" classes (?!) - i would probably move it out of the services directory in this situation to lib or some such ?!

jsugarman: There is also a gem i found btw, not that i would advocate its use, but interesting nonetheless
https://github.com/kenn/age_calculator

PR #4626, opened 2022-11-25
Remove IE conditional comments

naseberry:
## What
Remove IE conditional comments

Conditional comments are conditional statements interpreted by Microsoft Internet Explorer versions 5 through 9 in HTML source code.
Browser viersions that are no longer support by MS nor us.
In addition, these source paths do not exist, therefore would return a 404 if browsing on the relevant browsers.

## Checklist


jsugarman (approved):
🔥 👍

cpjmcquillan (approved):

PR #4638, opened 2022-11-29
Always enable `Setting.enhanced_bank_upload?`

cpjmcquillan:
Before, the enhanced bank upload setting was configurable so the work could be released incrementally to different environments. Decoupling deploying the changes from releasing the functionality to users made it easier to manage risk and test behaviour extensively in production-like environments.

Now that the feature has been released to all users with bank statement upload permission without issue, we can remove the feature from the configurable settings.

This changes the  `Setting.enhanced_bank_upload?` method so that it only returns `true`. Going forwards, this means that any legal aid application that is uploading bank statements (i.e the provider has bank statement upload permissions, and the client has not given open banking consent or bank statement evidence exists) will now also be using the enhanced bank upload journey.

Subsequent work will remove the setting and any related logic entirely.


jsugarman (commented):
Looks good 🧹 

I have left suggestions that are probably better to do in this PR which amount to moving the  feature scenarios currently in `features/providers/enhanced_bank_upload/` to replace those deleted in 
- `features/providers/bank_statement_flow.feature`
- `features/providers/check_your_answers,feature`

And delete the  `features/providers/enhanced_bank_upload/` directory

cpjmcquillan:
> Looks good 🧹
> 
> I have left suggestions that are probably better to do in this PR which amount to moving the feature scenarios currently in `features/providers/enhanced_bank_upload/` to replace those deleted in
> 
> * `features/providers/bank_statement_flow.feature`
> * `features/providers/check_your_answers,feature`
> 
> And delete the `features/providers/enhanced_bank_upload/` directory

Personally I find it difficult to understand what is going on when there is a large feature test file containing lots of tests. 

I think it makes sense to break these down into smaller, more specific, and thereby more manageable and maintainable chunks.

For now I've gone with `features/providers/bank_statement_upload`, but I'm happy for us to iterate on how we want to slice things.

jsugarman:
> For now I've gone with features/providers/bank_statement_upload, but I'm happy for us to iterate on how we want to slice things.

Sounds good to me. I think there is a discussion about directories reflecting journeys and sub-journeys (like bank statement upload) generally

jsugarman (approved):
👍

PR #4651, opened 2022-12-01
AP-3526: CYA update for full-loop

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3526)

This updates the CYA for looped proceedings
It has been agreed that providers will have to step through each page of the proceeding rather than us attempting to re-set answers affected by changed responses.

Therefore the CYA page will have a single, proceeding-level, change link that takes the user to the CIT page and then through the appropriate pages thereafter.  They will be shown the limitations page (as it relies on DF, LOS and SL responses before returning to the CYA page

## Checklist


kmahern (approved):
lgtm

PR #4652, opened 2022-12-01
Remove `enhanced_bank_upload` column from settings

cpjmcquillan:
In #4638 the `enhanced_bank_upload` column in the `Setting` model was ignored, and `Setting.enhanced_bank_upload?` always returns `true`.

This safely removes the `enhanced_bank_upload` column from the `settings` table.


colinbruce (approved):
👍

PR #4653, opened 2022-12-01
Remove `enhanced_bank_upload?` methods

cpjmcquillan:
Now that the enhanced bank upload functionality has been released to all
providers with bank statement upload permissions and the flag always returns
`true`, the `enhanced_bank_upload?` method no longer means anything.

This replaces its usage with calls to `uploading_bank_statements?`.

While there are some applications that have not been completed that used the
original, non-enhanced bank statement upload journey, these are nominal.

Any issues that arise will be dealt with manually, although it is likely the
small number of affected applications (~ 5) will remain inactive.


colinbruce (approved):
LGTM :+1:

PR #4654, opened 2022-12-01
ap-3393: Add print button

kmahern:
## What

## Checklist


app/views/providers/open_banking_guidances/show.html.erb
```diff
@@ -24,45 +24,61 @@
...
+      <p class="govuk-body"><%= t('.you_need_to') %></p>
+    </div>
+
+    <div class="govuk-!-padding-bottom-2">
+      <%= print_button t('.print_guidance_button') %>
+    </div>
+
+    <!--    TODO add this back in later -->
+        <%#= print_button t('.download_button') %>
+    <!--    <div class="govuk-!-padding-2"></div>-->
```
jsugarman, line 73: not you i know but can we get rid of this? if it just does the same thing (print) then the functionality seems redundant in any event.

jsugarman (approved):
LGTM codewise although you might want Rose to check the css class additions.

Just one suggestion while you are in the area

app/views/providers/open_banking_guidances/show.html.erb
```diff
@@ -12,7 +12,7 @@
     <p class="govuk-body"><%= t('.your_client_can') %></p>
     <p class="govuk-body"><%= t('.its_a_fast') %></p>
 
-    <details class="govuk-details" data-module="govuk-details">
+    <details class="govuk-details no-print" data-module="govuk-details">
```
naseberry, line 15: May I suggest using the [GOV.UK Design System display override](https://design-system.service.gov.uk/styles/layout/#override-how-elements-display) `govuk-!-display-none-print` instead of `no-print`

app/views/shared/forms/_next_action_buttons.html.erb
```diff
@@ -3,7 +3,7 @@
       continue_button_text,
       id: continue_id,
       name: 'continue_button',
-      class: 'govuk-button form-button',
+      class: 'govuk-button form-button no-print',
```
naseberry, line 6: Ditto, regarding the Design Systems display overrides.

app/views/shared/forms/_next_action_buttons.html.erb
```diff
@@ -13,7 +13,7 @@
       draft_button_text,
       name: 'draft_button',
       id: 'draft_button',
-      class: "govuk-button form-button govuk-secondary-button #{left_margin}",
+      class: "govuk-button form-button govuk-secondary-button #{left_margin} no-print",
```
naseberry, line 16: Ditto, regarding the Design Systems display overrides.

app/views/providers/open_banking_guidances/show.html.erb
kmahern: Thanks! Have amended.

app/webpack/stylesheets/helpers.scss
```diff
@@ -23,6 +23,13 @@
   .only-print, .only-print * {
     display: inherit !important;
   }
+  .pagebreak {
+    page-break-before: always;
+  }
```
naseberry, line 28: As the `page-break-before` property is used by print media, I would suggest adding this to the file `app/webpack/stylesheets/print-styles.scss`

naseberry:
Aside from my comments, this is ok.

app/webpack/stylesheets/helpers.scss
kmahern: @naseberry thanks. Following your comment I spotted that there was already existing page_break_before css so I have added a commit to use that instead. I have moved the open-banking-guidance-img css to print-styles.scss, if that is OK?

PR #4661, opened 2022-12-02
Address Rubocop erblint failures in `app/views/shared/forms`

cpjmcquillan:
Before, all templates in `app/views/shared/forms` were excluded from Rubocop linting.

All failures were resolved with the `--autocorrect` flag, and are all trivial.


jsugarman (commented):
👍 LGTM

jsugarman (approved):

PR #4662, opened 2022-12-02
Address Rubocop erblint failures in `app/views/shared/check_answers`

cpjmcquillan:
Before, all templates in `app/views/shared/check_answers` were excluded from Rubocop linting.

All failures were resolved with the `--autocorrect` flag and are trivial.

### DIFF
--- a/app/views/shared/check_answers/_vehicles.html.erb
+++ b/app/views/shared/check_answers/_vehicles.html.erb
@@ -1,4 +1,4 @@
-<% read_only = local_assigns[:read_only] ? read_only : false %>
+<% read_only = false unless local_assigns[:read_only] %>
 <h3 class="govuk-heading-m">
   <%= t(".#{journey_type}.heading") %>
 </h3>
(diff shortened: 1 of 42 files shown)


app/views/shared/check_answers/_vehicles.html.erb
```diff
@@ -9,38 +9,46 @@
...
-      ) if @legal_aid_application.own_vehicle? %>
+  <%= if @legal_aid_application.own_vehicle?
+        check_answer_link(
+          name: :vehicles_regular_uses,
+          url: check_answer_url_for(journey_type, :vehicles_regular_uses, @legal_aid_application),
+          question: t(".#{journey_type}.used_regularly"),
+          answer: yes_no(@legal_aid_application.vehicle.used_regularly),
+          read_only:,
+        )
+      end %>
```
jsugarman, line 53: given that these are all testing the one condition and are already sequential we would it work to have one condition with a all of the content yielded or somesuch.

jsugarman (approved):
LGTM 👍 

Just a comment

PR #4664, opened 2022-12-02
Remove redundant dependabot ignore rule

cpjmcquillan:
We don't use the `sprockets` gem anymore, so we don't need to ignore it anymore.


jsugarman (approved):
👍

PR #4674, opened 2022-12-06
[AP-3315] refactor and ensure provider/team only notified once

jsugarman:
## What
Handle duplicate mail sending and refactor

Multiple emails were being sent (upto 4/5 times) to
the apply team inbox and, in this PR, to the
provider, for undeliverable emails.

This appears to have been the result of a race condition
whereby caused by the EmailMonitorJob being enqueued
both by an initiating `ScheduledMailingsDeliveryJob` and by
itself (`EmailMonitorJob`)

### DIFF
--- a/app/services/govuk_emails/monitor.rb
+++ b/app/services/govuk_emails/monitor.rb
@@ -26,7 +26,7 @@ module GovukEmails
     end
 
     def send_undeliverable_alerts
-      return unless HostEnv.production?
+      # return unless HostEnv.production?
 
       AlertManager.capture_message("Unable to deliver mail to #{scheduled_mail.addressee} - ScheduledMailing record #{scheduled_mail.id}")
 
--- a/app/services/govuk_emails/monitor.rb
+++ b/app/services/govuk_emails/monitor.rb
@@ -22,9 +22,44 @@ module GovukEmails
       AlertManager.capture_message("Unable to deliver mail to #{@scheduled_mail.addressee} - ScheduledMailing record #{@scheduled_mail.id}")
       ScheduledMailing.send_now!(mailer_klass: UndeliverableEmailAlertMailer,
                                  mailer_method: :notify_apply_team,
-                                 legal_aid_application_id: @scheduled_mail.legal_aid_application_id,
+                                 legal_aid_application_id: legal_aid_application.id,
                                  addressee: Rails.configuration.x.support_email_address,
                                  arguments: [@scheduled_mail.id])
+
+      notify_provider_email_failed if @scheduled_mail.mailer_method == "citizen_start_email"
+    end
+
+    def notify_provider_email_failed
+      ScheduledMailing.send_now!(mailer_klass: UndeliverableEmailAlertMailer,
+                                 mailer_method: :notify_provider,
+                                 legal_aid_application_id: legal_aid_application.id,
+                                 addressee: provider_email_or_support,
+                                 arguments: mailer_args)
+    end
+
+    def mailer_args
+      [
+        provider.email,
+        legal_aid_application.application_ref,
+        applicant.full_name,
+        applicant.email,
+      ]
+    end
+
+    def provider_email_or_support
+      HostEnv.staging? ? Rails.configuration.x.support_email_address : provider.email
+    end
+
+    def legal_aid_application
+      @legal_aid_application = @scheduled_mail.legal_aid_application
+    end
+
+    def applicant
+      @applicant ||= legal_aid_application.applicant
+    end
+
+    def provider
+      @provider ||= legal_aid_application.provider
     end
   end
 end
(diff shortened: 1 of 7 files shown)


spec/services/govuk_emails/monitor_spec.rb
```diff
@@ -1,108 +1,181 @@
...
 RSpec.describe GovukEmails::Monitor do
-  let(:scheduled_mailing) { create(:scheduled_mailing, :processing) }
-  let(:response) { double GovukEmails::Email, status: }
-
   before do
     allow(GovukEmails::Email).to receive(:new).with(scheduled_mailing.govuk_message_id).and_return(response)
     Setting.setting.update(alert_via_sentry: true)
   end
 
+  let(:response) { double GovukEmails::Email, status: }
```
cpjmcquillan, line 9: Can we use `instance_double` here?

I'm surprised Rubocop didn't complain, although this file might be ignored!

spec/services/govuk_emails/monitor_spec.rb
```diff
@@ -1,108 +1,181 @@
...
+
   describe ".call" do
-    subject { described_class.call(scheduled_mailing.id) }
+    subject(:call) { described_class.call(scheduled_mailing.id) }
+
+    shared_examples "non-failure handling job" do
```
cpjmcquillan, line 14: Would it make sense to describe these as "successful" jobs, or is "non-failure" more descriptive?

colinbruce (approved):

spec/services/govuk_emails/monitor_spec.rb
```diff
@@ -1,108 +1,181 @@
...
+          expect { call }.to change(ScheduledMailing, :count).by(1)
+
+          undelivered_mailing = ScheduledMailing.find_by(mailer_klass: "UndeliverableEmailAlertMailer")
+
+          expect(undelivered_mailing)
+            .to have_attributes(
+              legal_aid_application_id: scheduled_mailing.legal_aid_application_id,
+              mailer_klass: "UndeliverableEmailAlertMailer",
+              mailer_method: "notify_apply_team",
+              addressee: Rails.configuration.x.support_email_address,
```
cpjmcquillan, line 70: Is this config variable set via an environment variable?

I wonder if it is worth stubbing the config, or using something like [`climate_control`](https://github.com/thoughtbot/climate_control) so the test can be more specific and potentially less flaky?

app/services/govuk_emails/monitor.rb
```diff
@@ -8,33 +13,52 @@ def initialize(scheduled_mail_id)
...
                                  legal_aid_application_id: legal_aid_application.id,
                                  addressee: provider_email_or_support,
                                  arguments: mailer_args)
+
+      cancel_citizen_start_email_and_reminders!
+    end
+
+    def cancel_citizen_start_email_and_reminders!
+      scheduled_mail.cancel!
+      legal_aid_application.scheduled_mailings.where(mailer_klass: "SubmitCitizenFinancialReminderMailer").map(&:cancel!)
```
cpjmcquillan, line 61: Do we need to return the array of outputs from `cancel!`?

If not, should we use `.each(&:cancel!)` instead?

cpjmcquillan (approved):
I added a couple of comments about testing.

The main concern I have is whether we should be resetting `ActiveJob::Base.queue_adapter` or not?

Other than that, it looks good to me 👍

spec/services/govuk_emails/monitor_spec.rb
jsugarman: its for anything which is not the `-failure` ones below, and also when not on production 😬 even it is a failure so 🤷  but "successful" does not really match either. open to suggestions

```
    "waiting", # record has been written, but mail not yet sent by ScheduledMailingsDeliveryJob
    "processing", # ScheduledMailingsDeliveryJob is in the process of sending the mail
    "created", # The mail has been sent to GOVUK Notify, but the status has not yet been queries
    "sending", # GOVUK Notify is still trying to send the mail
    "delivered", # GOVUK Notify has confirmed that the mail has been delivered
    "technical-failure", # GOVUK Notify has reported that the mail cannot  be sent due to a technical failure
    "permanent-failure", # GOVUK Notify has reported a permanent failure (most likely, non-existent email address)
    "temporary-failure", # GOVUK Notify has reported a temporary failure
    "cancelled", # Scheduled mail cancelled because it is no longer needed to be sent.
```

spec/services/govuk_emails/monitor_spec.rb
jsugarman: was in the original spec and does not seem to have problems.

jsugarman:
am closing in favour of https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4675 due to branch name deployment issues

PR #4683, opened 2022-12-08
[AP-3685] Modify applicant#age to apply non means tested logic

jsugarman:
## What
Modify applicant#age to address non means tested application logic

Ensure means and merits reports can be generated for 
Non-means-tested (NMT) applications

For NMT applications the age should be based on the value
stored in the db. This is set at the relevant point which
for NMT is when providers continue past the first check 
your answers page.

## Checklist

### DIFF
--- a/spec/models/applicant_spec.rb
+++ b/spec/models/applicant_spec.rb
@@ -102,16 +102,47 @@ RSpec.describe Applicant do
   end
 
   describe "#age" do
-    subject { legal_aid_application.applicant.age }
+    subject(:age) { legal_aid_application.applicant.age }
 
-    let(:legal_aid_application) { build(:legal_aid_application, :with_transaction_period, :with_applicant) }
+    let(:applicant) { build(:applicant, date_of_birth: 48.years.ago) }
 
-    it "returns the age of the applicant" do
-      expect(subject).to be_a(Integer)
+    let(:legal_aid_application) do
+      build(:legal_aid_application, :with_transaction_period, applicant:)
     end
 
-    it "sets the child? value" do
-      expect(legal_aid_application.applicant.child?).to eq legal_aid_application.applicant.age < 16
+    it "returns the calculated age of the applicant" do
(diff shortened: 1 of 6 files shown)


colinbruce (approved):
Aside from hating date calculations like `18 years.ago + 1.day` (which my brain always as interprets as go back 18 years and then back - not come forward -  another day!) this looks okay to me :+1:

jsugarman:
> Aside from hating date calculations like `18 years.ago + 1.day` (which my brain always as interprets as go back 18 years and then back - not come forward - another day!) this looks okay to me 👍

my brain too. One alternative is to freeze time to 1.day.ago and make them 18 today - but i think that is probably even worse for my/our brains?!

spec/models/applicant_spec.rb
```diff
@@ -102,16 +102,47 @@
...
-    let(:legal_aid_application) { build(:legal_aid_application, :with_transaction_period, :with_applicant) }
+    let(:applicant) { build(:applicant, date_of_birth: 48.years.ago) }
+
+    let(:legal_aid_application) do
+      build(:legal_aid_application, :with_transaction_period, applicant:)
+    end
 
     it "returns the age of the applicant" do
-      expect(subject).to be_a(Integer)
+      expect(age).to be_a(Integer)
```
cpjmcquillan, line 114: I'm not convinced this tests anything useful. 

Consider this example

```ruby
# spec/models/applicant_spec.rb
describe "#age" do
  let(:applicant) { build(:applicant, date_of_birth: 48.years.ago) }

  it "returns the age of the applicant" do
    expect(age).to be_a(Integer)
  end 
end

# app/models/applicant.rb
def age
  17
end
```

The test would still pass, but the age is clearly wrong.

Expecting the method to return an integer might have been sufficient when the method contained no logic besides calling `AgeCalculator` (which is unit tested itself), but I think we should be more specific now the method actually does something.

PR #4684, opened 2022-12-08
AP-3668: Update Client Undertaking question

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3668)

Ensure the providers enter a response for the client undertakings question regardless of whether they answer yes or no
This extrapolates and interpolates the values into split inputs on the page and merges into the single, additional_information, field in the database 

## Checklist

### DIFF
--- a/app/forms/providers/application_merits_task/client_offered_undertakings_form.rb
+++ b/app/forms/providers/application_merits_task/client_offered_undertakings_form.rb
@@ -3,28 +3,43 @@ module Providers
     class ClientOfferedUndertakingsForm < BaseForm
       form_for ::ApplicationMeritsTask::Undertaking
 
-      attr_accessor :offered, :additional_information
+      attr_accessor :offered, :additional_information, :additional_information_true, :additional_information_false
 
-      validate :offered_presence
-      validates :additional_information, presence: true, if: :requires_additional_information?
+      validates :offered, inclusion: { in: %w[true false] }
+      validates :additional_information_true, presence: true, if: :requires_additional_information?
+      validates :additional_information_false, presence: true, unless: :requires_additional_information?
+
+      def initialize(*args)
+        super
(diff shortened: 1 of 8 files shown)


app/forms/providers/application_merits_task/client_offered_undertakings_form.rb
```diff
@@ -3,28 +3,43 @@ module ApplicationMeritsTask
...
+      validates :additional_information_false, presence: true, unless: :requires_additional_information?
+
+      def initialize(*args)
+        super
+        extrapolate_additional_information
+      end
 
       def save
-        additional_information&.clear if offered?
+        attributes[:additional_information] = requires_additional_information? ? additional_information_true : additional_information_false
```
cpjmcquillan, line 18: What's the benefit of us storing the answer to two separate questions in the same place?

It seems like we're adding a fair amount of complexity when we could use specific columns in the database for each question.

Does this simplify things significantly elsewhere in the application?

cpjmcquillan (approved):
I added a question about the general approach, because I'm unsure what we gain from it.

But if we've really thought about it and decided this is worth it, then that's fine with me 👍

app/forms/providers/application_merits_task/client_offered_undertakings_form.rb
colinbruce: It's a weird circumstance, the data provided is always additional information about the client's undertakings.
The reason for two questions is so we can provide a different prompt for the provider, roughly speaking. `What are their undertakings?` and `Why have they not provided any undertakings?`

colinbruce: I picked this pattern as it is already in use on the Domestic Abuse Summary page for police notification

PR #4690, opened 2022-12-09
[AP-3662] Add stylesheet linting

naseberry:
## What

[Add stylesheet linting](https://dsdmoj.atlassian.net/browse/AP-3662)

Add stylesheet linter and configure it using GOV.UK standards to make the code consistent and  follows the GOV.UK standards.

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing: `bundle exec rake`
- Github should not be reporting conflicts; you should have recently run `git rebase main`.
- There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- The PR description should say what you changed and why, with a link to the JIRA story.
- You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
- You should have checked that the commit messages say why the change was made.

PR #4694, opened 2022-12-12
Remove unused styles

cpjmcquillan:
While reviewing #4690, these unused styles surfaced.

This removes the stylesheets and associated imports, since they are not used anywhere in the application.


naseberry (approved):
💯

PR #4695, opened 2022-12-12
AP-3083: Replace secure header gem

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3083)

Replace the secure_header gem with Rail's built-in version

## Checklist


config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+                       "https://www.google-analytics.com",
+                       "https://*.justice.gov.uk"
+  else
+    policy.script_src :self,
+                      "https://www.google-analytics.com",
+                      "https://www.googletagmanager.com"
+    policy.connect_src :self,
+                       "https://www.google-analytics.com",
+                       "https://*.justice.gov.uk",
+                       "http://localhost:*"
```
cpjmcquillan, line 28: Is this supposed to be added in development instead?

The previous config had 

```ruby
connect_src = %w['self' https://www.google-analytics.com https://*.justice.gov.uk] 
connect_src << "http://localhost:*" if Rails.env.development?
```

cpjmcquillan (approved):

jsugarman (commented):
getting a 

```
Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'unsafe-inline' 'nonce-Q5vYJvDZtUYhntemn29lOg=='". Note that 'unsafe-inline' is ignored if either a hash or nonce value is present in the source list.
```

Screenshot 2022-12-12 at 16 57 54


...on the chrome JS console.

I have had this happen to me before so 👀

jsugarman:
> getting a
> 
> ```
> Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' 'unsafe-inline' 'nonce-Q5vYJvDZtUYhntemn29lOg=='". Note that 'unsafe-inline' is ignored if either a hash or nonce value is present in the source list.
> ```
>...
> I have had this happen to me before so 👀

This looks like it fixes the problem BUT by explicitly not using the nonce for the `style-src` directive  - which `main` and `production` do not - nor for `script-src` in development. 

This is required because `unsafe-inline` is ignored if a nonce is specified, and rails CSP defaults to adding a nonce to `style-src` directive and `script-src` directive.
 
```
# set nonce only for directives you _do not want_ :unsafe_inline specified for
# https://github.com/rails/rails/pull/36534/commits/09d55b302266cf002a4b307f8d37a105d2838a18
if Rails.env.development?
  Rails.application.config.content_security_policy_nonce_directives = []
else
  Rails.application.config.content_security_policy_nonce_directives = %w(script-src)
end

```

jsugarman (approved):
Looks good. Have not UAT'd as yet though

colinbruce:
I wasn't going to smash merge this. I will ask Mike to regression test everything including Truelayer :+1:

PR #4699, opened 2022-12-13
Ap 3678 remove passported permissions

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3678)

Removed seeds for passported and non-passported permissions, as these are now both defaults.
Updated code and tests to remove dependencies on these permissions.
Removed redundant use_ccms_arbiter class.
Removed redundant permissions_updater task helper class.

## Checklist


app/views/providers/check_benefits/index.html.erb
```diff
@@ -1,5 +1 @@
-<% if should_use_ccms? %>
-  <%= render 'use_ccms' %>
```
naseberry, line 2: By removing the redundant `use_ccms_arbiter` service class is the `_use_ccms.html.erb` partial also redundant?

app/views/providers/check_benefits/index.html.erb
```diff
@@ -1,5 +1 @@
...
-<% else %>
-  <%= render 'check_benefits_result' %>
-<% end %>
+<%= render 'check_benefits_result' %>
```
cpjmcquillan, line 1: Can we remove the partial and move the content into this template?

As far as I can tell the partial isn't used anywhere else.

spec/requests/admin/roles/permissions_spec.rb
```diff
@@ -21,7 +21,7 @@
 
     it "shows the correct permissions" do
       subject
-      expect(response.body).to include("application.passported._")
+      expect(response.body).not_to include("application.passported._")
```
cpjmcquillan, line 24: Does this test make sense still now that passported permissions have been removed?

cpjmcquillan (approved):
🧹

spec/requests/admin/roles/permissions_spec.rb
kmahern: 👍 have removed this test

colinbruce (approved):
LGTM :+1:

spec/requests/providers/check_benefits_spec.rb
```diff
@@ -218,7 +218,7 @@
       let(:application) { create(:legal_aid_application, :with_positive_benefit_check_result, :checking_applicant_details, applicant:, provider:) }
 
       context "with permissions passported" do
-        let(:provider) { create(:provider, :with_passported_permissions) }
```
jsugarman, line 221: expanding this out i notice there are now two tests, doing the same thing.

jsugarman: the `context`s at lines 220 and 229 can probably be removed and replaced with a single `it` example

jsugarman:
```
        it "displays positive DWP result content" do
          get "/providers/applications/#{application.id}/check_benefits"
          expect(response.body).to include("DWP records show that your client receives a passporting benefit")
        end
```

jsugarman (approved):
💯 LGTM

 possibly a few repeated spec still lurking, like the one i commented but looks great other than that

jsugarman:
@kmahern from the sentry alerts on UAT I am thinking that [this alert](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/models/provider.rb#L36) will no longer need to be captured. I guess on production the firms will still have the passported and non-passported permissions in any event and so would not be raised there but that raise the question of whether we are going to have a follow up PR or bit of work to remove existing firm permissions that is no longer required.

spec/requests/providers/check_benefits_spec.rb
kmahern: 👍 have replaced with single test as suggested.

kmahern:
> @kmahern from the sentry alerts on UAT ([example](https://sentry.io/organizations/ministryofjustice/issues/3811658205/events/9f1e4dd8dbfb49efafdb7134a4661623/?project=5416052)) I am thinking that [this alert](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/models/provider.rb#L36) will no longer need to be captured. I guess on production the firms will still have the passported and non-passported permissions in any event and so would not be raised there but that raises the question of whether we are going to have a follow up PR or bit of work to remove existing firm permissions that is no longer required.

@jsugarman thanks for spotting this, I have removed the alert. FYI @colinbruce has created a tidy function in db/seeds/permissions.rb as part of an earlier pr that will remove unused permissions from providers - which is why it was alerting in UAT as the default is now that a provider will have no permissions.

PR #4663, opened 2022-12-02
Update Ruby to v3.1.3

cpjmcquillan:
Ruby 3.1.3 was [released on 24 November
2022](https://www.ruby-lang.org/en/news/2022/11/24/ruby-3-1-3-released/).

This updates the Ruby version to this latest patch.

PR #4700, opened 2022-12-13
Generate PDF reports from HTML using `grover` gem

cpjmcquillan:
Before, the [`wicked_pdf` gem][wicked] was used to generate PDF files for the means and merits reports from HTML templates. This gem relies on [`wkhtmltopdf`][wk] to generate PDFs from HTML.

As part of the [work to upgrade to Ruby version to 3.1.3][ruby], we discovered that `wkhtmltopdf` was no longer available as part the Alpine image.

As outlined in the repo, the project is no longer actively maintained —so we took a decide to migrate away from that tool and consider alternatives.

The [`grover` gem][grover] makes use of [`puppeteer`][puppeteer] to render HTML in a headless browser and save a PDF file. Other approaches were identified (e.g writing PDFs [using `prawn`][prawn]), but ultimately `grover` was chosen as a drop-in replacement.

This replaces `wicked_pdf` with `grover` for means and merits report creation (which are subsequently sent to CCMS), and updates the respective controllers so that PDF files can be rendered in the browser for testing purposes.

[wicked]: https://github.com/mileszs/wicked_pdf
[wk]: https://github.com/wkhtmltopdf/wkhtmltopdf
[ruby]: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4663#issuecomment-1340628604
[grover]: https://github.com/Studiosity/grover
[puppeteer]: https://github.com/puppeteer/puppeteer
[prawn]: https://github.com/prawnpdf/prawn


.circleci/config.yml
```diff
@@ -14,9 +15,10 @@ executors:
     docker:
       - image: cimg/ruby:3.1.2-browsers
         environment:
-          - RAILS_ENV=test
-          - TZ: "Europe/London"
-          - CI: true
+          RAILS_ENV: test
+          TZ: "Europe/London"
+          CI: true
```
jsugarman, line 20: probably not needed?!

.circleci/config.yml
```diff
@@ -34,6 +36,7 @@ executors:
         environment:
           GITHUB_TEAM_NAME_SLUG: laa-apply-for-legal-aid
           CI: true
```
jsugarman, line 38: probably not needed?!

features/providers/means_report.feature
```diff
@@ -1,385 +0,0 @@
-@javascript
```
jsugarman, line 1: why are we removing this feature test or not replacing it with something? These were intended to exercise the conditional logic that controls differing display for the fundamental journey types. 

I was thinking to extend these tests as part of the non-means-tested work which requires more conditional display logic.

jsugarman (commented):
See comment re removal of integration tests. 

Plus, what is the means_report.pdf binary at the root of the application for?

colinbruce (approved):
LGTM ⭐ 
Will probably need extensive UAT though :+1:

cpjmcquillan:
> Plus, what is the means_report.pdf binary at the root of the application for?

@jsugarman it wasn't supposed to be there, it was leftover from some manual testing.

features/providers/means_report.feature
cpjmcquillan: Previously we relied on the `show_as_html` configuration in `wicked_pdf` to render the view in tests rather than generate a PDF file.

If we wanted to replicate the same behaviour now, we would need to wrap functionality in a conditional like this.

```ruby
def show
  @cfe_result = @legal_aid_application.cfe_result
  @manual_review_determiner = CCMS::ManualReviewDeterminer.new(@legal_aid_application)
  
  unless params.key?(:debug)
    html = render_to_string("show", layout: "pdf")
    pdf = Grover.new(html).to_pdf
    
    send_data pdf, filename: "means_report.pdf", type: "application/pdf", disposition: "inline"
  else
    render "show", layout: "pdf"
  end
end
```

It seemed roundabout, so I removed the tests completely.

#4725 explores how we might take a different approach to testing the rendered content of reports.

jsugarman: If that will just work then it does sound more sensible given it unblocks ruby 3.1.3 and alpine 3.17 and we need to cover the non-means-tested changes in the coming sprint, so yes.

cpjmcquillan:
After many trials and tribulations this is finally ready to go.

It isn't perfect, but it unblocks Ruby upgrades and removes `wkhtmltopdf` as a dependency.

Two issues in particular it would be nice to resolve are:

1. We need to maintain Puppeteer/Chromium versions and compatibility across several places.

- Dev dependency in `package.json` for generating PDFs locally
- CircleCI steps when generating PDFs in unit tests in CI
- Apply base docker image for building and pushing image to environments (UAT/Staging/Production)

2. Generated PDFs look slightly different between environments/operating system

- Locally
Screenshot 2023-01-04 at 16 25 52

- In UAT
Screenshot 2023-01-04 at 16 26 06

In my opinion it isn't worth sinking more time into solving them right now. If they cause issues going forward, we can reevaluate.

PR #4707, opened 2022-12-14
Remove `regexp-examples` gem

cpjmcquillan:
Before, the [`regexp-examples`][regexp] gem was used to generate application
reference numbers.

The reference numbers we need to generate are trivial, so we can generate the
number by sampling the character set instead.

This removes the `regexp-examples` gem, and reimplements the private method
`#random_reference_number` to ensure it returns valid, random reference numbers.

The tests have been refactored to test the uniqueness and randomness of
reference number generation in a deterministic way without stubbing, by
leveraging `Kernel.srand` to seed random number generation.

The public API has also been refactored slightly to `ReferenceNumber.generate`,
which is a deviation from the strict "service-object" pattern, but provides a
more readable interface for users to interact with.

Reference number generation still utilises a loop to cycle through random
numbers to find a unique one. This will become more expensive over time, but
there are 427,518,000 permutations —so we should be ok for a reasonably
forever-shaped length of time.

[regexp]: https://github.com/tom-lord/regexp-examples

Unit tests are somewhat tied to the implementation in that they stub calls to
`LegalAidApplication.find_by(application_ref:)`. This is not ideal, but it
removes a source of flakiness that would currently prove difficult to unpick.

When creating a `LegalAidApplication` record with a given reference number, the
factory subequently makes lots of calls to `Faker` which results in lots of
calls to `.rand`. This means the generated reference number is much further
ahead in the sequence. This also means, the test is prone to being flaky, since
any addition/removal of "fake" attributes in the legal aid application or
associated factories would change the reference number that was generated.

The current approach avoids this by simply not creating any records. That way,
the sequence of random numbers generated is deterministic and corresponds
directly with the behaviour you would see outside the test (for example calling
`Kernel.srand(1234)` and `ReferenceNumber.generate` several times in the
terminal.

This approach (and the journey to get there) is outlined in more detail in this
[comment on the pull request][comment].

[comment]: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4707#discussion_r1050542830


spec/services/reference_number_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      it "returns the number" do
+        expect(reference_number).to eq("L-FL6-NCM")
+      end
+    end
+
+    context "when the first generated number is taken" do
+      it "generates a different number" do
+        _legal_aid_application = create(
+          :legal_aid_application,
+          application_ref: "L-1BL-BVJ",
```
colinbruce, line 19: This seems odd to me, I get that the srand gives us a known seed value, but won't the randomness of sequence cause flickers?
Why is line 19 not `L-FL6-NCM` too?  Shouldn't running seed_rand in the before block have reset it?
When running ` ReferenceNumber.generate` locally I get a very different sequence of codes 😕 
```ruby
Kernel.srand(1234)
 => 1234
3.1.2 :009 > ReferenceNumber.generate
 => "L-FL6-NCM"
3.1.2 :010 > ReferenceNumber.generate
 => "L-TUF-JR9"
3.1.2 :011 > ReferenceNumber.generate
 => "L-VBC-XVP"
```
So I'd expect line 22 to be `L-TUF-JR9`

cpjmcquillan: I agree it is slightly confusing, and I'm not sure exactly how to explain what's happening between the two different tests.

In the `"when the first generated number is not taken"` example, the sequence of random numbers generated is as you describe. If we put a `binding.pry` inside the loop.

```ruby
    15: def unique_reference_number
    16:   loop do
    17:     reference_number = random_reference_number
    18:     binding.pry
 => 19:     return reference_number unless LegalAidApplication.find_by(application_ref: reference_number)
    20:   end
    21: end

[1] pry(#<ReferenceNumber>)> reference_number
=> "L-FL6-NCM"
[2] pry(#<ReferenceNumber>)> random_reference_number
=> "L-TUF-JR9"
[3] pry(#<ReferenceNumber>)> random_reference_number
=> "L-VBC-XVP"
[4] pry(#<ReferenceNumber>)> random_reference_number
=> "L-VTH-X5H" 
```

However in the `"when the first generated number is taken example"`, we get a different set of numbers.

```ruby
    15: def unique_reference_number
    16:   loop do
    17:     reference_number = random_reference_number
    18:     binding.pry
 => 19:     return reference_number unless LegalAidApplication.find_by(application_ref: reference_number)
    20:   end
    21: end

[1] pry(#<ReferenceNumber>)> reference_number
=> "L-1BL-BVJ"
[2] pry(#<ReferenceNumber>)> quit

From: /Users/connor.mcquillan/projects/moj/laa-apply-for-legal-aid/app/services/reference_number.rb:19 ReferenceNumber#unique_reference_number:

    15: def unique_reference_number
    16:   loop do
    17:     reference_number = random_reference_number
    18:     binding.pry
 => 19:     return reference_number unless LegalAidApplication.find_by(application_ref: reference_number)
    20:   end
    21: end

[1] pry(#<ReferenceNumber>)> reference_number
=> "L-EL7-XAB"
[2] pry(#<ReferenceNumber>)> random_reference_number
=> "L-EXJ-RDM"
[3] pry(#<ReferenceNumber>)> random_reference_number
=> "L-0C5-YJ5"
```

You should be able to verify the same locally.

I thought it might be something to do with how tests are seeded, but I'm not so sure.

I experimented with a single test, but interestingly we see the same behaviour.

```ruby
require "rails_helper"

RSpec.describe ReferenceNumber do
  describe ".generate" do
    it "generates unique numbers" do
      Kernel.srand(1234)
      reference_number1 = ReferenceNumber.generate
      expect(reference_number1).to eq("L-FL6-NCM")

      _legal_aid_application = create(
        :legal_aid_application,
        application_ref: "L-TUF-JR9",
      )

      reference_number2 = ReferenceNumber.generate
      expect(reference_number2).to eq("L-VBC-XVP")
    end
  end
end
```

```sh
Randomized with seed 1175

ReferenceNumber
  .generate
    generates unique numbers (FAILED - 1)

Failures:

  1) ReferenceNumber.generate generates unique numbers
     Failure/Error: expect(reference_number2).to eq("L-VBC-XVP")
       Expected "L-1BL-BVJ" to eq "L-VBC-XVP".
     # ./spec/services/reference_number_spec.rb:16:in `block (3 levels) in <top (required)>'

Finished in 0.41442 seconds (files took 5.68 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/services/reference_number_spec.rb:5 # ReferenceNumber.generate generates unique numbers
```

I'm fairly certain the underlying reason for the difference in behaviour is because of how many times `.rand` is called indirectly and behind the scenes when we create a new `LegalAidApplication` with the factory.

If we refactor the test slightly, we see the same sequence of numbers you expected to begin with.

```ruby
require "rails_helper"

RSpec.describe ReferenceNumber do
  describe ".generate" do
    it "generates unique numbers" do
      Kernel.srand(1234)
      reference_number1 = ReferenceNumber.generate
      expect(reference_number1).to eq("L-FL6-NCM")

      allow(LegalAidApplication).to receive(:find_by).and_return(nil)

      allow(LegalAidApplication)
        .to receive(:find_by)
        .with(application_ref: "L-TUF-JR9")
        .and_return(true)

      reference_number2 = ReferenceNumber.generate
      expect(reference_number2).to eq("L-VBC-XVP")
    end
  end
end
```

```sh
Randomized with seed 33919

ReferenceNumber
  .generate
    generates unique numbers

Finished in 0.29996 seconds (files took 5.64 seconds to load)
1 example, 0 failures
```

I'm not sure which test is preferable.

The final (stubbed) test is probably easier to understand and matches your expectations when stepping through generated random numbers. Having said that, I would probably look at it and think "why aren't we just creating the application".

Either way, I'll be sure to include much of this context in the commit message.

I hope that was helpful, and sorry for the lengthy response.

What do you think?

colinbruce: As long as it's stable and doesn't start flickering I guess it's okay.  Hopefully, future-colin will find this conversation useful when he's trying to debug it! 😆

colinbruce (approved):

spec/services/reference_number_spec.rb
colinbruce: The number generator itself is 🏆, it was only the test that confused me!

PR #4708, opened 2022-12-14
AP-3705: Fix missing substantive date

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3705)

Update the handling of Delegated functions being stored using the DelegatedFunctionsDateService

After the mini-loop was enabled the sequencing was off, it would call the service before saving the form, so the delegated function dates were always missing

This updates the controller to save the form, then call the service and render or redirect

While investigating it also highlighted some issues with the proceeding scopes so these have been tightened and tests updated to ensure that we use the used_delegated_function boolean rather than checking for a date presence

## Checklist


kmahern (approved):
lgtm. Do we need any sort of data fix for applications with missing df dates and/or incorrect scope limitations in production?

colinbruce:
Good shout @kmahern 
I'll pass it for UAT and check on Prod for anything affected

colinbruce:
I'm still checking prod - I've merged to prevent any extension of the issue and will raise a separate PR if a fix rake task is needed :+1:

PR #4724, opened 2022-12-19
Clean up factory associations

cpjmcquillan:
Before, some factory associations were _always_ created, even 
when the parent record was not.

This behaviour was unexpected (to me at least), and I suspect 
unintentional.

Now, factory associations are not explicitly created, and instead 
defer to the parent object strategy (e.g if a `cfe_result` is built 
rather than created, its associated `submission` will also be built).


jsugarman (approved):
Makes sense to me and the test all pass 👍

colinbruce (approved):
Ditto Joel's comment 😄

PR #4759, opened 2022-12-30
Fix: Flaky features

colinbruce:
## What

This PR address some recent feature flickers:
* The features/providers/check_single_employment.feature (and probably others) would intermittently fail becuase the seed would generate a value below 500 and cause the validation to fail and leave it on the previous page

  This updates the `other_assets_declaration` factory to ensure a valuable_items_value cannot be below the validation threshold

* We saw an instance where a test failed because faker randomly picked `rose` as a username but there is an existing provider in the seeded table that caused it to fail

  This adds a zero-padded three-digit suffix to the username field in the `provider` factory to ensure this does not recur, e.g. `rose_056`

## Checklist


kmahern (approved):
lgtm

PR #4774, opened 2023-01-04
[AP-3646] Use govuk pagination component

naseberry:
## What

[Use pagination component](https://dsdmoj.atlassian.net/browse/AP-3646)

Update existing pagination to use the GOV.UK Design System pagination component

## Checklist

[AP-3646]: https://dsdmoj.atlassian.net/browse/AP-3646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/spec/requests/admin/feedback_spec.rb
+++ b/spec/requests/admin/feedback_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe Admin::FeedbackController do
 
       it "does not show navigation links" do
         subject
-        expect(parsed_response_body.css(".pagination-container nav")).to be_empty
+        expect(parsed_response_body.css(".govuk-pagination")).to be_empty
       end
 
       context "and more applications than page size" do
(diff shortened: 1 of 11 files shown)


spec/requests/admin/feedback_spec.rb
```diff
@@ -39,7 +39,7 @@
 
       it "does not show navigation links" do
         subject
-        expect(parsed_response_body.css(".pagination-container nav")).to be_empty
+        expect(parsed_response_body.css(".govuk-pagination")).to be_empty
```
cpjmcquillan, line 42: While you're here, how do you feel about updating this to use Capybara matchers?

```ruby
expect(page).not_to have_css(".govuk-pagination")
```

spec/requests/admin/feedback_spec.rb
```diff
@@ -53,7 +53,7 @@
 
         it "shows pagination" do
           subject
-          expect(parsed_response_body.css(".pagination-container nav").text).to match(/Previous\s+1\s+2\s+Next/)
+          expect(parsed_response_body.css(".govuk-pagination").text).to match(/12Next page/)
```
cpjmcquillan, line 56: Similarly, this could be updated to something like this:

```ruby
expect(page).to have_css(".govuk-pagination", text: "Next page")
```

cpjmcquillan (approved):
:shipit: 

I added a comment about updating some of the request specs to use Capybara matchers, but you can take it or leave it.

PR #4776, opened 2023-01-04
Fix: Make all features accessible

colinbruce:
## What

This is a spare-time PR that changes how the AXE (accessibility) checks are run in features.  It checks each page whenever you click a link (think CYA links) or a button (save and continue, etc)

### **PRO**
* It provides greater coverage
* It finds issues that the previous explicit tests missed
* It is only enabled locally if you set an ENVVAR

### **CONS**
* It increases the time taken to run the features
* Pages will be tested multiple times

## Time increase
To ameliorate the time taken you if you want to run the tests locally, 
```ruby
bundle exec cucumber
```
will still _just_ run the tests, running your cucumber tests locally with an ENV var:
```ruby
INC_AXE=true bundle exec cucumber
```
will run the additional AXE checks

To offset the increase in time when running in Circle, the parallelism has been increased to, currently, 6 pods and should be monitored for potentially increasing that amount

## Discussion point

Should we run more, slower AXE tests (like this) or investigate other methods of running tests on all pages in all states?
If we decide not to implement this, we can still add the accessibility fixes and remove the cucumber updates

## Checklist


app/views/providers/bank_statements/_uploaded_files.html.erb
```diff
@@ -17,7 +17,7 @@
       <tr class="govuk-table__row">
         <th class="govuk-table__header" scope="col"><%= t('.filename') %></th>
         <th class="govuk-table__header" scope="col"><%= t('.status') %></th>
-        <td class="govuk-table__header" scope="col"></td>
+        <th class="govuk-table__header" scope="col"><span class="govuk-table__caption govuk-visually-hidden">This column contains delete buttons per application</span></th>
```
colinbruce, line 20: wrong text

cpjmcquillan (approved):
I like it a lot, nice work 🏅 

I like how it automates the axe checks, and reduces the overhead on developers.

You mention the test time issue, and this is a concern of mine too. Personally I feel like the time it takes to run our test suite is quite long, so I'm wary of us doing anything that increases it significantly.

I think @jsugarman suggested running the axe checks once a day, and I think that approach might work well.

That way we get the benefits of automated accessibility testing that "just works", but we can reduce test time (by keeping the parallelism bump).

What do other people think?

.circleci/config.yml
```diff
@@ -219,7 +219,7 @@ jobs:
           - coverage_results
 
   integration_tests:
-    parallelism: 3
+    parallelism: 10
```
colinbruce, line 222: 😮

agoldstone93 (approved):

PR #4786, opened 2023-01-05
AP-3331 save vehicles remaining payments answer

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3331)

Save answer when returning from the CYA page
Add feature tests to check that answer is saved on return

## Checklist

### DIFF
--- a/app/forms/vehicle_form/remaining_payment_form.rb
+++ b/app/forms/vehicle_form/remaining_payment_form.rb
@@ -17,7 +17,7 @@ module VehicleForm
     end
 
     def save
-      attributes[:payment_remaining] = 0 if valid? && !payments_remain?
+      attributes[:payment_remaining] = nil if valid? && !payments_remain?
       super
     end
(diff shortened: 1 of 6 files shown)


app/forms/vehicle_form/remaining_payment_form.rb
```diff
@@ -17,7 +17,7 @@ def payments_remain?
     end
 
     def save
-      attributes[:payment_remaining] = 0 if valid? && !payments_remain?
+      attributes[:payment_remaining] = nil if valid? && !payments_remain?
```
RoseSAK, line 20: This is so we can just check if this value is nil when returning to the page so we know which radio should be selected, instead of having to check whether it's nil OR zero, I don't see any reason for saving a 0 value if they select 'No', but I may be wrong

kmahern (approved):
lgtm

PR #4813, opened 2023-01-10
Ap 3710 means report non means tested

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3710)

Amended means report to remove everything beneath the client details section for non means tested applications
Amended means_report.feature to reflect/test for this change

## Checklist

### DIFF
--- a/app/views/providers/means_reports/show.html.erb
+++ b/app/views/providers/means_reports/show.html.erb
@@ -13,5 +13,7 @@
         applicant: @legal_aid_application.applicant,
         read_only: true) %>
 
-  <%= render "with_cfe_result_details" %>
+  <% if !@legal_aid_application.non_means_tested? %>
+    <%= render "with_cfe_result_details" %>
+  <% end %>
 <% end %>
--- a/app/views/providers/means_reports/show.html.erb
+++ b/app/views/providers/means_reports/show.html.erb
@@ -13,7 +13,6 @@
         applicant: @legal_aid_application.applicant,
         read_only: true) %>
 
-  <% if !@legal_aid_application.non_means_tested? %>
-    <%= render "with_cfe_result_details" %>
-  <% end %>
+  <%= render "with_cfe_result_details" unless @legal_aid_application.non_means_tested? %>
+
 <% end %>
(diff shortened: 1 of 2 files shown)


config/locales/en/shared.yml
```diff
@@ -263,6 +263,7 @@ en:
         address: Correspondence address
         age: "%{years} years old"
         age_question: Age at computation date
+        means_tested: Was the client means-tested?
```
jsugarman, line 266: I do not know if a content person or others have changed the requested text but the ticket specifies the text to be "Does the client need a means test?"

jsugarman (approved):
LGTM 👍 

Just a question on the content for the question itself versus the ticket

app/views/providers/means_reports/show.html.erb
```diff
@@ -9,9 +9,11 @@
 
   <%= render(
         'shared/check_answers/client_details',
-        attributes: %i[first_name last_name date_of_birth age national_insurance_number],
+        attributes: %i[first_name last_name date_of_birth age means_test national_insurance_number],
         applicant: @legal_aid_application.applicant,
         read_only: true) %>
 
-  <%= render "with_cfe_result_details" %>
+  <% if !@legal_aid_application.non_means_tested? %>
```
cpjmcquillan, line 16: The double negative here is a little confusing to me.

I wonder whether it would be more readable to say:

```erb
<% unless @legal_aid_application.non_means_tested? %>
```

Or perhaps if we defined another predicate method, something like this might work:

```erb
<% if @legal_aid_application.means_tested? %>
```

What do you think?

cpjmcquillan (approved):

app/views/providers/means_reports/show.html.erb
```diff
@@ -9,9 +9,11 @@
...
+    <%= render "with_cfe_result_details" %>
```
colinbruce, line 17: To slightly extend Connor's suggestion, how about, as it's a oneliner
```suggestion
    <%= render "with_cfe_result_details" unless @legal_aid_application.non_means_tested? %>
```

config/locales/en/shared.yml
colinbruce: It was reviewed on 3711 by content - at the time this is shown it is in the past 👍

app/views/providers/means_reports/show.html.erb
kmahern: Thanks @colinbruce @cpjmcquillan I agree and have implemented this suggestion

PR #4821, opened 2023-01-11
Means test review declaration content changes

cpjmcquillan:
Once the means test review work has been completed, providers 
will be able to apply for legal aid on the behalf of clients under the
age of 18.

This change means the wording of the declaration needs to be
different depending on whether the applicant is under the age of
18 or not.

Currently, the concept of a `non_means_tested?` legal aid 
application is synonymous with an applicant being under the age
of 18 for the purposes of the means test (i.e 
`age_for_means_test_purposes < 18`).

I'm not convinced this coupling is particularly future-proof though
since there will likely be future scenarios where an application is 
not means tested, but the applicant is 18 or over.

I think it would be worthwhile the team thinking about whether this
is a problem worth solving now, or whether we should address it
when the issue arises.

<details>
<summary>Screenshots</summary>

### Before
Before

### Means tested
Means tested

### Non-means tested
Non-means tested

</details>

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3712)

### DIFF
--- a/features/providers/check_multiple_employment.feature
+++ b/features/providers/check_multiple_employment.feature
@@ -146,7 +146,7 @@ Feature: Check multiple employment
     When I click 'Save and continue'
     Then I should be on a page showing "Confirm the following"
 
-    When I check "I confirm the above is correct and that I'll obtain a signed declaration from my client."
+    When I check "I confirm the above is correct and that I'll get a signed declaration from my client"
     And I click 'Save and continue'
     Then I should be on a page showing "Review and print your application"
 
--- a/features/providers/check_pending_employment.feature
+++ b/features/providers/check_pending_employment.feature
@@ -138,7 +138,7 @@ Feature: Check pending employment
     When I click 'Save and continue'
     Then I should be on a page showing "Confirm the following"
 
-    When I check "I confirm the above is correct and that I'll obtain a signed declaration from my client."
+    When I check "I confirm the above is correct and that I'll get a signed declaration from my client"
     And I click 'Save and continue'
     Then I should be on a page showing "Review and print your application"
 
--- a/spec/forms/legal_aid_applications/confirm_client_declaration_form_spec.rb
+++ b/spec/forms/legal_aid_applications/confirm_client_declaration_form_spec.rb
@@ -3,23 +3,51 @@ require "rails_helper"
 RSpec.describe LegalAidApplications::ConfirmClientDeclarationForm do
   subject(:form) { described_class.new(params) }
 
-  let(:client_declaration_confirmed) { true }
-  let(:params) do
-    {
-      client_declaration_confirmed:,
-    }
-  end
+  describe "#validates" do
+    context "when the client declaration is confirmed" do
+      let(:params) { { client_declaration_confirmed: true } }
 
-  describe "valid?" do
-    it "returns true when validations pass" do
-      expect(form).to be_valid
+      it { is_expected.to be_valid }
     end
 
-    context "when invalid" do
-      let(:client_declaration_confirmed) { nil }
+    context "when the client declaration is not confirmed" do
+      let(:params) { { client_declaration_confirmed: "" } }
 
-      it "returns false when validations fail" do
-        expect(form).not_to be_valid
+      it "is invalid" do
+        expect(form).to be_invalid
+        expect(form.errors).to be_added(:client_declaration_confirmed, :accepted)
+      end
+    end
+  end
+
+  describe "#save" do
+    let(:legal_aid_application) do
+      create(:legal_aid_application, client_declaration_confirmed_at: nil)
+    end
+
+    before do
+      freeze_time
+      form.save!
+    end
+
+    context "when the client declaration is confirmed" do
+      let(:params) do
+        { client_declaration_confirmed: true, model: legal_aid_application }
+      end
+
+      it "updates the application" do
+        expect(legal_aid_application.client_declaration_confirmed_at)
+          .to eq Time.current
+      end
+    end
+
+    context "when the client declaration is not confirmed" do
+      let(:params) do
+        { client_declaration_confirmed: "", model: legal_aid_application }
+      end
+
+      it "does not update the application" do
+        expect(legal_aid_application.client_declaration_confirmed_at).to be_nil
       end
     end
   end
--- a/app/controllers/providers/confirm_client_declarations_controller.rb
+++ b/app/controllers/providers/confirm_client_declarations_controller.rb
@@ -5,14 +5,10 @@ module Providers
     end
 
     def update
-      return continue_or_draft if draft_selected?
-
       @form = LegalAidApplications::ConfirmClientDeclarationForm.new(form_params)
-      if @form.valid?
-        legal_aid_application.update!(client_declaration_confirmed_at: Time.zone.now)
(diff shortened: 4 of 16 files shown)


features/providers/check_multiple_employment.feature
```diff
@@ -146,7 +146,7 @@ Feature: Check multiple employment
     When I click 'Save and continue'
     Then I should be on a page showing "Confirm the following"
 
-    When I check "I confirm the above is correct and that I'll obtain a signed declaration from my client."
+    When I select the first checkbox
```
jsugarman, line 149: be good to have a more descriptive step here, like the old one. Have tried and this works

```suggestion
    When I check "I confirm the above is correct and that I'll get a signed declaration from my client."
```

features/providers/check_pending_employment.feature
```diff
@@ -138,7 +138,7 @@ Feature: Check pending employment
     When I click 'Save and continue'
     Then I should be on a page showing "Confirm the following"
 
-    When I check "I confirm the above is correct and that I'll obtain a signed declaration from my client."
+    When I select the first checkbox
```
jsugarman, line 141: [ditto](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4821/files#r1068345409)

spec/forms/legal_aid_applications/confirm_client_declaration_form_spec.rb
```diff
@@ -1,25 +1,57 @@
...
-  subject(:form) { described_class.new(params) }
+  let(:params) { { client_declaration_confirmed: } }
 
-  let(:client_declaration_confirmed) { true }
-  let(:params) do
-    {
-      client_declaration_confirmed:,
-    }
+  describe "#validates" do
+    subject(:form) { described_class.new(params) }
```
jsugarman, line 7: **nit** but is it simple enough to have a single let for the form at the top level and change params at the describe/context level  with a subject if suitable.

```ruby
let(:form) { described_class.new(params) }

describe "#validates" do
  context "when the client declaration is confirmed" do
    let(:params) { client_declaration_confirmed: true }
  
    it { expect(form).to be_valid }
  end

  context "when the client declaration is not confirmed" do
    let(:params) { client_declaration_confirmed: "" }
  
    it "is invalid" do
      expect(form).to be_invalid
      expect(form.errors).to be_added(:client_declaration_confirmed, :accepted)
    end
  end
end

describe "#save" do
  before { login }

  let(:legal_aid_application) do
    create(:legal_aid_application, client_declaration_confirmed_at: nil)
  end

  context "when the client declaration is confirmed" do
    let(:params) { client_declaration_confirmed: true, model: legal_aid_application }

    it "updates the application" do
      freeze_time do
        expect { form.save }.to change { legal_aid_application.client_declaration_confirmed_at }
          .from(nil)
          .to(Time.current)
      end
    end
  end
end
```

I also changed the `freeze_time`, as i thought you need to ensure you `travel_back` (via and around or after block)  if not passing a block

features/providers/check_multiple_employment.feature
cpjmcquillan: The reason I changed these was because it didn't feel like the specific content was important.

But I can change these back if we prefer 👍

spec/forms/legal_aid_applications/confirm_client_declaration_form_spec.rb
cpjmcquillan: > i thought you need to ensure you travel_back (via and around or after block) if not passing a block?!

_**TLDR: time stubs are removed automatically after using helpers like `freeze_time` and `travel_to`**_

---

I've been living my RSpec life under the assumption that ["the stubs are automatically removed at the end of the test"](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to), but then this convinced a tired me my life has been a lie and this wasn't the case in RSpec after all. 

[`TimeHelper` defines an `after_teardown` method to remove the stubs](https://github.com/rails/rails/blob/b824c20b5afa1286fece234dcac35a0a97a1ba95/activesupport/lib/active_support/testing/time_helpers.rb#L70-L73), but I couldn't see where this was called in RSpec.

So, I did an _investigation_...

In this example branch, I defined this spec in `spec/time_helper_spec.rb`:

```ruby
require "rails_helper"

RSpec.describe "ActiveSupport::Testing::TimeHelpers", type: :model, order: :defined do
  it "freezes time and sets a global variable" do
    puts Time.current.strftime("\n%Y-%m-%d %H:%M:%S\n")

    freeze_time
    $time = Time.current

    sleep(1)

    expect(Time.current).to eq $time

    puts Time.current.strftime("\n%Y-%m-%d %H:%M:%S\n")
  end

  it "removes time stubs" do
    puts Time.current.strftime("\n%Y-%m-%d %H:%M:%S\n")

    sleep(1)

    expect($time).not_to be_nil
    expect(Time.current).not_to eq $time

    puts Time.current.strftime("\n%Y-%m-%d %H:%M:%S\n")
  end
end
```

which outputs

```
ActiveSupport::Testing::TimeHelpers

2023-01-17 11:39:34

2023-01-17 11:39:34
  freezes time and sets a global variable

2023-01-17 11:39:35

2023-01-17 11:39:36
  removes time stubs

Finished in 2.21 seconds (files took 6.38 seconds to load)
2 examples, 0 failures
```

This confirmed that time is unfrozen/we do travel back after using Rails time helpers in RSpec specs.

[`RSpec::Rails::RailsExampleGroup` is included for each Rails example group](https://github.com/search?q=repo%3Arspec%2Frspec-rails%20include%20RSpec%3A%3ARails%3A%3ARailsExampleGroup&type=code), and it includes [`RSpec::Rails::MinitestLifecycleAdapter`](https://github.com/rspec/rspec-rails/blob/40261bb72875c00a6e4a0ca2ac697b660d4e8d9c/lib/rspec/rails/adapters.rb#L66), which ensures [`after_teardown` is called around each example](https://github.com/rspec/rspec-rails/blob/40261bb72875c00a6e4a0ca2ac697b660d4e8d9c/lib/rspec/rails/adapters.rb#L73-L77).

As it turns out, this is also included for non-Rails example groups since `RSpec::Rails::FixtureSupport` (which includes `RSpec::Rails::ExampleGroup`) is [included by default](https://github.com/rspec/rspec-rails/blob/40261bb72875c00a6e4a0ca2ac697b660d4e8d9c/lib/rspec/rails/configuration.rb#L82) to address an issue.

You can test this out by removing `type: :model` from the spec definition and seeing the tests pass.

Is that helpful?

features/providers/check_multiple_employment.feature
jsugarman: @cpjmcquillan for me the use of "I select the first checkbox" is not a pattern/style I like for reasons mentioned and would change elsewhere too if found. 

can we change this and get this in?

spec/forms/legal_aid_applications/confirm_client_declaration_form_spec.rb
jsugarman: Thanks for the check. interesting. This is not a blocker so resolving

jsugarman (approved):
👍 

needs a rebase now due to civil_application_journey.feature having been split up

PR #4823, opened 2023-01-12
Move client declaration information to component

cpjmcquillan:
Before, conditional logic that determined which set of content was rendered for
a given application (i.e means-tested vs non-means-tested) resided in the view.

This made it difficult to test in isolation. While it is possible to write view
specs, their use is not widespread (and we have none in our test suite).

Currently we lean on integration tests (e.g request specs or cucumber features)
to assert that given content is rendered. This is useful, but they are
expensive —which prohibits unit style testing.

This introduces `view_component` as an explicit dependency (we use it implicitly
already through the `govuk-components` gem), and refactors the provider client
declaration confirmation page so that it renders a component.

This component encapsulates the logic for rendering means-tested and
non-means-tested specific content.

For now, `i18n-tasks.yml` excludes components files. I spent a couple of hours
trying to piece together a pattern mapper so that `i18n-tasks` didn't categorise
the keys as missing, but it was turning into a time-sink.

Given we have `config.raise_on_missing_translations = true` in development and
test, and the components are thoroughly unit-tested, I don't think this is a
major issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,8 +94,9 @@ gem "grover"
 # DFE formbuilder
 gem "govuk_design_system_formbuilder"
 
-# DFE ViewComponent library
+# ViewComponent
 gem "govuk-components"
+gem "view_component"
 
 # Catching unsafe migrations in development
 gem "strong_migrations"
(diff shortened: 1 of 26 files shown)


cpjmcquillan:
Closing because I haven't found time to compelling advocate that team members discuss the proposed change in the forum of this PR

PR #418, opened 2019-04-02
AP-488 Fix issue when redirecting from https to http

Kalagan:
## What

[AP-488](https://dsdmoj.atlassian.net/browse/AP-488)

Make the portal and our fake portal redirect to https instead of http if we're in production.

I've deployed this to an alternate staging env:
https://applyforlegalaid-staging-wip.apps.cloud-platform-live-0.k8s.integration.dsd.io
and to UAT :
https://julien-wip-applyforlegalaid-uat.apps.cloud-platform-live-0.k8s.integration.dsd.io/providers

and it works on safari and chrome.

as well as in localhost

## Checklist

Before you ask people to review this PR:

- [X] Tests and rubocop should be passing: `bundle exec rake`
- [X] Github should not be reporting conflicts; you should have recently run `git rebase master`.
- [X] There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- [X] The PR description should say what you changed and why, with a link to the JIRA story.
- [X] You should have looked at the diff against master and ensured that nothing unexpected is included in your changes.
- [X] You should have checked that the commit messages say why the change was made.

PR #4834, opened 2023-01-16
Remove `success_message` from page template helper

cpjmcquillan:
Rubocop raised a self assignment issue in the `_uploaded_files` partials
regarding the `success_message` variable.

On investigation, it looks like the success message pattern is used in only one
place (the feedback page). Its use was unconventional, and didn't seem like
something the top level `page_template` helper should be concerned with either.

This replaces the current usage with a more conventional flash notice.

The behaviour is the same (i.e providers are redirected to the feedback page and
shown a message after signing out), but the style of the message has changed
slightly.

Following discussion with design, it was agreed that the [notification
component] would be used for all flash notice messages.

This also brings mock saml and non-mock saml behvaiour inline, so that both sets
of use cases mean the user is redirected to the feedback page.

[notification component]: https://design-system.service.gov.uk/components/notification-banner/

<details>
<summary>Screenshots</summary>

### Before

image

image

### After

image

image

</details>

### DIFF
--- a/app/controllers/saml_sessions_controller.rb
+++ b/app/controllers/saml_sessions_controller.rb
@@ -17,11 +17,8 @@ class SamlSessionsController < Devise::SamlSessionsController
     session["signed_out"] = true
     session["feedback_return_path"] = destroy_provider_session_path
     sign_out current_provider
-    if IdPSettingsAdapter.mock_saml?
-      redirect_to providers_root_url
-    else
-      redirect_to new_feedback_path
-    end
+
+    redirect_to new_feedback_path, notice: t(".notice")
   end
 
   def after_sign_in_path_for(_provider)
--- a/app/views/providers/gateway_evidences/_uploaded_files.html.erb
+++ b/app/views/providers/gateway_evidences/_uploaded_files.html.erb
@@ -1,7 +1,3 @@
-<%
-success_message = success_message
-%>
-
 <table class="govuk-table <%= 'hidden' if attachments.empty? %>">
   <caption class="govuk-table__caption"><%= t('.title') %></caption>
   <thead class="govuk-table__head">
@@ -13,8 +9,6 @@ success_message = success_message
   </tr>
   </thead>
   <tbody class="govuk-table__body">
-
-  <span alert-live="assertive" class="govuk-visually-hidden"><%= success_message %></span>
 
   <% attachments.order(:created_at).each do |attachment| %>
     <tr class="govuk-table__row">
--- a/app/views/providers/gateway_evidences/_uploaded_files.html.erb
+++ b/app/views/providers/gateway_evidences/_uploaded_files.html.erb
@@ -1,10 +1,10 @@
-<table class="govuk-table <%= 'hidden' if attachments.empty? %>">
-  <caption class="govuk-table__caption"><%= t('.title') %></caption>
+<table class="govuk-table <%= "hidden" if attachments.empty? %>">
+  <caption class="govuk-table__caption"><%= t(".title") %></caption>
   <thead class="govuk-table__head">
   <tr class="govuk-table__row">
-    <th class="govuk-table__header" scope="col"><%= t('.filename') %></th>
-    <th class="govuk-table__header" scope="col"><%= t('.size') %></th>
-    <th class="govuk-table__header" scope="col"><%= t('.status') %></th>
+    <th class="govuk-table__header" scope="col"><%= t(".filename") %></th>
+    <th class="govuk-table__header" scope="col"><%= t(".size") %></th>
+    <th class="govuk-table__header" scope="col"><%= t(".status") %></th>
     <th class="govuk-table__header" scope="col"></th>
   </tr>
   </thead>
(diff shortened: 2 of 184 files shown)


app/controllers/saml_sessions_controller.rb
```diff
@@ -17,11 +17,8 @@ def destroy
     session["signed_out"] = true
     session["feedback_return_path"] = destroy_provider_session_path
     sign_out current_provider
-    if IdPSettingsAdapter.mock_saml?
-      redirect_to providers_root_url
-    else
-      redirect_to new_feedback_path
-    end
```
cpjmcquillan, line 24: I can't think of a reason why we shouldn't be redirecting all providers to the feedback page in all environments, so I've removed this extra complexity.

The commit history didn't convince me that this was something worth keeping, but someone might have some context that was missing.

cpjmcquillan:
This also addresses a bug where all users see a `You are signed out` message on the feedback page, even if they have never been signed in 🤷

app/controllers/saml_sessions_controller.rb
jsugarman: Have you tried it out and seen odd behaviour? 

This is something i have seen in other LAA/MoJ apps too. One reason might be convenience for devs/testers, another might be to prevent/avoid submission of feedback from non-prod environments which gets mixed in with real feedback 🤷 

I am happy to remove however!

spec/requests/saml_sessions_spec.rb
```diff
@@ -29,28 +29,11 @@
...
-
-      it "redirects to providers root" do
-        subject
-        expect(response).to redirect_to(providers_root_url)
-      end
+    it "redirects to the feedback page" do
+      subject
+      expect(response).to redirect_to(new_feedback_path)
+      follow_redirect!
+      expect(page).to have_content(I18n.t("saml_sessions.destroy.success"))
```
jsugarman, line 36: i'd decouple the test from the translation

cpjmcquillan: I don't really care either way. 

I think there are benefits to using translations and benefits to not using them.

In this instance, I referred to the translation since that is what the existing test did already —but I can change it if we prefer.

app/views/providers/gateway_evidences/_uploaded_files.html.erb
```diff
@@ -14,8 +10,6 @@ success_message = success_message
   </thead>
   <tbody class="govuk-table__body">
 
-  <span alert-live="assertive" class="govuk-visually-hidden"><%= success_message %></span>
```
naseberry, line 17: ARIA live regions are generally used for accessibility purposes, are you sure this no longer required?
If required, it is best to move it to elsewhere in DOM as this isn't a valid location.

app/views/providers/uploaded_evidence_collections/_uploaded_files.html.erb
```diff
@@ -23,8 +20,6 @@
     </thead>
     <tbody class="govuk-table__body">
 
-    <span alert-live="assertive" class="govuk-visually-hidden"><%= success_message %></span>
```
naseberry, line 26: ARIA live regions are generally used for accessibility purposes, are you sure this no longer required?
If required, it is best to move it to elsewhere in DOM as this isn't a valid location.

app/views/providers/gateway_evidences/_uploaded_files.html.erb
cpjmcquillan: There is no scenario where `success_message` is anything other than `nil`, so I'm not sure what use an empty and invisible (except to screen readers) span is?

@RoseSAK do you have any idea why these were here in the first place?

app/controllers/saml_sessions_controller.rb
cpjmcquillan: I'm not sure what you mean by "odd behaviour"?

As I said, the commit history doesn't indicate that this distinction is important —so I removed it.

In the interests of brevity and not boring people to death, I omitted the details of the commit history (everyone can look if they need to).

I didn't expect this to be a particularly controversial change, but here is what I gleaned from the commit history to inform my decision.

---

In #185 we redirected to the `providers_root_url`, then in #418 we took the decision to log providers out of the portal when they logged out of our application (in production environments).

In #1193 we started redirecting providers to the feedback page instead when they logged out.

Before, the conditional was logical because it made no sense to log users out of the portal if they weren't logged in that way (i.e `mock_saml` is true).

As soon as we're no longer doing that (i.e we're just redirecting to the feedback page), the conditional adds no value.

app/controllers/saml_sessions_controller.rb
```diff
@@ -17,11 +17,8 @@ def destroy
...
+
+    redirect_to new_feedback_path, notice: t(".success")
```
naseberry, line 21: Design: the flash notice message uses the govuk warning component, may be better to use another component, for example the notification component. Check component usage or speak with design profession.

app/views/providers/gateway_evidences/_uploaded_files.html.erb
cpjmcquillan: I don't think these are needed at all.

app/controllers/saml_sessions_controller.rb
cpjmcquillan: Discussed this with design, and they have confirmed we can use the notification component in place of the warning text for flash notice messages.

PR #4838, opened 2023-01-16
[AP-3744] Move secret_key_base to values files (remove rails_master_key)

jsugarman:
## What
Use secret_key_base in values files directly

**WARNING:** will log users out on production.

[Jira task](https://dsdmoj.atlassian.net/browse/AP-3744)

This both moves and rotates the `secret_key_base` providing the following advantages:

- different secret key bases between environments
- is consistent with Apply's sister applications/APIs
- removes the `rails_master_key` which was only being
  used to decrypt/encrypt the `secret_key_base` from the
`credentials.yml.env` in any event.

## Checklist


jsugarman:
out of date and replaced

PR #4872, opened 2023-01-23
ap-3224: Do not display CCMS reference if this doesn't yet exist

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-3224)

Only display the CCMS reference: heading if the application has a CCMS reference.

The question_and_answers partial is shared and the application might not yet have been submitted (and therefore might not yet have a CCMS reference).

## Checklist

### DIFF
--- a/app/views/shared/review_application/_questions_and_answers.html.erb
+++ b/app/views/shared/review_application/_questions_and_answers.html.erb
@@ -10,10 +10,12 @@
     </dt>
     <dd><%= @legal_aid_application.application_ref %></dd>
 
-    <dt>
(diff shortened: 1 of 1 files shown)


cpjmcquillan (commented):
I'm not convinced something this trivial is worth the cost of two integration tests.

What do you think?

kmahern:
> I'm not convinced something this trivial is worth the cost of two integration tests.
> 
> What do you think?

👍 you are probably right @cpjmcquillan - I have deleted the two tests

cpjmcquillan (approved):

PR #4852, opened 2023-01-18
Ap 3681 civil app journey feature tidy

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3681)

Split up civil_application.feature. This feature file had become very large. It also tested a whole range of different functionality. This commit attempts to separate it into smaller feature files.

Also fixed one of the cookies feature test scenarios that was failing when ran locally. This was failing when ran in chrome headless browser because it couldn't click on a link. Forced a window resize before clicking the link as a workaround.

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing: `bundle exec rake`
- Github should not be reporting conflicts; you should have recently run `git rebase main`.
- There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- The PR description should say what you changed and why, with a link to the JIRA story.
- You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
- You should have checked that the commit messages say why the change was made.

PR #4873, opened 2023-01-23
Fix flaky test

cpjmcquillan:
Before, the feedback page used the `.enum_radio_buttons` class method from the `TranslatableModelAttribute` concern to build options for the difficulty and satisfaction feedback questions.

While these provided a nice API for building the options, it added unnecessary complexity and opened cache-sized door of opportunity for bugs.

This was highlighted by the [flaky test] which started failing intermittently (but with very high frequency) since tests were broken down in #4852.

When the feedback form was completed in Welsh, the translations were cached and subsequent tests of the feedback form in English would fail because the "Easy" option was missing (it was "ysaE" instead).

This fixes the flaky test by replacing the use of `.enum_radio_buttons` with an inline translation lookup —removing the cache and source of the leak.

[flaky test]: https://app.circleci.com/pipelines/github/ministryofjustice/laa-apply-for-legal-aid/20171/workflows/ba644a60-8077-42db-97fe-f1d733948eb1/jobs/104965

### DIFF
--- a/app/models/concerns/translatable_model_attribute.rb
+++ b/app/models/concerns/translatable_model_attribute.rb
@@ -14,37 +14,8 @@
 module TranslatableModelAttribute
   extend ActiveSupport::Concern
 
-  RadioButtonStruct = Struct.new(:name, :label)
-
   def enum_t(attribute)
     model = model_name.i18n_key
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
🔥

naseberry (approved):

PR #4874, opened 2023-01-23
Remove `TranslatableModelAttribute` concern

cpjmcquillan:
Since #4873, this concern only includes a single helper method to lookup translations for enums.

This is fairly straightforward, and didn't feel like it should necessitate the use of its own concern.

Usage of `enum_t` has been replaced with scoped calls to `I18n.t`.


kmahern (approved):
👍 lgtm

PR #4876, opened 2023-01-23
Remove redundant styles

cpjmcquillan:
These styles are not used (and in some instances not imported) anywhere, so they can be safely removed.


naseberry (approved):
💯

PR #4892, opened 2023-01-25
Add hint text when a new substantive cost limit is requested

cpjmcquillan:
This adds some clarifying hint text to the substantive cost limit request question when a provider selects "Yes" to request a higher substantive cost limit.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3227)


jsugarman (approved):
👍

PR #4894, opened 2023-01-25
Use Snyk to scan docker image daily

cpjmcquillan:
This adds a GitHub action to scan the application docker image 
using Snyk each day. The workflow runs each day at 05:45am, 
and the results are uploaded to GitHub CodeQL.

### DIFF
--- a/.github/workflows/docker-image-scan.yml
+++ b/.github/workflows/docker-image-scan.yml
@@ -1,0 +1,42 @@
+name: Scan docker image
+
+on:
+  schedule:
+    - cron:  '45 5 * * *'
+
+jobs:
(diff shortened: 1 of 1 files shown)


cpjmcquillan:
I don't think my Synk account has the correct permissions to be able to create service accounts, so somebody else will need to either give me the correct permissions or create the account and add the token as a secret to GitHub Actions.

cpjmcquillan:
The `SNYK_TOKEN` secret has now been added

jsugarman (approved):
config wise looks fine.

I would wait for merge until `SNYK_TOKEN` is ready.

PR #4904, opened 2023-01-26
AP-3735: Investigate Application Digest failures

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3735)

This increases the logging details to provide more data to help diagnose issues, it also raises an error when the spreadsheet is empty so that, in theory, the job will rerun and hopefully succeed

It extends the tests to cover the error generation

## Checklist


jsugarman (approved):
👍 

given what we know now do we need all the extra logging. but either way!

colinbruce:
> 👍
> 
> given what we know now do we need all the extra logging. but either way!

It may help with the next issue!
And I'm only 70% sure it will fix the issue!

PR #4905, opened 2023-01-26
Ignore Puppeteer dependabot updates

cpjmcquillan:
Puppeteer (and Chromium) are used by the Grover gem to generate PDF reports.

In development Puppeteer downloads the relevant Chromium version when installed, but we need to install Chromium separately in CircleCI and when building a docker image.

Since each Alpine image only supports a specific Chromium version, we need to install a specific version of Puppeteer.

If we want to use the same version across all environments (production, test, development), then we shouldn't update until we move to a new Alpine image.

This pins the version in `package.json` and ignores Puppeteer in the dependabot config.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,3 +25,5 @@ updates:
   open-pull-requests-limit: 10
   reviewers:
   - "ministryofjustice/laa-apply-for-legal-aid"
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
👍

PR #4921, opened 2023-01-30
Farewell webpacker

cpjmcquillan:
Rails 7 brought with it some pretty wholesale changes to how Rails 
applications work with JavaScript and assets by default.

Modern Rails applications often don't need Node at all, and the default
trifecta of [importmaps], [turbo], and [stimulus] gives teams all they need.

[importmaps]: https://guides.rubyonrails.org/working_with_javascript_in_rails.html#import-maps
[turbo]: https://turbo.hotwired.dev
[stimulus]: https://stimulus.hotwired.dev

While HTML-over-the-wire is cool, and I'd love to never install Node ever
again, our use case as a Government Digital Service is slightly different.

We lean heavily on the design system and [@alphagov/govuk-frontend](https://github.com/alphagov/govuk-frontend)
for almost all our JS and CSS.

Because of that, we need to take a slightly different approach. 

Here is what I would recommend based on personal experience, and from 
what I have seen work well in other government departments (e.g DfE).

---

1. Remove webpacker, and replace it with the new default Rails gems for 
managing JS and CSS: [jsbundling-rails] and [cssbundling-rails]

[jsbundling-rails]: https://github.com/rails/jsbundling-rails
[cssbundling-rails]: https://github.com/rails/cssbundling-rails

2. Use [propshaft](https://github.com/rails/propshaft) to manage assets. It
might feel like a return to sprockets, but it is far simpler and likely to be the
default approach in future versions of Rails.

3. Replace webpack with [esbuild](https://github.com/evanw/esbuild) which is
fast, simple, and easy to use.

---

### Todo

- [x] Identify IE11 users and notify them that we will be dropping support


package.json
```diff
@@ -3,30 +3,35 @@
   "private": true,
   "scripts": {
     "standardjs": "standard",
-    "stylelint": "stylelint **/*.scss",
-    "test": "jest"
+    "stylelint": "stylelint app/assets/stylesheets/**/*.scss",
+    "test": "jest",
+    "build": "esbuild app/javascript/*.* --bundle --sourcemap --outdir=app/assets/builds --public-path=assets --target=es6",
```
cpjmcquillan, line 8: By default, the "earliest" `esbuild` can transpile to is ES6 JavaScript. Since
IE11 is officially end-of-life and not supported, we took the decision to drop
support for it (i.e ES6 is fine).

If we decide we want to bundle IE11 compatible JS, then we can add an `esbuild`
config file, and use `babel` to transpile compatible code. We might also need to
re-add `regenerator-runtime` for compatible async JS.

app/javascript/application.js
```diff
@@ -0,0 +1,55 @@
...
+
+    if (requestCount) {
+      const requestsOutstanding = parseInt(requestCount.dataset.requestsOutstanding)
+      requestCount.dataset.requestsOutstanding = (requestsOutstanding - 1).toString()
+    }
+  }
+  return response
+}, function (error) {
+  return Promise.reject(error)
+})
```
cpjmcquillan, line 55: Before, we relied on `jquery` to tell us if there were any ongoing AJAX requests
in tests, so could wait for them to resolve before making assertions against
asynchronously rendered content.

We shouldn't need to do this, and Capybara should wait appropriately. We might
want to revisit the tests that make use of the `wait_for_ajax` helper to ensure
they are not making network requests, and that Capybara selectors wait long
enough.

Removing jquery means we cannot rely on `$.active` or `AJAX.activeRequestCount`.
Since we are using `axios`, we can configure some crude
[interceptors](https://github.com/axios/axios#interceptors) that increment and
decrement a data attribute on an empty `div` to determine whether requests are
ongoing in development mode.

I would expect us to discuss and iterate on this approach. Alternatives might be
to rethink how our tests work (i.e we expect rendered content or certain
behvaiour after the request), or to use a different driver (e.g Cuprite) that
allows us to easily wait for network idle in our tests.

cpjmcquillan:
As an aside, does anybody know why we don't use `govuk-frontend` assets and instead have copies of them in our repo?

Procfile.dev
```diff
@@ -0,0 +1,4 @@
+web: bin/rails server -p 3002
+worker: bundle exec sidekiq
+js: yarn build --watch
```
naseberry, line 3: Make command descriptive, i.e `yarn build:js`

package.json
```diff
@@ -3,30 +3,35 @@
...
+    "build:css": "sass ./app/assets/stylesheets/application.scss:./app/assets/builds/application.css --no-source-map --load-path=node_modules"
   },
   "dependencies": {
-    "@rails/activestorage": "^7.0.4",
     "@rails/ujs": "^7.0.4",
```
naseberry, line 12: Is `@rails/ujs` dependency still required?

package.json
```diff
@@ -3,30 +3,35 @@
...
+    "postcss": "^8.4.21",
     "puppeteer": "19.2.0",
     "standard": "^17.0.0",
     "stylelint": "^14.16.1",
     "stylelint-config-gds": "^0.2.0",
-    "stylelint-order": "^6.0.1",
-    "webpack-dev-server": "^4.11.1"
+    "stylelint-order": "^6.0.1"
   },
   "resolutions": {
```
naseberry, line 36: Tidy: review & remove resolutions related to webpacker

app/javascript/application.js
naseberry: Refactoring the applications' Javascript is part of the larger objectives:

1. Reduce dependencies
2. Use core JS (fetch API)
3. Review JS testing strategy

This would mean not relying on jquery, axios etc.

naseberry:
> As an aside, does anybody know why we don't use `govuk-frontend` assets and instead have copies of them in our repo?

This pre-dates my time on the team, could you replace these local assets with those from `govuk-frontend`

naseberry:
```
1. Remove webpacker, and replace it with the new default Rails gems for 
managing JS and CSS: [jsbundling-rails] and [cssbundling-rails]

[jsbundling-rails]: https://github.com/rails/jsbundling-rails
[cssbundling-rails]: https://github.com/rails/cssbundling-rails

2. Use [propshaft](https://github.com/rails/propshaft) to manage assets. It
might feel like a return to sprockets, but it is far simpler and likely to be the
default approach in future versions of Rails.

3. Replace webpack with [esbuild](https://github.com/evanw/esbuild) which is
fast, simple, and easy to use.
```

Great work, lets discuss this next week.

package.json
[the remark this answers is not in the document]
cpjmcquillan: > Make command descriptive, i.e yarn build:js

This command is mostly autogenerated (I added the `--target` flag).

It's added as part of the install step with [@rails/jsbundling-rails](https://github.com/rails/jsbundling-rails), and it [needs to be called `build` so that the relevant Rails tasks call it](https://github.com/rails/jsbundling-rails/blob/main/lib/tasks/jsbundling/build.rake).

---

> Can improve performance by ... Minifying compiled file

Sure, we can add a `--minify` flag.

---

> Can improve performance by ... Adding content hash to file name

[@rails/propshaft handles digest stamping for us](https://github.com/rails/propshaft#readme) already, if that's what you're referring to. 

I'm not familiar with cloud platform, so I'm not sure how or where we configure cache related headers etc. Are we using cloudfront?

package.json
cpjmcquillan: Yeah, it is. 

You can see the [related test failures when I tried to remove it](https://app.circleci.com/pipelines/github/ministryofjustice/laa-apply-for-legal-aid/20382/workflows/19c5c664-e4a9-4cc7-86b9-05653cb87835/jobs/106031) 😢

cpjmcquillan:
> This pre-dates my time on the team, could you replace these local assets with those from govuk-frontend

Sure. I just wondered if anyone knew why we did it this way, but I'm happy for us to remove them 👍

app/javascript/application.js
cpjmcquillan: We decided to leave `jQuery` in for now, and we will revisit this as part of a wider JS-focused review and refactor.

colinbruce:
> > This pre-dates my time on the team, could you replace these local assets with those from govuk-frontend
> 
> Sure. I just wondered if anyone knew why we did it this way, but I'm happy for us to remove them 👍

Could this service pre-date the govuk-frontend packaged versions?

naseberry:
Can remove webpacker output directories `/public/packs` and `/public/packs-test` from `.gitignore`

naseberry:
Update linter include paths in `.overcommit.yml`

package.json
```diff
@@ -3,30 +3,36 @@
   "private": true,
   "scripts": {
     "standardjs": "standard",
-    "stylelint": "stylelint **/*.scss",
-    "test": "jest"
+    "stylelint": "stylelint app/assets/stylesheets/**/*.scss",
+    "test": "jest",
+    "build": "esbuild app/javascript/*.* --bundle --sourcemap --minify --outdir=app/assets/builds --public-path=assets --target=es6",
+    "build:css": "sass ./app/assets/stylesheets/application.scss:./app/assets/builds/application.css --no-source-map --load-path=node_modules"
```
naseberry, line 9: Could we minify the css output?

.circleci/config.yml
```diff
@@ -209,6 +209,9 @@ jobs:
         command: |
           curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./tmp/cc-test-reporter
           chmod +x ./tmp/cc-test-reporter
+    - run:
+        name: Precompile assets
```
colinbruce, line 213: Could this be a shared job, rather than be repeated?

.circleci/config.yml
```diff
@@ -247,6 +250,9 @@ jobs:
     - *restore_js_packages_cache
     - *install_js_packages
     - *save_js_packages_cache
+    - run:
+        name: Precompile assets
+        command: bundle exec rails assets:precompile
```
colinbruce, line 255: See above

.circleci/config.yml
cpjmcquillan: What do we gain from making it a shared job?

Personally I find it harder to jump around the config file (it's over 500 lines now 😬).

How do you feel about a 1 line command? I think it's pretty self-explanatory so we could lose the `name` argument 🤷‍♂️

```yml
- run: bundle exec rails assets:precompile
```

package.json
cpjmcquillan: Sure 👍

.circleci/config.yml
colinbruce: If we remove duplication we may be able to get it to 499 lines! 😆 
Fair point, I use an IDE that lets me click through re-used yaml tags *most* of the time so I get less affected by the file size

package.json
```diff
@@ -3,30 +3,36 @@
   "private": true,
   "scripts": {
     "standardjs": "standard",
-    "stylelint": "stylelint **/*.scss",
-    "test": "jest"
+    "stylelint": "stylelint app/assets/stylesheets/**/*.scss",
```
naseberry, line 6: Stylelint path with glob issue: this only runs against sub-directories due to a path and glob problem, use `app/assets/stylesheets/` instead, this should target all files including sub-directories

cpjmcquillan: Done and tested with violations in nested and top level files 👍

naseberry (approved):

PR #4925, opened 2023-01-31
Update clamav image to most recent build

jsugarman:
## What
Update clamav build to latest available

Pin to most recent tagged version, rather than use `latest` for
testing and stability reasons.

see available tags

https://github.com/ministryofjustice/hmpps-tools-images/pkgs/container/hmpps-clamav

## Checklist


jsugarman:
Sanity tested with normal and malware file and all looks fiine

Screenshot 2023-01-31 at 15 28 15

cpjmcquillan (approved):

PR #4938, opened 2023-02-02
[AP-3782] Tweak and extend application DoB specs

jsugarman:
## What
Amend and extend DoB specs

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3782)

Came of investigating bug AP-3782

Previous use of `3.days.from_now`, which returns a Timestamp, in
the test was actually excercising an invalid date error. Actual behaviour
in the app for dates in the future is to raise date_is_in_future error
with message "Date of birth must be in the past". Replacing with
`Date.tomorrow` excercises the intended logic.

Extended tests to cover far in past dates too.

## Checklist

### DIFF
--- a/spec/forms/applicants/basic_details_form_spec.rb
+++ b/spec/forms/applicants/basic_details_form_spec.rb
@@ -53,7 +53,7 @@ RSpec.describe Applicants::BasicDetailsForm, type: :form do
       end
     end
 
-    context "with an invalid date" do
+    context "with an invalid date_of_birth string" do
       let(:attributes) { attributes_for(:applicant).merge(date_of_birth: "invalid-date") }
 
       it "does not persist model" do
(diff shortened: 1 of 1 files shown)


spec/forms/applicants/basic_details_form_spec.rb
```diff
@@ -66,8 +66,21 @@
...
-    context "with dob in the future" do
-      let(:attributes) { attributes_for(:applicant).merge(date_of_birth: 3.days.from_now) }
+    context "with date_of_birth in the future" do
+      let(:attributes) { attributes_for(:applicant).merge(date_of_birth: Date.tomorrow) }
+
+      it "does not persist model" do
+        expect { subject.save }.not_to change(Applicant, :count)
+      end
+
+      it "errors to be present" do
```
colinbruce, line 76: This reads a bit weirdly... how about 
> it "generates errors"

So that the rspec output will read
> with date_of_birth in the future it generates errors

?

colinbruce (approved):
One, non-blocking, observation but otherwise LGTM 👍

spec/forms/applicants/basic_details_form_spec.rb
jsugarman: sure, was a copy and paste, from previous which i will amend too - i was not keen on the language of the spec either

PR #4950, opened 2023-02-06
ap-3640: implement gov.uk header component

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3644)

- Amended header.html.erb to use govuk header component https://govuk-components.netlify.app/components/header
- Introduced application_helper spec to test header content (mocked devise helper methods to allow this)
- Removed now redundant app/views/shared/_provider_header_link.html.erb

## Checklist


app/helpers/application_helper.rb
```diff
@@ -1,10 +1,6 @@
 module ApplicationHelper
   def home_link
-    link_url = "#"
-    link_url = providers_legal_aid_applications_url if request.path_info.include?("providers")
-    link_to_accessible(t("layouts.application.header.title"),
-                       link_url,
-                       class: "govuk-heading-m govuk-!-margin-bottom-0 govuk-header__link govuk-header__service-name")
+    request.path_info.include?("providers") ? providers_legal_aid_applications_url : "#"
```
naseberry, line 3: Concern: Is this correct, there is no link if a `citizen`?

naseberry (approved):

app/helpers/application_helper.rb
kmahern: @naseberry I believe that is the current functionality for admins and citizens

app/helpers/application_helper.rb
```diff
@@ -42,31 +32,26 @@ def current_journey
...
-    provider_header_link
+    provider_header_navigation(header)
   end
 
-  def provider_header_link
+  def provider_header_navigation(header)
     if provider_signed_in?
-      render "shared/provider_header_link"
+      header.with_navigation_item(text: current_provider.username, href: providers_provider_path, active: false)
+      header.with_navigation_item(text: t("layouts.logout.provider"), href: destroy_provider_session_path, active: false, options: { method: :delete }, html_attributes: { method: :delete })
```
naseberry, line 45: `html_attributes: { method: :delete }`, this attribute is added to the `<li>` and may not required

kmahern: @naseberry without this I get a Page not found error when I click on the Logout link. I think because by default the link generates a GET action and saml_sessions_controller.rb cannot direct this to the right action (as it requires a DELETE action).

kmahern: 👍 thanks @naseberry have removed html_attributes: { method: :delete }

naseberry:
Looks good 👍

spec/rails_helper.rb
```diff
@@ -80,6 +80,7 @@
   config.include Devise::Test::ControllerHelpers, type: :controller
   config.include Devise::Test::IntegrationHelpers, type: :request
   config.include Capybara::RSpecMatchers, type: :request
+  config.include GovukComponent, type: :helper
```
cpjmcquillan, line 83: I'm confused by this bit (because I've not seen it before). What's the reason we need it?

app/helpers/application_helper.rb
cpjmcquillan: Is there a reason we're checking the path rather than something like `provider_signed_in?`?

Also, is there a reason we're passing the entire url rather than using the `_path` helper?

spec/rails_helper.rb
kmahern: You are right @cpjmcquillan I thought I needed it for the reference to GovukComponent::HeaderComponent in the spec to work but it isn't required - have removed 👍

PR #4951, opened 2023-02-06
AP-3794: Update digest-exporter

colinbruce:
## What

This PR updates the generation of the digest

As data-studio requires a static sheet name of `Sheet1`, previously
this would delete rows and re-create, this has started to cause errors
where the header row would get overwritten and cause the data-studio to break

This now renames the current Sheet1 to `Archive` and the date it was
originally generated and recreates Sheet1 from scratch

## Checklist


jsugarman (approved):
👍 

No blockers but be good to break this up a bit from my point of view. Coming to it for the first time there are some big methods with hard coded values and vars that could be clearer named and possibly turned into attr_reader etc.

But lets fix it first  💯

naseberry (approved):

PR #4952, opened 2023-02-06
[AP-3776] Add pre-MTR phase one under 16 interrupt page

jsugarman:
## What
Add pre-MTR phase one under 16 interrupt page

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3776)

Under 16s cannot apply through Apply, for section 8
or any other matter, until MTR phase one is enabled. The
provider should be redirected to CCMS.

This is because under 16s require a special means test that
apply does not implement, but CCMS does.

## Checklist


colinbruce (approved):
LGTM :+1:

PR #4959, opened 2023-02-07
Use `.html.erb` templates

cpjmcquillan:
This updates some partial templates to `.html.erb` files rather than `.erb` files.

They were probably left over from a Rails upgrade once upon a time.


app/views/shared/_vehicle_results.erb
```diff
@@ -1,11 +1,11 @@
 <% if @cfe_result.vehicles? %>
```
jsugarman, line 1: should this not be an `.html.erb` too then?

jsugarman (approved):
LGTM, one file in this PR is still an `.erb` - `app/views/shared/_vehicle_results.erb`

PR #4960, opened 2023-02-07
AP-3794: Update DigestExporter... again

colinbruce:
## What

This PR is another attempt to make the digest export more reliable, the SpreadsheetError has been renamed so it can be re-used and the delete_rows method (which could only be called for row 2+) has been replaced with a lower level Google::Apis call that allows the deletion of all rows and columns, leaving only cell A1 behind

The header rows have been incorporated into the @rows variable so that a single write function now exports all data in a single hit (rather than write 1 row, then rows 2 to 25K!)

This still needs refactoring but I would like to get it functioning before making more refinements!

## Checklist

### DIFF
--- a/app/services/digest_exporter.rb
+++ b/app/services/digest_exporter.rb
@@ -1,5 +1,5 @@
 class DigestExporter
-  class SpreadsheetEmpty < StandardError; end
+  class SpreadsheetError < StandardError; end
 
   def self.call
     new.call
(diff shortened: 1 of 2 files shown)


cpjmcquillan (approved):

app/services/digest_exporter.rb
```diff
@@ -46,36 +47,38 @@ def digest_ids
...
+    return if @worksheet.max_rows == 1 && @worksheet.max_cols == 1
 
-    log_message "Writing #{ApplicationDigest.column_headers.count} column headers + extraction date"
-    # Overwrite the 1 remaining row with column headers
-    @worksheet.update_cells(1, 1, [ApplicationDigest.column_headers + extraction_date])
-    @worksheet.save
+    log_message "reset_worksheet"
+    clear_sheet("ROWS") unless @worksheet.max_rows == 1
+    clear_sheet("COLUMNS") unless @worksheet.max_cols == 1
+    @worksheet.reload
```
naseberry, line 78: Should this be `.save` instead of `.reload`

naseberry, line 78: Should this be `.save` instead of `.reload`

colinbruce: No. Because we use the batch_update of google api calls in `clear_sheet` these are saved automatically *but* the cached version of @worksheet does not get updated without a reload :+1:

PR #4966, opened 2023-02-08
ap-3472: add employment status to your application

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/jira/software/c/projects/AP/boards/233?modal=detail&selectedIssue=AP-3472)

Added Employment status to the "Your application" page (the page the provider sees when clicking into the application while waiting for the client to finish their part). 

## Checklist

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -98,4 +98,8 @@ class Applicant < ApplicationRecord
 
     sprintf("%<amount>.2f", amount: cfe_result.mortgage_per_month || 0)
   end
+
+  def employment_status
+    employed? ? "Employed" : "Not employed"
+  end
 end
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -100,6 +100,6 @@ class Applicant < ApplicationRecord
   end
 
   def employment_status
-    employed? ? "Employed" : "Not employed"
+    employed? ? I18n.t("activemodel.attributes.applicant.employed") : I18n.t("activemodel.attributes.applicant.not_employed")
   end
 end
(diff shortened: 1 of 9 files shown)


app/views/shared/check_answers/_client_details.html.erb
```diff
@@ -58,6 +58,16 @@
         ) %>
   <% end %>
 
+  <% if :employment_status.in?(attributes) && !applicant.employed.nil? %>
+    <%= check_answer_link(
+          name: :employment_status,
+          url: nil,
```
naseberry, line 64: Should there be a link to the applicant employment status view?

app/models/applicant.rb
```diff
@@ -98,4 +98,8 @@ def mortgage_per_month
 
     sprintf("%<amount>.2f", amount: cfe_result.mortgage_per_month || 0)
   end
+
+  def employment_status
+    employed? ? "Employed" : "Not employed"
```
naseberry, line 103: Localization: should these values be translated?

jsugarman (commented):
code aside, the employment status is only select by the provider AFTER this check your answers page so i do not understand the ticket?

kmahern:
> code aside, the employment status is only selected by the provider AFTER this check your answers page so i do not understand the ticket?

The client_details partial is shared in several places, but this ticket is to add employment status to the the "Your application" page (the page the provider sees when clicking into the application from the list of applications while waiting for the client to finish their part) - this is after the user has selected the employment status although it did initially confuse me.

app/views/shared/check_answers/_client_details.html.erb
kmahern: I don't think so no. There aren't links to any of the other client details on this page and I believe it would be too late in the journey for the user to change the employment status. However, I have removed url: nil for clarity.

naseberry (approved):

PR #4967, opened 2023-02-08
Use `ActiveRecord::Encryption` to encrypt TrueLayer tokens

cpjmcquillan:
Rails 7 introduced an API for encrypting ActiveRecord model attributes at the application level.

[ActiveRecord Encryption](https://guides.rubyonrails.org/active_record_encryption.html)

We can use this high level API in place of our bespoke `SecureData` pattern.

This begins encrypting TrueLayer token data for Applicants (alongside the existing SecureData encryption).

Subsequent work will address migration of existing data and ultimately the removal of all `SecureData` code, but this is a nice first step.


config/application.rb
```diff
@@ -30,6 +30,11 @@ class Application < Rails::Application
       g.test_framework :rspec
     end
 
+    # ActiveRecord::Encryption keys generated with `bin/rails db:encryption:init`
+    config.active_record.encryption.primary_key = ENV.fetch("ENCRYPTION_PRIMARY_KEY", "fake-primary-key")
+    config.active_record.encryption.deterministic_key = ENV.fetch("ENCRYPTION_DETERMINISTIC_KEY", "fake-deterministic-key")
+    config.active_record.encryption.key_derivation_salt = ENV.fetch("ENCRYPTION_KEY_DERIVATION_SALT", "fake-key-derivation-salt")
```
cpjmcquillan, line 36: The guides recommend using the built-in `credentials.yml.enc` (it would pick up the keys automatically) but we're not using it.

Initially I configured the application to require these environment variables to be set in production and development, with fake values used in test.

That was clunky though, and meant we had to pass dummy environment variables when doing a whole bunch of stuff (e.g building docker images, precompiling assets) outside of `RAILS_ENV=test`.

It feels strange to fallback to values when we definitely need these to be set in production environments. Perhaps we could explicitly fetch without a fallback in `config/production.rb` —but we're not doing that for anything else.

What do people think?

colinbruce: Agree it feels clunky, can we drop a tech debt ticket in the backlog to review after the CloudPlatform secret manager update?

colinbruce (approved):
LGTM 👍

config/application.rb
cpjmcquillan: https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4967/files#r1101957684

PR #4968, opened 2023-02-08
Migrate applicant SecureData true layer tokens

cpjmcquillan:
Before, the bespoke `SecureData` pattern was used to store applicant TrueLayer tokens.

In #4967, a new `encrypted_true_layer_token` encrypted attribute was introduced, and data was written to it as part of the TrueLayer flow.

This adds a temporary idempotent task to copy encrypted token data to the new attribute, and updates existing methods to read from the new attribute if it is present.

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -34,11 +34,19 @@ class Applicant < ApplicationRecord
   end
 
   def true_layer_token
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM :+1:

PR #4973, opened 2023-02-09
[AP-3778] Under 18 content changes

jsugarman:
## What
[Link to story](https://dsdmoj.atlassian.net/browse/AP-3788)

Content change following review by content designer and others

- [X] Amends the client declaration form
- [X] Amends the review and print your application form
- [X] Amends the printed application page
- [x] Amends the application complete page
- [X] does this need to be for under 18 pre-MTR phase one to?*

\* in which case all existing content can use the `under_18?` method to handle the conditional logic, but this impacts all these changes plus previous work done on the client declaration as well as naming considerations [partials and translation impacts]  - see notes below.

## NOTE:

This [commit switches to using under 18 logic](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4973/commits/3c4625be67ce902ea91bbc2e5de131ccefb41f80) rather than non-means-tested logic. If BA and content agree it can remain in but if not then it can just be removed.

- [x] Fix "Printable declaration" for "LAA privacy policy" link issue - additional line break in printed version
- [x] ??? Fix [or not] "Printable declaration" for "LAA privacy policy" link issue - href visible in printed version - @naseberry 
- [x] A single line of content which is not applicable pre-MTR phase one, as below - conditionalise or amend to be applicable now? - @cpjmcquillan 

> if the case continues after their 18th birthday, we may reassess their income and capital to check if they should contribute towards legal costs

## Checklist

### DIFF
--- a/app/assets/stylesheets/helpers.scss
+++ b/app/assets/stylesheets/helpers.scss
@@ -21,6 +21,7 @@
 
 .app-print-only {
   display: none !important;
+  page-break-before: always;
 
   @include govuk-media-query($media-type: print) {
     display: inherit !important;
--- a/app/assets/stylesheets/helpers.scss
+++ b/app/assets/stylesheets/helpers.scss
@@ -19,6 +19,14 @@
   display: none !important;
 }
 
+.app-print-only {
+  display: none !important;
+
+  @include govuk-media-query($media-type: print) {
+    display: inherit !important;
+  }
+}
+
 @media print {
   .no-print,
   .no-print * {
--- a/app/assets/stylesheets/print-styles.scss
+++ b/app/assets/stylesheets/print-styles.scss
@@ -41,20 +41,6 @@
     font-weight: bold;
   }
 
-  .govuk-list.govuk-list--bullet > li > strong:before {
-    content: "● ";
-  }
-
-  .print-add-bullet {
-    position: relative;
-
-    & > li:before {
-      content: "●";
-      position: absolute;
-      left: 0;
-    }
-  }
-
   .print-bottom-gap {
     padding-bottom: 3.5em;
   }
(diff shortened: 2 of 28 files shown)


naseberry:
@jsugarman re "href visible in print version"

The Design System includes link destination when printing, see: https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/helpers/_links.scss#L411

We have enable [global styles](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-global-styles) that explicitly treats all links as a govuk-link and this calls the above link destination printing.
Reference: https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/webpack/stylesheets/govuk-frontend.scss#L7

colinbruce:
So, having done some digging and discussing with BA and Content people, to answer @jsugarman 's questions...
>does this need to be for under 18 pre-MTR phase one to?*
>* in which case all existing content can use the under_18? method to handle the conditional logic, but this impacts all these changes plus previous work done on the client declaration as well as naming considerations [partials and translation impacts] - see notes below.
>
>NOTE:
This [commit switches to using under 18 logic](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/4973/commits/3c4625be67ce902ea91bbc2e5de131ccefb41f80) rather than non-means-tested logic. If BA and content agree it can remain in but if not then it can just be removed.
> A single line of content which is not applicable pre-MTR phase one, as below - conditionalise or amend to be applicable now? - @cpjmcquillan

I am going to investigate and revert that commit as these changes should only_occur when the MTR feature flag is true and the application is under 18

colinbruce (approved):

colinbruce (approved):
:+1:

PR #4974, opened 2023-02-09
Remove `data_migrate` gem

cpjmcquillan:
The [data_migrate](https://github.com/ilyakatz/data-migrate) gem was introduced in #2955 as a way of migrating data alongside schema migrations.

Since then, the gem has only been used sporadically —and not at all in over a year. Old data migrations also cannot run without raising errors in the current state of the application.

This removes the gem as a dependency.


colinbruce (approved):
LGTM :+1:

PR #4975, opened 2023-02-09
Use `rails db:prepare` to prepare the database

cpjmcquillan:
Since Rails 6, the [`db:prepare` task][task] has provided a simple API for managing databases in Rails applications.

Running `bin/rails db:prepare` will prepare the database. This means

If the database does not exist:
  1. Create the database
  2. Load the schema if it exists, or run migrations
  3. Seed the database

If the database exists already, then only migrations are run.

[task]: https://github.com/rails/rails/blob/9fcfbe5306cf632f5f47fa8bb5dfa36545ab1be6/activerecord/lib/active_record/tasks/database_tasks.rb#L188

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ executors:
     docker:
       - image: ministryofjustice/apply-ci:latest-3.2.0
         environment:
+          RAILS_ENV: test
           GITHUB_TEAM_NAME_SLUG: laa-apply-for-legal-aid
           CACHE_VERSION: v1
           NODE_OPTIONS: --openssl-legacy-provider
@@ -121,13 +122,6 @@ references:
       paths:
         - node_modules
 
-  setup_database: &setup_database
-    run:
-      name: Database Setup
-      command: |
-        bundle exec rake db:create db:schema:load
-        bundle exec rake db:migrate db:seed
-
 jobs:
   lint_checks:
     executor: linting-executor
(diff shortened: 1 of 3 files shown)


colinbruce:
I'm not sure it's running schema load _yet_ https://blog.saeloun.com/2023/02/02/rails-allow-db-prepare-to-load-schema-if-database-exists

.circleci/config.yml
```diff
@@ -186,7 +180,9 @@ jobs:
         key: apply-repo-{{ .Environment.CIRCLE_SHA1 }}
         paths:
           - ~/project
-    - *setup_database
+    - run:
```
colinbruce, line 183: Is there a reason you duplicated the steps, rather than replacing the setup_database commands?

colinbruce (approved):
One non blocking query but otherwise looks okay :+1:

.circleci/config.yml
colinbruce: I saw the comment on the other PR - please ignore - asked and answered!

cpjmcquillan:
> I'm not sure it's running schema load _yet_ https://blog.saeloun.com/2023/02/02/rails-allow-db-prepare-to-load-schema-if-database-exists

Currently it loads the schema when the database doesn't exist.

However in Rails 7.1 the command will load the schema instead of running migrations if the database exists but hasn't yet been populated with tables 👍

cpjmcquillan:
In UAT the database will already exist, so `db:seed` isn't run (but we want it to).

We could always seed the database, but that might not be much better.

```sh
bundle exec rails db:prepare db:seed
```

Let's discuss

PR #4980, opened 2023-02-10
AP-3794: Update digest exporter 

colinbruce:
## What

Previously we were using a mixture of `google_drive` gem methods (that were slow) and `Google::Api::SheetService`
calls

This consolidates all methods into the single, SheetService type, for speed of use and simplicity

It removes the google_drive gem and uses only the standard google supported gem

## Checklist


Gemfile
```diff
@@ -8,7 +8,7 @@ gem "active_model_serializers", "~> 0.10.13"
 gem "data_migrate"
 gem "discard", "~> 1.2"
 gem "geckoboard-ruby"
-gem "google_drive", ">= 3.0.7"
+gem "google-apis-sheets_v4"
```
naseberry, line 11: :white_check_mark:

naseberry (approved):

PR #4991, opened 2023-02-14
Ignore the `SecureData` true layer token

cpjmcquillan:
Before #4698, the `true_layer_secure_data` association was used to retrieve secure TrueLayer tokens when interacting with the TrueLayer API to import client's bank data.

Now that existing tokens have been migrated to the new `encrypted_true_layer_token` column and the application is using this data, we can begin to remove the legacy true layer token code.

This ignores the `true_layer_secure_data_id` column as a first step, and that change was used to drive the work —fixing (and refactoring) failing tests as they arose.

As part of the work, bank providers no longer store the TrueLayer `token` and `token_expires_at` attributes. These were not used anywhere, and can be accessed through the associated applicant if required in future.

When fixing the `ImportProviderService` specs, a uniqueness validation (and corresponding index) was added to ensure duplicate bank providers cannot be added for a given applicant. #1748 did the initial work to resolve a bug, but this validation was omitted.


db/migrate/20230213141824_add_true_layer_provider_id_applicant_index_to_bank_providers.rb
```diff
@@ -0,0 +1,7 @@
+class AddTrueLayerProviderIdApplicantIndexToBankProviders < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
```
cpjmcquillan, line 2: Our applicant table is small, so it probably wouldn't be a problem, but adding indexes blocks writes —so we want to `disable_ddl_transaction` and use the `concurrently` algorithm.

https://thoughtbot.com/blog/how-to-create-postgres-indexes-concurrently-in

spec/support/true_layer_mock_data.rb
```diff
@@ -7,75 +7,81 @@ module TrueLayerHelpers
         provider_id: "lloyds",
       },
     },
-    account_holders: [{
-      full_name: Faker::Name.name,
-      date_of_birth: Faker::Date.backward.to_time.utc.iso8601,
-      addresses: [{
-        address: Faker::Address.street_address,
-        city: Faker::Address.city,
+    account_holders: [
```
cpjmcquillan, line 10: These are alignment changes. Personally I find it much easier to follow now, but feel free to object if you find it more difficult 👍

app/services/true_layer/importers/import_provider_service.rb
```diff
@@ -36,8 +34,6 @@ def mapped_resource
         {
           true_layer_response: provider,
           credentials_id: provider[:credentials_id],
-          token: applicant.true_layer_secure_data_id,
-          token_expires_at:,
```
cpjmcquillan, line 40: I couldn't find any instance of us doing anything with this data, besides asserting against it in tests.

Perhaps we intended to use it when retrieving bank data, or perhaps we wanted to reconcile failed imports that had expired tokens?

I wasn't able to glean anything from my digging, but maybe someone has some missing context about _why_ this data was stored on the bank provider record as well as the applicant.

colinbruce: No missing context from me 🤷‍♂️

db/migrate/20230213141824_add_true_layer_provider_id_applicant_index_to_bank_providers.rb
colinbruce: 👍

spec/support/true_layer_mock_data.rb
colinbruce: Fine by me :+1:

colinbruce (approved):
LGTM :+1:

PR #5001, opened 2023-02-15
Remove ignored true layer columns

cpjmcquillan:
In #4991, some TrueLayer related columns were ignored in the Applicant and BankProvider models.

Now that those columns are ignored, they can be safely removed from their respective tables.


naseberry (approved):
LGTM

PR #5003, opened 2023-02-15
Stop ignoring removed columns

cpjmcquillan:
In #5001 these columns were removed.

We can stop ignoring them now!


naseberry (approved):
LGTM

PR #5010, opened 2023-02-17
Use `ActiveRecord::Encryption` for citizen URLs

cpjmcquillan:
Similarly to #4967, we can use this new API to encrypt citizen URLs rather than our bespoke `SecureData` pattern.

This begins deterministically encrypting a random UUID that is stored against the LegalAidApplication alongside an expiry date.

The UUID is encrypted deterministically (rather than non-deterministically) so that applications can be queried by the `citizen_url_id` when requests are made to the Citizen application page.

A new `CITIZEN_URL_EXPIRES_AFTER_IN_DAYS` constant is used to define the number of days the URL is valid for.

For example, if the URL is generated on 2023-01-01, it will be valid for 7 days, expiring on the 8th day (i.e 2023-01-09).

This is consistent with the existing behaviour, although the implementation differs slightly.

Currently, if a URL is generated at *any time* on 2023-01-01, it will be valid until the end of the day, 7 days later (i.e 2023-01-08 23:59.999999) —after which it will have expired.

Subsequent work will migrate previously generated URLs before removing `SecureData` code.


colinbruce (approved):
LGTM :+1:

PR #5011, opened 2023-02-17
AP-3873: Split opponent model

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3873)

PR 1 of X
We need to implement multiple opponents but the current configuration stores the three, separate, sets of questions in a single, opponent, model.

This is the first PR in a set, it splits the models and tables so that each page of questions is saved to a unique location.  This should allow further work in the future with the minimum of work.

It contains a rake task that migrates the data from the current, Opponent, table into the two new tables Domestic Abuse Summary and Parties Mental Capacity.

A follow-up PR will then remove the fields from the opponent table

## Checklist


app/models/application_merits_task/parties_mental_capacity.rb
```diff
@@ -0,0 +1,5 @@
+module ApplicationMeritsTask
+  class PartiesMentalCapacity < ApplicationRecord
+    belongs_to :legal_aid_application
+  end
+end
```
cpjmcquillan, line 5: I'm slightly confused by the modelling here, but I'm probably missing context.

What does a `PartiesMentalCapacity` represent? 

Can an application have many different parties? I'm assuming it can, and that an `Opponent` is a particular type of party. But what are the different types of parties?

colinbruce: A court case has _many_ parties, at base level and applicant and respondent - in civil law.
This question used to be phrased, `Does the Opponent have the mental capacity to understand the case?`  
It was changed a few months ago, when we implemented Client Involvement Type to, `Do all parties have the mental capacity to understand the case?` to avoid having to have lots of branching locales for `Does your client, the Applicant...`, `Does the respondent...`, etc

This is why I renamed it to PartiesMentalCapacity, rather than OpponentsMentalCapacity

cpjmcquillan: Ok, that makes sense. Thanks for explaining 👍

cpjmcquillan (approved):

PR #5018, opened 2023-02-20
Use encrypted `citizen_url_id` to retrieve applications

cpjmcquillan:
In #5010, unique encrypted citizen URL IDs were generated alongside `SecureData` records and stored on the LegalAidApplication.

This uses the new column to generate unique citizen URLs and updates the `SecureApplicationFinder` service so that it can retrieve applications by citizen URL id *and* by SecureData id.

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -217,19 +217,11 @@ class LegalAidApplication < ApplicationRecord
   end
 
   def generate_secure_id
-    transaction do
-      update!(
-        citizen_url_id: SecureRandom.uuid,
-        citizen_url_expires_on: CITIZEN_URL_EXPIRES_AFTER_IN_DAYS.days.from_now,
-      )
-
-      SecureData.create_and_store!(
-        legal_aid_application: { id: },
-        expired_at: (Time.current + SECURE_ID_DAYS_TO_EXPIRE.days).end_of_day,
-        # So each secure data payload is unique
-        token: SecureRandom.hex,
-      )
-    end
+    update!(
+      citizen_url_id: SecureRandom.uuid,
+      citizen_url_expires_on: CITIZEN_URL_EXPIRES_AFTER_IN_DAYS.days.from_now,
+    )
+    citizen_url_id
   end
 
   def set_transaction_period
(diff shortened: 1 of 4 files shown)


colinbruce (approved):
LGTM :+1:

jsugarman (approved):
👍

app/services/secure_application_finder.rb
```diff
@@ -24,6 +26,10 @@ def check_for_errors
   end
 
   def expired?
-    secure_data[:expired_at] && secure_data[:expired_at] < Time.current
+    expires_on && expires_on < Time.current
+  end
+
+  def expires_on
+    legal_aid_application.citizen_url_expires_on.presence || secure_data[:expired_at]
```
jsugarman, line 33: can we/should we be using  presence for the secure_data too

```suggestion
    legal_aid_application.citizen_url_expires_on.presence || secure_data[:expired_at].presence
```

app/models/legal_aid_application.rb
```diff
@@ -217,19 +217,11 @@ def outgoing_types?
...
-        expired_at: (Time.current + SECURE_ID_DAYS_TO_EXPIRE.days).end_of_day,
-        # So each secure data payload is unique
-        token: SecureRandom.hex,
-      )
-    end
+    update!(
+      citizen_url_id: SecureRandom.uuid,
+      citizen_url_expires_on: CITIZEN_URL_EXPIRES_AFTER_IN_DAYS.days.from_now,
+    )
+    citizen_url_id
```
cpjmcquillan, line 224: I thought this would be a nice way to migrate to using the new approach, but it turns out we are currently calling `generate_secure_id` multiple times when we send an email to the applicant.

Before this worked because each URL would correspond to a new `SecureData` record that referenced the application.

However with the new approach, this means we send emails with URLs that contain `citizen_url_id` values that have since been overwritten).

app/services/secure_application_finder.rb
cpjmcquillan: `object.presence` is another way of writing

```ruby
object.present? ? object : nil
```

so it wouldn't mean anything if we added it to `secure_data[:expired_at]`

https://api.rubyonrails.org/classes/Object.html#method-i-presence

PR #5031, opened 2023-02-22
Introduce encrypted Citizen access tokens 

cpjmcquillan:
In https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5010 we started encrypting a `citizen_url_id` and storing it alongside the
date the corresponding Citizen application link would expire on the
`LegalAidApplication` model.

As work to migrate from the `SecureData` approach continued in https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5018, a bug was
discovered in manual testing.

Rather than generating a token for the Citizen URL once and passing that around,
tokens are generated multiple times for an application in the process of sending
an email to the client.

This means there could be X different `SecureData` records in the database for
an application, and we don't know which correspond to links sent in emails to
Citizens.

In order to accommodate this behaviour, and to migrate existing data so we can
continue to handle existing links appropriately -a new approach was needed.

This introduces the concept of a Citizen access token, that can be sent to
Citizens in emails, expire, and used to retrieve a client's legal aid
application.

This approach will also us to accommodate existing links (including expired
links), however we might later decide that initial requirements laid out in
the [initial ticket] and [discussion] or decisions taken in https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/73 are no longer
appropriate.

[initial ticket]: https://dsdmoj.atlassian.net/browse/AP-148
[discussion]: https://dsdmoj.atlassian.net/wiki/spaces/ATPPB/pages/1106608346/2018-09-17+Securing+Citizen+access+to+the+service

---

A temporary task can be used to migrate existing SecureData records to
Citizen access tokens so that the application can continue to handle existing
links gracefully.

This migrates data that looks like this:

```
SecureData
----------
- id <uuid>
- data <encrypted JWT token>
```

where the JWT data looks like:

```
{
  legal_aid_application: {
    "id": <uuid>
  },
  expired_at: <timestamp> (2020-01-01 23:59:59 +0000),
  token: <hex>
}
```

into data that looks like this:

```
Citizen::AccessToken
---------
- id <uuid>
- token <encrypted uuid string>
- expires_on <date> (2020-01-02)
- legal_aid_application_id <uuid>
```

Links are currently valid for 7 days, but Citizens can request new links if they
try to access an application using an expired link.

This migration will allows us to continue to retrieve applications from valid
and expired links.

### DIFF
--- a/app/models/citizen/access_token.rb
+++ b/app/models/citizen/access_token.rb
@@ -1,0 +1,21 @@
+module Citizen
+  class AccessToken < ApplicationRecord
+    EXPIRES_AFTER_IN_DAYS = 8
+
+    self.table_name = "citizen_access_tokens"
+
+    belongs_to :legal_aid_application
+
+    encrypts :token, deterministic: true
+
+    validates :expires_on, comparison: { greater_than: ->(_record) { Date.current } }
+
+    def self.generate_for(legal_aid_application:)
+      create!(
+        token: SecureRandom.uuid,
+        expires_on: EXPIRES_AFTER_IN_DAYS.days.from_now,
+        legal_aid_application:,
+      )
+    end
+  end
+end
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -1,4 +1,6 @@
 class LegalAidApplication < ApplicationRecord
+  self.ignored_columns += %w[citizen_url_id citizen_url_expires_on]
+
   include Discard::Model
   include DelegatedFunctions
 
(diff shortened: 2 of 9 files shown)


app/models/citizen/access_token.rb
```diff
@@ -0,0 +1,21 @@
+module Citizen
+  class AccessToken < ApplicationRecord
+    EXPIRES_AFTER_IN_DAYS = 8
+
+    self.table_name = "citizen_access_tokens"
+
+    belongs_to :legal_aid_application
+
+    encrypts :token, deterministic: true
```
cpjmcquillan, line 9: Before, all the "data" was JWT encoded for a `SecureData` record.

Now, we encrypt only the `token`. This is what is used in links in emails to Citizens. It is encrypted deterministically so we can query by it.

colinbruce (approved):
LGTM :+1:

jsugarman (approved):
👍 and me

PR #5032, opened 2023-02-23
Remove citizen URL columns

cpjmcquillan:
The new approach taken in #5031 means these columns are no longer relevant and can be safely removed.


colinbruce (approved):
LGTM :+1:

PR #5034, opened 2023-02-24
Retrieve applications by Citizen access token

cpjmcquillan:
This builds on the work in #5031.

Before, `LegalAidApplication` records were retrieved in the `Citizen::LegalAidApplicationsController` by an ID that corresponded to a `SecureData` record.

These `SecureData` records contain JWT encoded tokens that refer to the legal aid application ID and have an expiry time.

In itself, the `SecureData` ID does not expose any sensitive information, and only reveals the legal aid application ID if decoded with the relevant secret key.

Now that all existing `SecureData` records have been migrated to `Citizen::AccessToken` records, the application can be refactored to use these new tokens.

Now, a `Citizen::AccessToken` is retrieved by its encrypted token, and this provides an interface to access the associated `LegalAidApplication`.

These new tokens are encrypted at rest, which means the legal aid application they belong to can only be revealed if you also know the relevant secret keys.

The `SecureApplicationFinder` has been refactored to reflect this new approach, while `LegalAidApplication#generate_secure_id` has been renamed to `LegalAidApplication#generate_citizen_access_token!` to better reflect its intention.

Related email services and controllers have had smaller, related refactors and tidy ups.

Tests have been updated, and in some instances refactored to improve readability or address Rubocop issues.


spec/requests/citizens/consents_controller_spec.rb
```diff
@@ -3,22 +3,23 @@
...
-      get citizens_consent_path
-    end
+    before { get citizens_consent_path }
 
     it "returns http success" do
       expect(response).to have_http_status(:ok)
-      expect(unescaped_response_body).to include("Do you agree to share 3 months of bank statements with " \
-                                                 "the <abbr title='Legal Aid Agency'>LAA</abbr> via TrueLayer?")
+      expect(page).to have_content(
+        "Do you agree to share 3 months of bank statements with the LAA via " \
```
colinbruce, line 14: Does this line need splitting? 
We no longer have line length limits 🤷‍♂️

PR #5035, opened 2023-02-24
Remove `SecureData`

cpjmcquillan:
Now that we are no longer using the bespoke `SecureData` pattern to encrypt sensitive date (i.e Applicant TrueLayer tokens and Citizen LegalAidApplication access tokens), we can remove the code and drop the tables.


Gemfile
```diff
@@ -55,9 +55,6 @@ gem "browser"
 # Used to mock saml request in UAT
 gem "ruby-saml-idp", github: "dev-develop/ruby-saml-idp", branch: "master"
 
-# Used to encrypt JSON stored in SecureData
-gem "jwt"
```
cpjmcquillan, line 59: We're actually using this in `TrueLayer::SampleTransactionsCsvParser`, but `jwt` is pulled in as a dependency by a few different gems —so no tests failed.

Let's see if we can stop relying on `JWT` in that service (it's only used to build mock test data), otherwise we should add `jwt` back as an explicit dependency.

cpjmcquillan: Added a ticket to the backlog to discuss this https://dsdmoj.atlassian.net/browse/AP-3900

colinbruce (approved):
LGTM :+1:

PR #5053, opened 2023-03-02
AP-3864: Enable multiple opponents

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3864)

This PR updates the application -> opponent relationship from has_one to has_many
It then implements the `has-another` pattern for the application merits task opponents' name page

This required a re-write of the ccms/xml matchers to allow multiple values to be matched.  
Currently these only work in to assure all values are present, not their sequence, this may cause an issue in future with boolean values if they are dissimilar. e.g. expecting `[true, false]` it will ensure that one of each is present, but not necessarily that they are assigned to the expected linked blocks

Future work will expand the opponent to allow differentiation between individuals and organisations 

## Checklist

### DIFF
--- a/app/helpers/task_list_helper.rb
+++ b/app/helpers/task_list_helper.rb
@@ -18,7 +18,7 @@ module TaskListHelper
 private
 
   def _task_url(name, legal_aid_application, status)
-    url = if application_has_no_involved_children?(legal_aid_application) && name.eql?(:children_application)
+    url = if display_new_page?(legal_aid_application, name)
             "new_providers_legal_aid_application_#{url_fragment(name)}_path".to_sym
           else
             "providers_legal_aid_application_#{new_url_fragment(name, status, legal_aid_application)}_path".to_sym
@@ -28,6 +28,10 @@ private
 
   def application_has_no_involved_children?(legal_aid_application)
     legal_aid_application.involved_children.empty?
+  end
+
+  def application_has_no_opponents?(legal_aid_application)
+    legal_aid_application.opponents.empty?
   end
 
   def proceeding_task_url(name, application, ccms_code)
@@ -42,6 +46,7 @@ private
 
   def new_url_fragment(name, status, application)
     name = "has_other_involved_children" if name.eql?(:children_application) && (status.eql?(:complete) || application.involved_children.any?)
+    name = "has_other_opponent" if name.eql?(:opponent_name) && (status.eql?(:complete) || application.opponents.any?)
     I18n.t("providers.merits_task_lists.task_list_item.urls.#{name}")
   end
 
@@ -50,5 +55,22 @@ private
 
     proceeding = application.proceedings.find_by(ccms_code:)
     proceeding.id
+  end
+
+  def display_new_page?(legal_aid_application, task_name)
+    return false unless %i[children_application opponent_name].include?(task_name)
+
+    [
+      starting_children_application_for_first_time?(legal_aid_application, task_name),
+      starting_opponents_name_for_first_time?(legal_aid_application, task_name),
+    ].any?(true)
+  end
+
+  def starting_children_application_for_first_time?(legal_aid_application, task_name)
+    task_name.eql?(:children_application) && application_has_no_involved_children?(legal_aid_application)
+  end
+
+  def starting_opponents_name_for_first_time?(legal_aid_application, task_name)
+    task_name.eql?(:opponent_name) && application_has_no_opponents?(legal_aid_application)
   end
 end
--- a/app/services/ccms/requestors/case_add_requestor.rb
+++ b/app/services/ccms/requestors/case_add_requestor.rb
@@ -123,7 +123,9 @@ module CCMS
       end
 
       def generate_other_parties(xml)
-        generate_opponent(xml, @legal_aid_application.opponents.first)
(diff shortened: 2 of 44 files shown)


app/helpers/task_list_helper.rb
```diff
@@ -18,7 +18,8 @@ def task_list_includes?(legal_aid_application, task_name)
 private
 
   def _task_url(name, legal_aid_application, status)
-    url = if application_has_no_involved_children?(legal_aid_application) && name.eql?(:children_application)
+    url = if (application_has_no_involved_children?(legal_aid_application) && name.eql?(:children_application)) ||
+        (application_has_no_opponents?(legal_aid_application) && name.eql?(:opponent_name))
```
naseberry, line 22: Complexity: I know our domain is complicated and this feels complex, is it be possible to make this less complex?

app/services/ccms/requestors/case_add_requestor.rb
```diff
@@ -291,7 +291,8 @@ def generate_opponent_other_parties_means_instance(xml, other_party, config)
       end
 
       def other_parties
-        [opponent] + involved_children
+        # [opponent] + involved_children
```
naseberry, line 296: Comment: Do we need to keep this comment?

app/services/ccms/requestors/case_add_requestor.rb
```diff
@@ -123,7 +123,7 @@ def generate_case_docs(xml)
       end
 
       def generate_other_parties(xml)
-        generate_opponent(xml, @legal_aid_application.opponent)
+        generate_opponent(xml, @legal_aid_application.opponents.first)
```
naseberry: With limited knowledge of the payload, i may be talking squat, but as there can be multiple opponents, should this be an array?

naseberry (commented):
I would require a second review from a ruby-ist.

PS - loving the GoT references in specs.

app/services/ccms/requestors/case_add_requestor.rb
colinbruce: This looks like a rebase issue to me - I am going to return the PR to draft for now and address all of your -very valid- points!

colinbruce: Yeah, a later commit had resolved this 😕 
Not sure how this was in the review state 😖

app/helpers/task_list_helper.rb
colinbruce: Have a look at the latest version @naseberry - happy to discuss

spec/requests/providers/application_merits_task/opponents_names_controller_spec.rb
```diff
@@ -8,50 +8,70 @@ module ApplicationMeritsTask
...
+            login_provider
+            get_existing_opponent
+          end
+
+          it "returns success" do
             expect(response).to have_http_status(:ok)
           end
 
-          it "displays opponent data" do
+          it "displays child details" do
```
kmahern, line 53: Is this description correct? Should it be "displays opponent details"

app/services/flow/flows/provider_merits.rb
```diff
@@ -60,10 +60,40 @@ class ProviderMerits < FlowSteps
           check_answers: :check_merits_answers,
         },
         opponents_names: {
-          path: ->(application) { urls.providers_legal_aid_application_opponents_name_path(application) },
-          forward: ->(application) { Flow::MeritsLoop.forward_flow(application, :application) },
+          path: ->(application) { urls.new_providers_legal_aid_application_opponents_name_path(application) },
+          forward: :has_other_opponents, # ->(application) { Flow::MeritsLoop.forward_flow(application, :application) },
```
kmahern, line 64: can the comment be removed?

spec/services/ccms/requestors/case_add_requestor_spec.rb
```diff
@@ -205,6 +206,21 @@ module Requestors
...
+          let(:opponent_two) { create(:opponent, first_name: "Sansa", last_name: "Opponent-Test") }
+          let(:opponents) { [opponent_one, opponent_two] }
+
+          before do
+            allow(opponent_one).to receive(:generate_ccms_opponent_id).and_return("123456")
+            allow(opponent_two).to receive(:generate_ccms_opponent_id).and_return("654321")
+          end
+
+          it "generates expected xml with multiple opponents" do
+            expect(expected_xml).to match(/Sansa Opponent-Test/)
```
kmahern, line 221: Is this correct? (that only the Sansa Opponent-Test is included  in the xml and not Joffrey Test-Opponent)? Could the description be clearer? (e.g. it "generates expected xml with first opponent")

colinbruce: Good catch, it was my initial test that the _new_ value was at least found.  Extended to ensure it matches both :+1:

colinbruce:
> I would require a second review from a ruby-ist.
> 
> PS - loving the GoT references in specs.

There were there to start, I just extended them! 😀

naseberry (approved):
lgtm

I would have preferred new views to use the govuk-components view helpers, however, we have tickets in the backlog to action this.

PR #5058, opened 2023-03-03
Fix: CCMS Submissions with split opponent data

colinbruce:
## What

When the tables were spilt, handlers were put in place
for when the objects existed, but with the new section 8
applications, sometimes they are not even created

This handles missing domestic abuse summary and parties mental
capacity by not submitting the blocks if the objects are missing

## Checklist


jsugarman (approved):
👍 awesome

colinbruce:
I just did a rebase and remove an additional safe navigation operator

PR #5066, opened 2023-03-06
AP-3915: Add CCMS queue analysis to admin

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3915)

This proof of concept will, in its current form, allow admin users to assess the CCMS queue 

Currently, a developer has to access the console to monitor the steps that have been taken and then assess the next best steps 

The hope is that this is the first step in allowing a non-dev admin to see the current state and then re-start submissions
Index page
image

Individual submission page with both options expanded
image

## Checklist

### DIFF
--- a/app/controllers/admin/ccms_queues_controller.rb
+++ b/app/controllers/admin/ccms_queues_controller.rb
@@ -1,0 +1,46 @@
+module Admin
+  class CCMSQueuesController < AdminBaseController
+    before_action :authenticate_admin_user!
+
+    def index
+      @in_progress = CCMS::Submission.where.not(aasm_state: %w[completed abandoned]).order(created_at: :desc)
+    end
+
+    def show
+      submission
+      legal_aid_application
+    end
+
+    def reset_and_restart
+      result = submission.complete_restart! # return from service
+      status = result ? :notice : :error
+      core = "submission #{submission_id}"
+      message = status.eql?(:error) ? "Restarting #{core} failed" : "Reset and restarted #{core}"
+      flash[status] = message
+      redirect_to action: :show
+    end
+
+    def restart_current_submission
+      result = submission.restart_existing_submission!
+      status = result ? :notice : :error
+      core = "submission #{submission_id}"
+      message = status.eql?(:error) ? "Restarting #{core} failed" : "Restarted #{core}"
+      flash[status] = message
+      redirect_to action: :show
+    end
+
+  private
+
+    def submission
+      @submission ||= CCMS::Submission.find(submission_id)
+    end
+
+    def legal_aid_application
+      @legal_aid_application ||= submission.legal_aid_application
+    end
+
+    def submission_id
+      params.require(:id)
+    end
+  end
+end
--- a/app/controllers/admin/ccms_queues_controller.rb
+++ b/app/controllers/admin/ccms_queues_controller.rb
@@ -34,11 +34,8 @@ module Admin
     end
 
     def execute_submission_method(method)
-      result = submission.send(method)
-      status = result ? :notice : :error
-      message = status.eql?(:error) ? "Restarting submission #{submission_id} failed" : "#{method.to_s.humanize} submission #{submission_id}"
-      flash[status] = message
-      redirect_to action: :show
+      submission.public_send(method)
+      redirect_to admin_ccms_queue_path(submission.id), notice: "#{method.to_s.humanize.sub('!', '')} submission #{submission_id}"
     end
   end
 end
--- a/app/controllers/admin/ccms_queues_controller.rb
+++ b/app/controllers/admin/ccms_queues_controller.rb
@@ -12,21 +12,11 @@ module Admin
     end
 
     def reset_and_restart
-      result = submission.complete_restart! # return from service
-      status = result ? :notice : :error
-      core = "submission #{submission_id}"
-      message = status.eql?(:error) ? "Restarting #{core} failed" : "Reset and restarted #{core}"
-      flash[status] = message
-      redirect_to action: :show
+      execute_submission_method(:restart_from_beginning!)
     end
 
     def restart_current_submission
-      result = submission.restart_existing_submission!
-      status = result ? :notice : :error
-      core = "submission #{submission_id}"
-      message = status.eql?(:error) ? "Restarting #{core} failed" : "Restarted #{core}"
-      flash[status] = message
-      redirect_to action: :show
+      execute_submission_method(:restart_current_step!)
     end
 
   private
@@ -41,6 +31,14 @@ module Admin
 
     def submission_id
       params.require(:id)
+    end
+
+    def execute_submission_method(method)
+      result = submission.send(method)
+      status = result ? :notice : :error
+      message = status.eql?(:error) ? "Restarting submission #{submission_id} failed" : "#{method.to_s.humanize} submission #{submission_id}"
+      flash[status] = message
+      redirect_to action: :show
     end
   end
 end
--- a/app/models/ccms/submission.rb
+++ b/app/models/ccms/submission.rb
@@ -40,7 +40,7 @@ module CCMS
     def process_async!
       SubmissionProcessWorker.perform_async(id, aasm_state)
     end
-    alias_method :restart_existing_submission!, :process_async!
+    alias_method :restart_current_step!, :process_async!
 
     def sidekiq_running?
       return :running if Sidekiq::Workers.new.map(&:to_s).to_s.scan(id).count.positive?
@@ -49,7 +49,7 @@ module CCMS
       false
     end
 
-    def complete_restart!
+    def restart_from_beginning!
       update!(aasm_state: "initialised", case_poll_count: 0, applicant_poll_count: 0)
       process_async!
     end
--- a/app/models/ccms/submission.rb
+++ b/app/models/ccms/submission.rb
@@ -44,7 +44,7 @@ module CCMS
 
     def sidekiq_running?
       return :running if Sidekiq::Workers.new.map(&:to_s).to_s.scan(id).count.positive?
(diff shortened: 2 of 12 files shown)


jsugarman:
looking good @colinbruce. are you wanting to get it in or just UAT it for now

colinbruce:
> looking good @colinbruce. are you wanting to get it in or just UAT it for now
Either to be fair - it's an MVP 🤷‍♂️ 
It's kind of tricky to get a UAT instance in a state where jobs are in the right place...
I did wonder about merging it and a dev tries the next submission to fail and monitor the outcome?

jsugarman:
> > looking good @colinbruce. are you wanting to get it in or just UAT it for now
> 
> Either to be fair - it's an MVP/Proof of Concept 🤷‍♂️ It's kind of tricky to get a UAT instance in a state where jobs are in the right place... I did wonder about merging it and a dev tries the next submission to fail and monitor the outcome?

Makes sense to me, will give it a once over for approval

app/controllers/admin/ccms_queues_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      @in_progress = CCMS::Submission.where.not(aasm_state: %w[completed abandoned]).order(created_at: :desc)
+    end
+
+    def show
+      submission
+      legal_aid_application
+    end
+
+    def reset_and_restart
+      result = submission.complete_restart! # return from service
```
jsugarman, line 15: do we need the comment

app/controllers/admin/ccms_queues_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      status = result ? :notice : :error
+      core = "submission #{submission_id}"
+      message = status.eql?(:error) ? "Restarting #{core} failed" : "Reset and restarted #{core}"
+      flash[status] = message
```
jsugarman, line 19: nit - refactor, extract to method?

jsugarman (approved):
Looks great, should probably use govuk component going forward for some of that html would be my primary observation

app/views/admin/ccms_queues/show.html.erb
```diff
@@ -0,0 +1,127 @@
...
+      <%= link_to_accessible("CCMS Queue", admin_ccms_queues_path, class: "govuk-breadcrumbs__link") %>
+    </li>
+    <li class="govuk-breadcrumbs__list-item">
+      <%= @legal_aid_application.application_ref %>
+    </li>
+  </ol>
+</div>
+
+<h2 class="govuk-heading-l"><%= t(".submission") %></h2>
+<div class="govuk-summary-card">
```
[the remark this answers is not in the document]
colinbruce: Summary cards aren't available yet #4832 is still blocking the govuk-component update nneded for the summary card helper :(

app/models/ccms/submission.rb
```diff
@@ -40,5 +40,18 @@ def process!(options = {})
...
+    alias_method :restart_existing_submission!, :process_async!
+
+    def sidekiq_running?
+      return :running if Sidekiq::Workers.new.map(&:to_s).to_s.scan(id).count.positive?
+      return :in_retry if Sidekiq::RetrySet.new.map { |job| job.args.include?(id) }.any?(true)
+
+      false
+    end
+
+    def complete_restart!
```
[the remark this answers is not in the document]
colinbruce: #NamingThingsIsHard!

This was the third rename and it still wasn't that clear! 😆

config/routes.rb
```diff
@@ -65,6 +65,12 @@
       delete :destroy_all, on: :collection
     end
     resource :settings, only: %i[show update]
+    resources :ccms_queues, only: %i[index show] do
+      member do
+        get "reset_and_restart"
+        get "restart_current_submission"
+      end
```
cpjmcquillan, line 72: Did you consider having `reset_submission` and `restart_submission` (or similar) resourceful routes instead?

There would be a couple extra controllers, but it might help keep things simpler in the long run.

spec/requests/admin/ccms_queues_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+    let!(:ccms_submission) { create(:ccms_submission, :case_ref_obtained) }
+
+    before { sign_in admin_user }
+
+    it "calls submission.restart_current_submission" do
+      expect(get_restart_current_submission).to redirect_to(admin_ccms_queue_path(ccms_submission.id))
+      expect(flash[:notice]).to match(ccms_submission.id)
+    end
+  end
```
cpjmcquillan, line 82: This is something for wider discussion (outside the scope of this PR) but it's something I've been thinking about for a while and this is a nice example.

It feels like sometimes our tests are a means to an end (e.g 100% code coverage), but I think we could perhaps be writing more valuable specs to test the behaviour of our application.

In this instance, I think a system spec would be more helpful.

In my opinion, it would naturally
- exercise more functionality (e.g we don't need to explicitly test the index page)
- reflect more accurately the use case
- help us write more accessible code (since we are being more specific and intentional with how we interact with the page, and considering how functionality affects the user directly)

```ruby
require "rails_helper"

RSpec.describe "Admin users can manage the CCMS submission queue", type: :system do
  context "when restarting a submission" do
    it "restarts the submission at the current step" do
      admin_user = create(:admin_user)
      ccms_submission = create(:ccms_submission)
      sign_in admin_user

      visit admin_ccms_queues_path

      within "table" do  # I'd use a more specific helper 
        click_link ccms_submission.case_reference
      end
      
      click_button "Restart at current step"
      
      expect(current_path).to eq(admin_ccms_queue_path(ccms_submission.id))
      expect(page).to have_flash_notice("What the message was")
    end
  end
end
```

I'd be interested to hear what the rest of the team think.

cpjmcquillan (approved):
Awesome 🧙 

The only small concern I have is whether this will _actually_ be used by non-developers to manage the CCMS queue.

If it isn't used, then I'm worried we're adding some nice functionality (now ) but also complexity and code we need to maintain (forever).

If we think it will be used and we're confident it solves the problem of failed/stuck submissions being a developer issue —let's ship it!

app/models/ccms/submission.rb
```diff
@@ -40,5 +40,18 @@ def process!(options = {})
     def process_async!
       SubmissionProcessWorker.perform_async(id, aasm_state)
     end
+    alias_method :restart_current_step!, :process_async!
+
+    def sidekiq_running?
+      return :running if Sidekiq::Workers.new.map(&:to_s).to_s.scan(id).count.positive?
+      return :in_retry if Sidekiq::RetrySet.new.map { |job| job.args.include?(id) }.any?(true)
```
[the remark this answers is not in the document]
colinbruce: I had a lot of issues with the Sidekiq objects not exposing some of the enumerable functions and, to be honest, this probably started as a copy/paste/extension of the line above! 😄 
Updating

app/controllers/admin/ccms_queues_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      params.require(:id)
+    end
+
+    def execute_submission_method(method)
+      result = submission.send(method)
+      status = result ? :notice : :error
+      message = status.eql?(:error) ? "Restarting submission #{submission_id} failed" : "#{method.to_s.humanize} submission #{submission_id}"
+      flash[status] = message
+      redirect_to action: :show
+    end
```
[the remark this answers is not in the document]
colinbruce: Happy with `public_send` :+1:

slight tweak to your suggestion...
```ruby
if submission.public_send(method)
  redirect_to admin_ccms_queue_url(submission.id), notice: "#{method.to_s.humanize} submission #{submission_id}"
else
  redirect_to admin_ccms_queue_url(submission.id), flash: { error: "Restarting submission #{submission_id} failed" }
 end
```
Apparently `redirect_to action: :name` strips the flash notifications, whereas  using the url_helpers lets them through

Also, only notice and alert have [built-in helpers](https://api.rubyonrails.org/v7.0.4.2/classes/ActionDispatch/Flash/FlashHash.html) so the error has to be in a flash hash setter

config/routes.rb
colinbruce: It feels like that would trip _all_ the duplicate code warnings (🙄 @ sonarcloud) as the internals would be identical apart from the public_send method, then it would get extracted to a base controller and that seems like a whole chuck of extra stuff for a proof of concept that, as you say above, may not end up being used very much! 🤷‍♂️

app/controllers/admin/ccms_queues_controller.rb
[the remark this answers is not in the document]
cpjmcquillan: Are we able to use `_path` helpers instead of `_url`?

Also, if we wanted to we could add this line to `ApplicationController`

```ruby
add_flash_types :error
```

config/routes.rb
cpjmcquillan: I just ignore SonarCloud 😂.

I think this is fine for a proof of concept. If we end up needing to change things much, we would probably revisit then anyway 👍

app/controllers/admin/ccms_queues_controller.rb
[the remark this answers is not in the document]
colinbruce: Turns out - the error return was optimistic... `process_async!` will always return a value and then fail (and raise errors) elsewhere

Therefore it is currently being removed :+1:

spec/requests/admin/ccms_queues_controller_spec.rb
colinbruce: Added a link to this conversation to a future dev-time discussion

jsugarman: outside of scope for this PR as you mention but would be happy to try system tests out, in place of cucumber.

PR #5079, opened 2023-03-09
AP-3943: Fix for citizens locked out of applications

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3943)

When a citizen reaches the CYA page, the state of the application changes. If they close the browser and return later, the application tries to run the same starting code that involves resetting the state.  The state machine does not allow this and, even if it did, would mean they update their bank statements twice.

This checks for the change of state and routes them back to the CYA page

A ticket to investigate a proper save and return function for citizens will investigate a better long term solution 🤞

## Checklist

### DIFF
--- a/spec/requests/citizens/legal_aid_applications_controller_spec.rb
+++ b/spec/requests/citizens/legal_aid_applications_controller_spec.rb
@@ -38,6 +38,25 @@ RSpec.describe Citizens::LegalAidApplicationsController do
       expect(session["page_history_id"]).not_to be_nil
     end
 
+    context "when the applicant has reached the check your answers page" do
+      let(:legal_aid_application) do
+        create(
+          :application,
+          :with_applicant,
+          :with_non_passported_state_machine,
+          :checking_citizen_answers,
+          completed_at:,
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
👍

spec/requests/citizens/legal_aid_applications_controller_spec.rb
```diff
@@ -38,6 +38,25 @@
...
+    context "when the applicant has reached the check your answers page" do
+      let(:legal_aid_application) do
+        create(
+          :application,
+          :with_applicant,
+          :with_non_passported_state_machine,
+          :checking_citizen_answers,
+          completed_at:,
+          applicant: build(:applicant, first_name: "Test", last_name: "Applicant"),
+          provider: build(:provider, firm: build(:firm, name: "Test Firm")),
```
cpjmcquillan, line 50: Do we need to specify any of this data explicitly? 

I don't see any relationship between the test data and the test itself, but perhaps I'm missing something 🤔

cpjmcquillan (approved):

PR #5044, opened 2023-02-28
Add Partner means assessment feature flag

cpjmcquillan:
[Link to story](https://dsdmoj.atlassian.net/browse/AP-3816)

PR #5080, opened 2023-03-09
Refactor Settings

cpjmcquillan:
Before, new methods needed to be defined each time a new Setting was added to the application.

This leans on `method_missing` to delegate to the Setting record.

Adding a new Setting now only involves adding the attribute to `Setting::ATTRIBUTES`, defining labels and hint text, and running a migration to add the new field to the Settings table.

Compare #5081 to #5044 or #4577 to see how it might make your life easier.

### DIFF
--- a/config/locales/en/admin.yml
+++ b/config/locales/en/admin.yml
@@ -42,7 +42,10 @@ en:
           means_test_review_phase_one: Enable changes for MTR Phase 1
           partner_means_assessment: Enable Partner Means Assessment
         hints:
-          mock_true_layer_data: Select Yes and TrueLayer data will be replaced by mock data from %{bank_transaction_filename}
+          mock_true_layer_data: >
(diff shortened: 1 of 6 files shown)


config/locales/en/admin.yml
```diff
@@ -42,7 +42,10 @@ en:
           means_test_review_phase_one: Enable changes for MTR Phase 1
           partner_means_assessment: Enable Partner Means Assessment
         hints:
-          mock_true_layer_data: Select Yes and TrueLayer data will be replaced by mock data from %{bank_transaction_filename}
+          mock_true_layer_data: >
```
cpjmcquillan, line 45: Updating this hint text to be more generic seemed like a reasonable tradeoff to make to me —especially since `bank_transaction_filename` cannot be updated through the Settings UI.

PR #5083, opened 2023-03-10
Bump Devise to v4.9.0

cpjmcquillan:
The new version of Devise includes support for Hotwire/Turbo.

We're not using Turbo, so we shouldn't need to change anything.

[Changelog](https://github.com/heartcombo/devise/blob/main/CHANGELOG.md#490---2023-02-17)


cpjmcquillan:
Ready to UAT on staging

agoldstone93 (approved):

PR #5126, opened 2023-03-27
AP-3949: partners address

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3949)

This adds the screens required to record the partner's address:
* asking if they share an address with the client
* postcode input
* select from a list
* manual input

It refactors a number of controllers and forms to use shared base forms or concerns to prevent code duplication

## Checklist


spec/factories/partners.rb
```diff
@@ -26,5 +26,13 @@
...
       date_of_birth { 16.years.ago + 1.day }
     end
+
+    trait :with_address do
+      address_line_one { Faker::Address.building_number }
+      address_line_two { Faker::Address.street_address }
+      city { Faker::Address.city }
+      county { Faker::Address.city }
+      postcode { ["SW10 9LB", "W6 0LQ", "SW1A 1AA", "RG2 7PU", "BH22 7HR"].sample }
+    end
```
naseberry, line 36: Do we wish to test the `organisation`?

app/lib/address_handling.rb
```diff
@@ -0,0 +1,33 @@
+module AddressHandling
```
jsugarman, line 1: looks like this is a controller concern, so just thinking whether this would be better (more idiomatically) located in `app/controllers/concerns`?

app/views/providers/partners/address_lookups/show.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    ) do |form| %>
+
+  <%= page_template(
+        page_title: t(".page_title"),
+        template: :basic,
+        form:,
+      ) do %>
+
+    <%= form.govuk_fieldset legend: { size: "xl", tag: "h1", text: t(".heading") }  do %>
+      <div class="govuk-!-padding-2"></div>
```
naseberry, line 16: Best practice: the relevant spacing should already be applied by the design system, if additional spacing is required, best to add padding to the fieldset or margin to the text field.

app/views/providers/partners/address_selections/show.html.erb
```diff
@@ -0,0 +1,47 @@
...
+              suffix: "postcode",
+            ) %>
+      </p>
+      <%= form.hidden_field :postcode %>
+
+      <% @addresses.each_with_index do |address, index| %>
+        <%= hidden_field_tag "address_selection[list][]", address.to_json, id: "address_selection_list_#{index}" %>
+      <% end %>
+
+      <div class="govuk-!-padding-bottom-2"></div>
```
naseberry, line 35: Best practice: the relevant spacing should already be applied by the design system, if additional spacing is required, best to add it using [spacing overrides](https://design-system.service.gov.uk/styles/spacing/#overriding-spacing) on components

naseberry (approved):
Aside from my non-blocking comments, the address update and refactoring looks much better 👍

app/forms/partners/details_form.rb
```diff
@@ -1,5 +1,8 @@
 module Partners
   class DetailsForm < BaseForm
```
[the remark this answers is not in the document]
colinbruce: I'd be more inclined to remove basic from the applicant form! 😆 
In my head, the first devs thought:
> We'll start with the basic details and then add more pages...

and the other pages were named as address. 
> Oh, wait, that's it there are no more applicant details!

But I could be wrong! 🤷‍♂️

app/controllers/providers/address_selections_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Providers
   class AddressSelectionsController < ProviderBaseController
```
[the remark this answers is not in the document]
colinbruce: If you want to review another round of file names on this PR? 🤔 🙂 
I think namespacing helps reduce the complexity of the provider controllers folder but I'll raise a new ticket :+1:

app/views/providers/partners/address_lookups/show.html.erb
colinbruce: These were duplicated from the original views and I'm not sure why these were added in the first place!
image
If I remove the padding entirely...
image
I'm not sure I see a need for this, do you @naseberry?

PR #5129, opened 2023-03-28
Update GOVUK dependencies

cpjmcquillan:
This updates `govuk-components` and `govuk-formbuilder` to the latest release
candidate versions, which unblocks the `pagy` upgrade.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -94,10 +94,10 @@ gem "wdm", ">= 0.1.0" if Gem.win_platform?
 gem "grover"
 
 # DFE formbuilder
-gem "govuk_design_system_formbuilder"
+gem "govuk_design_system_formbuilder", "4.0.0a1"
 
 # ViewComponent
-gem "govuk-components"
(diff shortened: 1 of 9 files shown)


colinbruce:
I don't have (many) issues with RC's for the GOVUK stuff.  Does view component need to be that advanced too?  That seems a bigger risk for bad actors to slip code into 🤔

cpjmcquillan:
> I don't have (many) issues with RC's for the GOVUK stuff. Does view component need to be that advanced too? That seems a bigger risk for bad actors to slip code into 🤔

`govuk-components` is holding off on a release until ViewComponent 3.0 is released.

If we don't want to upgrade `view_component` yet we can hold out and use the earlier `govuk-component` alpha release 👍

naseberry:
Like Colin, I don't have any issues with RC's for the GOVUK dependencies, however, I would wait for the official release.

colinbruce:
replaced by non-rc releases in #5242

PR #5185, opened 2023-04-12
Repoint UAT to CCMS UAT

naseberry:
## What

[Repoint UAT to CCMS UAT](https://dsdmoj.atlassian.net/browse/AP-4026)

As part of the EBS migration work to verify submissions

## Checklist


jsugarman (approved):
👍 if this is what we should be using now and assuming it all works as expected

naseberry:
> 👍 if this is what we should be using now and assuming it all works as expected

New env creds provided by the CCMS team, confirmed that it is solid to use, plus QAT confirmed working connection during the cutover checks 👍

PR #5198, opened 2023-04-17
AP-3642 details component

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3642)

Use GOV.UK components library to render all instances of the details component
Also removed some code in the types_of_income and types_of_outgoings forms which was no longer used as the citizen is not asked questions about payment types anymore

***I rebased on @agoldstone93 's  AP-3991 summary list component branch so that's why there are so many changes showing at the moment***

## Checklist

### DIFF
--- a/app/views/citizens/consents/show.html.erb
+++ b/app/views/citizens/consents/show.html.erb
@@ -11,33 +11,23 @@
         <p class="govuk-body govuk-!-padding-bottom-2"><%= t(".inset_text") %></p>
       <% end %>
 
-      <details class="govuk-details govuk-!-margin-top-4" data-module="govuk-details">
-        <summary class="govuk-details__summary">
-          <span class="govuk-details__summary-text"
-            title="<%= t(".summary_heading") %>"
-            aria-label="<%= t(".summary_heading") %>">
-            <%= t(".summary_heading") %>
-          </span>
-        </summary>
+      <%= govuk_details(summary_text: t(".summary_heading")) do %>
+        <p><%= t(".able_to_see.heading") %></p>
 
-        <div class="govuk-details__text">
-          <p><%= t(".able_to_see.heading") %></p>
+        <ul class="govuk-list govuk-list--bullet">
+          <% t(".able_to_see.list").each_line do |item| %>
+            <li><%= item %></li>
+          <% end %>
+        </ul>
 
-          <ul class="govuk-list govuk-list--bullet">
-            <% t(".able_to_see.list").each_line do |item| %>
-              <li><%= item %></li>
-            <% end %>
-          </ul>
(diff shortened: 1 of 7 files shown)


app/views/citizens/consents/show.html.erb
```diff
@@ -11,33 +11,23 @@
...
-          <p><%= t(".unable_to_see.heading") %></p>
-
-          <ul class="govuk-list govuk-list--bullet">
-            <% t(".unable_to_see.list").each_line do |item| %>
-              <li><%= item %></li>
-            <% end %>
-          </ul>
-        </div>
-      </details>
+      <%= govuk_details(summary_text: t(".summary_heading")) do %>
```
naseberry, line 14: Layout: is the layout ok without the spacing override?

app/views/providers/open_banking_guidances/show.html.erb
```diff
@@ -12,17 +12,10 @@
...
-          <span class="govuk-details__summary-text">
-            <%= t(".details_summary") %>
-          </span>
-      </summary>
-      <div class="govuk-details__text govuk-!-padding-bottom-1">
-        <%= t(".details_content_html") %>
-        <%= list_from_translation_path(".open_banking_consents.show.details") %>
-      </div>
-    </details>
+    <%= govuk_details(summary_text: t(".details_summary")) do %>
```
naseberry, line 15: Layout: is the layout ok without the override class?

naseberry:
Browser support: As we older browsers are no longer supported, do we still need `app/javascript/src/details-polyfill.js`?

app/views/citizens/consents/show.html.erb
```diff
@@ -11,33 +11,23 @@
...
+        <p><%= t(".able_to_see.heading") %></p>
```
agoldstone93, line 15: Very minor - and it might not make any difference - but you could stick the `govuk-body` class into a paragraph. A quick scan of the codebase makes it look there's already a mix so it's entirely up to you

agoldstone93 (approved):
One very minor comment but otherwise looks good

naseberry:
Clean as you code (CYC): This seems redundant can it be removed https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/assets/stylesheets/providers/legal_aid_applications.scss#L53

app/views/providers/open_banking_guidances/show.html.erb
RoseSAK: I have added this back in! Thank you

app/views/citizens/consents/show.html.erb
RoseSAK: Yes, it's identical

PR #5208, opened 2023-04-19
Fix: Speed up CircleCI integration test steps

colinbruce:
## What

The `rails assets:precompile` in unit tests takes <10s but in integration tests it was taking 2m+

This PR resequences the integration test steps - precompile now runs before installing chromedrivers, etc

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -231,8 +231,6 @@ jobs:
     executor: test-executor
     steps:
     - checkout
-    - browser-tools/install-chrome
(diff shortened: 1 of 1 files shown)


colinbruce:
> Can't say i understand why (more assets to precompile due to browser tools?) but can see that it is fast to precompile so 👍

I _suspect_ it's because it added chrome drivers, after restroing the cache and that invalidated the cache so it had to pull everything down again. 🤷‍♂️

PR #5217, opened 2023-04-21
Fix: Update dependabot time

colinbruce:
## What

This is to prevent failures when servers are shutdown overnight

## Checklist


naseberry (approved):
👍  why the staggered scheduled time?

colinbruce:
> 👍 why the staggered scheduled time?

🤷‍♂️ Give circle a bit of chance to catch up! :) 
I have seen slow down when too many PR's are opened all at once, not that it should affect us at 9pm! 😄

PR #5227, opened 2023-04-25
[AP-3647] Use govuk panel component

skdejong:
## What

[Use panel component](https://dsdmoj.atlassian.net/browse/AP-3647)

Update the views to use the govuk_panel component

## Checklist


app/views/citizens/means_test_results/show.html.erb
```diff
@@ -1,22 +1,26 @@
...
-    <h1 class="govuk-panel__title govuk-!-padding-top-4">
-      <%= t(".completed_financial_assessment") %>
-    </h1>
-    <div class="govuk-panel__body">
-      <dl class="govuk-list inline-list">
-        <dt>
-          <strong><%= t(".name") %>:</strong>
+  <%= govuk_panel(
+    title_text: t(".completed_financial_assessment")) do %>
+    <dl class="govuk-summary-list govuk-summary-list--no-border govuk-list inline-list">
```
agoldstone93, line 4: There's a govuk_summary_list component, like the govuk_panel component. [Info here](https://govuk-components.netlify.app/components/summary-list/)

I've not tried using it within a panel but I'm assuming it'll work

app/views/providers/delegated_confirmation/index.html.erb
```diff
@@ -1,14 +1,9 @@
...
-  <div class="govuk-panel govuk-panel--confirmation">
-    <h1 class="govuk-panel__title">
-      <%= t ".page-heading" %>
-    </h1>
-    <div class="govuk-panel__body">
-    <%= t ".sub_title" %>
-    <br>
+  <%= govuk_panel(
+    title_text: t(".page-heading")) do %>
+    <%= t ".sub_title" %><br>
```
agoldstone93, line 4: A very minor thing, and it's not your code because it was already in here, but I think having `<%= t(".sub_title") %>` is a little more readable than without brackets and it seems to be what we're using most of the time elsewhere in the code. Might be worth changing just for consistency

skdejong: Thank you. I have fixed this now.

app/views/citizens/means_test_results/show.html.erb
skdejong: Thank you for the suggestion. Do you think it would be best if this was done as part of the summary_list component ticket(s)?

PR #5230, opened 2023-04-26
Remove details-polyfill and details styling

RoseSAK:
## What

We do not need details-polyfill as we are no longer supporting older browsers, we do not need the details white text class as this is no longer used

## Checklist


agoldstone93 (approved):

app/forms/providers/confirm_dwp_non_passported_form.rb
```diff
@@ -0,0 +1,9 @@
+module Providers
```
naseberry, line 1: Change doesn't seem related to the commit, was this meant to be committed?

agoldstone93 (approved):

PR #5238, opened 2023-04-27
Update SubmissionBuilder

colinbruce:
## What

Improve the submission_histpry creation, only build the object by default and save when the @save_result is set to true

## Checklist

### DIFF
--- a/app/services/cfe_civil/submission_builder.rb
+++ b/app/services/cfe_civil/submission_builder.rb
@@ -67,9 +67,10 @@ module CFECivil
     def query_cfe_service
       raw_response = post_request
       parsed_response = parse_json_response(raw_response.body)
+      history = build_submission_history(raw_response)
       if @save_result
+        history.save!
         submission.save!
-        history = write_submission_history(raw_response)
       end
 
       case raw_response.status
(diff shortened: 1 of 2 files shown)


app/services/cfe_civil/submission_builder.rb
```diff
@@ -94,7 +95,7 @@ def process_response
       submission.cfe_result = @response.body
       if @save_result
         submission.results_obtained!
-        write_submission_history(@response, "POST")
+        build_submission_history(@response, "POST").save!
```
naseberry, line 98: nit: as the second param `http_method` has the same value as the default, do we need to specify it here?

naseberry (approved):

app/services/cfe_civil/submission_builder.rb
colinbruce: Probably not, this is a copy paste from the current CFE submission where it does vary.
In this brave new world the _only_ submission will be a `POST` so the entire thing could be refactored! 
🤔

colinbruce: Have removed the superfluous call for now.  The *_submission_history pattern is in use in a number of places.  Extracting and centralising it may be more useful than slightly altering the pattern in one place :+1:

PR #5254, opened 2023-05-02
Update: CFECivil comparison

colinbruce:
## What

Tidy the cfe_civil urls to prevent double slashes (//) when constructing endpoints
Refactor the employment component
* only generate when there is an employment record with payments
* tidy the employment_income_payload method to remove breakpoint and use the preferred each_with_object pattern

## Checklist


jsugarman (approved):
👍

PR #5255, opened 2023-05-02
Update: CFE Comparison

colinbruce:
## What

Ensure data types are consistent
Populated data is returned with `.to_json` the empty version should be too

## Checklist

### DIFF
--- a/app/services/cfe_civil/components/employments.rb
+++ b/app/services/cfe_civil/components/employments.rb
@@ -4,9 +4,11 @@ module CFECivil
       delegate :employments, to: :legal_aid_application
 
       def call
-        return {} unless employment_and_payments?
-
-        { employment_income: employment_income_payload }.to_json
+        if employment_and_payments?
+          { employment_income: employment_income_payload }.to_json
+        else
+          {}.to_json
+        end
       end
 
     private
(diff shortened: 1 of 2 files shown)


spec/services/cfe_civil/components/employments_spec.rb
```diff
@@ -9,7 +9,7 @@
     before { create(:applicant, :not_employed) }
 
     it "renders the expected, empty hash" do
-      expect(call).to eq({})
+      expect(call).to eq({}.to_json)
```
jsugarman, line 12: could just use "{}" i believe, but either way

jsugarman (approved):
👍

app/services/cfe_civil/components/employments.rb
```diff
@@ -4,9 +4,11 @@ class Employments < BaseDataBlock
...
 
       def call
-        return {} unless employment_and_payments?
-
-        { employment_income: employment_income_payload }.to_json
+        if employment_and_payments?
+          { employment_income: employment_income_payload }.to_json
+        else
+          {}.to_json
+        end
```
colinbruce, line 11: I toyed with the idea of 
```ruby
      def call
        if employment_and_payments?
          { employment_income: employment_income_payload }
        else
          {}
        end
      end.to_json

```
but I couldn't handle the nausea it caused! 😄 

I also tried 
```ruby
      def call
        result = if employment_and_payments?
                   { employment_income: employment_income_payload }
                 else
                   {}
                 end
        result.to_json
      end
```
but it just seemed a bit noisy

spec/services/cfe_civil/components/employments_spec.rb
colinbruce: That seemed like it would be less clear in six months! 😄

PR #5272, opened 2023-05-09
Update CFECivil handling of outgoings

colinbruce:
## What

CFE-Civil requires a valid housing_cost_type when submitting transactions flagged as rent_or_mortgage, this feeds into a calculation for board and lodging that was never implemented into Apply and has been ignored by CFE Legacy.

This implements the correct logic and can be expanded if the correct board and lodging handling is implemented in the future

## Checklist


jsugarman (approved):
👍

jsugarman (approved):
👍

PR #5279, opened 2023-05-11
AP-3902: What you need to do next page (before financial means assessment)

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3902)

add what you need to do page
- add view and controller
- put content in translation file
- reroute accordingly
- update tests

To do
- [ ] Consider renaming to match pattern `capital_introductions`, maybe `means_financial_introductions`?

## Checklist

### DIFF
--- a/app/controllers/providers/about_financial_means_controller.rb
+++ b/app/controllers/providers/about_financial_means_controller.rb
@@ -2,8 +2,11 @@ module Providers
   class AboutFinancialMeansController < ProviderBaseController
     def show
       @name = full_name
-      @application = legal_aid_application
-      @applicant = legal_aid_application.applicant
+      @applicant = applicant
+    end
+
+    def update
+      continue_or_draft
     end
 
   private
--- a/app/controllers/providers/about_financial_means_controller.rb
+++ b/app/controllers/providers/about_financial_means_controller.rb
@@ -1,0 +1,15 @@
+module Providers
+  class AboutFinancialMeansController < ProviderBaseController
+    def show
+      @name = full_name
+      @application = legal_aid_application
+      @applicant = legal_aid_application.applicant
+    end
+
+  private
+
+    def full_name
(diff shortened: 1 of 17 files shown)


app/controllers/providers/about_financial_means_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      @applicant = applicant
+    end
+
+    def update
+      continue_or_draft
+    end
+
+  private
+
+    def full_name
```
colinbruce, line 14: There is already an `applicant.full_name` helper method on the model, you should be able to use that

app/views/providers/about_financial_means/show.html.erb
```diff
@@ -0,0 +1,24 @@
+<%= page_template page_title: t(".h1-heading") do %>
+
+<h2 class='govuk-heading-l'>
+  <% if @applicant.has_partner? %>
+    <%= t(".partner_heading", name: @name) %>
```
colinbruce, line 5: As above, using `@applicant.full_name` should work here

spec/requests/providers/about_financial_means_controller_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe Providers::AboutFinancialMeansController do
+  let(:legal_aid_application) { create(:legal_aid_application, :applicant_details_checked, applicant:) }
+  let(:provider) { legal_aid_application.provider }
+  let(:applicant) { create(:applicant) }
```
colinbruce, line 6: As the applicant is not explicitly used, you could use the `:with_applicant` trait on the legal_aid_application?
```suggestion
  let(:legal_aid_application) { create(:legal_aid_application, :applicant_details_checked, :with_applicant) }
  let(:provider) { legal_aid_application.provider }
```

colinbruce (approved):
👍

PR #5295, opened 2023-05-17
AP-4095: Update the existing HMRC playback pages for client

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4095)

- Updated to match designs in ticket
- Replaced table with govuk component table
- Added custom sass to remove bottom border on one table row. This is done in [here](https://govuk-components.netlify.app/components/table/#complex-tables) using `no-border-bottom` class but I couldn't get it work without custom css

To do:
- [x] Make sure caption is in right place when there's an error on the page

## Checklist


app/views/providers/means/full_employment_details/show.html.erb
```diff
@@ -2,6 +2,8 @@
               url: providers_legal_aid_application_means_full_employment_details_path(@legal_aid_application),
               method: :patch,
               local: true) do |form| %>
+
+  <span class="govuk-caption-xl"><%= t("providers.means.client_caption") %></span>
```
colinbruce, line 6: The pages I'm working on have the caption as `govuk-caption-l`
Just checking. If this page is different, we should check with the design team!

config/locales/en/providers.yml
```diff
@@ -819,12 +819,16 @@ en:
       employment_incomes:
         show:
           hmrc-information: HMRC have provided the information on this page.
-          inset_text: "We've used your client's National Insurance number to get this information from HMRC."
+          inset_text: We used your client’s National Insurance number to get this information from HMRC.
```
colinbruce: The apostrophe here is different to the one on line 828

colinbruce (approved):
A couple of minor observations, but otherwise LGTM :+1:

config/locales/en/providers.yml
agoldstone93: This is already fixed in a later commit. Good spot though!

app/views/providers/means/full_employment_details/show.html.erb
agoldstone93: Good spot. I've updated these to be `govuk-caption-l`

PR #5326, opened 2023-05-30
Update CFE::CompareSubmission

colinbruce:
## What

Some of the arrays of hashes were getting returned in a different order When there are multiple payments in a month there is no way of guaranteeing the sequence.  This checks for multiple values in a hash and searches for a string match on both...

It is hacky and nasty but hopefully shortlived until we switch over when this, and all the fixtures, can be deleted

## Checklist


RoseSAK (approved):
Looks good 👍

naseberry (approved):

PR #5361, opened 2023-06-09
CFE-Civil handling

colinbruce:
## What

There are two `level of help` options in CFE-Civil, `certificated` and `controlled`

The schema documentation says that `certificated` is the default and Apply should always be certificated but, just in case that ever changes, this ensures we send the correct value each time.

### Controlled definition from DFab:
Controlled = Provider does not apply for Legal Aid - they self grant.
Civil Apply's scope is only for certificated applications

## Checklist


jsugarman (approved):
👍

PR #5365, opened 2023-06-12
AP-4069: Create partner's financial assessment intro page

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4069)

Add partner about_financial_means page
- add page
- update flow
- add specs and cucumber feature tests

## Checklist

### DIFF
--- a/app/controllers/providers/partners/about_financial_means_controller.rb
+++ b/app/controllers/providers/partners/about_financial_means_controller.rb
@@ -1,0 +1,11 @@
+module Providers
+  module Partners
+    class AboutFinancialMeansController < ProviderBaseController
+      prefix_step_with :partner
+
+      def update
+        continue_or_draft
+      end
+    end
+  end
+end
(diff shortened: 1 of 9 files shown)


colinbruce (approved):
Code wise this looks okay but...
The position in the flow (while matching the ticket) seems a bit odd.  
The housing benefit page is still for the client, so we interrupt client questions for the partner, then return to the client's Housing Benefits?
Or does that page get updated with a "your client or their partner" version? 🤔 
We should get the partner figma and confirm :)

agoldstone93:
> Code wise this looks okay but... The position in the flow (while matching the ticket) seems a bit odd. The housing benefit page is still for the client, so we interrupt client questions for the partner, then return to the client's Housing Benefits? Or does that page get updated with a "your client or their partner" version? 🤔 We should get the partner figma and confirm :)

Confirmed that this is how it is structured in the figma file

PR #5372, opened 2023-06-13
CFE: Switch non prod environments to use CFE-Civil

colinbruce:
## What

[WIP] Try switching all non-prod environments to use CFE Civil

Benefits:
* increased confidence in the new CFEVersion

Risks:
* if something breaks we will need to fix asap
* Devs will need to update their .env files to point at the new URL

## Checklist


app/models/cfe/base_result.rb
```diff
@@ -152,7 +156,7 @@ def vehicle_assessed_amount
     ################################################################
 
     def pensioner_capital_disregard
-      if version_4? || version_5? || version_empty?
+      if version_4? || version_5? || version_6? || version_empty?
```
jsugarman, line 159: not sure of context here but I assume there is a theoretical version 3, and below, so would `version >= 4` version be neater, with maybe a rewrite of version method
```
def version
  result_hash.fetch(:version, "1").sub(/empty/i, "4").to_i
end
```

but as **nit** as test suite explosion and anxiety could ensue 😄

jsugarman (approved):
looks good to me 👍 

one comment on version method

colinbruce:
Did you remove my SonarCloud security hotspot @naseberry? 😍

colinbruce:
I will merge this on Monday 19th as I am on leave. This will allow monitoring and support if needed

PR #5394, opened 2023-06-19
Fix: Benefit breakdown view

colinbruce:
## What

The benefit changes were intended to roll out as a single hit, but some providers pause before submitting.  
We have seen, so far, at least one application that has a state benefit without a description.

This fix will allow that application to complete, generate a means report, and be submitted to CCMS.  it should allow applications in a similar state to proceed as well

## Checklist

### DIFF
--- a/app/views/shared/check_answers/_benefits_breakdown.html.erb
+++ b/app/views/shared/check_answers/_benefits_breakdown.html.erb
@@ -4,7 +4,7 @@
 <%= govuk_summary_list(actions: false) do |summary_list| %>
   <% @legal_aid_application.applicant.state_benefits.each do |benefit| %>
     <% summary_list.with_row do |row|
-         row.with_key { benefit.description }
+         row.with_key { benefit.description || "" }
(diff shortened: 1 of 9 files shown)


app/views/shared/check_answers/_benefits_breakdown.html.erb
```diff
@@ -4,7 +4,7 @@
 <%= govuk_summary_list(actions: false) do |summary_list| %>
   <% @legal_aid_application.applicant.state_benefits.each do |benefit| %>
     <% summary_list.with_row do |row|
-         row.with_key { benefit.description }
+         row.with_key { benefit.description || "" }
```
jsugarman, line 7: would safe operator work, but either way 
```suggestion
         row.with_key { benefit&.description }
```

colinbruce: The summary list helpers break if you try and pass `nil` that's why I ended up on the empty string :+1:

PR #5397, opened 2023-06-20
AP-4149: CCMS page for partner means assessment

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4149)

- Add Use CCMS page for partners when self employed or in armed forces
- Add partial for use across all Use CCMS pages
- Remove code for accessing the Use CCMS page when employed as this is no longer possible
- Rename `use_ccms_employed` to `use_ccms_employment` to reflect this change

## Checklist


spec/factories/legal_aid_applications.rb
```diff
@@ -47,6 +47,14 @@
     end
 
     trait :with_applicant_and_partner do
+      transient do
+        with_bank_accounts { 0 }
+      end
+      applicant { build(:applicant, :with_address, :with_partner, with_bank_accounts:) }
+      partner { build(:partner) }
```
colinbruce, line 58: Do you need `partner { build(:partner) }` I would have expected `:with_partner` to have created one

agoldstone93: Looks like `:with_partner`  just adds `has_partner { true }` to the applicant

RoseSAK (approved):
LGTM 💯

PR #5400, opened 2023-06-21
CFE: Comparison sheet filter update

colinbruce:
## What

The overnight comparison tool has been erroring and looping because of the number of API calls to google sheets 

I added a method of clearing the filter but it was being called on each line addition, in retrospect it should only be called once, at the start of the run, and this PR implements that.

This code is still quite inefficient, but should only be running for another two weeks until we cut over permanently

## Checklist


jsugarman (approved):
👍

PR #5401, opened 2023-06-21
Change anti-virus script and tests

jsugarman:
## What
Proposition to change the clamav install script, plus tests

- Change clamav install script to use TCP instead of LocalSocket configuration
- Add  more specific test for virus definition

This comes out of work on laa-assure-hmrc-data that identified false positive tests resulting from
failure to communicate with local clamav resulting in virsu_found? returning true, not because there
was a virus but because the scan returned an error.

The fix for assure was, firstly to ensure the tests fail for the right reason (a virus was found or not) and
secondly to amend the clamav install script to configure clamav to function as it does in a production like
environment - namely the App's clamav "client" communicates with the clamav "server" over TCP on localhost:3310.

**Pros IMO:**
1. it functions locally in a production like manner
2. the clamav config itself states that it binds to INADDR_ANY by default and that this is not wise
```
# TCP address.
# By default we bind to INADDR_ANY, probably not wise.
# Enable the following to provide some degree of protection
# from the outside world. This option can be specified multiple
# times if you want to listen on multiple IPs. IPv6 is now supported.
# Default: no
```
3. Use of LocalSocket seems to fail "sometimes"

**Cons**
 I am not sure whether using TCP, even on `localhost:3310` only, is less secure than using `LocalSocket` for our machines?

## Testing (locally)
- run the test, it may succeed or fail depending on your current clamav installation
```shell
rspec -fd spec/services/malware_scanner_spec.rb
```
- [optional] try disabling LocalSocket in config and run tests again, it should fail
```shell
# check your TCPSocket is closed
cat $(brew --prefix)/etc/clamav/clamd.conf | grep TCP
=>
# TCP port address.
#TCPSocket 3310
# TCP address.
#TCPAddr localhost

# open config in editor,
vim  $(brew --prefix)/etc/clamav/clamd.conf
# comment out this line 
...
LocalSocket /usr/local/var/run/clamav/clamd.socket
...

# restart (stop and start)
sudo launchctl unload /Library/LaunchDaemons/clamav.clamd.plist
sudo launchctl load /Library/LaunchDaemons/clamav.clamd.plist
```
- run the script
```shell
bin/install_clamav_on_mac
```
- run tests again and they should pass

## Checklist

### DIFF
--- a/bin/install_clamav_on_mac
+++ b/bin/install_clamav_on_mac
@@ -5,6 +5,7 @@
 
 CONFIG_FOLDER=$(brew --prefix)/etc/clamav
 CLAMD_CONFIG_FILE=$CONFIG_FOLDER/clamd.conf
+CLAMD_TCP_SOCKET=3310
 FRESHCLAM_CONFIG_FILE=$CONFIG_FOLDER/freshclam.conf
 DB_FOLDER=$(brew --prefix)/var/lib/clamav
 TMP_FOLDER=$DB_FOLDER/tmp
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
LGTM :+1:

PR #5406, opened 2023-06-22
Fix: Update CFE Comparison

colinbruce:
## What

The handling of employment has changed in CFE Civil this has been accepted and we no longer send empty employment blocks to the new version unless actual pay is included.

In this instance, there were two empty employment
blocks so Legacy adds multiple employment remarks
but Civil has no employments and adds no remarks

This adds an override handler so it will add an
observation to the comparison spreadsheet for
review and, ultimately, to be ignored

## Checklist


jsugarman (approved):
👍

PR #5417, opened 2023-06-27
AP-4263: Switch production CFE to use civil

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4263)

This PR removes the calls to CFE::SubmissionRouter and directs _all_ calls to CFE to use CFECivil

It removes the optional save parameter from CFECivil::SubmissionBuilder, as we should always save the result, and amend the calling methods and tests.
It also duplicates the ObtainStateBenefitTypesService class from CFELegacy as this was overlooked in the original implementation but is a much simpler set of values to return

## Checklist


RoseSAK (approved):
LGTM! 💯

app/services/cfe_civil/base_service.rb
```diff
@@ -0,0 +1,32 @@
...
+    def headers
+      {
+        "Content-Type" => "application/json",
+        "Accept" => "application/json;version=#{cfe_version}",
+        "User-Agent" => "CivilApply/#{HEADER_VERSION} #{HostEnv.environment.to_s || 'missing'}",
+      }
+    end
+
+    def cfe_version
+      "6"
```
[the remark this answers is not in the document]
colinbruce: I just copied from the existing versions but you are right, it _could_ be.  I will update :)

app/services/cfe/compare_results.rb
```diff
@@ -17,7 +17,7 @@ def call
   private
 
     def compare_values_for(legal_aid_application)
-      v6 = CFECivil::SubmissionBuilder.new(legal_aid_application, save_result: false)
+      v6 = CFECivil::SubmissionBuilder.new(legal_aid_application)
       v6.call
```
jsugarman, line 21: **nit** and assuming it works fine to use class method direct

```suggestion
      CFECivil::SubmissionBuilder.call(legal_aid_application)
```

app/services/cfe/submission_router.rb
```diff
@@ -12,7 +12,7 @@ def call
       if HostEnv.production?
         CFE::SubmissionManager.call(@legal_aid_application.id)
       else
-        CFECivil::SubmissionBuilder.call(@legal_aid_application, save_result: true)
+        CFECivil::SubmissionBuilder.call(@legal_aid_application)
```
jsugarman, line 15: lack of context info on my part so feel free to ignore, but should this not be doing the same thing on any host now?

jsugarman (approved):
👍 question on submission router but i assume its ok and approving coz i lack context knowledge on that aspect and otherwise looks good

app/services/cfe/compare_results.rb
colinbruce: 1) This is also a stub that will no longer be called after cutover and is due for deletion in the very near future and;
2) No, Because in the next line we pass the v6 object, containing the response to the comparison tool. Making the call object return the CFE Result invalidated it's use in the main flow!

PR #5418, opened 2023-06-27
Remove db backup cronjobs

naseberry:
## What

[Remove hourly db backup cronjobs](https://dsdmoj.atlassian.net/browse/AP-4061)

Remove the cron jobs involved in storing and clearing the db backups

## Why

Amazon, in theory, allows a point-in-time recovery making the backup files a resource drain that we are unlikely to need

## TO-DO

- [x] Manually run the backup cleaner to remove any legacy backups

## Checklist


kmahern:
Do we need to first check with CP that we able to carry out point in time restores before merging this pr? The ticket says that AWS in theory allows point-in-time restores but my understanding is that this needs to be configured. I have looked in the CP user guide but this only mentions one overnight backup.

colinbruce (approved):
LGTM :+1:

_Maybe_ we could schedule a point-in-time recovery, the same as the original DR test?

naseberry:
> Do we need to first check with CP that we able to carry out point in time restores before merging this pr? The ticket says that AWS in theory allows point-in-time restores but my understanding is that this needs to be configured. I have looked in the CP user guide but this only mentions one overnight backup.

Quoting CP - "Point-in-time recovery for RDS relies on snapshots, either automated or manual. You can do point-in-time recovery to any point in time within the backup retention period; which is (by default) 7 days."

So yes, PITR is possible with CP's automated snapshots (although we may require CP support)

jsugarman:
> Do we need to first check with CP that we able to carry out point in time restores before merging this pr? The ticket says that AWS in theory allows point-in-time restores but my understanding is that this needs to be configured. I have looked in the CP user guide but this only mentions one overnight backup.

It does look and sound like a combination of an automated daily backup plus point in time recovery (PITR) can allow us to restore an instance of the DB from up to 7 days ago until up to ~5 minutes ago.

I have used the cli to confirm LatestRestorableTime is showing as ~5 minutes ago for Apply production RDS instance. Documentation confirms that a [daily backup is taken](https://aws.amazon.com/rds/features/backup/#:~:text=Amazon%20RDS%20automatically%20creates%20a,known%20as%20the%20backup%20window.) so it sounds like we do not need to take a manual snapshot. But when was the last time a DR scenario was test?

PR #5421, opened 2023-06-28
Remove use of LOCAL_CLAMAV

jsugarman:
## What
Remove use of LOCAL_CLAMAV

The bin/install_clamav_on_mac was altered to install and
configure local claim to use TCP instead of Local[unix]Socket
This means local clamav server and client should work in the same way
as the production clamav server client - namely itwill use the
`clamd.container.conf` settings for TCP.

Therefore we do not need, and should no longer use, the LOCAL_CLAMAV var
and its assoicated command which uses `--fdpass` (for improved speed?!).

```
--fdpass
Pass the file descriptor permissions to clamd. This is useful if clamd is
running as a different user as it is faster than streaming the file to clamd.
Only available if connected to clamd via local(unix) socket.
```

## Checklist

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -39,8 +39,6 @@ ADMIN_ALLOW_RESET=<true or false>
 ADMIN_PASSWORD=<password>
 ADMIN_SHOW_FORM=<true or false>
 
-LOCAL_CLAMAV=true
-
 # Connection to CCMS SOA
 CCMS_SOA_AWS_GATEWAY_API_KEY=<aws-gateway-api-key-for-environment>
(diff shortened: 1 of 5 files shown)


skdejong (approved):
I pulled this branch, tried it locally and it works.

jsugarman:
I have sanity tested this on UAT branch and malware still triggers form errors as expected.

Screenshot 2023-06-28 at 11 35 04

colinbruce (approved):
Like @skdejong, pulled locally and confirmed it works as expected :+1:

PR #5438, opened 2023-07-03
Fix: Update DB export

colinbruce:
## What

The download anonymised DB export often failed with an `EOF` issue
This allows the `cp` command to retry
It has been tested locally and allows the download to restart multiple times and still allow the file to completely download and restore

## Checklist

### DIFF
--- a/scripts/db_export.sh
+++ b/scripts/db_export.sh
@@ -25,7 +25,7 @@ esac
 echo "Finding pod for $environment"
 echo "Connecting to $POD, anonymizing, compressing and exporting DB"
 kubectl -n laa-apply-for-legalaid-$environment -c web exec "$POD" -- rake db:export
-kubectl -n laa-apply-for-legalaid-$environment cp -c web "laa-apply-for-legalaid-$environment/$POD:tmp/temp.sql.gz" "./tmp/$environment.anon.sql.gz"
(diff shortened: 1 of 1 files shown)


jsugarman (approved):
👍

colinbruce:
> 👍 - apparently -1 will result in infinite retries - nothing like an infinite loop

The example suggested 999 so it would, eventually, timeout! 
In practice, what do you think... a smaller number like... 5(?) or infinite

jsugarman:
> > 👍 - apparently -1 will result in infinite retries - nothing like an infinite loop
> 
> The example suggested 999 so it would, eventually, timeout! In practice, what do you think... a smaller number like... 5(?) or infinite

i think a finite number makes more sense/is safer - maybe 5 to 50 🤷 , but lacking context [as usual].

colinbruce:
> > > 👍 - apparently -1 will result in infinite retries - nothing like an infinite loop
> > 
> > 
> > The example suggested 999 so it would, eventually, timeout! In practice, what do you think... a smaller number like... 5(?) or infinite
> 
> i think a finite number makes more sense/is safer - maybe 5 to 50 🤷 , but lacking context [as usual].

It's always going to be a bit `try it and see what happens`! I have set it to 10 for now, that should allow the DB size to increase and still have enough retries to complete the download.

If that starts failing we can increase it 👍

PR #5442, opened 2023-07-04
Update rubocop.govuk

colinbruce:
## What

Update the rubocop govuk gem and then step through each update in turn addressing issues that arose

Three exemptions added to `.rubocop.yml`:

* FactoryBot/FactoryAssociationWithStrategy, exclude spec/factories/legal_aid_applications.rb
  This is because there are too many traits that create objects that loop back

* RSpec/IndexedLet, set the max number of spec variables to 2

  This will allow up to 2 number let variables e.g.
  ```ruby
      let!(:transaction1) { create(:bank_transaction, :benefits, happened_at: 10.days.ago) }
      let!(:transaction2) { create(:bank_transaction, :benefits, happened_at: 40.days.ago) }
  ```
  But trying to create `transaction3` will raise an error

  The aim is to slowly reduce the number over time to 1 as we phase out the use of large numbered blocks

* Exclude spec/services/cfe/* from RSpec/IndexedLet as the CFE classic integration is due to be deleted soon

## Checklist


spec/services/true_layer/bank_data_import_service_spec.rb
```diff
@@ -34,7 +34,7 @@
           .to change(BankAccount, :count).by(mock_data[:accounts].count)
 
         expect(bank_provider.bank_accounts.pluck(:true_layer_id))
-          .to contain_exactly(*mock_data[:accounts].pluck(:account_id))
+          .to match_array(mock_data[:accounts].pluck(:account_id))
```
jsugarman, line 37: Seems contrary to the MatchArray cop previously fixed but does not matter unless the test needs to ensure no additional array elements exist (this would pass but contains_exactly would not i believe)

jsugarman (approved):
LGTM 👍 - one comment re: contrary cops MatchArray versus ContainExactly. No blockers

spec/services/true_layer/bank_data_import_service_spec.rb
colinbruce: Thanks for the detail check. I think, after some digging, that it will be okay. MatchArray seems to be an exact match running the following
`> fake_array = mock_data[:accounts].pluck(:account_id)`

```
=> ["b95b007f11d9cc230d939f2bad508823", "53f74085702ee77c4ff46248446a2351"]
```

`> fake_array << "Test"`

```
=> ["b95b007f11d9cc230d939f2bad508823", "53f74085702ee77c4ff46248446a2351", "Test"]
```
`> expect(bank_provider.bank_accounts.pluck(:true_layer_id)).to match_array(fake_array)`
```RSpec::Expectations::ExpectationNotMetError:            Expected ["53f74085702ee77c4ff46248446a2351", "b95b007f11d9cc230d939f2bad508823"]
to match array with "b95b007f11d9cc230d939f2bad508823", "53f74085702ee77c4ff46248446a2351" and "Test"
```

spec/models/aggregated_cash_income_spec.rb
```diff
@@ -85,12 +85,12 @@
     end
 
     context "when cash income transaction records exist" do
-      let!(:pension1) { create(:cash_transaction, :credit_month1, legal_aid_application: application, transaction_type: pension) }
-      let!(:pension2) { create(:cash_transaction, :credit_month2, legal_aid_application: application, transaction_type: pension) }
-      let!(:pension3) { create(:cash_transaction, :credit_month3, legal_aid_application: application, transaction_type: pension) }
-      let!(:maintenance_in1) { create(:cash_transaction, :credit_month1, legal_aid_application: application, transaction_type: maintenance_in) }
-      let!(:maintenance_in2) { create(:cash_transaction, :credit_month2, legal_aid_application: application, transaction_type: maintenance_in) }
-      let!(:maintenance_in3) { create(:cash_transaction, :credit_month3, legal_aid_application: application, transaction_type: maintenance_in) }
+      let!(:pension_first_month) { create(:cash_transaction, :credit_month1, legal_aid_application: application, transaction_type: pension) }
```
naseberry, line 88: You actually renamed them all 👏 🏆

naseberry (approved):

spec/models/aggregated_cash_income_spec.rb
colinbruce: Only the ones where the numbers are greater than 2! :)
There are still a lot of variable1, variable2 pairs!

colinbruce: There are still 82 offences, so 41 occurrences of numbered variables
Happy to discuss if the team are happy with more 1&2 being created but 1,2&3 will get blocked?

PR #5443, opened 2023-07-04
AP-3805 Part 2 add owner to employments

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3805)

Update Employment model and DB table to have owner as a polymorphic column. 
The owner will be either the applicant or partner. At the moment the owner is just the applicant.
Added rake task to set owner of all existing employments to the applicant.
Updated rspecs and feature tests to create the employment object with an owner

NB: At the moment all employments have to be explicitly created in each setup step for rspec/features by passing the owner (in all cases the applicant right now). There is probably a more efficient way of doing this but I couldn't figure it out. It would be ideal if in the employment factory the owner could be created/added there, but the three-way dependency between LegalAidApplication, Employment, and Applicant/Partner makes it kind of awkward and circular

## Checklist

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -17,6 +17,7 @@ class Applicant < ApplicationRecord
   has_many :bank_transactions, through: :bank_accounts
   has_many :regular_transactions, as: :owner
   has_many :hmrc_responses, class_name: "HMRC::Response", as: :owner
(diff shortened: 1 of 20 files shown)


colinbruce (approved):
LGTM :+1:

PR #5458, opened 2023-07-07
Fix: Add missing close dl tag

colinbruce:
## What

This keeps being flagged by my IDE and is annoying! :)

## Checklist


jsugarman (approved):
👍 can't tell if that is indented to the right level though. it looks off by one

colinbruce:
> 👍 can't tell if that is indented to the right level though. it looks off by one

No, it's on the left margin and matches the opening tag 👍

PR #5479, opened 2023-07-12
AP-4321: Replace references to unused/maintained clamav images

jsugarman:
## What
Replace laa-apply-for-legal-aid/clamav with ministryofjustice/hmpps-clamav for docker-compose

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4321)

Remove reference to old, archived and unused clamav image.

The old clamav is archived and not used any longer, plus its ECR repo is to be deleted. This
updates to use the same exact image and commit as production.

I also:

1. updated postgres to production version
2. changed to use the docker/run script as bin/run no longer exists

see cloud-platform-environments PR to [remove clamav ECR repo](https://github.com/ministryofjustice/cloud-platform-environments/pull/14563)

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,18 +2,18 @@ version: '3'
 
 services:
   postgres:
-    image: postgres:10.5
+    image: postgres:11.6
     volumes:
     - pgdata:/var/lib/postgresql/data
   clamav:
-    image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-apply-for-legal-aid/clamav:1.0.2
+    image: ghcr.io/ministryofjustice/hmpps-clamav:sha-0ae98e9
     ports:
     - "3310:3310"
-  api:
+  web:
     environment:
     - DB_HOST=postgres
     build: .
-    command: bin/run
+    command: docker/run
     ports:
     - "3002:3002"
     depends_on:


kmahern (approved):
LGTM

docker-compose.yml
```diff
@@ -2,18 +2,18 @@ version: '3'
 
 services:
   postgres:
-    image: postgres:10.5
+    image: postgres:11.6
     volumes:
     - pgdata:/var/lib/postgresql/data
   clamav:
-    image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/laa-apply-for-legal-aid/clamav:1.0.2
+    image: ghcr.io/ministryofjustice/hmpps-clamav:sha-0ae98e9
```
colinbruce, line 9: This may not be the best SHA
We locked it at this version when HMPPS broke latest
See the comment on the Jira ticket

jsugarman: This is just to use the same as production currently uses. The primary aim is to simply remove the reference to the old image which does/will not work at all.

If we are to bump the production version as a PR I would hope 🤞 we also bump this inline for testing purposes, but IMO that is beyond the scope of this. 

ps: there are a few other issues with docker compose I am trying to resolve locally as well, like DB timeout, clamav port binding conflicts (with local clamav install) 😬

colinbruce: Does anyone use docker-compose for Apply would be my next question...
I didn't realise we had one and now it should spin up all the microservices too shouldn't it?
The most recent change was in 2019, most of it is from 2018!

jsugarman: indeed, we could just get rid of. but it can be useful occasionally for testing things like a clamav bumps and docker image changes. Other services could be added.

i do not mind either way?!

colinbruce: Still approve :+1: 
Merge this and the team can decide on updates later 👍

PR #5480, opened 2023-07-12
Bump hmpps-clamav to latest commit shah

jsugarman:
## What
Bump anti virus softare, [hmpps-clamav](https://github.com/ministryofjustice/hmpps-utility-container-images/pkgs/container/hmpps-clamav), to latest version

From sha-0ae98e9 to sha-ae9a953

## Test
I have UAT'd an upload of malware with successful blocking.

## Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
     volumes:
     - pgdata:/var/lib/postgresql/data
   clamav:
-    image: ghcr.io/ministryofjustice/hmpps-clamav:sha-0ae98e9
+    image: ghcr.io/ministryofjustice/hmpps-clamav:sha-ae9a953
     ports:
     - "3310:3310"
   web:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     volumes:
     - pgdata:/var/lib/postgresql/data
   clamav:
-    image: ghcr.io/ministryofjustice/hmpps-clamav:sha-0ae98e9
+    image: ghcr.io/ministryofjustice/hmpps-clamav:sha-ae9a953
```
colinbruce, line 9: Maybe we should revert to using `:latest`
Historically we used it, but HMPPS broke their build pipeline and took a few days to recover it.  We locked at a known SHA so we could still run tests and deploy.
Using latest pros:
* up-to-date build
Cons
* fragile if build pipeline builds again

Using locked sha pros
* known good build

Cons
* quickly out of date
* requires 'regular' updates

colinbruce: 🤦‍♂️ Just read that jira ticket!  This looks fine 🙂

kmahern (approved):
I guess we should also do this on laa-assure-hmrc-data

PR #5485, opened 2023-07-13
AP-3806: Implement HMRC playback for partner

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3806)

- add partner_status_analyzer
- update status_analyzer to differentiate
- add flow logic to partner_bank_statements
- copy LegalAidApplication methods to partner, applicant
- add and update specs

To do:
- [x] Make sure all tests are working
- [x] Make sure pages are redirecting to right place
- [x] Create ticket to handle extra employment info for partner

## Checklist

### DIFF
--- a/app/controllers/providers/partners/unexpected_employment_incomes_controller.rb
+++ b/app/controllers/providers/partners/unexpected_employment_incomes_controller.rb
@@ -1,28 +1,28 @@
 module Providers
-  module Means
+  module Partners
     class UnexpectedEmploymentIncomesController < ProviderBaseController
       before_action :employments_and_payments
 
       def show
-        @applicant = applicant
+        @partner = partner
         @form = LegalAidApplications::UnexpectedEmploymentIncomeForm.new(model: legal_aid_application)
       end
 
       def update
-        @applicant = applicant
+        @partner = partner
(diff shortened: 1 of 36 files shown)


app/controllers/providers/partners/unexpected_employment_incomes_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    class UnexpectedEmploymentIncomesController < ProviderBaseController
+      before_action :employments_and_payments
+
+      def show
+        @partner = partner
+        @form = LegalAidApplications::UnexpectedEmploymentIncomeForm.new(model: legal_aid_application)
+      end
+
+      def update
+        @partner = partner
```
RoseSAK, line 12: I think this could be simplified by just called the `partner` method which you've defined below, it seems like all that's happening here is that you're assigning the variable again, but that's already happening in the method (unless I've misunderstood!)
```suggestion
        partner
```

RoseSAK (approved):
Looks good to me! Only had one minor comment. I like the refactoring/extending you've done on the employment income pages 😍 🎉 💯

PR #5486, opened 2023-07-13
Add opponent organisations feature flag

skdejong:
## What

[Add organisation feature flag to Apply](https://dsdmoj.atlassian.net/browse/AP-4278)

## Checklist


nly
```diff
@@ -0,0 +1,4 @@
+Inspecting 1 file
+.
+
+0 files inspected, no offenses detected
```
naseberry, line 4: Should this file be ignored?

jsugarman: looks like output from rubocop or other linter/tool. will need to be removed I think

spec/models/setting_spec.rb
```diff
@@ -54,6 +54,7 @@
       expect(described_class.alert_via_sentry?).to be true
```
kmahern, line 54: @skdejong I think you could add a test for opponent_organisations? to the "when there is no setting record" and "when a record already exists with non-default values" contexts in this spec - give me a shout if you want to pair 🙂

spec/models/setting_spec.rb
```diff
@@ -54,6 +54,7 @@
...
       expect(described_class.means_test_review_phase_one?).to be false
       expect(described_class.partner_means_assessment?).to be false
+      expect(described_class.opponent_organisations?).to be false
```
jsugarman, line 57: do we also need tests above - for example, wherever `partner_means_assessment` is tested we should have equivalent test for this flag I think

jsugarman: I would follow along with changes from this PR to make sure everything is covered in terms of specs 😄 
https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5044/files

jsugarman (commented):
See comments regarding specs plus do we need to amend `lib/tasks/setup_db_for_testing.rake`? I am not sure whether we even need that task anymore though

nly
skdejong: I have deleted this file now. Thanks for pointing it out.

spec/models/setting_spec.rb
skdejong: Thank you, Katharine. I have added these tests now.

spec/models/setting_spec.rb
skdejong: Thank you, Joel. I have added these tests now.

jsugarman (approved):
LGTM 👍

PR #5487, opened 2023-07-13
Remove docker compose

jsugarman:
Remove docker compose

Agreed with team devs that this can be removed because
nobody uses it and it is not runnable currently due to (at least):

- postgres DB timeout errors (possibly due to env var setup
or db config issues)
- clamav port binding confllicts (with
local clamav binding to the same port)
- out of date or non-existent api service references via env vars

_PS: I could not find any documentation to update/remove in the repo._

## What

## Checklist


colinbruce (approved):
:boom: :+1:

PR #5511, opened 2023-07-20
AP-4284: Add db structure for polymorphic opponent individuals

jsugarman:
## What
Add database structures necessary to migrate opponents to polymorphic individual opponents

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4284)

-  Create individuals table
-  Add opposable to opponents table, but null values allowed

First part of database changes required to migrate opponents to being a polymorphic
relationship, so that we can extend opponents to include organisations

## Checklist


kmahern (approved):
👍

PR #5512, opened 2023-07-20
AP-4343: warn and redirect under 16s to CCMS post MTR phase one

jsugarman:
## What
Warn and block under 16s post MTR phase one too

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4343)

Under 16s will not be able to use the service even after MTR phase
one goes live. This is because they must provide linked cases information
as a minimum and this is not yet implemented. Instead we need to redirect
their providers to CCMS.

## Checklist

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -49,8 +49,7 @@ class Applicant < ApplicationRecord
   end
 
   def under_16_blocked?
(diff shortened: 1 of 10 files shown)


naseberry (approved):
:simples

PR #5517, opened 2023-07-24
AP-4264: Remove legacy CFE code

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4264)

Now that we have migrated fully to using CFE Civil we can start to de-commission the CFE Legacy code base.

This PR removes the code from `app/services/cfe` that called the legacy version of CFE.
It also removes any tests, fixtures, and cassettes that were used by the legacy code calls.
Finally, it removes the sidekiq worker job that was running the overnight comparisons.  
It had been stopped from running when we cut over the calls, so this code was superfluous.

## Checklist


agoldstone93 (approved):

naseberry:
nit: minor reference here: https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/ap-4264/remove-cfe-legacy-code/.rubocop.yml#L37

PR #5518, opened 2023-07-24
AP-4302: Add page for opponent type question

kmahern:
## What

[Add “Is the opponent an individual or an organisation?” page to Apply](https://dsdmoj.atlassian.net/browse/AP-4302)

## Checklist

### DIFF
--- a/app/services/flow/flows/provider_merits.rb
+++ b/app/services/flow/flows/provider_merits.rb
@@ -78,9 +78,6 @@ module Flow
         opponent_types: {
           path: ->(application) { urls.providers_legal_aid_application_opponent_type_path(application) },
           forward: :opponents_names,
-          # TODO: currently this always forwards to the individual opponent page
-          # as part of AP-4301 it should forward to the organisation opponent page
-          # if the user selects organisation.
         },
         has_other_opponents: {
           path: ->(application) { urls.providers_legal_aid_application_has_other_opponent_path(application) },
--- a/app/services/flow/flows/provider_merits.rb
+++ b/app/services/flow/flows/provider_merits.rb
@@ -86,7 +86,7 @@ module Flow
           path: ->(application) { urls.providers_legal_aid_application_has_other_opponent_path(application) },
           forward: lambda { |application, has_other_opponent|
             if has_other_opponent
-              Setting.opponent_organisations? ? urls.providers_legal_aid_application_opponent_type_path(application) : :opponents_names
+              Setting.opponent_organisations? ? :opponent_types : :opponents_names
             else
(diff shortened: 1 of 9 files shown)


app/controllers/providers/application_merits_task/opponents_types_controller.rb
```diff
@@ -0,0 +1,32 @@
+module Providers
+  module ApplicationMeritsTask
+    class OpponentsTypesController < ProviderBaseController
```
jsugarman, line 3: **nit** why are we using a double plural here? would not `OpponentTypesController` make more sense. I know its inline with some other opponent related controllers but they too seem weird

app/services/flow/flows/provider_merits.rb
```diff
@@ -71,15 +71,22 @@ class ProviderMerits < FlowSteps
...
             else
-              urls.new_providers_legal_aid_application_opponents_name_path(application)
+              Setting.opponent_organisations? ? urls.providers_legal_aid_application_opponents_types_path(application) : urls.new_providers_legal_aid_application_opponents_name_path(application)
             end
           end,
         },
+        opponents_types: {
+          path: ->(application) { urls.providers_legal_aid_application_opponents_types_path(application) },
+          forward: :opponents_names,
+          # TODO: currently this always forwards to the individual opponent page
```
jsugarman, line 81: I don't think we need this todo. It will necessarily need to change as part of the ticket to add the organisation page I'd say

jsugarman (approved):
Looks good. One issue on controller naming, which i also faced - and changed to use singular for `OpponentWhateversController` in AP-4284.

I am not sure about all the TODOs. I think the one in the flow is not needed at least but would be tempted to remove the others too.

app/controllers/providers/application_merits_task/opponents_types_controller.rb
jsugarman: I renamed the `OpponentsNamesController` [here](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5519/files#diff-3c6b193c4beb4d633c7dfa6c7f82b748737cc173f843207d66cbeaded946c877R3)

kmahern: 👍 I had added a commit to rename the controller to be inline with other opponent related controllers but have now dropped this commit so that it is opponent_types_controller.rb

app/services/flow/flows/provider_merits.rb
kmahern: 👍 removed

PR #5521, opened 2023-07-24
Fix: DB Export clear production flag on restore

colinbruce:
## What

The suggested script shown to the user restores the file to the local DB but the internal metadata leaves the DB flagged as production.  This can cause issues if the user wants to reset or rebuild their local DB.

This PR adds the `&& rails db:environment:set RAILS_ENV=development` command to the psql restore line.  This ensures rails is set to be in development mode when the database is successfully restored

## Checklist


jsugarman (approved):
👍 

How did the "UPDATE ar_internal_metadata..." go?

colinbruce:
> 👍
> 
> How did the "UPDATE ar_internal_metadata..." go?

It did not... :) 
The warning said the table did not exist 🤷‍♂️

PR #5529, opened 2023-07-25
Fix: Opponent bug in anonymizer

colinbruce:
## What

This occurred a few times when it was set up and I had not noticed the recurrence, or needed to use the affected tables

This adds an explanation to the anonymizer ruleset and updates the two affected tables

## Checklist


jsugarman (approved):
👍 

How does column ordering impact this? just thinking if you add a newline to a column value that is not the last what would the impact be. But worth a go regardless

PR #5555, opened 2023-07-31
AP-4105: Create partner student finance page

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4105)

Replicate student finance question for partner. As student finance is always annual only a monetary value is needed. Redundant fields have been removed

The partner student finance page is not currently connected in the flow to any others as they have not been built yet

On merging, this command should be run on production: `rails db:migrate && rake migrate:student_finance_responses`

What's been done:
- move student finance info from LegalAidApplication to Applicant and Partner models
- add migration for new columns
- add new StudentFinanceForm for partner
- add view partial to be used by both applicant and partner
- update specs
- move translation data to shared
- add rake task to move existing student finance info onto the Applicant
- update cfe submission to reflect changes

## Checklist

[AP-3804]: https://dsdmoj.atlassian.net/browse/AP-3804?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/forms/applicants/student_finance_form.rb
```diff
@@ -0,0 +1,16 @@
+module Applicants
+  class StudentFinanceForm < BaseForm
+    form_for Applicant
```
colinbruce, line 3: Could these inherit from a BaseStudentFinanceForm to reduce duplication?
Or even share the same form?

app/models/irregular_income.rb
```diff
@@ -4,14 +4,6 @@ class IrregularIncome < ApplicationRecord
   validates :income_type, inclusion: %w[student_loan]
   validates :frequency, inclusion: %w[annual]
 
-  def as_json(_options = nil)
```
colinbruce, line 7: Is the plan to remove this whole model in a future PR?

app/services/cfe_civil/components/irregular_incomes.rb
```diff
@@ -1,15 +1,37 @@
 module CFECivil
   module Components
     class IrregularIncomes < BaseDataBlock
-      delegate :irregular_incomes, to: :legal_aid_application
-
       def call
         {
           irregular_incomes: {
```
colinbruce, line 6: Have you checked this with a CFE submission?  I'm not sure how it will handle 
```ruby
{
  irregular_incomes: {
    []
  }
}
```
I have seen Civil fail with empty arrays, we may have to not send _anything_ but I'm not sure.  Happy to pair and investigate :+1:

spec/factories/applicants.rb
```diff
@@ -7,6 +7,8 @@
     has_national_insurance_number { true }
     national_insurance_number { "JA123456D" }
     employed { false }
+    student_finance { nil }
```
colinbruce, line 10: Should the partner factory be updated with defaults too?

colinbruce (commented):
Some minor observations around duplication - happy to discuss the CFE stuff too :+1:

app/services/cfe_civil/components/irregular_incomes.rb
agoldstone93: Won't this return:
```
{
    irregular_incomes: {
        payments: [],
    }
}
```

That's what the original first spec was returning. I haven't made any changes to that. Either way, pairing to see how it will handle it sounds like a good idea

app/models/irregular_income.rb
agoldstone93: Yeah, I was going to do a tidy up PR. I just removed this block now to get the coverage to pass

app/forms/applicants/student_finance_form.rb
agoldstone93: I did do that at first, and then reverted to to separate ones because it didn't feel like there was much of a benefit. I've made some changes since then though, so I'll have another go

app/services/cfe_civil/components/irregular_incomes.rb
colinbruce: Cool, if you are matching existing tests then it should be fine! :) 
I just remember when I was writing the components that if we sent (I think employment) as 
```ruby
 { employment: [] } 
```
that cfe would error, so we ended up with skipping the whole component :)

colinbruce (approved):
LGTM :+1:

PR #5580, opened 2023-08-04
Update: Update DB restoration

colinbruce:
## What

Make a few updates to the methods of restoring anonymised databases to environments

* Prevent production restoration
* allow staging restoration
* zip the upload file before uploading to speed up the step
* re-seed the staging environment so that providers exist that can be access via the staging portal
* reduce the message output when deleting the existing version

## Checklist

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -20,8 +20,8 @@ namespace :db do
   task import_to_uat: :environment do
     raise(ArgumentError, "Cannot construct DB connection string") if build_postgres_url.length < 25
 
-    drop_existing_schema = "PGOPTIONS='--client-min-messages=warning' psql -q  #{build_postgres_url} -c 'drop schema public cascade'"
-    create_new_schema = "PGOPTIONS='--client-min-messages=warning' psql -q  #{build_postgres_url} -c 'create schema public'"
+    drop_existing_schema = "PGOPTIONS='--client_min_messages=warning' psql -q  #{build_postgres_url} -c 'drop schema public cascade'"
+    create_new_schema = "PGOPTIONS='--client_min_messages=warning' psql -q  #{build_postgres_url} -c 'create schema public'"
     restore_anonymised_database = "psql #{build_postgres_url} -f ./tmp/anonymised_db.sql"
 
     `#{drop_existing_schema}`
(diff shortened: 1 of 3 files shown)


lib/tasks/db.rake
```diff
@@ -20,8 +20,8 @@ namespace :db do
   task import_to_uat: :environment do
     raise(ArgumentError, "Cannot construct DB connection string") if build_postgres_url.length < 25
 
-    drop_existing_schema = "psql #{build_postgres_url} -c 'drop schema public cascade'"
-    create_new_schema = "psql #{build_postgres_url} -c 'create schema public'"
+    drop_existing_schema = "PGOPTIONS='--client-min-messages=warning' psql -q  #{build_postgres_url} -c 'drop schema public cascade'"
+    create_new_schema = "PGOPTIONS='--client-min-messages=warning' psql -q  #{build_postgres_url} -c 'create schema public'"
```
naseberry, line 24: query: should it be `client-min-messages` or `client_min_messages`?

scripts/restore_anonymised_db.sh
```diff
@@ -4,20 +4,63 @@
...
+  ;;
+"staging")
+  environment=$1
+  echo "Finding pod in $environment"
+  POD=$(kubectl -n "laa-apply-for-legalaid-$environment" get pods | grep -o -m4 "apply-for-.*2/2" | tail -n1 | cut -d' ' -f 1)
+  ;;
+*)
+  environment='uat'
+  echo "Finding pod for branch $1 in $environment"
+#  POD=$(kubectl -n laa-apply-for-legalaid-$environment get pods | grep -o -m4 "apply-$1-.*2/2" | head -n1 | cut -d' ' -f 1)
```
naseberry, line 38: nit: is there a purpose for leaving this comment, considering we could copy and paste L33 for UAT?

naseberry (approved):
LGTM

minor comments

lib/tasks/db.rake
colinbruce: 🤔 Weird.  You are right that the docs say underscore, but the blog post (and stackoverflow example I found) suggest dashes.
As it seems to accept either,  I will update both this and the db_export to use the underscored version from the postgres docs in case it's an old affectation that gets removed :+1:

scripts/restore_anonymised_db.sh
colinbruce: Nope, that was a mistake, line 33 has a slightly different grep command.
I was restoring to staging so missed this!

PR #5590, opened 2023-08-08
[AP-3650] Apply GOVUK Table component to admin views

naseberry:
## What

[Use table component](https://dsdmoj.atlassian.net/browse/AP-3650)

Update all admin views to use GOVUK Table component

## Checklist

### DIFF
--- a/config/locales/en/admin.yml
+++ b/config/locales/en/admin.yml
@@ -27,7 +27,9 @@ en:
         status: Status
         provider_username: Provider's username
         discarded: Discarded by user on %{date}
-        data_view: Submission details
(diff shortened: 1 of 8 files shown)


config/locales/en/admin.yml
```diff
@@ -106,6 +109,7 @@ en:
         satisfaction: Satisfaction rating
         suggestion: Improvement suggestion
         source: Source
+        page_title: Feedback
```
colinbruce, line 112: Does this replace heading_1?  Should we use the same entry for both?

colinbruce (approved):
It's a lot to take in, and I'd like to UAT to confirm nothing major, but looks good from reading!

One minor nit about the feedback label repetition

PR #5597, opened 2023-08-09
Rubocop: reduce RSpec/ContextWording exclusions

colinbruce:
## What

Address the issues in spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb

This also had some weirdly nested tests that I have flattened but this makes it harder to review

## Checklist

### DIFF
--- a/spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
+++ b/spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
@@ -3,7 +3,7 @@ require "rails_helper"
 module CCMS
   module Requestors
     RSpec.describe NonPassportedCaseAddRequestor, :ccms do
-      context "XML request" do
+      describe "XML request" do
         let(:expected_tx_id) { "201904011604570390059770666" }
         let(:firm) { create(:firm, name: "Firm1") }
         let(:office) { create(:office, firm:) }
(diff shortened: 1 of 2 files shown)


spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
```diff
@@ -190,8 +190,8 @@ module Requestors
             end
           end
 
-          context "conditionally present land attributes" do
-            context "applicant has land" do
+          context "when conditionally present land attributes" do
```
agoldstone93, line 193: I don't think I understand what this line means

agoldstone93 (approved):
One context I don't understand but otherwise all good

spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
colinbruce: Does
```suggestion
          context "when the conditionally present 'land' attribute is set" do
```
read any better @agoldstone93 ?

agoldstone93: Yep, makes sense. Thanks @colinbruce

PR #5598, opened 2023-08-09
AP-3654: Use button helpers

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3654)

Replace instances of `button_to_accessible` with `govuk_button_to` so that they are maintained by the govuk components library

- update js to remove govuk-button--disabled class from button when it becomes visible
- remove outdated gateway_evidence information, but leave in model for future reference
- update tests to use `uploaded_evidence_collection` factory rather than `gateway_evidence` factory
- update css so that `button-as-link` class styles links correctly

## Checklist


colinbruce (commented):
This feels like it bloated slightly - can the gateway evidence bits be split into a separate PR?

There are still records in the live service that did use the gateway evidence so some of the removals seem a bit over eager, removing the check answers section for example if they look at a completed application will no longer show their data

naseberry:
Update to use button helpers look good, however, see Colin comment.

PR #5607, opened 2023-08-10
AP-4374: Update proceeding flow

colinbruce:
## What

When users use the back button to move back past the proceedings search page they get stuck having to
search again.  If they have already added the single proceeding they need, this is very counter intuitive.

This PR matches the newer, add another, implementations and skips to the 'list and add another proceeding' page

It has been checked with the design team and they are happy with the change to the flow

## Checklist


jsugarman (approved):
Nice, ye had noticed this problematic flow 👍

PR #5608, opened 2023-08-10
AP-4373: Add internally routed LFA url

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4373)

We have seen a number of slow calls to LFA from Apply, this caused slow page transitions and possibly errors for providers while also polluting the logs and alerts channels for devs

This PR adds a JS host variable that allows https, externally routed, calls to LFA for the Javascript search
The CSP did not need updating as it still uses the https, service.gov.uk, url as the CSP prohibits calling the internal routing http url from a secure, https, page

## Checklist


jsugarman (commented):
Visible code looks fine but I pulled locally and noticed the values-production `host` looks to be pointing to staging internal cluster namespace. Is that right?

colinbruce:
Updated with correction, thanks for the comprehensive check @jsugarman ⭐

jsugarman (approved):
👍 looks good

PR #5617, opened 2023-08-14
AP-4350: Ignore unneeded opponent model attributes

jsugarman:
## What
Ignore unneeded `opponents` model attributes as prelude to dropping

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4350)

Following migration to using opponent individuals and
organisations via a polymorphic relation these attributes
are no longer needed.

- [x] ignore the attributes
- [x] update dependant code - should just be the before_save hook 

There will be a follow up PR to remove/drop the opponents.first_name and opponents.last_name columns in a migration. This follows the [strong_migrations style for removing columns](https://github.com/ankane/strong_migrations#removing-a-column)

## Checklist

### DIFF
--- a/app/models/application_merits_task/opponent.rb
+++ b/app/models/application_merits_task/opponent.rb
@@ -2,6 +2,8 @@ module ApplicationMeritsTask
   class Opponent < ApplicationRecord
     include CCMSOpponentIdGenerator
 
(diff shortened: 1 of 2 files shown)


colinbruce (approved):
LGTM :+1:

PR #5624, opened 2023-08-15
Test: Encrypted and logged console access

colinbruce:
## What

This is something that always comes up in risk registers, about logging access to production consoles.

It uses the basecamp gems [audits1984](https://github.com/basecamp/audits1984) and [console1984](https://github.com/basecamp/console1984) which will force all users to input a reason for accessing production and then record all commands entered by users.  These will be viewable in a web console (that we may have to lock down!)

We have been unable to install them due to Postgres and Rails being outdated, but now we are up to date i.e. psql 14+, I have started investigating again

- [x] get proof of concept working
- [x] get auditing working
- [ ] get user name logging working
- [x] fix 500 favicon error - temp hack (re-skin may fix)
- [ ] Skin the audit page within our layout

## Checklist


colinbruce:
@naseberry are you happy for me to delete this PR and branch?  We can always recover it if needed

PR #5631, opened 2023-08-16
Modify chances_of_successes FK

jsugarman:
## What
Alter foreign key constraint on chances_of_successes/proceedings association

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4378)

Change the FK constraint between chances_of_successes and proceedings
to add "on delete cascade" behaviour. This to address a failure with 
seen in the `rake purge:destroy` which means applications with a 
chances_of_success cannot be destroyed, raising the error.

```
PG::ForeignKeyViolation: ERROR:  update or delete on table "proceedings" violates foreign key constraint "fk_rails_7549c96c78" on table "chances_of_successes" (ActiveRecord::InvalidForeignKey)
DETAIL:  Key (id)=(ae1f1999-1048-43b6-b194-14f115774c5a) is still referenced from table "chances_of_successes".
```

This started happening on the 6th August 2023 and is most likley because
we are only now seeing purgable applications where the chances of success
are attached to the proceeding rather than the legal_aid_application -
this changed on June 2021.

Error Scenario: If you attempt to destroy a parent record using destroy!
(with a bang ! to raise an exception on failure), Rails will first attempt
to destroy the associated child records. However, if there are other
records in the database that reference the child records you're trying
to destroy, you will encounter a PG::ForeignKeyViolation error.
This error occurs because the foreign key constraint prevents the child
records from being destroyed due to their association with other records.

### Also
Add rake task to log SQL to stdout, useful for checking SQL in migration

e.g. 
```sh
rails db:log db:migrate
```

## Checklist

### DIFF
--- a/db/migrate/20230815155824_validate_foreign_key_on_chances_of_successes.rb
+++ b/db/migrate/20230815155824_validate_foreign_key_on_chances_of_successes.rb
@@ -1,0 +1,5 @@
+class ValidateForeignKeyOnChancesOfSuccesses < ActiveRecord::Migration[7.0]
+  def change
+    validate_foreign_key :chances_of_successes, :proceedings
+  end
+end
--- a/db/migrate/20230815154938_change_foreign_key_on_chances_of_successes.rb
+++ b/db/migrate/20230815154938_change_foreign_key_on_chances_of_successes.rb
@@ -1,0 +1,6 @@
+class ChangeForeignKeyOnChancesOfSuccesses < ActiveRecord::Migration[7.0]
+  def change
+    remove_foreign_key :chances_of_successes, :proceedings
+    add_foreign_key :chances_of_successes, :proceedings, on_delete: :cascade, validate: false
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_08_15_072458) do
+ActiveRecord::Schema[7.0].define(version: 2023_08_15_155824) do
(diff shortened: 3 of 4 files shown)


db/migrate/20230815155824_validate_foreign_key_on_chances_of_successes.rb
```diff
@@ -0,0 +1,5 @@
+class ValidateForeignKeyOnChancesOfSuccesses < ActiveRecord::Migration[7.0]
+  def change
+    validate_foreign_key :chances_of_successes, :proceedings
```
colinbruce, line 3: Could this be part of the first migration or does it need to be separate?

colinbruce (approved):
LGTM :+1:

One question around migrations, but feel free to ignore 😄

db/migrate/20230815155824_validate_foreign_key_on_chances_of_successes.rb
jsugarman: @colinbruce This comes from a [strong_migrations recommendation](https://github.com/ankane/strong_migrations#good-10). 

Effectively because both tables are write locked while the new FK constraint is added, and validation make it more time consuming, they recommend adding the constraint without validating then validating separately. `

I suppose there is a minor risk of the validation migration failing - have not seen locally, or on the branch, I migrated with a prod dump on it - leaving some FKs being invalid but it otherwise seemed like a sensible approach to me.

jsugarman: incidentally it took it 4.7 seconds to run the migrations on ~25k records.

colinbruce: Cool :+1:

PR #5633, opened 2023-08-16
Anonymise success prospect details

jsugarman:
## What
Anonymise `chances_of_successes#success_prospect_details` during dump

Because this field may contain sensitive information

Also, amend the connect details mechanism for anonymised database dumping

## Testing

I have tested this by performing a database dump both locally with rails `db:export` and via the script
on the branch and both work. The `chances_of_successes#success_prospect_details` is anonymised only
when it is not null.

## Checklist

### DIFF
--- a/db/anonymise/rules.rb
+++ b/db/anonymise/rules.rb
@@ -91,7 +91,11 @@ NINO_REGEXP = /^[A-CEGHJ-PR-TW-Z]{1}[A-CEGHJ-NPR-TW-Z]{1}[0-9]{6}[A-DFM]{1}$/
   cfe_submissions: {
     cfe_result: nil,
   },
-  chances_of_successes: {},
+  chances_of_successes: {
+    success_prospect_details: lambda do |original_value, row_context|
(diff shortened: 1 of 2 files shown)


db/anonymise/rules.rb
```diff
@@ -91,7 +91,11 @@
   cfe_submissions: {
     cfe_result: nil,
   },
-  chances_of_successes: {},
+  chances_of_successes: {
+    success_prospect_details: lambda do |original_value, row_context|
+      row_context[:field_1_val] = Faker::Lorem.paragraph(sentence_count: 2) unless original_value.eql?('\N')
+    end,
```
colinbruce, line 97: Oooh, is this to handle the last column problem?

colinbruce (approved):

db/anonymise/rules.rb
jsugarman: sadly not. its just a weird way of setting a fields value conditionally - :field_1_val in this context is the first field _mentioned_ in the hash for the chances_of_successes table COPY statements. If I added another column/field to be anonymised above it i would have to amend this to  `field_1_val`.

I don't like it but it seems to be the only way to not fill a field that would have been null

PR #5650, opened 2023-08-22
AP-4225: cleanup following WSDL investigation

jsugarman:
## What
Cleanup WSDL and CCMS integration

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4225)

Following investigation into WSDLs 

- AP-4303: remove commented out code
- AP-4303: Amend abstractWSDLS

## Checklist


colinbruce (approved):
This looks okay, although I am nervous about updating the production WSDLs in the same PR as the non-live ones...
I assume it has been tested on staging/UAT?

app/services/ccms/wsdls/CaseServicesPRODWsdl.xml
```diff
@@ -7,7 +7,7 @@
   xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/" 
   xmlns:wsse="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd">
   <wsdl:documentation>
-    <abstractWSDL>http://ccmsebsprod.lab.gov:80/soa-infra/services/default/CaseServices!1.0/CaseProxyService.wsdl</abstractWSDL>
+    <abstractWSDL>https://ccmssoagateway.stg.legalservices.gov.uk/ccmssoa/soa-infra/services/default/CaseServices!10.0/CaseProxyService.wsdl</abstractWSDL>
```
jsugarman, line 10: remove `stg`

jsugarman:
> This looks okay, although I am nervous about updating the production WSDLs in the same PR as the non-live ones... I assume it has been tested on staging/UAT?

This was asked of Muraliwho said
> The [production] abstract WSDL’s are still pointing to 6DG SOA which doesn’t exist, as all the SOA composites are now migrated to AWS...I don’t think this matters and can be ignored.

jsugarman:
> I assume it has been tested on staging/UAT?

Since I have only changed the abstractWSDLs on production there is not really a way to test any impact on UAT/Staging. I will keep an eye out but it seems unlikely as a. the old domain no longer exists and has not caused issues b. they are documentary only, in nature. 

BTW the "new" dev environment does not work, returning an internal server error, just as the old dev environment does not (no longer exists). I am talking to Murali and Sahid in an attempt to resolve this today

PR #5668, opened 2023-08-25
Test: Group Jest dependencies together in dependabot

colinbruce:
## What

Rather than have two npm dependabot PRs for jest and jest-environment-jsdom try and get them grouped together to reduce circle runs when the first is merged and the second has to be rebased

Also, late addition, try grouping the aws-* PRs for the same reason

## Checklist


jsugarman (approved):
👍 nice, was not aware it could do this!

PR #5728, opened 2023-09-11
AP-4441: Fix the overnight purge job

colinbruce:
## What

The overnight purge job had been failing because the LegalFrameworkMeritsTaskList was not being cascade deleted.  This PR updates the foreign key link between the two models and has been tested with an anonymised DB backup.

Before the fix, the job failed with the same error seen in production. After the fix it successfully completed

## Checklist


jsugarman (approved):
👍  will there be a follow up PR to validate the foreign key again?

PR #5744, opened 2023-09-13
AP-4442: Ensure dependant pages show save as draft button

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4442)

One did not have it at all, the other was conditional reverting to false.  It was agreed that these should _always_ be present, so this PR sets the value to always be true on both pages

## Checklist


RoseSAK (approved):
💯

PR #5783, opened 2023-09-22
Rubocop: address RSpec/NamedSubject exclusions

colinbruce:
## What

Address the exclusions in the folders in `spec/requests/providers/` starting with `a` and `b`

## Checklist

### DIFF
--- a/spec/requests/providers/address_lookups_controller_spec.rb
+++ b/spec/requests/providers/address_lookups_controller_spec.rb
@@ -6,10 +6,10 @@ RSpec.describe Providers::AddressLookupsController do
   let(:provider) { legal_aid_application.provider }
 
   describe "GET /providers/applications/:legal_aid_application_id/address_lookup" do
-    subject { get providers_legal_aid_application_address_lookup_path(legal_aid_application) }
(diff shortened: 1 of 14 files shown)


spec/requests/providers/address_lookups_controller_spec.rb
```diff
@@ -28,7 +28,7 @@
   end
 
   describe "PATCH/providers/applications/:legal_aid_application_id/address_lookup" do
-    subject { patch providers_legal_aid_application_address_lookup_path(legal_aid_application), params: }
+    subject(:path_request) { patch providers_legal_aid_application_address_lookup_path(legal_aid_application), params: }
```
jsugarman, line 31: should that be `patch_request`

jsugarman (approved):
you have a `path_request` which looks like it should be `patch_request` in one spec. otherwise 👍

spec/requests/providers/address_lookups_controller_spec.rb
colinbruce: Yes it should! 🤦‍♂️

PR #5792, opened 2023-09-26
Ap 3655 use link helpers part 1

skdejong:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3655)

Replaced all instances of `link_to_accessible` with the `govuk_link_to` component in the helper, admin, citizens, feedback and problem files. 

## Checklist

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -6,7 +6,7 @@ module AdminHelper
                             [path, nil]
                           end
 
-    link = link_to_accessible name, link_path, class: "govuk-link govuk-link--no-visited-state"
+    link = govuk_link_to name, link_path, class: "govuk-link govuk-link--no-visited-state"
     content_tag :li, link, class: li_class
   end
 end
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -6,7 +6,7 @@ module AdminHelper
                             [path, nil]
                           end
 
-    link = govuk_link_to name, link_path, class: "govuk-link govuk-link--no-visited-state"
+    link = govuk_link_to name, link_path, { no_visited_state: true }
     content_tag :li, link, class: li_class
   end
 end
--- a/app/views/citizens/accounts/index.html.erb
+++ b/app/views/citizens/accounts/index.html.erb
@@ -41,7 +41,7 @@
 
     <% end %>
 
-    <p><%= link_to_accessible t(".link_other_account"), citizens_banks_path, class: "govuk-link" %></p>
+    <p><%= govuk_link_to t(".link_other_account"), citizens_banks_path, class: "govuk-link" %></p>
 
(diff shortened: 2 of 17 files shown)


app/helpers/admin_helper.rb
```diff
@@ -6,7 +6,7 @@ def admin_nav_to(name, path)
                             [path, nil]
                           end
 
-    link = link_to_accessible name, link_path, class: "govuk-link govuk-link--no-visited-state"
+    link = govuk_link_to name, link_path, class: "govuk-link govuk-link--no-visited-state"
```
naseberry, line 9: I think the class `govuk-link` can be removed as it is applied by default by the helper `govuk_link_to`

Also, the visited class should be set using the [helpers option](https://govuk-components.netlify.app/helpers/link/#other-link-styles) `no_visited_state`

In short, 
`link = govuk_link_to name, link_path, class: "govuk-link govuk-link--no-visited-state"`

becomes 

`link = govuk_link_to name, link_path, { no_visited_state: true }`

app/views/citizens/check_answers/_bank_accounts.html.erb
```diff
@@ -24,5 +24,5 @@
 <% end %>
 
 <div class="govuk-body govuk-!-padding-bottom-4">
-  <%= link_to_accessible t("citizens.check_answers.index.link_other_account"), citizens_banks_path, class: "govuk-link" %>
+  <%= govuk_link_to t("citizens.check_answers.index.link_other_account"), citizens_banks_path, class: "govuk-link" %>
```
naseberry, line 27: I think the class `govuk-link` can be removed as it is applied by default by the helper `govuk_link_to`

app/views/citizens/accounts/index.html.erb
```diff
@@ -41,7 +41,7 @@
 
     <% end %>
 
-    <p><%= link_to_accessible t(".link_other_account"), citizens_banks_path, class: "govuk-link" %></p>
+    <p><%= govuk_link_to t(".link_other_account"), citizens_banks_path, class: "govuk-link" %></p>
```
naseberry, line 44: I think the class `govuk-link` can be removed as it is applied by default by the helper `govuk_link_to`

app/helpers/admin_helper.rb
skdejong: Thank you for the tip. I have made these suggested changes now.

app/views/citizens/accounts/index.html.erb
skdejong: Thank you for the tip. I have made these suggested changes now.

app/views/citizens/check_answers/_bank_accounts.html.erb
skdejong: Thank you for the tip. I have made these suggested changes now.

naseberry (approved):
LGTM

PR #5809, opened 2023-09-29
AP-4141: Partner offline bank account page

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4141)

Update partner bank account handling, add additional fields to the savings amount table 
Add the new view and routing needed for a partner claim to proceed

## Checklist

### DIFF
--- a/app/models/concerns/capital.rb
+++ b/app/models/concerns/capital.rb
@@ -4,6 +4,10 @@ module Capital
   end
 
   def amount_attributes
-    attributes.except("id", "legal_aid_application_id", "updated_at", "created_at", "none_selected", "no_account_selected")
+    shared_attributes.except("partner_offline_current_accounts", "partner_offline_savings_accounts")
+  end
+
+  def shared_attributes
+    attributes.except("id", "legal_aid_application_id", "updated_at", "created_at", "none_selected", "no_account_selected", "no_partner_account_selected")
   end
 end
--- a/app/views/providers/partners/bank_accounts/show.html.erb
+++ b/app/views/providers/partners/bank_accounts/show.html.erb
@@ -8,7 +8,7 @@
 
     <%= form.govuk_check_boxes_fieldset :savings_amount,
                                         legend: { text: page_title, size: "xl", tag: "h1" },
-                                        hint: { text: t("generic.select_all_that_apply") },
+                                        hint: { text: t(".select_all_that_apply") },
                                         classes: "govuk-!-padding-bottom-4" do %>
       <div class="govuk-checkboxes" data-module="govuk-checkboxes">
         <div class="deselect-group" data-deselect-ctrl="#savings-amount-no-partner-account-selected-true-field">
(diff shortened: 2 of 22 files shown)


app/models/concerns/capital.rb
```diff
@@ -4,6 +4,10 @@ def positive?
   end
 
   def amount_attributes
-    attributes.except("id", "legal_aid_application_id", "updated_at", "created_at", "none_selected", "no_account_selected")
+    shared_attributes.except("partner_offline_current_accounts", "partner_offline_savings_accounts")
```
colinbruce, line 7: The next ticket may need to extract only the partner fields, so I separated this rather than merging the list into a single entity and then extracting again immediately

config/locales/en/shared.yml
```diff
@@ -564,11 +566,22 @@ en:
...
+            partners:
+              bank_accounts:
+                check_box_partner_offline_current_accounts: Current account
+                check_box_partner_offline_savings_accounts: Savings account
+                hint:
+                  revealing_check_box_partner_offline_current_accounts: Include accounts with a negative balance or no money in them.
+                  revealing_check_box_partner_offline_savings_accounts: Include accounts with a negative balance or no money in them.
+                partner_offline_current_accounts: Enter the total in all current accounts
+                partner_offline_savings_accounts: Enter the total in all savings accounts
             savings_and_investments:
```
RoseSAK, line 578: I find it confusing that this part is within `savings_and_investments` - I think I have actually removed all of this in my current PR. So there may be some merge conflicts for one/both of us at some point 😬 not sure where is best for it to go though!

app/views/providers/partners/bank_accounts/show.html.erb
```diff
@@ -0,0 +1,29 @@
...
+        local: true,
+      ) do |form| %>
+
+    <%= form.govuk_check_boxes_fieldset :savings_amount,
+                                        legend: { text: page_title, size: "xl", tag: "h1" },
+                                        hint: { text: t("generic.select_all_that_apply") },
+                                        classes: "govuk-!-padding-bottom-4" do %>
+      <div class="govuk-checkboxes" data-module="govuk-checkboxes">
+        <div class="deselect-group" data-deselect-ctrl="#savings-amount-no-partner-account-selected-true-field">
+          <%= render partial: "shared/forms/revealing_checkbox/attribute",
```
RoseSAK, line 15: In AP-4143 I have removed the use of this partial from all the other capital question pages - not sure whether it makes sense to do the same here or if it's still needed?

RoseSAK (approved):
Looks good, only a couple of questions which are non-blocking!

app/views/providers/partners/bank_accounts/show.html.erb
colinbruce: Let me take a look at your PR while reviewing and then re-assess this code :+1:

config/locales/en/shared.yml
colinbruce: I think I had to add these for the CYA/review pages, but I can't remember now!

app/views/providers/partners/bank_accounts/show.html.erb
colinbruce: This has been refactored and removed - can you take a look at the changes?

RoseSAK (approved):
LGTM!

PR #5813, opened 2023-10-02
Chore: remove public keys

naseberry:
## What

Remove public gpg keys and the related documentation.
We have since moved to using symmetric key and therefore this is no longer used.

## Checklist


colinbruce (changes requested):
I'm pretty sure we still need the .gitattributes file. The symmetric key uses it to encrypt expected files.

colinbruce (approved):

PR #5824, opened 2023-10-03
Rubocop: Address RSpec/NamedSubjects in services

colinbruce:
## What

First of two PRs addressing RSpec/NamedSubject exclusions in the services folder

This addresses those in services where the folder or class starts with `a` to `m`

## Checklist

### DIFF
--- a/spec/services/banking/bank_transactions_trimmer_spec.rb
+++ b/spec/services/banking/bank_transactions_trimmer_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 
 RSpec.describe Banking::BankTransactionsTrimmer do
   describe ".call" do
-    subject { described_class.call(application) }
+    subject(:bank_transaction_trimmer) { described_class.call(application) }
 
     let(:application) { create(:legal_aid_application, :with_applicant) }
     let(:applicant) { application.applicant }
@@ -21,7 +21,7 @@ RSpec.describe Banking::BankTransactionsTrimmer do
       let(:period_end) { Date.parse("2021-01-07").beginning_of_day }
 
       it "does not delete any transactions" do
-        expect { subject }.not_to change(BankTransaction, :count)
+        expect { bank_transaction_trimmer }.not_to change(BankTransaction, :count)
       end
     end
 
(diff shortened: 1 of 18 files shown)


spec/services/banking/bank_transactions_trimmer_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe Banking::BankTransactionsTrimmer do
   describe ".call" do
-    subject { described_class.call(application) }
+    subject(:bank_transaction_trimmer) { described_class.call(application) }
```
jsugarman, line 5: nit, could just use the described (class) method, but either way its an improvement

spec/services/metrics/send_metrics_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe Metrics::SendMetrics do
   describe "#call" do
-    subject { described_class.call }
+    subject(:call_send_metrics) { described_class.call }
```
jsugarman, line 5: nit, could just use `call`, but either way

spec/services/metrics/sidekiq_queue_sizes_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 RSpec.describe Metrics::SidekiqQueueSizes do
   describe "#call" do
-    subject { described_class.call(prometheus_client) }
+    subject(:call_queue_sizes) { described_class.call(prometheus_client) }
```
jsugarman, line 5: nit,  could just `call` but either way

spec/services/mock_benefit_check_service_spec.rb
```diff
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 RSpec.describe MockBenefitCheckService do
-  subject { described_class.call(application) }
+  subject(:mock_benefit_check) { described_class.call(application) }
```
jsugarman, line 4: nit, could just use `call`, but either way

jsugarman (approved):
👍 ✅   just some take-it-or-leave-it nits, all of which amount to "could just use `call`...)

colinbruce:
> 👍 ✅ just some take-it-or-leave-it nits, all of which amount to "could just use `call`...)

I toyed with the idea of call but it becomes less clear when you read it in the tests, e.g. in a long test
```ruby
        it "marks the transaction as a state benefit" do
          call
          tx = legal_aid_application.reload.bank_transactions.first
          expect(tx.transaction_type_id).to eq included_benefit_transaction_type.id
        end
```
`call` seemed a bit generic,`call_object_name` felt clearer to me 🤷‍♂️

PR #5832, opened 2023-10-04
AP-4468: Update feature step susceptible to false positives

jsugarman:
## What
Amend step susceptible to false positives and use, correct or replace in features

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4468)

- Amend offending step to use within so as to prevent false positives
- Correct newly failing feature step call by correct the section name used
- Correct use of step for statement of case where expected text is not actually
  connected to the DOM row, wrongly or rightly?!
- Add new step to allow CYA list items to exhausively check row questions and answers

## Checklist

### DIFF
--- a/features/providers/passported_journey.feature
+++ b/features/providers/passported_journey.feature
@@ -90,8 +90,8 @@ Feature: passported_journey completes application
     Then I enter the application merits task statement of case statement field 'This is some test data for the statement of case'
     Then I click 'Save and continue'
     Then I should be on a page showing "Check your answers"
-    Then I should be on a page showing "hello_world.pdf (15.7 KB)"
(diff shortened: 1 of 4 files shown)


features/providers/passported_journey.feature
```diff
@@ -91,7 +91,8 @@ Feature: passported_journey completes application
     Then I click 'Save and continue'
     Then I should be on a page showing "Check your answers"
     Then I should be on a page showing "hello_world.pdf (15.7 KB)"
```
jsugarman, line 93: delete this line as being tested below

colinbruce (approved):
LGTM 🏆

PR #5838, opened 2023-10-05
Rubocop: Final removal of NamedSubject exclusions

colinbruce:
This is the final PR addressing RSpec/NamedSubject exclusions.

It addresses those in services where the folder or class starts with n to x

It switches to using `subject(:call)` for brevity

## What

## Checklist

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -110,32 +110,6 @@ RSpec/LetSetup:
     - 'spec/workers/bank_holiday_update_worker_spec.rb'
     - 'spec/workers/true_layer_banks_update_worker_spec.rb'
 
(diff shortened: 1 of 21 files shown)


jsugarman (approved):
LGTM 👍 

A habit/pattern I generally use for `subject`s that are calling `new` is just to call them `instance`, for brevity and because the name of the class may change, but as ever, either way.

colinbruce:
> LGTM 👍
> 
> A habit/pattern I generally use for `subject`s that are calling `new` is just to call them `instance`, for brevity and because the name of the class may change, but as ever, either way.

I just want to get the rubocop_todo volume down! 😆

PR #5870, opened 2023-10-13
AP-4481: Fix opponent individual CCMS injection payload

jsugarman:
## What
Amend opponent individual payload to fix CCMS/PUI amendment bug

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4481)

- Switch to using NONE/none instead of UNKNOWN/Unknown, then test
- Add a dummy title, then test

It was discovered that Individual Opponents injected by the Apply service were not being displayed on the Amend case screen in PUI. As the link states ‘Edit Opponents and other parties’ we believe Individual should appear on the Amend case (this is the screenshot with the ‘Edit opponents and other parties’ link) but only organisations are displayed as expected.

image-20230927-091223

## Checklist

### DIFF
--- a/app/models/application_merits_task/individual.rb
+++ b/app/models/application_merits_task/individual.rb
@@ -13,7 +13,7 @@ module ApplicationMeritsTask
     end
 
     def ccms_relationship_to_client
-      "UNKNOWN"
+      "NONE"
     end
 
     def ccms_child?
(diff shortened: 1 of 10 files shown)


jsugarman:
- Switch to using NONE/none instead of UNKNOWN/Unknown, then test
- Add a dummy title, then test

The providing of NONE/none instead of UNKNOWN/Unknown had no effect. The providing of a dummy Tile fixed the issue. Ways forward added to [ticket AP-4499](https://dsdmoj.atlassian.net/browse/AP-4559)

PR #5871, opened 2023-10-13
AP-4359: Add cfe_civil partner submission

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4359)

Update the CFE submission builder to add a partner object, if needed.

Update the, current supported, individual components to build an applicant or partner data set if an owner tpe of partner is submitted.

Some of the components are not needed, and some will need updating in the future when the truelayer flow is implemented for the partners

## Checklist

### DIFF
--- a/app/services/cfe_civil/components/partner.rb
+++ b/app/services/cfe_civil/components/partner.rb
@@ -7,14 +7,15 @@ module CFECivil
         return {}.to_json if partner.blank?
 
         result = base_partner
-        result.merge! JSON.parse(Components::CashTransactions.new(@legal_aid_application, "Partner").call) # so far... all other models are stored as Applicant/Partner
-        result.merge! JSON.parse(Components::IrregularIncomes.new(@legal_aid_application, "Partner").call) # so far... all other models are stored as Applicant/Partner
-        result.merge! JSON.parse(Components::Vehicles.new(@legal_aid_application, "partner").call) # this is stored as client/partner
-        result.merge! JSON.parse(Components::Employments.new(@legal_aid_application, "Partner").call) # this is stored as Applicant/Partner
-        result.merge! JSON.parse(Components::RegularTransactions.new(@legal_aid_application, "Partner").call) # this is stored as Applicant/Partner
-        result.merge! JSON.parse(Components::Capitals.new(@legal_aid_application, "Partner").call) # this is stored as Applicant/Partner
-        # employment_details # skipping this is a ccq datatype, we don't use it
-        # self_employment_details # skipping this is a ccq datatype, we don't use it
+        # Most records are stored with an Owner field that is tied to Applicant or Partner models, using their IDs
+        # Vehicles are stored with the, optional, output of a radio button as client or partner (lower cased)
+        # For this reason the call to Components::Vehicles.new is slightly different from the rest of the generators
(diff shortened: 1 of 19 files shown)


spec/services/cfe_civil/components/vehicles_spec.rb
```diff
@@ -33,5 +34,46 @@
...
+
+    before do
+      create(:vehicle,
+             legal_aid_application:,
+             estimated_value: 2345.0,
+             payment_remaining: 321.0,
+             purchased_on: Date.new(2020, 5, 18),
+             used_regularly: true,
+             owner: "partner")
+      create(:vehicle, :populated, legal_aid_application:)
```
jsugarman, line 65: i prefer the explicit value setting and test thereof here to the factory (i..e hidden values) method employed above at line 39. I would change the values to be different from those from the factory too to differentiate.

spec/services/cfe_civil/components/capitals_spec.rb
```diff
@@ -17,26 +23,56 @@
...
+          },
+        }.to_json)
+      end
+    end
+
+    context "when invoked with partner as owner" do
+      subject(:call) { described_class.call(legal_aid_application, "Partner") }
+
+      it "returns json in the expected format with the partner and joint bank values merged" do
+        expect(call).to match_json_expression({
```
agoldstone93, line 64: Out of interest, why have you used `match_json_expression` here but `eq` above, on line 41?

app/services/cfe_civil/components/partner.rb
```diff
@@ -0,0 +1,40 @@
...
+    class Partner < BaseDataBlock
+      delegate :partner, to: :legal_aid_application
+
+      def call
+        return {}.to_json if partner.blank?
+
+        result = base_partner
+        result.merge! JSON.parse(Components::CashTransactions.new(@legal_aid_application, "Partner").call) # so far... all other models are stored as Applicant/Partner
+        result.merge! JSON.parse(Components::IrregularIncomes.new(@legal_aid_application, "Partner").call) # so far... all other models are stored as Applicant/Partner
+        result.merge! JSON.parse(Components::Vehicles.new(@legal_aid_application, "partner").call) # this is stored as client/partner
```
agoldstone93, line 12: I don't think it makes a difference because you call downcase, but this is the only line where "partner" is all lower case

agoldstone93 (approved):
Only minor comments, otherwise LGTM

spec/services/cfe_civil/components/capitals_spec.rb
colinbruce: It was a debugging issue, match_json_expression doesn't care about the sequence, I'll revert it to `eq` now to match all the others   :+1:

for reference, given a json block of:
```ruby
{ "foo": "bar", "bar": "baz" }

expect({ "foo": "bar", "bar": "baz" }).to eq({ "bar": "baz", "foo": "bar" }) # will fail
expect({ "foo": "bar", "bar": "baz" }).to match_json_expression({ "bar": "baz", "foo": "bar" }) # will pass, as the key/value pairs match
```

app/services/cfe_civil/components/partner.rb
colinbruce: Yeah, the vehicles block is keyed off the radio button so it stores client/partner rather than the class name of Applicant/Partner

I can try updating the comments to make that clearer?

PR #5885, opened 2023-10-18
ap-4508: Add class to handle linked_application_types

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4508)

## Checklist

### DIFF
--- a/app/models/linked_application_type.rb
+++ b/app/models/linked_application_type.rb
@@ -2,11 +2,11 @@ class LinkedApplicationType
   LinkTypeStruct = Struct.new(:code, :description)
 
   LINK_TYPES = [
-    { code: "FAMILY", description: "Family" },
-    { code: "LEGAL", description: "Legal" },
+    LinkTypeStruct.new(code: "FAMILY", description: "Family"),
+    LinkTypeStruct.new(code: "LEGAL", description: "Legal"),
   ].freeze
 
   def self.all
-    LINK_TYPES.map { |link_type_hash| LinkTypeStruct.new(code: link_type_hash[:code], description: link_type_hash[:description]) }
+    LINK_TYPES
   end
 end
--- a/app/models/linked_application_type.rb
+++ b/app/models/linked_application_type.rb
@@ -1,0 +1,12 @@
+class LinkedApplicationType
+  LinkTypeStruct = Struct.new(:code, :description)
+
+  LINK_TYPES = [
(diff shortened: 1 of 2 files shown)


app/models/linked_application_type.rb
```diff
@@ -0,0 +1,12 @@
+class LinkedApplicationType
+  LinkTypeStruct = Struct.new(:code, :description)
+
+  LINK_TYPES = [
+    { code: "FAMILY", description: "Family" },
+    { code: "LEGAL", description: "Legal" },
+  ].freeze
+
+  def self.all
+    LINK_TYPES.map { |link_type_hash| LinkTypeStruct.new(code: link_type_hash[:code], description: link_type_hash[:description]) }
```
jsugarman, line 10: you could just 

```ruby
LINK_TYPES = [
  LinkTypeStruct.new(code: "FAMILY", description: "Family"),
  LinkTypeStruct.new(code: "LEGAL", description: "Legal"),
].freeze

def self.all
  LINK_TYPES
end
```

spec/models/linked_application_type_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe LinkedApplicationType do
+  describe ".link_types" do
```
jsugarman, line 4:
```suggestion
  describe ".all" do
```

jsugarman: Possible tests to lock down the duck type and values

```ruby
it "returns a collection of Link types" do
   # expect(described_class.all).to all(be_instance_of(LinkTypeStruct))
   expect(described_class.all).to all(respond_to?(:code))
   expect(described_class.all).to all(respond_to?(:description))
end

it "returns expected collection attribute values" do
   expect(described_class.all).to match_array....
end
```

jsugarman (commented):
Looks good, some simplifications and test name correction and further/different test approach suggestions

spec/models/linked_application_type_spec.rb
kmahern: 👍 Had to tweak this suggestion slightly to get tests passing but have added tests for respond_to?(:code) and respond_to?(:description)

app/models/linked_application_type.rb
kmahern: 👍 have made suggested change

jsugarman (approved):
👍

PR #5894, opened 2023-10-20
AP-4427: Add alert and area-live=assertive for screen readers and reformat sign in page

skdejong:
## What

[Announce error message via screen reader ](https://dsdmoj.atlassian.net/browse/AP-4427)

Added role="alert" aria-live="assertive" to sign in page so that error message will be announced via screen readers.

## Checklist

### DIFF
--- a/app/controllers/saml_idp_controller.rb
+++ b/app/controllers/saml_idp_controller.rb
@@ -1,4 +1,6 @@
 class SamlIdpController < SamlIdp::IdpController
+  layout "application"
+
   def create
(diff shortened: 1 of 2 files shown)


jsugarman (approved):
nice, tried it out and it seems to work fine.

I see you have based this off the [gems new template](https://github.com/dev-develop/ruby-saml-idp/blob/master/app/views/saml_idp/idp/new.html.erb). I wonder if it could be further govuk'erised but it a lot better than it was generally anyway?

PR #5895, opened 2023-10-20
AP-4380: Add copy case feature

jsugarman:
## What
Add copy case feature

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4380)

The ability to copy case details from one application to another
is considered a prerequisite for the linking cases feature.

## TODO
- [ ] do we really need the `copy_case` boolean attribute to be stored - try to use un-persisted option
- [ ] prevent or validate against "finding" of applications that are from other firms
- [ ] prevent or validate against "finding" of applications that are not submitted
- [ ] add the actual copying of the case
- [ ] consider state and step to leave case in after copy
- [ ] consider ability to amend substantive cost limitation and delegated function cost limitation
- [ ] elements from the legal aid application that relate to substantive and emergency costs may/will need to be copied over too  -  see *1

*1
```
# legal_aid_application attributes that are derived from proceedings
# and will therefore require copying too
t.boolean "emergency_cost_override"
t.decimal "emergency_cost_requested"
t.string "emergency_cost_reasons"
...
t.boolean "substantive_cost_override"
t.decimal "substantive_cost_requested"
t.string "substantive_cost_reasons"
```

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,7 @@ ruby "3.2.2"
 
 gem "aasm", "~> 5.5.0"
 gem "active_model_serializers", "~> 0.10.14"
+gem "deep_cloneable", "~> 3.2.0"
 gem "discard", "~> 1.3"
 gem "geckoboard-ruby"
 gem "google-apis-sheets_v4"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -213,6 +213,8 @@ GEM
     date (3.3.3)
     debug_inspector (1.1.0)
     declarative (0.0.20)
+    deep_cloneable (3.2.0)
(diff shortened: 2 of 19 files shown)


kmahern (approved):
👍 I think this branch meets the criteria for "agreed approach for copying an existing application's detail to a new application". in terms of providing a poc for how to achieve this with the deep_cloneable gem.

It looks like it only tests the copying of proceedings and not other application associations like dependants or opponents.

I do think there are some concerns about how well the deep_cloneable gem is supported but this is probably also true of alternatives (ie amoeba).

jsugarman:
see implmentation branch https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5932

PR #5916, opened 2023-10-27
Chore: enable sentry session replay

naseberry:
## What

Task to enable sentry session replay.

## Checklist


jsugarman (commented):
is it ok to expose the sentry public key on every page like this? in which case does it need to be a secret?

And needing the last commit removed before merge of course.

naseberry:
> is it ok to expose the sentry public key on every page like this? in which case does it need to be a secret?
> 
> And needing the last commit removed before merge of course.

Based on the guide, yes, it is ok to expose the public key, for security, we have restricted the allowed domains from which we can receive notifications.
 
 As its public, then no its not a secret, but based on our setup, how would you load the env vars?
 
 Yes, last commit would be removed before merging

PR #5921, opened 2023-10-30
Update Rails 7 gem and node

colinbruce:
## What

This PR updates the required Rails changes:
- gem and node options to Rails 7.1.1
- removing legacy handler block
- setting load defaults, etc; 

as well as:
- A couple of the `require` blocks seem to no longer autoload from the precise location so these were updated to load explicitly 
- `Serialize` calls now require am explicit `coder:` type and this has been set to YAML
- The action_dispatch.show_exceptions setting was updated as the original setting made the cucumber tests very noisy
- A change in rails 7.1 now raises errors when _any_ i18n lookup fails.  Previously this only failed on views so this adds some error handling in the application code and changes some failing tests, where previously, a deleted key was expected not to be on the page was passing because it didn't find a translation missing error on the view!

## Checklist


spec/requests/providers/check_capital_answers_controller_spec.rb
```diff
@@ -65,18 +65,18 @@
...
       it "displays first vehicle question" do
         expect(response.body).to include(I18n.t("shared.check_answers.vehicles.providers.own", individual: "your client"))
       end
 
       it "does not display other vehicle questions" do
-        expect(response.body).not_to include(I18n.t("shared.check_answers_vehicles.providers.estimated_value"))
-        expect(response.body).not_to include(I18n.t("shared.check_answers_vehicles.providers.payment_remaining"))
-        expect(response.body).not_to include(I18n.t("shared.check_answers_vehicles.providers.purchased_on"))
-        expect(response.body).not_to include(I18n.t("shared.check_answers_vehicles.providers.used_regularly"))
+        expect(response.body).not_to include("What is the estimated value of the vehicle?")
```
jsugarman, line 76: nice

app/forms/legal_aid_applications/has_evidence_of_benefit_form.rb
```diff
@@ -9,7 +9,7 @@ class HasEvidenceOfBenefitForm < BaseForm
   private
 
     def passporting_benefit_error
-      ".shared.forms.received_benefit_confirmation.form.providers.received_benefit_confirmations.#{model.passporting_benefit}"
+      ".shared.forms.received_benefit_confirmation.form.providers.received_benefit_confirmations.#{model.passporting_benefit.squish.downcase.tr(' ', '_')}"
```
jsugarman, line 12: this is just an extra I assume. not to do with rails bump?

config/locales/en/transaction_types.yml
```diff
@@ -22,6 +22,7 @@ en:
         child_care: Childcare payments
         excluded_benefits: Disregarded benefits
         friends_or_family: Financial help from friends or family
+        housing_benefit: Housing Benefit
```
jsugarman, line 25: looks like the new config for translations caught a missing one?!

jsugarman (approved):
👍 Codewise looks fine. much prefer the use of plain text in place of translations in specs.

I imagine some mobbing and/or extensive manual testing would be necessary to feel more confident. On Assure there was an issue with devise that passed all tests but prevented login on staging/production due to AR encryption.

app/forms/legal_aid_applications/has_evidence_of_benefit_form.rb
colinbruce: This code was a safety check and a bit weird, it would send the checkbox value, e.g. `Universal credit` to find an error.
So the string interpolation would try and find `.shared.forms.received_benefit_confirmation.form.providers.received_benefit_confirmations.Universal Credit` and fail

This part of the i18n update now, correctly, raised an error!

PR #5926, opened 2023-10-31
AP-4567: Add non-passported to list of reasons on means report

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4567)

Add "Non-Passported application" to list of review reasons on means report if the application is non-passported. This is to avoid a situation whereby caseworkers are flagged to review an application but the list of review reasons is empty

## Checklist

### DIFF
--- a/spec/services/ccms/manual_review_determiner_spec.rb
+++ b/spec/services/ccms/manual_review_determiner_spec.rb
@@ -2,13 +2,14 @@ require "rails_helper"
 
 module CCMS
   RSpec.describe ManualReviewDeterminer, :ccms do
-    let(:setting) { Setting.setting }
-    let!(:cfe_submission) { create(:cfe_submission, legal_aid_application:) }
     let(:determiner) { described_class.new(legal_aid_application) }
     let(:legal_aid_application) { create(:legal_aid_application, :with_applicant) }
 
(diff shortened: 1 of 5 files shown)


jsugarman (approved):
LGTM 👍  and no blockers, however the `ManualReviewDeterminer#review_reasons` spec does not seem to test for the `passported` or `non-means-tested` route not having any review reasons as I would have hoped for. It may be worth adding a context to that effect if you have time

spec/services/ccms/manual_review_determiner_spec.rb
```diff
@@ -257,60 +257,72 @@ module CCMS
...
+            expect(review_reasons_result).to include(:ineligible)
+          end
         end
       end
 
-      context "with cfe result ineligible" do
-        let(:cfe_result) { double "CFE Result", remarks: cfe_remarks, ineligible?: true }
+      context "with passported application" do
+        let(:legal_aid_application) { create(:legal_aid_application, :passported) }
+        let(:review_reasons) { [] }
```
jsugarman, line 322: LGTM but is this let needed.

jsugarman: see last commit i added

PR #5929, opened 2023-11-01
AP-4583: Extend upload evidence for partner

skdejong:
## What

[Extend upload evidence for partner](https://dsdmoj.atlassian.net/browse/AP-4583)

Add a new evidence type for when the client's _partner_ needs to provide evidence of their employment as part of the Partner Means Assessment.  

## Checklist


config/locales/en/providers.yml
```diff
@@ -1580,7 +1581,7 @@ en:
       multi_proceeding: Proceeding %{position} of %{total}
       client_involvement_type:
         show:
-          question: What is your client’s role in this proceeding?
+          question: What is your client's role in this proceeding?
```
RoseSAK, line 1584: There are a lot of changes which look like they are just to do with the apostrophe - is this because these were previously backticks instead of apostrophes? Finding it a bit hard to tell just by looking!

db/schema.rb
```diff
@@ -1071,8 +1071,8 @@
   add_foreign_key "legal_aid_applications", "providers"
   add_foreign_key "legal_framework_merits_task_lists", "legal_aid_applications", on_delete: :cascade
   add_foreign_key "legal_framework_submissions", "legal_aid_applications"
-  add_foreign_key "linked_applications", "legal_aid_applications", column: "associated_application_id"
-  add_foreign_key "linked_applications", "legal_aid_applications", column: "lead_application_id"
+  add_foreign_key "linked_applications", "legal_aid_applications", column: "associated_application_id", validate: false
+  add_foreign_key "linked_applications", "legal_aid_applications", column: "lead_application_id", validate: false
```
RoseSAK, line 1075: How come this has been added? This might be due to running migrations locally through rails

RoseSAK (approved):
Looks great! I had only a couple of comments/questions :) great work! 🚀 💯

config/locales/en/providers.yml
skdejong: Yes. I was trying to fix some of the tests that were failing because of the `’` vs `'` and I just ended up changing all of them rather than only the specific ones that I needed to make the specific failing tests pass.

db/schema.rb
skdejong: Yes, I think that's what must have happened.

PR #5931, opened 2023-11-01
Amend example for RedisScanner utility class

jsugarman:
## What

Amend example for RedisScanner utility class

The scanner requires an instance of the Redis gem's Redis client
rather than the Redis connection return by Sidekiq.redis.

## Checklist

### DIFF
--- a/lib/utilities/redis_scanner.rb
+++ b/lib/utilities/redis_scanner.rb
@@ -9,13 +9,11 @@ require "redis"
 # scanner.results
 #
 # Example usage:
-# redis = Sidekiq.redis { |r| r }
 #
 # require_relative 'lib/utilities/redis_scanner'
 #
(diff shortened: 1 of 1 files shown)


lib/utilities/redis_scanner.rb
```diff
@@ -9,13 +9,11 @@
 # scanner.results
 #
 # Example usage:
-# redis = Sidekiq.redis { |r| r }
 #
 # require_relative 'lib/utilities/redis_scanner'
 #
+# redis  = Redis.new(url: Rails.configuration.x.redis.base_url)#
```
agoldstone93, line 15: Has this # got in the end of the line by accident?

agoldstone93 (approved):
Only a minor comment but otherwise looks good

PR #5933, opened 2023-11-01
AP-4512: add page flows for linking applications

agoldstone93:
## What

Link case feature

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4512)

The linking cases feature allows a provider to link certain cases together. This PR adds this basic functionality. Additional validations ([AP-4510](https://dsdmoj.atlassian.net/browse/AP-4510)) and the check your answers page ([AP-4601](https://dsdmoj.atlassian.net/browse/AP-4601)) will be added in subsequent tickets.

## Checklist

[AP-4576]: https://dsdmoj.atlassian.net/browse/AP-4576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AP-4601]: https://dsdmoj.atlassian.net/browse/AP-4601?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[AP-4510]: https://dsdmoj.atlassian.net/browse/AP-4510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/models/linked_application_type.rb
```diff
@@ -1,9 +1,19 @@
...
+  LinkTypeStruct = Struct.new(:code, :description, :label, :hint)
 
   LINK_TYPES = [
-    LinkTypeStruct.new(code: "FAMILY", description: "Family"),
-    LinkTypeStruct.new(code: "LEGAL", description: "Legal"),
+    LinkTypeStruct.new(
+      code: "FAMILY",
+      description: "Family",
+      label: I18n.t("providers.linking_case_confirmations.show.options.family_link.text"),
+      hint: I18n.t("providers.linking_case_confirmations.show.options.family_link.hint"),
```
jsugarman, line 9: ideally these should not be in a model but located in the translations for views, to be more idiomatic

app/models/linked_application_type.rb
```diff
@@ -1,9 +1,19 @@
...
+    ),
+    LinkTypeStruct.new(
+      code: "LEGAL",
+      description: "Legal",
+      label: I18n.t("providers.linking_case_confirmations.show.options.legal_link.text"),
+      hint: I18n.t("providers.linking_case_confirmations.show.options.legal_link.hint"),
```
jsugarman, line 15: ideally these should not be in a model but located in the translations for views, to be more idiomatic

jsugarman (commented):
just a general comment on naming before you progress even further. 

It feels to me like we should be naming these modules, forms, controllers `LinkedCaseBlah`  or `LinkedApplicationBlah` instead of Link**ing**CaseBlah

app/forms/linking_case/confirmation_form.rb
```diff
@@ -0,0 +1,16 @@
+module LinkingCase
```
jsugarman, line 1: To follow the pattern for copy_case and inline with controller naming this module should be called `LinkCase`

app/forms/linking_case/invitation_form.rb
```diff
@@ -0,0 +1,9 @@
+module LinkingCase
```
jsugarman, line 1: To follow the pattern for copy_case and inline with controller naming this module should be called `LinkCase`

app/forms/linking_case/search_form.rb
```diff
@@ -0,0 +1,34 @@
+module LinkingCase
```
jsugarman, line 1: To follow the pattern for copy_case and inline with controller naming this module should be called `LinkCase`

app/forms/linking_case/search_form.rb
```diff
@@ -0,0 +1,34 @@
...
+      model.associated_application_id = legal_aid_application.id
+      model.save!(validate: false)
+    end
+    alias_method :save!, :save
+
+    def case_exists
+      errors.add(:search_ref, :not_found) unless case_found?
+    end
+
+    # TODO: restrict to firm / provider applications
```
jsugarman, line 29: I believe we agreed that with TODOs that are to remain in the codebase we should add ticket reference

app/models/linked_application_type.rb
```diff
@@ -1,9 +1,15 @@
 class LinkedApplicationType
-  LinkTypeStruct = Struct.new(:code, :description)
+  LinkTypeStruct = Struct.new(:code, :description, :label, :hint)
```
jsugarman, line 2: I don't think we need the label and hint anymore.

app/views/providers/link_case/confirmations/show.html.erb
```diff
@@ -0,0 +1,32 @@
...
+          end
+
+          summary_list.with_row do |row|
+            row.with_key { t(".table.client") }
+            row.with_value { @linked_application.lead_application.applicant.full_name }
+          end
+        end %>
+
+    <%= form.govuk_radio_buttons_fieldset(:link_type_code, legend: { size: "l", tag: "h2", text: t(".question", application_ref: @linked_application.lead_application.application_ref) }) do %>
+      <% LinkedApplicationType.all.each do |f| %>
```
jsugarman, line 24: Extract query to controller form more idiomatic rails:

```suggestion
      <% @linked_application_types.each do |f| %>
```

```
# confirmations controller
def show
   ...
  @linked_application_types = LinkedApplicationType.all
end
```

spec/requests/providers/link_case/searches_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        expect(page).to have_error_message("Enter a valid application reference to search for")
+      end
+
+      it "does not create a linked application" do
+        expect { patch_request }.not_to change(LinkedApplication, :count).from(0)
+      end
+    end
+
+    context "when application reference was not found" do
+      let(:params) { { linked_application: { search_ref: "L-17B-RT1" }, continue_button: "Save and continue" } }
```
jsugarman, line 86: the value supplied to `continue_button` is irrelevant, so in the interests of having tests reflect that I tend to add something that makes that clear(er), either ""/blank or "somewordtoclarifyirrelevance"
```suggestion
      let(:params) { { linked_application: { search_ref: "L-17B-RT1" }, continue_button: "irrelevant" } }
```

spec/requests/providers/link_case/searches_controller_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+        expect(page).to have_error_message("The application reference entered cannot be found")
+      end
+
+      it "does not create a linked application" do
+        expect { patch_request }.not_to change(LinkedApplication, :count).from(0)
+      end
+    end
+
+    context "when form submitted using Save as draft button" do
+      let(:params) { { linked_application: { search_ref: "L-TVH-U0T" }, draft_button: "Save and come back later" } }
```
jsugarman, line 100: the value supplied to `draft_button` is irrelevant, so in the interests of having tests reflect that I tend to add something that makes that clear(er), either ""/blank or "somewordtoclarifyirrelevance"
```suggestion
      let(:params) { { linked_application: { search_ref: "L-TVH-U0T" }, draft_button: "irrelevant" } }
```

jsugarman (commented):
Looking good, I think we need to consider renaming some things. see comments. probably best to talk though

app/controllers/providers/link_case/confirmations_controller.rb
```diff
@@ -0,0 +1,55 @@
+module Providers
+  module LinkCase
+    class ConfirmationsController < ProviderBaseController
+      prefix_step_with :link_case
+
+      def show
+        @linked_application_types = LinkedApplicationType.all
+
+        if legal_aid_application.copy_case?
+          destroy_linked_application
```
jsugarman, line 10: what scenario does this cover, do we need this?

app/forms/link_case/confirmation_form.rb
```diff
@@ -0,0 +1,16 @@
+module LinkCase
+  class ConfirmationForm < BaseForm
+    form_for LinkedApplication
+
+    attr_accessor :link_type_code
+
+    validates :link_type_code, presence: true, unless: proc { draft? }
+
```
jsugarman, line 8: As discussed we could potentially destroy the link here using 

```ruby
after_validation do
  linked_application.destroy! unless link_type_code
end
```

app/forms/link_case/confirmation_form.rb
```diff
@@ -0,0 +1,16 @@
...
+    def save
+      return if invalid? || draft?
+
+      model.update!(link_type_code:)
```
jsugarman, line 12: I think we need to update the legal_aid_application in here as well since we jump over the question "Do you want to link the case?" when they have copied so it is not set. This would require retrieving the legal_aid_application, if not already available, then below should do it

```ruby
  legal_aid_application.update!(link_case: link_type_code.present?)
```
tests need to cover this too if we do this

app/controllers/providers/link_case/confirmations_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+          @form = ::LinkCase::ConfirmationForm.new(model: linked_application)
+        end
+      end
+
+      def update
+        @form = ::LinkCase::ConfirmationForm.new(form_params)
+
+        if @form.link_type_code.eql?("false")
+          destroy_linked_application
+        end
```
jsugarman, line 24: I think this logic should be owned by the form and perform as part of save (after_validation perhaps, see comment below)

app/controllers/providers/link_case/confirmations_controller.rb
jsugarman: to answer my own question i think this is to cover the flow of a copied case being direct to this page, skipping the invitation and search pages. In some looping scenarios this could result in an existing linked application which we would need to delete here.

app/controllers/providers/link_case/searches_controller.rb
```diff
@@ -16,12 +18,16 @@ def update
 
     private
 
+      def set_search_ref
+        @search_ref = existing_linked_application&.lead_application&.application_ref
```
jsugarman, line 22: This will need to be..

```ruby
   ... || form_params[:search_ref]
```
...to capture invalid params and redisplay them

app/forms/link_case/confirmation_form.rb
agoldstone93: If I destroy the model after validation, won't it fail when it gets to the `model.update` line in the `save` method?

app/forms/link_case/confirmation_form.rb
jsugarman: changes made in pairing session

app/controllers/providers/link_case/searches_controller.rb
jsugarman: actually you do not need

app/forms/link_case/confirmation_form.rb
```diff
@@ -0,0 +1,28 @@
+module LinkCase
+  class ConfirmationForm < BaseForm
+    form_for LinkedApplication
+
+    attr_accessor :link_type_code
+
+    validates :link_type_code, presence: true, unless: proc { draft? }
```
jsugarman, line 7: nit, this does the same thing as `proc { draft? }`

```suggestion
    validates :link_type_code, presence: true, unless: :draft?
```

config/routes.rb
```diff
@@ -165,6 +165,12 @@
         resource :search, only: %i[show update]
         resource :confirmation, only: %i[show update]
       end
+      namespace :link_case do
+        resource :invitation, only: %i[show update]
+        resource :search, only: %i[show update]
+        resource :confirmation, only: %i[show update]
+      end
+      resource :copy_case_confirmation, only: %i[show update]
```
jsugarman, line 173: this will need rebasing as i removed this old unnamespaced route after the fact

jsugarman: or just remove that line again

spec/forms/link_case/confirmation_form_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+
+  describe "#save" do
+    subject(:call_save) { instance.save }
+
+    context "with family link chosen" do
+      let(:link_type_code) { "FC_LEAD" }
+
+      it "updates the link_type_code" do
+        expect { call_save }.to change(linked_application, :link_type_code).from(nil).to("FC_LEAD")
+      end
```
jsugarman, line 30: we should probably have a context or contexts to capture the flows whereby `legal_aid_applicaiton#link_case` is updated too:

```ruby
context "when linking a copied case" do
  ...setup...
  
  it "updates link_case to true" do
    expect { call_save }.to change { legal_aid_applicaiton.link_case }.from(nil).to(true)
  end
end
```

spec/forms/link_case/confirmation_form_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    context "with legal link chosen" do
+      let(:link_type_code) { "LEGAL" }
+
+      it "updates the link_type_code" do
+        expect { call_save }.to change(linked_application, :link_type_code).from(nil).to("LEGAL")
+      end
```
jsugarman, line 38: [ditto](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5933/files#r1399292688)

spec/forms/link_case/confirmation_form_spec.rb
```diff
@@ -0,0 +1,125 @@
...
+    end
+
+    context "with no chosen" do
+      let(:link_type_code) { "false" }
+
+      it "does not update the link_type_code" do
+        expect { call_save }.not_to change(linked_application, :link_type_code).from(nil)
+      end
```
jsugarman, line 46: [ditto](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5933/files#r1399292688) but for update to nil

spec/forms/link_case/search_form_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+      search_ref:,
+    }
+  end
+
+  describe "#save" do
+    subject(:call_save) { instance.save }
+
+    context "with a valid application reference" do
+      let(:search_ref) { "L-TVH-U0T" }
+      let(:source_application) do
```
jsugarman, line 22: nit - linebreak before multine blocks

spec/forms/link_case/confirmation_form_spec.rb
jsugarman: plus we should have an example to exercise the destruction of the linked_application model instance.

```ruby
     it "does destroys the linked_application" do
        expect { call_save }.to change(LinkedApplication, :count).by(-1)
        # AND/OR
        call_save
        expect { linked_application.reload }.to raise_error(ActiveRecord::RecordNotFound)
      end
```

[examples](https://stackoverflow.com/questions/66108221/rspec-how-to-verify-if-a-record-has-been-deleted)

jsugarman (approved):
👍 great!

PR #5942, opened 2023-11-02
AP-4596: Amend modsec rule to prevent false positive

jsugarman:
## What
Amend modsec rule to prevent false positive

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4596)

The postcode TF1 5RG returns addresses on Union Street, Telford.

We believe the Word "Union" raises the anomaly score and thereby
violates ModSec SecRule 949110, "Inbound Anomaly Score Exceeded",
due to it being a possible SQL injection attack(?!).

The rule has been removed ONLY when the request URI contains the
relevant URI path name.

## Checklist


colinbruce (approved):
:+1:

PR #5948, opened 2023-11-03
Share the have_error_message helper

jsugarman:
## What
Mixin the have_error_message helper to all request specs

[Came out of story](https://dsdmoj.atlassian.net/browse/AP-4575)

use an extensible mixin for request spec capybara based helpers. Currently
there is only the have_error_message helper but there could be others in
the future or already existing that could be shared via this mechanims.

## Checklist


naseberry (approved):
Good spot and nice cleanup

PR #5968, opened 2023-11-08
AP-4597:Update required evidence text,add line break,and add accessib…

skdejong:
…ilty for delete button

## What

[Fix error and accessibility messages on upload evidence page](https://dsdmoj.atlassian.net/browse/AP-4597)

Fix error and accessibility messages on upload evidence page so that:

- The error message at the top of the page lists out _every_ document that is missing which needs to be uploaded. Due to this design requirement, customised code was needed instead of the GOVUK Form Builder component. 
- In situations when the client has failed to upload more than one of the required documents, in the “Uploaded files” drop zone each required document is listed on a separate line rather than as a run-on sentence. 
- When using screen readers, such as VoicOver, the delete link reads out "Delete filename.doc" instead of just “delete.”

## Checklist

### DIFF
--- a/app/views/providers/uploaded_evidence_collections/_upload_evidence.html.erb
+++ b/app/views/providers/uploaded_evidence_collections/_upload_evidence.html.erb
@@ -7,7 +7,7 @@
       local: true,
     ) do |form| %>
 
-  <%#= form.govuk_error_summary %>
+  <% if @uploaded_evidence_collection.errors.any? %>
     <div class="govuk-error-summary" data-module="govuk-error-summary">
       <div role="alert">
         <h2 class="govuk-error-summary__title">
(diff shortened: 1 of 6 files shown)


app/views/providers/uploaded_evidence_collections/_upload_evidence.html.erb
```diff
@@ -43,8 +60,8 @@
         <% if @mandatory_evidence_errors %>
             <p id="dropzone-mandatory-error" class='govuk-error-message dropzone-error govuk-!-margin-bottom-1'>
               <% @mandatory_evidence_errors.each do |error| %>
-                <span class="govuk-visually-hidden">Error:</span>
-                <%= error.message %>
+                  <span class="govuk-visually-hidden">Error:</span>
```
colinbruce, line 63: These two lines have been indented an extra step

colinbruce (approved):
LGTM 🏆 
One, very minor, suggestion.

app/views/providers/uploaded_evidence_collections/_upload_evidence.html.erb
skdejong: Thanks for spotting that. I have made the change.

colinbruce (approved):
Final updaye LGTM

PR #5982, opened 2023-11-13
Ap 4503 correction to ineligible result

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4503)

I have added code to handle the scenario where CFE returns 'ineligible' for multiple categories out of gross income, disposable income and capital and added a 'fake' CFE result to test this scenario, however, in practice this is not currently happening, see https://dsdmoj.atlassian.net/browse/AP-4503?focusedCommentId=342986 and https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/5982/files#r1392418804. This has been added as it is the desired way for CFE to work that is shown in the designs for this ticket but I am not sure if it is the best/correct approach?

## Checklist

### DIFF
--- a/app/helpers/ineligible_reasons_helper.rb
+++ b/app/helpers/ineligible_reasons_helper.rb
@@ -1,0 +1,40 @@
+module IneligibleReasonsHelper
+  def ineligible_reasons(cfe_result)
+    return "" unless cfe_result.assessment_result == "ineligible"
+
+    reasons = ineligible_reasons_array(cfe_result)
+    return reasons.first if reasons.length == 1
+
+    ineligible_reasons = ":<ul class='govuk-list govuk-list--bullet'>"
+    reasons.each { |reason| ineligible_reasons += "<li>#{reason}</li>" }
+    ineligible_reasons += "</ul>"
+  end
+
+private
+
+  def ineligible_reasons_array(cfe_result)
+    ineligible_reasons_array = []
+    ineligible_reasons_array << "gross income" if ineligible_gross_income?(cfe_result)
+    ineligible_reasons_array << "disposable income" if ineligible_disposable_income?(cfe_result)
+    ineligible_reasons_array << "disposable capital" if ineligible_disposable_capital?(cfe_result)
+    ineligible_reasons_array
+  end
+
+  def ineligible_gross_income?(cfe_result)
+    return false unless (cfe_result.gross_income_results - %w[ineligible]).empty?
+
+    true
+  end
+
+  def ineligible_disposable_income?(cfe_result)
+    return false unless (cfe_result.disposable_income_results - %w[ineligible]).empty?
+
+    true
+  end
+
+  def ineligible_disposable_capital?(cfe_result)
+    return false unless (cfe_result.capital_results - %w[ineligible]).empty?
+
+    true
+  end
+end
--- a/app/helpers/ineligible_reasons_helper.rb
+++ b/app/helpers/ineligible_reasons_helper.rb
@@ -14,27 +14,9 @@ private
 
   def ineligible_reasons_array(cfe_result)
     ineligible_reasons_array = []
-    ineligible_reasons_array << "gross income" if ineligible_gross_income?(cfe_result)
-    ineligible_reasons_array << "disposable income" if ineligible_disposable_income?(cfe_result)
-    ineligible_reasons_array << "disposable capital" if ineligible_disposable_capital?(cfe_result)
+    ineligible_reasons_array << "gross income" if cfe_result.ineligible_gross_income?
+    ineligible_reasons_array << "disposable income" if cfe_result.ineligible_disposable_income?(cfe_result)
+    ineligible_reasons_array << "disposable capital" if cfe_result.ineligible_disposable_capital?(cfe_result)
     ineligible_reasons_array
-  end
-
-  def ineligible_gross_income?(cfe_result)
-    return false unless (cfe_result.gross_income_results - %w[ineligible]).empty?
-
-    true
-  end
-
-  def ineligible_disposable_income?(cfe_result)
-    return false unless (cfe_result.disposable_income_results - %w[ineligible]).empty?
-
-    true
-  end
-
-  def ineligible_disposable_capital?(cfe_result)
-    return false unless (cfe_result.capital_results - %w[ineligible]).empty?
-
-    true
   end
 end
--- a/app/helpers/ineligible_reasons_helper.rb
+++ b/app/helpers/ineligible_reasons_helper.rb
@@ -3,20 +3,26 @@ module IneligibleReasonsHelper
     return "" unless cfe_result.assessment_result == "ineligible"
 
     reasons = ineligible_reasons_array(cfe_result)
-    return reasons.first if reasons.length == 1
+    return ineligible_reason(reasons.first) if reasons.length == 1
 
     ineligible_reasons = ":<ul class='govuk-list govuk-list--bullet'>"
-    reasons.each { |reason| ineligible_reasons += "<li>#{reason}</li>" }
+    reasons.each do |reason|
+      ineligible_reasons += "<li>#{ineligible_reason(reason)}</li>"
+    end
     ineligible_reasons += "</ul>"
   end
 
 private
 
+  def ineligible_reason(reason)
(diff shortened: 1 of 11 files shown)


spec/factories/cfe_results/v6/mock_results.rb
```diff
@@ -0,0 +1,619 @@
...
+          combined_disputed_capital: 0,
+          combined_non_disputed_capital: 0,
+          capital_contribution: 0.0,
+          pensioner_capital_disregard: 0.0,
+          combined_assessed_capital: 0.0,
+        }
+        result
+      end
+
+      # This result has has been faked, to cover an example where CFE returns 'ineligible'
```
kmahern, line 525: Is this approach correct?

app/helpers/ineligible_reasons_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+    ineligible_reasons_array << "gross income" if ineligible_gross_income?(cfe_result)
+    ineligible_reasons_array << "disposable income" if ineligible_disposable_income?(cfe_result)
+    ineligible_reasons_array << "disposable capital" if ineligible_disposable_capital?(cfe_result)
+    ineligible_reasons_array
+  end
+
+  def ineligible_gross_income?(cfe_result)
+    return false unless (cfe_result.gross_income_results - %w[ineligible]).empty?
+
+    true
```
jsugarman, line 26: using a positive would be much clearer as ineligible..., false unless...empty? is pretty confusing.

```ruby
def ineligible_gross_income?(cfe_result)
  cfe_result.gross_income_results.all? { |val| val == 'ineligible' }
end
```

app/helpers/ineligible_reasons_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+
+private
+
+  def ineligible_reasons_array(cfe_result)
+    ineligible_reasons_array = []
+    ineligible_reasons_array << "gross income" if ineligible_gross_income?(cfe_result)
+    ineligible_reasons_array << "disposable income" if ineligible_disposable_income?(cfe_result)
+    ineligible_reasons_array << "disposable capital" if ineligible_disposable_capital?(cfe_result)
+    ineligible_reasons_array
+  end
```
jsugarman, line 21: **not a blocker**: as mentioned below, even this method could belong to the backend/model and use keys rather than displayable values (such as `gross_income`). The presentation layer can then "translate" those keys. This maintains a "separation of concerns" IMO.

app/helpers/ineligible_reasons_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  def ineligible_reasons(cfe_result)
+    return "" unless cfe_result.assessment_result == "ineligible"
+
+    reasons = ineligible_reasons_array(cfe_result)
+    return reasons.first if reasons.length == 1
+
+    ineligible_reasons = ":<ul class='govuk-list govuk-list--bullet'>"
+    reasons.each { |reason| ineligible_reasons += "<li>#{reason}</li>" }
+    ineligible_reasons += "</ul>"
+  end
```
jsugarman, line 11: **not a blocker**, you could use `.html_safe` here since we know that no injection of html is being done from user input in this method. This should allow you to not need to `sanitize` in the view, as further below

jsugarman (commented):
looking good, some comments as we discussed.

app/helpers/ineligible_reasons_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+    return false unless (cfe_result.disposable_income_results - %w[ineligible]).empty?
+
+    true
+  end
+
+  def ineligible_disposable_capital?(cfe_result)
+    return false unless (cfe_result.capital_results - %w[ineligible]).empty?
+
+    true
+  end
```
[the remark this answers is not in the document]
kmahern: 👍 have moved to app/models/cfe/v6/result.rb

app/helpers/ineligible_reasons_helper.rb
kmahern: I agree, thank you, have made suggested change. 👍

app/helpers/ineligible_reasons_helper.rb
kmahern: Have moved the ineligible reasons into translations

app/helpers/ineligible_reasons_helper.rb
kmahern: I tried using html_safe, but it didn't seem to work particularly easily with this use case - also it might be that using sanitise rather than html_safe is the preferred solution? - see https://apidock.com/rails/String/html_safe. I have left as is for now...

spec/requests/providers/capital_income_assessment_results_controller_spec.rb
```diff
@@ -38,14 +38,36 @@
...
+            let(:cfe_result) { create(:cfe_v6_result, :ineligible_gross_income) }
 
-          it "returns http success" do
-            expect(response).to have_http_status(:ok)
+            it "returns http success" do
+              expect(response).to have_http_status(:ok)
+            end
+
+            it "displays the correct result" do
+              expect(unescaped_response_body).to include(I18n.t("ineligible.heading", name: applicant_name, scope: locale_scope))
```
jsugarman, line 49: **nit** I think it is clear and simpler and less coupled to simple expect the text/string that you expect

spec/requests/providers/capital_income_assessment_results_controller_spec.rb
```diff
@@ -38,14 +38,36 @@
...
-            expect(unescaped_response_body).to include(I18n.t("ineligible.heading", name: applicant_name, scope: locale_scope))
+          context "when ineligible and disposable income is above the upper threshold" do
+            let(:cfe_result) { create(:cfe_v6_result, :ineligible_disposable_income) }
+
+            it "returns http success" do
+              expect(response).to have_http_status(:ok)
+            end
+
+            it "displays the correct result" do
+              expect(unescaped_response_body).to include(I18n.t("ineligible.heading", name: applicant_name, scope: locale_scope))
```
jsugarman, line 65: **nit** I think it is clear and simpler and less coupled to simple expect the text/string that you expect

spec/requests/providers/capital_assessment_results_controller_spec.rb
```diff
@@ -46,6 +46,10 @@
           it "displays the correct result" do
             expect(unescaped_response_body).to include(I18n.t("ineligible.heading", name: applicant_name, scope: locale_scope))
```
jsugarman, line 47: **nit** not you but could refactor as for other comments "I think it is clear and simpler and less coupled to simple expect the text/string that you expect"

spec/factories/cfe_results/v6/mock_results.rb
```diff
@@ -0,0 +1,619 @@
...
+      # for more than one proceeding category, which it appears is currently not possible.
+      # Currently if an applicant is ineligible for disposable income and capital,
+      # CFE returns the disposable income result for proceedings as 'ineligible',
+      # but the capital result for proceedings as 'pending'.
```
jsugarman, line 529: You could add that this is due to a current "shortcut" in CFE that  may be removed in the future  - see [LEP-349](https://dsdmoj.atlassian.net/browse/LEP-349)

jsugarman (approved):
👍  A couple of test nits which I personally have been trying to address whenever i see them, plus you could add reference https://dsdmoj.atlassian.net/browse/LEP-349 in the code comment you added on the fake result payload.

spec/requests/providers/capital_assessment_results_controller_spec.rb
kmahern: 👍 have amended to use string instead of translation in spec expectation

spec/requests/providers/capital_income_assessment_results_controller_spec.rb
kmahern: 👍 have amended to use string instead of translation in spec expectation

spec/requests/providers/capital_income_assessment_results_controller_spec.rb
kmahern: 👍 have amended to use string instead of translation in spec expectation

PR #6000, opened 2023-11-16
Chore: Amend babel dependabot group

colinbruce:
## What

Slight amendment to the current babel group name, but today
16/Nov/23 we had two `@babel` dependabots opened that the
existing babel group did not trigger, adding the * prefix
should allow it to group wether it has the @ prefix or not

## Checklist


agoldstone93 (approved):
LGTM

PR #6001, opened 2023-11-16
Linked application amendments

jsugarman:
## What
Amend the code for family link type, refactor and amend specs

[Came out of story](https://dsdmoj.atlassian.net/browse/AP-4511)

The code require for a family link is FC_LEAD. Once injected into
CCMS the lead application has a link added to it of type FC_ASSOC.
Apply does not need to know about FC_ASSOC as we use a join table
instead.

## Checklist


kmahern (approved):
lgtm

PR #6010, opened 2023-11-17
AP-4610: Update partner means report

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4610)

Update the means report to ensure that the data is presented in the best way for caseworkers, splitting headings to client/partner and with the requested grouping

## Checklist

### DIFF
--- a/app/views/shared/means_report/_item.html.erb
+++ b/app/views/shared/means_report/_item.html.erb
@@ -1,12 +1,25 @@
 <% addendum = local_assigns[:addendum] || nil %>
 <% suppress_border ||= false %>
 <% css_border_klass = suppress_border ? "govuk-summary-list__row--no-border" : "" %>
-<div class="<%= "govuk-summary-list__row  normal-word-break #{css_border_klass}" %>" id="<%= "means-merits-report__#{name}" %>">
-  <dt class="<%= "govuk-summary-list__key govuk-!-width-one-half" %>">
-    <%= t(".#{scope}.#{name}", addendum:) %>
+<div class="govuk-summary-list__row  normal-word-break" id="<%= "means-merits-report__#{name}" %>">
+  <dt class="govuk-summary-list__key govuk-!-width-one-half">
+    <%= t(".#{scope}.client.#{name}", addendum:) %>
   </dt>
   <!--  in-line styling due to issues around pdf generation in our production envs -->
   <dd class="govuk-summary-list__value govuk-!-text-align-right">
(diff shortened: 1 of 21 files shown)


colinbruce:
@naseberry I assume we are okay to ignore sonarcloud?  It doesn't like the `<dt>` tags in a partial not being surrounded by a `<dl>` tag but that is in the parent page.  It renders like this:
image

Also, just spotted duplicate IDs so I'll get those refactored! 😬

naseberry:
> @naseberry I assume we are okay to ignore sonarcloud? It doesn't like the `<dt>` tags in a partial not being surrounded by a `<dl>` tag but that is in the parent page. It renders like this: image
> 
> Also, just spotted duplicate IDs so I'll get those refactored! 😬

Yes, in this case it can be ignored in SonarCloud.

app/views/shared/means_report/_item.html.erb
```diff
@@ -1,12 +1,25 @@
 <% addendum = local_assigns[:addendum] || nil %>
 <% suppress_border ||= false %>
 <% css_border_klass = suppress_border ? "govuk-summary-list__row--no-border" : "" %>
-<div class="<%= "govuk-summary-list__row  normal-word-break #{css_border_klass}" %>" id="<%= "means-merits-report__#{name}" %>">
+<div class="<%= "govuk-summary-list__row  normal-word-break" %>" id="<%= "means-merits-report__#{name}" %>">
```
naseberry, line 4: nit: `class="<%= "govuk-summary-list__row normal-word-break" %>"` are the erb opening and closing tags required or can they be removed?

app/views/shared/means_report/_item.html.erb
```diff
@@ -1,12 +1,25 @@
...
   <dt class="<%= "govuk-summary-list__key govuk-!-width-one-half" %>">
```
naseberry, line 5: ditto on erb tags

config/locales/en/shared.yml
```diff
@@ -213,6 +213,11 @@ en:
         deductions:
           dependants_allowance: Dependants allowance
           disregarded_state_benefits: Income from benefits excluded from calculation
+          partner_allowance: Partner Allowance
```
naseberry, line 216: nit: following the existing content case types, should `Partner Allowance` be `Partner allowance`

naseberry (approved):

PR #6017, opened 2023-11-20
AP-4575 [EXT]: Display copy case search ref if exists

jsugarman:
## What
"Remember" already searched for copy case reference during back paging

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4575)

when backpaging we can display the existing
application reference of a previous search
and copy flow. This means the user does not have to
go and retrieve it again if they went back inadvertently
or are attempting to get further back to amend details.

## Checklist


agoldstone93 (approved):
LGTM

PR #6030, opened 2023-11-22
AP-4575 [EXT]: Allow whitespace and case-insensitive searching

jsugarman:
## What
Allow whitespace and case-insenstive search for copy and link of cases

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4575)

So inadvertent whitespace or character case use does
not prevent users from finding otherwise valid case
references for copying or linking.

The addition of the class method to the base_form
means this could be used for other form attributes, such
as postcode (BaseAddressLookupForm)

## Checklist

### DIFF
--- a/app/forms/base_form.rb
+++ b/app/forms/base_form.rb
@@ -34,9 +34,9 @@ class BaseForm
     end
 
     def normalizes(name, with:)
-      send(:before_validation, proc {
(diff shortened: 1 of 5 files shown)


agoldstone93 (approved):
Looks good to me

PR #6043, opened 2023-11-27
Rubocop: Address save_bang exclusions

colinbruce:
## What

Remove the SaveBang exclusions from rubocop todo

This meant ensuring that all forms accept `save!` as well as save (these could be renamed and aliases removed later) and updating the linked_children form to ensure save_as_draft works as expected

## Checklist


app/forms/providers/proceeding_merits_task/linked_children_form.rb
```diff
@@ -16,21 +16,25 @@ def value_list
...
-        end
-        true
-      rescue StandardError
-        false
+        update_involved_children_on_proceeding
+      end
+
+      def save_as_draft
+        @draft = true
+        save! unless update_involved_children_on_proceeding || all_entries_blank?
```
jsugarman, line 25: is this expected to call the save defined above? As without `alias :save!, :save` it won't

jsugarman: was looking at this commit in isolation, i see the alias_mehod was added in a subsequent commit. 👍

jsugarman (approved):
Looks good to me.

PR #6055, opened 2023-11-29
delete any PVC named after the release [long named branch]

jsugarman:
## What
delete any dependabot persistent volume claims after build and deploy of branch

we have too many PVCs sitting around and have been asked to tidy them up. This mechanism is looking to use something that is not dependant on the PVC name so as to decouple if from the complex naming logic.

## Checklist


jsugarman:
success. deleting PVCS for deployments named after long branch names work fine

PR #6062, opened 2023-11-29
AP-3811 partner means assessment

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3811)

Updated the means assessment result page to show the partner assessment details.
This required creating the CFE V6 result model which implements many method to retrieve/calculate the partner financial results, overriding these methods from CFE V4 while ensuring these updated methods also still work for client-only applications. 
CFE V4 also had to be updated to take option args (**) to avoid throwing a NoMethod error when the updated V6 methods are called with `partner: true` on an old CFE V4 result
A new separate rspec file was created to test all these methods for an application with a partner `spec/models/cfe/v6/with_partner.rb`. A new mock result `with_partner` has been added. 
Existing rspecs were udpated

## Checklist

### DIFF
--- a/app/views/providers/capital_income_assessment_results/_bank_statements.html.erb
+++ b/app/views/providers/capital_income_assessment_results/_bank_statements.html.erb
@@ -1,7 +1,8 @@
-<h3 class="govuk-heading-m"><%= t(".title") %></h3>
+<h3 class="govuk-heading-m"><%= t(".title", individual:) %></h3>
 <%= govuk_table do |table| %>
   <%= table.with_body do |body| %>
     <%= body.with_row do |row| %>
+      <p><%= individual %></p>
       <%= row.with_cell(header: true, text: t(".uploaded")) %>
       <%= row.with_cell(numeric: true) do %>
         <ul class="govuk-list">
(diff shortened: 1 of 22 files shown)


kmahern (approved):
Probably worth getting someone familiar with the partner work to review but LGTM 👍

app/views/providers/capital_income_assessment_results/_bank_statements.html.erb
```diff
@@ -1,7 +1,8 @@
-<h3 class="govuk-heading-m"><%= t(".title") %></h3>
+<h3 class="govuk-heading-m"><%= t(".title", individual:) %></h3>
 <%= govuk_table do |table| %>
   <%= table.with_body do |body| %>
     <%= body.with_row do |row| %>
+      <p><%= individual %></p>
```
colinbruce, line 5: Is this needed or just debug? 

We rarely seem to use just <p> tags

colinbruce (approved):
Tricky to say it's 💯 but it looks fine and has no obvious spelling mistakes! 😄 
UAT will be the making of it 🤞

app/views/providers/capital_income_assessment_results/_bank_statements.html.erb
RoseSAK: Ah yes good catch, it is just a debug

PR #6075, opened 2023-12-04
Rubocop: Remove verified doubles violations

colinbruce:
## What

Address the remaining VerifiedDouble infractions
* Change the slack alerter to use the correct double, ActionMailer::MessageDelivery, that actually accepts `deliver_now!`
* Remove the WageSlip section of AttributeValueGenerator, this model does not exist and was written speculatively in 2019.
  Remove the tests for handling it
* Update AttributeValueGenerator to use actual models and their existing variables. 
  These handlers were created early in the system days before a lot of the actual code exists.  It had examples of a car having a regno as that is what CCMS expects but we ended up not using
* Update DeliveryMan tests to use instance_doubles of actual objects rather than anonymous doubles that respond to anything passed
* Update MalwareScanner test to use instance double of Process::Status
* Update ccms_soa_configuration_context to use a struct instead of a double.  This allows returning named variables as needed by the configuration block

## Checklist


agoldstone93 (approved):
LGTM

PR #6081, opened 2023-12-05
Rubocop: Address StubbedMocks issues

colinbruce:
## What

This PR removes uses of expect...to_return in test blocks

It, generally, replaces them with allow...to_return as this is the correct function.  
Where we were _expecting_ the return value to be "`X`" I have updated the tests to reflect and set proper expectations on the return value, or... deleted the rogue expectations!

## Checklist


spec/jobs/dashboard/updater_job_spec.rb
```diff
@@ -10,8 +10,10 @@ module Dashboard
 
         context "and job is not in the suspended list" do
           it "instantiates widget with the specified parameter" do
-            expect(Dashboard::Widget).to receive(:new).with("MyWidget").and_return(instance_double(Dashboard::SingleObject::Feedback, run: nil))
+            widget = instance_double(Dashboard::SingleObject::Feedback, run: nil)
+            allow(Dashboard::Widget).to receive(:new).with("MyWidget").and_return(widget)
             described_class.perform_now("MyWidget")
+            expect(widget).not_to receive(:run)
```
jsugarman, line 16: don't you need to use `have_received` after a call, or put expectation`to receive` before the call

```suggestion
            expect(widget).not_to have_received(:run)
```

jsugarman (approved):
one comment otherwise 👍

spec/jobs/dashboard/updater_job_spec.rb
colinbruce: Ran the update and I made a mistake when I extracted it... Now changed to 
```expect(widget).to have_received(:run)```
The not_to was a mistake but as it wasn't exercised properly it wasn't failing the test

PR #6107, opened 2023-12-12
AP-4604: Validation so "none" and another checkbox cannot both be selected

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4606)

Add extra validation so "None of the above" and another checkbox cannot both be selected when JS is disabled

## Checklist

### DIFF
--- a/app/forms/base_employed_form.rb
+++ b/app/forms/base_employed_form.rb
@@ -4,20 +4,33 @@ class BaseEmployedForm < BaseForm
 
   attr_accessor(*CHECK_BOXES_ATTRIBUTES)
 
-  validate :any_checkbox_checked_or_draft
-
-  def any_checkbox_checked?
-    CHECK_BOXES_ATTRIBUTES.map { |attribute| __send__(attribute) }.any?(&:present?)
-  end
+  validate :validate_any_checkbox_checked, unless: :draft?
+  validate :validate_none_and_another_checkbox_not_both_checked, unless: :draft?
 
 private
+
+  def any_checkbox_checked?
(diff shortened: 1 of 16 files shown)


app/forms/base_employed_form.rb
```diff
@@ -4,20 +4,33 @@ class BaseEmployedForm < BaseForm
...
+    checkbox_hash.values.any?(&:present?)
   end
 
-private
+  def none_and_another_checkbox_checked?
+    checkbox_hash[:none_selected].present? && checkbox_hash.except(:none_selected).values.any?(&:present?)
+  end
+
+  def checkbox_hash
+    CHECK_BOXES_ATTRIBUTES.index_with { |attribute| __send__(attribute) }
```
colinbruce, line 21: Is the change from `.map` to `.index_with` by design?
Is there an advantage?

agoldstone93: Unless I'm mistaken, it means I can access `checkbox_hash[:none_selected]` in line 17

agoldstone93: Because the hash being returned will be:
```
{
  :check_box_partner_offline_current_accounts=>nil,
  :check_box_partner_offline_savings_accounts=>"true",
  :no_partner_account_selected=>"true"
}
```

rather than:

```
[nil, "true", "true"]
```

(This example is from a different page but should be the same)

colinbruce: Cool, thanks for saving me having to think and or test it! 😆

colinbruce (approved):

PR #6116, opened 2023-12-13
AP-4703: Use task list component, update feature tests

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4703)

Use task list component
Update feature tests
Remove unused code
Remove helper tests associated with removed helper code
Update the colours of the tags based on latest design system guidance

Updated design:
127 0 0 1_8234_providers_applications_ae583183-d70c-44b5-9aaf-5961f30c1fb4_merits_task_list (1)

## Checklist


app/views/providers/merits_task_lists/_task_list_item.html.erb
```diff
@@ -1,16 +0,0 @@
...
-    <% if status != :waiting_for_dependency %>
-      <%= govuk_link_to formatted_name, url, aria_describedby: "#{name}-status", "aria-label": formatted_name %>
-    <% else %>
-      <p><%= formatted_name %></p>
-    <% end %>
-  </span>
-  <%= govuk_tag(
-        text: t(".states.#{status}"),
-        colour: tag_colour,
-        classes: "app-task-list__tag",
```
naseberry, line 13: I cannot find any further reference to `app-task-list__tag`, could this be removed from the stylesheet?

Also, it may be worth checking if the custom classes removed from these deleted views can also be removed elsewhere.

app/views/providers/merits_task_lists/show.html.erb
```diff
@@ -13,17 +13,41 @@
...
+              title: t(".#{task.name}"),
+              href: _task_url(task.name, @legal_aid_application, task.state),
+              status: task.state == :complete ? t(".states.#{task.state}") : govuk_tag(text: t(".states.#{task.state}")),
+            )
+          end
+        end %>
+
+    <h2 class="govuk-heading-l"><%= t(".proceeding_heading") %></h2>
+    <% @merits_tasks[:proceedings].each_with_index do |(ccms_code, proceeding), index| %>
+      <h3 class="govuk-heading-m"><%= index + 2 %>. <%= proceeding[:name] %></h3>
```
naseberry, line 31: `<%= index + 2 %>`  count seems off for heading sections, should count start from `1`?

agoldstone93: There's already a sneaky "1" on line 17. If this isn't immediately obvious, have you got any suggestions on how to make it clearer?

agoldstone93: Thinking about this more, it should probably be a `<ol>`, but I don't know a way of doing this because the headers (e.g. "Enter general case details") would sit outside the list. Maybe the page should be redesigned to accommodate this. What do you think @naseberry

agoldstone93: I could do something like `<ol start="2">` but that doesn't feel right

naseberry: Good thought process, lets discuss.

My original comment was more about the logical order of number.
Why would the section start from any number but 1, it seemed odd.

naseberry (approved):

PR #6124, opened 2023-12-14
Fix: npm Dependabot grouping

colinbruce:
## What

The previous configuration meant that dependabot would try and group all wildcards into a single PR

This meant, that when a problem occurred, it was hard to isolate the cause

## Checklist


naseberry (approved):
Thanks, I was meant to correct my mistake

PR #6145, opened 2023-12-20
AP-4706: remove task list classes

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4706)

Remove custom css where possible and replace with standard govuk classes
Rename from `app-task-list` to `sort-transactions`

## Checklist


colinbruce (approved):
LGTM :+1:

app/views/providers/income_summary/_income_type_item.html.erb
```diff
@@ -8,29 +8,24 @@
...
-      <div class="govuk-hint" style="margin-bottom: 0px">
-        <%= t("transaction_types.hints.#{name}") %>
-      </div>
-    <% end %>
     <% if error_message %>
       <p id='<%= "error-#{name}" %>' class="govuk-error-message govuk-!-padding-top-2"><%= error_message %></p>
     <% end %>
   </h2>
-  <%= content_tag(:div, class: "app-task-list__items", id: "list-item-#{name}") do %>
+  <%= content_tag(:div, class: "govuk-!-padding-bottom-9 govuk-!-padding-left-0", id: "list-item-#{name}") do %>
```
naseberry, line 23: nit: can we remove `content_tag` and have just a `<div>` instead?
It seems redundant here to ruse the rails tag helper when we can write the html tag.

naseberry (approved):
aside from my comment, everything else looks ok

app/views/providers/outgoings_summary/_outgoing_type_item.html.erb
```diff
@@ -8,24 +8,19 @@
...
-      <div class="govuk-hint" style="margin-bottom: 0px">
-        <%= t("transaction_types.hints.#{name}") %>
-      </div>
-    <% end %>
     <% if error_message %>
       <p id='<%= "error-#{name}" %>' class="govuk-error-message govuk-!-padding-top-2"><%= error_message %></p>
     <% end %>
   </h2>
-  <%= content_tag(:div, class: "app-task-list__items", id: "list-item-#{name}") do %>
+  <%= content_tag(:div, class: "sort-transactions__items", id: "list-item-#{name}") do %>
```
naseberry, line 18: nit: can we remove `content_tag` and have just a `<div>` instead?

agoldstone93: Yep, sorry, should've checked for other instances. Will fix now

config/locales/en/providers.yml
```diff
@@ -820,21 +820,15 @@ en:
       show:
         error: Select yes if they have received other benefits
     income_summary:
-      income_type_item:
```
colinbruce, line 825: Are these handled elsewhere now?  
I always get the fear when locale stuff is deleted, not moved! :)

colinbruce (approved):
One non-blocking observation, but otherwise additional changes look good :+1:

config/locales/en/providers.yml
agoldstone93: Sorry, was probably a bit heavy handed of me. I couldn't see where they were being used, but I can add them back in to be on the safe side

naseberry (approved):
nice improvements to the partials

config/locales/en/providers.yml
colinbruce: We do run an i18n test for expected keys being missing and if that didn't trigger it's probably okay :+1:

PR #6164, opened 2023-12-29
AP-4716: Update applicant search response parser

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4716)

Following the recent discovery of issues at the CCMS end of our submissions, issues with identifying applicants were discovered.  This PR addresses the following, newly identified, issues with the ApplicantSearchResponseParser:
* when an applicant matches surname and DOB but has no NINO, it would return `Number Not held`
* If two records for the same values were returned with `Number Not Matched` the previous equality check would fail because it would resolve to does `Number Not Matched` equal `Number Not MatchedNumber Not Matched`

This adds tests for the new scenarios and updates the validation checks for the new scenarios

While investigating I discovered that the existing XML example files did not match the current returns from CCMS so I amended them to better match the data that is currently returned.

## Checklist

### DIFF
--- a/app/lib/message_logger.rb
+++ b/app/lib/message_logger.rb
@@ -1,0 +1,7 @@
+module MessageLogger
+  def log_message(message)
+    message = "#{self.class}:: #{message}"
+    message = Time.zone.now.strftime("%F %T.%L ") + message if Rails.env.development?
+    Rails.logger.info message
+  end
+end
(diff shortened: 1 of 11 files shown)


naseberry (approved):

naseberry (approved):
👍  on the logging

PR #6215, opened 2024-01-10
AP-4726-Add partner income to "Review and print" page

skdejong:
## What

[Add partner income to 'Review and print' page](https://dsdmoj.atlassian.net/browse/AP-4726)

## Checklist

### DIFF
--- a/app/views/shared/check_answers/_housing_benefit.html.erb
+++ b/app/views/shared/check_answers/_housing_benefit.html.erb
@@ -1,4 +1,3 @@
-<h2 class="govuk-heading-l">Housing Benefit</h2>
 <div class="govuk-grid-row" id="app-check-your-answers__housing_benefit">
   <div class="govuk-grid-column-two-thirds">
     <h3 class="govuk-heading-m"><%= t(".heading") %></h3>
--- a/app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
+++ b/app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
@@ -15,7 +15,7 @@
       <% if @legal_aid_application.applicant.has_multiple_employments? %>
         <%= render(
               "shared/check_answers/full_employment_details",
-              individual: "Your client",
+              individual: t("generic.client"),
               read_only: true,
               partner: false,
             ) %>
@@ -47,7 +47,7 @@
   <section class="print-no-break">
     <%= render(
           "shared/check_answers/payments_summary",
-          payment_type: t(".credits-section-heading", individual: "your client"),
+          payment_type: t(".credits-section-heading", individual: t("generic.client")),
           url: @legal_aid_application.uploading_bank_statements? ? :regular_incomes : :identify_types_of_incomes,
           transaction_types: TransactionType.credits.without_disregarded_benefits.without_benefits,
           read_only: true,
@@ -62,7 +62,7 @@
           "shared/check_answers/cash_payments",
           read_only: true,
           individual: "Applicant",
-          individual_with_determiner: "your client",
+          individual_with_determiner: t("generic.client"),
           type: :income,
         ) %>
   </section>
@@ -71,7 +71,7 @@
 
   <%= render(
         "shared/check_answers/student_finance",
-        individual_text: "your client",
+        individual_text: t("generic.client"),
         individual: @legal_aid_application.applicant,
         read_only: true,
       ) %>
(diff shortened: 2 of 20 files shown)


app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
```diff
@@ -18,7 +22,7 @@
       <% else %>
         <%= render "shared/check_answers/employed_income", partner: false, read_only: true %>
       <% end %>
-    <% elsif @legal_aid_application.applicant_employed? %>
+    <% elsif @legal_aid_application.applicant.employed? %>
```
naseberry, line 25: The removed method `applicant_employed?` was using safe operator `&.`, I presume to avoid `NoMethodError`

Should this be applied if we want to be safe and not risk a `nil` error or is this no longer a concern?

spec/requests/providers/substantive_applications_controller_spec.rb
```diff
@@ -17,7 +17,7 @@
 
   before do
     login_provider
-    allow(legal_aid_application).to receive(:applicant_employed?).and_return(false)
+    # allow(legal_aid_application.applicant).to receive(:employed?).and_return(false)
```
naseberry, line 20: Can we action this comment?

app/views/shared/check_answers/_housing_benefit.html.erb
```diff
@@ -1,4 +1,3 @@
-<h2 class="govuk-heading-l">Housing Benefit</h2>
```
naseberry, line 1: Accessibility: does removing this `h2` break the heading order or is there a relevant heading in the parent partial?

app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
```diff
@@ -28,34 +32,30 @@
...
                read_only: true) %>
   <% end %>
 
+  <!-- INCOME -->
+
   <section class="print-no-break">
     <%= render(
           "shared/check_answers/payments_summary",
-          payment_type: t(".credits-section-heading"),
+          payment_type: t(".credits-section-heading", individual: "your client"),
```
naseberry, line 50: Localization: should the value `your client` be contains with a locales file?

app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
```diff
@@ -67,26 +67,32 @@
...
+  <!--OUTGOINGS -->
+
   <section class="print-no-break">
     <h2 class="govuk-heading-l"><%= t(".outgoings-heading") %></h2>
 
     <%= render(
           "shared/check_answers/payments_summary",
-          payment_type: t(".debits-section-heading"),
-          url: @legal_aid_application.uploading_bank_statements? ? :regular_outgoings : :identify_types_of_outgoings,
+          payment_type: t(".debits-section-heading", individual: "your client"),
```
naseberry, line 86: Localization: same as the above comment

naseberry:
Well spotted on the missing section and actioning it, just a few minor questions from me.

config/locales/en/shared.yml
```diff
@@ -377,11 +377,9 @@ en:
         employment: Employment income
         employment_details: Details
         partner:
-          heading: Partner employment income
```
colinbruce, line 380: This is worrying, it was only added in the prior ticket to allow Case Workers to get clarity on the differences between Client and Partner sections,  can we have a chat about the reasons for reversion?

app/views/shared/check_answers/_housing_benefit.html.erb
RoseSAK: Having this h3 here matches the other heading levels for the Income section. Overall though, the heading order on this page is quite broken due to partials being reused in multiple different places but this will be addressed in the future work on this page. The removal of this heading in particular was a specific request from content/design hence why it's included here

app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
RoseSAK: I assumed this method was a old method leftover from before we migrated a lot of methods (and data) away from the `LegalAidApplication` model into the specific `Partner` and `Applicant` models. This page was the only location where this method was called, hence I removed it and replaced with the applicant/partner methods. 

There are _many, many_ instances where we chain method calls on the applicant (or the partner) through the `@legal_aid_application` instance and do not use the safe operator! IMO it is highly unlikely that the applicant would not exist but perhaps this is something we should look at, so thanks for pointing it out.

app/views/shared/review_application/_income_payments_and_assets_for_bank_statement_upload.html.erb
RoseSAK: Yes, it could be, I will update this

colinbruce (approved):
LGTM :+1: 
🎉

PR #6231, opened 2024-01-16
Chore: Implement Slack alerts

colinbruce:
## What

This was started as an LnD with the hope that it would allow Slack alerting when events are triggered, e.g. CCMS submission fails, rack-attack is triggered

I span it out into it's own PR to make it easier to review/discuss

Basic use in app would be `Slack::SendMessage.call({ "text":"Hello, again, World!" })`
The messages can be quite complex and tested via the https://app.slack.com/block-kit-builder
This may require app access, I'm not sure at this time.

It also requires webhooks having been created via the slack API e..g in dev in can point to your personal chat channel so you can debug and not annoy the whole team! 

## Checklist


naseberry (approved):
looks good

PR #6237, opened 2024-01-17
AP-4465: Order banks alphabetically on view

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4465)

Put banks in alphabetical order, regardless of case. Mock bank stays at the top as it is only used for testing

## Checklist

### DIFF
--- a/app/controllers/citizens/banks_controller.rb
+++ b/app/controllers/citizens/banks_controller.rb
@@ -4,9 +4,12 @@ module Citizens
 
     def index
       remember_me(current_applicant)
+      @ordered_banks = ordered_banks
     end
 
     def create
+      @ordered_banks = ordered_banks
+
       if params[:provider_id].present?
         session[:provider_id] = params[:provider_id]
         session[:locale] = I18n.locale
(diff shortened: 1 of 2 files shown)


app/controllers/citizens/banks_controller.rb
```diff
@@ -16,5 +19,13 @@ def create
...
       end
     end
+
+  private
+
+    def ordered_banks
+      mock = TrueLayerBank.available_banks.select { |b| b[:display_name] == "Mock Bank" }
+      real_banks = TrueLayerBank.available_banks.sort_by { |bank| bank[:display_name].downcase }
+      mock + (real_banks - mock)
+    end
```
colinbruce, line 29: Could possibly simplify with an excluding syle operator, but not a blocker
```
def ordered_banks
  mock = TrueLayerBank.available_banks.select { |b| b[:display_name] == "Mock Bank" }
  real_banks = TrueLayerBank.available_banks.excluding(mock).sort_by { |bank| bank[:display_name].downcase }
  mock + real_banks 
end
```

colinbruce (approved):
Looks good, one non-blocking, fake code, suggestion :+1:

PR #6239, opened 2024-01-17
Fix: remove duplicate DocumentIdRequestor call

colinbruce:
## What

This came out of the CCMS dance review.

The DocumentIdRequestor was called on line 57 when it was extracted into a separate method.

It was then called it again and the end of the `update_document_and_return_response` method and, best case, did nothing with the duplicate call.

Worst case, it (correctly) saved number 00001 into the database and then logged 00002 into the submission history

This would have made tracking issues harder, and created empty rows in the CCMS database, but not have been a significant error

## Checklist

### DIFF
--- a/app/services/ccms/submitters/obtain_document_id_service.rb
+++ b/app/services/ccms/submitters/obtain_document_id_service.rb
@@ -53,8 +53,8 @@ module CCMS
         raise
       end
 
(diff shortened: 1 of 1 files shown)


agoldstone93 (approved):
Good spot. Looks good to me!

PR #6041, opened 2023-11-24
AP-4617: Use puffing-billy to stub feature test javascript API calls

jsugarman:
## What
Implement puffing-billy, an intercepting proxy server for JS requests, to prevent actual LFA (and other) API calls during  feature test runs.

The aim is not to fix all of them but introduce pattern that can be used going forward.

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4617)

So test suite can pass when external APIs are unavailable. LFA is not available between 22:00 and 07:00, for example.

<details>
<summary> Ensure all http client requests are intercepted by VCR</summary>
<a href="https://www.rubydoc.info/gems/vcr/VCR%2FConfiguration:hook_into">VCR::Configuration#hook_into</a>
<a href="https://github.com/vcr/vcr/blob/master/features/configuration/hook_into.feature">hook_into documentation</a>

The following http client libraries are used and therefore we need VCR
to intercept them.

| API              | Library   |
| ---------------- | --------- |
| Address lookup   | faraday   |
| Bank holidays    | Net::HTTP |
| LFA              | faraday   |
</details>

<details>
<summary> Ensure JS script initiated requests are intercepted</summary>
The test suite should pass locally when wifi is disabled
</details>

## TODO

- [x] Get screenshot helper working with billy chrome driver
- [x] Get worst offending JS reliant features stubbed - all should pass with wifi off
- [x] Document the setup, creation of and/or recording of stubs

## Offending tests

- [x] features/providers/applicant_details.feature:3 *1
- [x] features/providers/applicant_details.feature:132 
- [x] features/providers/applicant_details.feature:191 
- [x] features/providers/applicant_details.feature:267 
- [x] features/providers/applicant_details.feature:320 
- [x] features/providers/applicant_details.feature:380 *1
- [x] features/providers/regressions/scope_limitations.feature:3 *3
- [x] features/providers/merits_task_list_opponent.feature:24 *3
- [x] features/providers/check_provider_answers.feature:83 *3
- [x] features/providers/non_means_tested_journey/under_18_application_journey.feature:4 *3
- [x] features/providers/cost_override.feature:3  *3
- [x] features/providers/search_proceedings.feature:4 *3
- [x] features/providers/search_proceedings.feature:25 *3
- [x] features/providers/non_means_tested_journey/without_delegated_functions.feature:4 *2
- [x] features/providers/non_means_tested_journey/with_delegated_functions.feature:4 *2
- [x] features/providers/search_opponent_organisations.feature:15
[...]

PR #6272, opened 2024-01-24
ap-3772: attempt to fix failures on proceeding search

kmahern:
The purpose of this pr is to try and resolve (some) of the circleci integration test failures we are seeing. Looking [here](https://app.circleci.com/insights/gh/ministryofjustice/laa-apply-for-legal-aid/workflows/merge_pr/tests?branch=main) it appears the most common test failures are still on applicant_details.feature. We [previously](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/6041) attempted to resolve these by introducing puffing billy to stub calls made in javascript to LFA and I do believe this has worked when the test failed due to network connectivity, as the applicant_details.feature test passes for me when run locally with wifi switched off.

I therefore think the remaining failures are a result of a timing issue/race condition. The failures we are seeing on circle seem to look like this:-

`expected to find visible css "#proceeding-list .proceeding-item" but there were no matches. Also found "", "", "", "Non-molestation order\nFamily (Domestic abuse)", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", "", which matched the selector but not all filters.  (RSpec::Expectations::ExpectationNotMetError)
`

I noticed running the test locally (bundle exec cucumber features/providers/applicant_details.feature) that although we define a default_max_wait_time of 10 seconds [here](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/9d53972897c1f67923fd74e9e25b64e219e29d1e/features/support/env.rb#L62), if I force the test to fail, e.g. by changing [this line](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/9d53972897c1f67923fd74e9e25b64e219e29d1e/features/providers/applicant_details.feature#L14) to

`Then I search for proceeding 'XXX'`

it doesn't seem to wait for the 10 seconds for the expectation to be met. Adding a binding.pry to one the integration test steps revealed that (at least when running locally with bundle exec cucumber) default_max_wait_time was being set to 2 - the Capybara default - rather than 10, possibly not long enough for some asynchronous processes to to complete. I have amended the config in features/support/env.rb to apply the default_max_wait_time of 10 for all browsers. Do we think this is worth a go to see if it resolves the integration test failures.

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-3772)

## Checklist

### DIFF
--- a/docs/puffing_billy_recording.md
+++ b/docs/puffing_billy_recording.md
@@ -51,7 +51,7 @@ If you want to record a request and response, either to use in a feature test or
 3. run the feature (with debug to view request being made of billy's proxy)
 
   ```shell
-  DEBUG_BILLY=true cuke features/providers/applicant_details.feature:3
+  DEBUG_BILLY=true cucumber features/providers/applicant_details.feature:3
   ```
 
 4. The requests should appear in the configured folder
@@ -76,5 +76,5 @@ To rerecord a request cache simply delete the file and rerun the feature.
 You can display all calls that puffing-billy proxied using an env var as below.
 
 ```shell
-DEBUG_BILLY=true cuke features/providers/applicant_details.feature:3
+DEBUG_BILLY=true cucumber features/providers/applicant_details.feature:3
 ```
--- a/features/step_definitions/civil_journey_steps.rb
+++ b/features/step_definitions/civil_journey_steps.rb
@@ -887,7 +887,7 @@ Then(/^proceeding suggestions has (results|no results)$/) do |results|
   wait_for_ajax
   case results
   when "results"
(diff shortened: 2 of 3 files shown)


agoldstone93:
Does making these changes increase the wait time to 10 seconds when testing locally? If so, even if this doesn't address the issue, I can't see any harm in merging this. At the very least, the wait time will be what we (I assume) previously intended it to be based on that setting

colinbruce (approved):
Nice! 
Let's get this merged and monitor :+1:

PR #6254, opened 2024-01-22
Fix: Partner means

colinbruce:
## What

After a partner run through we saw some issues and this fixes some of them
* Update the means report income summary to use the combined total instead of the client total
* Update asset partial to show truelayer bank accounts as well as the manually entered values
* Update the CFECivil component list to ensure that truelayer data is sent when the partner uploads bank statements

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing: `bundle exec rake`
- Github should not be reporting conflicts; you should have recently run `git rebase main`.
- The standards in the [Git Workflow document on Confluence](https://dsdmoj.atlassian.net/wiki/spaces/ATPPB/pages/4602855954/Git+Workflow) should be followed
- There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- The PR description should say what you changed and why, with a link to the JIRA story.
- You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
- You should have checked that the commit messages say why the change was made.

PR #6274, opened 2024-01-24
AP-4759: Update partner means

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4759)

Based off the work on #6254 which was undertaken before the stakeholder run through - feel free to review entirely but the first three commits have been reviewed on the original PR

### Original work 
After a partner run through we saw some issues and this fixes some of them

* Update the means report income summary to use the combined total instead of the client total
* Update asset partial to show truelayer bank accounts as well as the manually entered values
* Update the CFECivil component list to ensure that truelayer data is sent when the partner uploads bank statements

### Updated work
* Update applicant_income_assessment view - Only switch when the client is uploading bank statements not
either party
* Update asset view - ensure the heading is only shown when bank account data
is present and should be shown

## Checklist

### DIFF
--- a/app/models/cfe/submission.rb
+++ b/app/models/cfe/submission.rb
@@ -6,7 +6,7 @@ module CFE
     has_many :submission_histories, -> { order(created_at: :asc) }, inverse_of: :submission
     has_one :result, class_name: "BaseResult", dependent: :destroy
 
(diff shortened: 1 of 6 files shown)


kmahern (approved):
lgtm

PR #6284, opened 2024-01-29
AP-4437: Remove additional element

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4437)

Remove hidden element which duplicates the page heading as this could be navigated to using a screen reader. Page should look visually identical

image

## Checklist

### DIFF
--- a/app/views/citizens/consents/show.html.erb
+++ b/app/views/citizens/consents/show.html.erb
@@ -5,30 +5,7 @@
       local: true,
     ) do |form| %>
 
-    <%= page_template page_title: t(".title_html"), head_title: t(".head_title"), form: do %>
-
-      <%= govuk_inset_text do %>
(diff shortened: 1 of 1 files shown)


colinbruce (approved):
LGTM :+1:

Does the page_template helper add an H1 block if the legend is hidden? 
This looks fine, but I can't see where the duplication happened before! 👀

agoldstone93:
> LGTM 👍
> 
> Does the page_template helper add an H1 block if the legend is hidden? This looks fine, but I can't see where the duplication happened before! 👀

@colinbruce I think it was the `template: :basic`. 

From the `page_template_helper`:
```
  # Where a minimal wrapper is needed (for example when the heading needs
  # to be positioned within a form), use the :basic template
  #
  #   <%= page_template(page_title: 'Some Page', template: :basic) do %>
  #     <p>Some content</p>
  #   <% end %>
```

PR #6292, opened 2024-01-30
AP-4774: Fix accessibility issues with hidden labels

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4774)

Remove instances of visually hidden text, that is read out with a screenreader, which duplicates visual text

## Checklist

### DIFF
--- a/app/views/providers/legal_aid_applications/search.html.erb
+++ b/app/views/providers/legal_aid_applications/search.html.erb
@@ -4,16 +4,13 @@
 
     <%= render partial: "shared/error" if @error %>
 
-    <%= form.govuk_fieldset legend: { text: page_title, tag: "h1", size: "xl" } do %>
-      <div class="govuk-!-padding-bottom-4"></div>
-      <%= form.govuk_text_field(
-            :search_term,
-            label: { text: :search_term, hidden: true },
-            caption: { text: t(".hint") },
-            width: "two-thirds",
-            value: @search_term,
-          ) %>
-    <% end %>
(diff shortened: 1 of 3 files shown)


app/views/providers/legal_aid_applications/search.html.erb
```diff
@@ -4,16 +4,13 @@
...
-            label: { text: :search_term, hidden: true },
-            caption: { text: t(".hint") },
-            width: "two-thirds",
-            value: @search_term,
-          ) %>
-    <% end %>
+    <%= form.govuk_text_field(
+          :search_term,
+          label: { text: "Search applications", size: "xl", tag: "h1" },
+          hint: { text: t(".hint") },
```
naseberry, line 10: My mistake, should this value be `caption` or `hint`?

naseberry:
Only one question.

Good work CYC-ing `list_from_translation_path` and removing those spacing divs

app/views/providers/legal_aid_applications/search.html.erb
agoldstone93: No, it should be a hint. This is what it looked like before:

image

And now:

image

So I think it's always been a hint. The caption (which would be above some text) was being used as a hint

naseberry (approved):

app/views/providers/legal_aid_applications/search.html.erb
naseberry: All good then

PR #6312, opened 2024-02-05
Faraday soap call

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4767)
[Link to story](https://dsdmoj.atlassian.net/browse/AP-4771)

Class to allow Faraday to make SOAP calls - this will allow us to replace Savon but the aim is to add this and allow us to update the Savon calls individually before removing the gem.

It was developed during a mob session and has been overseen by all current developers

## Checklist

### DIFF
--- a/app/services/faraday/soap_call.rb
+++ b/app/services/faraday/soap_call.rb
@@ -1,4 +1,6 @@
 module Faraday
+  class SoapError < StandardError; end
+
   class SoapCall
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
Developed during mobbing and agreed that it should be merged as as and enhanced as we start using it for actual calls

PR #6340, opened 2024-02-09
AP-4767: Use faraday for CCMS soap calls

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4767)

Replace Savon calls in the CCMS namespace with Faraday::SoapCall
Set the faraday_soap class to load in test: This is eager loaded in dev and production but ignored in test so needs an explicit require

## Checklist

### DIFF
--- a/app/services/ccms/requestors/base_requestor.rb
+++ b/app/services/ccms/requestors/base_requestor.rb
@@ -46,22 +46,6 @@ module CCMS
 
     private
 
-      def soap_client
-        @soap_client ||= Savon.client(
-          headers: { "x-api-key" => config.aws_gateway_api_key },
-          env_namespace: :soap,
-          wsdl: wsdl_location,
-          namespaces:,
-          pretty_print_xml: true,
-          convert_request_keys_to: :none,
-          namespace_identifier: "ns2",
-          log: false,
-          log_level: :debug,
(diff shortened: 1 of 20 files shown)


kmahern (approved):
LGTM

naseberry (approved):
Looks good

app/services/ccms/requestors/base_requestor.rb
```diff
@@ -46,22 +46,6 @@ def transaction_request_id
...
-          headers: { "x-api-key" => config.aws_gateway_api_key },
-          env_namespace: :soap,
-          wsdl: wsdl_location,
-          namespaces:,
-          pretty_print_xml: true,
-          convert_request_keys_to: :none,
-          namespace_identifier: "ns2",
-          log: false,
-          log_level: :debug,
-          open_timeout: 180,
```
kmahern, line 60: @colinbruce had a thought after approving this pr, we are removing the timeout here, have we tested how the timeouts are handled with Faraday?

colinbruce: I _think_ they should be handled by the SoapCall class and passed back as Faraday::SoapErrors.

Let me do some testing 👍

PR #6344, opened 2024-02-12
AP-4771: replace benefit checker savon calls

kmahern:
## What

[AP-4771: Replace BenefitChecker Savon calls with Faraday](https://dsdmoj.atlassian.net/browse/AP-4771)

## Checklist


app/services/benefit_check_service.rb
```diff
@@ -1,7 +1,15 @@
 class BenefitCheckService
+  class BenefitCheckError < StandardError; end
+
   BENEFIT_CHECKER_NAMESPACE = "https://lsc.gov.uk/benefitchecker/service/1.0/API_1.0_Check".freeze
   USE_MOCK = ActiveModel::Type::Boolean.new.cast(Rails.configuration.x.bc_use_dev_mock)
   REQUEST_TIMEOUT = 30.seconds
```
naseberry, line 6: Unused var: Is this still used or required?

app/services/benefit_check_service.rb
```diff
@@ -1,7 +1,15 @@
 class BenefitCheckService
+  class BenefitCheckError < StandardError; end
+
   BENEFIT_CHECKER_NAMESPACE = "https://lsc.gov.uk/benefitchecker/service/1.0/API_1.0_Check".freeze
```
naseberry, line 4: Nit: If var used only once, would it be ideal to add it to  the below `NAMESPACES`

app/services/benefit_check_service.rb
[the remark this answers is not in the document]
kmahern: 👍 have removed

app/services/benefit_check_service.rb
kmahern: 👍 have added to NAMESPACES

spec/services/benefit_check_service_spec.rb
```diff
@@ -30,8 +51,9 @@
       end
 
       it "sends the right parameters" do
-        allow(Savon).to receive(:client).and_return(savon_client)
-        expect(savon_client).to receive(:call).with(:check, expected_params)
+        allow(Faraday::SoapCall).to receive(:new).and_return(faraday)
+        allow(faraday).to receive(:call).with(payload).and_return(response)
+        expect(faraday).to receive(:call).with(payload)
```
agoldstone93, line 56: These two lines are almost identical - is that intentional? Do they both need to be there?

kmahern: @agoldstone93 rubocop seemed to demand the two separate lines (basically it doesn't allow you to specify the and_return with an expect but it also wants the test to have an expect not just an allow).

colinbruce: You could try switching it to:
```
it "sends the right parameters" do
        allow(Faraday::SoapCall).to receive(:new).and_return(faraday)
        benefit_check_service.call
        allow(faraday).to have_received(:call).with(payload).and_return(response)
end
```
but it may cause other issues! 🤷‍♂️

agoldstone93: I'd guess that Rubocop won't be happy with no `expect` in the block @colinbruce

skdejong: Yes, I just tried that and rubocop is not happy with the fact that there are no expectations found in this example.

colinbruce: 🤦‍♂️ I changed the received verb and forgot to change allow to expect!

skdejong:
LGTM

spec/services/benefit_check_service_spec.rb
agoldstone93: I'm not seeing these changes. Have you pushed @kmahern?

kmahern: @agoldstone93 apologies, should be on last commit now.

agoldstone93 (approved):

PR #6378, opened 2024-02-19
AP-4800: Postcode finder search results for correspondence

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4800)

Change address selection from select to radio buttons
Update page for when there is only one option
Update specs

To do:
- [x] Check change links
- [x] Update cassette names
- [x] Update locales for sentence with postcode info in it
- [x] Check the URL name

## Checklist


app/views/shared/address_selection/_confirm_address.html.erb
```diff
@@ -0,0 +1,42 @@
...
+      page_title: t(".heading"),
+      template: :basic,
+      form:,
+    ) do %>
+
+      <h1 class="govuk-heading-xl"><%= t(".heading") %></h1>
+
+      <p class='govuk-body'>1 address found for
+        <span class="govuk-!-font-weight-bold"><%= @form.postcode.insert(-4, " ") %></span>
+        <% unless @form.model.building_number_name.empty? %>
```
colinbruce: This is an unless, but the other form is `if @form.model.building_number_name.present?`
Should they be consistent?

colinbruce (approved):
One, non-blocking, consistency, issue but otherwise LGTM 👍

app/controllers/concerns/address_handling.rb
```diff
@@ -3,9 +3,16 @@ module AddressHandling
 
 private
 
+  def filter_addresses
```
kmahern, line 6: Feels like it would be worth having a spec/specs to explicitly test the address filtering but I don't think I can see any? Possibly something to add to the address_selections_controller_request_spec?

app/views/shared/address_selection/_confirm_address.html.erb
agoldstone93: I think I've updated this in a subsequent commit. Should probably have done it in the same commit, but they are consistent now

app/controllers/concerns/address_handling.rb
agoldstone93: @kmahern I've added another spec in the second last commit

app/views/shared/address_selection/_no_address.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      form:,
+    ) do %>
+
+      <h1 class="govuk-heading-xl"><%= page_title %></h1>
+
+      <p class="govuk-body">
+        <% if @form.model.building_number_name.present? %>
+            <%= t(".addresses_found_long_html",
+                  number: @address_collection.count,
+                  postcode: @form.postcode.insert(-4, " "),
```
naseberry, line 13: Could we move `postcode.insert(-4, " ")` to the model or can we use the `pretty_postcode` method?

app/views/shared/address_selection/_no_address.html.erb
```diff
@@ -0,0 +1,27 @@
...
+                  building_name_number: @form.model.building_number_name) %>
+          <% else %>
+            <%= t(".addresses_found_html",
+                  number: @address_collection.count,
+                  postcode: @form.postcode.insert(-4, " ")) %>
```
naseberry, line 18: Could we move `postcode.insert(-4, " ")` to the model or can we use the `pretty_postcode` method?

app/views/shared/address_selection/_no_address.html.erb
agoldstone93: @naseberry I've updated this in the last commit

PR #6413, opened 2024-02-27
Fix: reduce Staging sentry alerts

colinbruce:
## What

The overnight jobs all require a database to function but, as we turn it off overnight, they all fail and cause errors which eat into the sentry allowance.
image

This PR tries adding a 7 hour offset to staging to allow them to run once the DB has restarted.

It currently adds a 1 hour offset to UAT so I can test if the theory works without deploying to staging and potentially breaking everything!

It sets an offset of 0 otherwise so production remain at the normal scheduled times

## Checklist


kmahern (approved):
👍

PR #6464, opened 2024-03-18
AP-4906: Update default country handling for addresses

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4906)

When adding Non-UK countries we forgot to set the default to GBR when storing a new address, this updates the creation of address and ensures they get set to the default value, GBR

## Checklist

### DIFF
--- a/app/controllers/providers/address_lookups_controller.rb
+++ b/app/controllers/providers/address_lookups_controller.rb
@@ -18,7 +18,7 @@ module Providers
     end
 
     def address
-      applicant.address || applicant.build_address(location: "correspondence")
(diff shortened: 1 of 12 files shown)


kmahern:
@colinbruce unfortunately I don't think this fix is working, at least not in the scenario I have tested where the user has used address lookups. I think this is because the applicant.address and applicant.home_address are already built in the address_lookups_controller and the home_address_lookups_controller so build_address won't get called in the address_selections and home_address_selections controllers.

kmahern (approved):
lgtm 👍

PR #6475, opened 2024-03-20
AP-4728: Update CCMS address handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4728)

This PR is to update the address handling when submitting data to CCMS
It updates the Applicant model:
* with a `home_address_for_ccms` method that checks for a home address and falls back to the correspondence address if one is not found
* it does not send a home address if the provider has said they have no fixed residence

It updates the ApplicantAddRequestor to:
* use the new method from the Applicant model
* to submit a calculated value to `clientbio:NoFixedAbode` rather than using a fixed, `false` response

It updates the Address model and AddressHelper to return the country value if it is anything other than GBR, this allows us to display the chosen country code on CYA, review pages etc for non-UK home addresses

## Checklist

### DIFF
--- a/app/helpers/address_helper.rb
+++ b/app/helpers/address_helper.rb
@@ -6,7 +6,8 @@ module AddressHelper
               address.address_line_two,
               address.city,
               address.county,
-              address.pretty_postcode].compact.compact_blank.join("<br>"), tags: %w[br]
+              address.pretty_postcode,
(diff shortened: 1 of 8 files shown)


kmahern (approved):
lgtm 👍

PR #6489, opened 2024-03-22
AP-4909: Improve CCMS client matching

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4909)

Update ApplicantSearchResponseParser
> Check actual provider values against CCMS provided values rather than checking 
for `Number Matched` and assuming that the record is correct.  
This has a specific set of rules to comply with, documented on the ticket

Update ObtainApplicantReferenceService
> Ensure the parser is sent the applicant for the current application
It also updates the tests to generate a named applicant that matches the new xml test data

Update BaseResponseParser
> Remove the `ni_number_match` method.  It was the initial "fix" for improving data matching and this whole PR is to replace it, and the original process, with a much more detailed matching process.  
As the only call is now removed, this method can be deleted

Update the xml examples
> Create new examples that make the new rules clearer
Delete the old examples that are no longer used

## Checklist


app/services/ccms/parsers/applicant_search_response_parser.rb
```diff
@@ -36,8 +49,34 @@ def extracted_record_count
...
+        end
+      end
+
+      def parse_client(client_struct)
+        client_struct.match = [
+          @applicant.first_name.first.casecmp?(client_struct.first_initial),
+          @applicant.last_name.casecmp?(client_struct.last_name),
+          @applicant.surname_at_birth.casecmp?(client_struct.last_name_at_birth),
+          @applicant.date_of_birth.strftime("%Y-%m-%d").eql?(client_struct.date_of_birth),
+          (@applicant.national_insurance_number.nil? ? "" : @applicant.national_insurance_number).casecmp?(client_struct.national_insurance_number),
```
kmahern, line 74: possibly a bit briefer/more readable to just have have (@applicant.national_insurance_number || "").casecmp?(client_struct.national_insurance_number)

kmahern (approved):
lgtm 👍

app/services/ccms/parsers/applicant_search_response_parser.rb
colinbruce: Thanks @kmahern I remember saying I would investigate the new `.presence` method and then forgot!
I think that would make it 
```ruby
(@applicant.national_insurance_number.presence || "").casecmp?(client_struct.national_insurance_number)
```
I'll check tomorrow :+1:

colinbruce: Reviewed and yours was even simpler than mine! 🎉 
Updated :+1:

naseberry (approved):

PR #6506, opened 2024-03-28
ap-4913: build a new home address if changing between overseas and uk home address

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4913)

The overseas address page was displaying a uk home address if one had been entered previously.

On further investigation it was found that that there was also a similar issue with displaying overseas addresses on pages that should only display uk addresses.

The country was also consequently not being updated correctly when swapping between uk and non-uk home addresses;  i.e. if an overseas address home address was entered and then a uk home address the home address still had the overseas country.

Fix - introduce current_non_uk_home_address in non-uk-home_addresses_controller to check whether a home address exists and whether it is an overseas home address, if both are true use the existing home address if not build a new home address.

Introduce similar (reversed) logic in the home_address/home_address_lookups_controller and home_address/addresses_controller with a  new current_uk_home_address method.

## Checklist

### DIFF
--- a/app/controllers/providers/home_address/addresses_controller.rb
+++ b/app/controllers/providers/home_address/addresses_controller.rb
@@ -32,7 +32,13 @@ module Providers
       end
 
       def address
-        @address ||= applicant.home_address || build_address
+        @address ||= current_uk_home_address || build_address
+      end
+
+      def current_uk_home_address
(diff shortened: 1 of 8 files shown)


colinbruce (approved):
LGTM 👍

PR #6508, opened 2024-03-28
AP-4910: Interrupt flow in progress

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4910)

WIP - checking coverage

## Checklist

### DIFF
--- a/app/views/providers/interrupt/blocks/show.html.erb
+++ b/app/views/providers/interrupt/blocks/show.html.erb
@@ -1,0 +1,19 @@
+<div class="interruption-panel">
+  <%= page_template(page_title: t(".title_html"), column_width: "full", template: :basic) do %>
+    <%= page_heading(size: "l") %>
+
+    <div class="maximize-text-width">
+      <p class="govuk-body"><%= t(".body_text") %></p>
+      <p class="govuk-body"><%= t(".options") %></p>
+      <ul class="govuk-list govuk-list--bullet">
+        <% t(".option_bullets").each do |bullet| %>
+          <li><%= bullet %></li>
+        <% end %>
+      </ul>
+    </div>
+
+    <div class='govuk-button-group'>
+      <%= govuk_button_link_to t(".applications"), providers_legal_aid_applications_path, class: "white-button", role: "button" %>
+    </div>
+  <% end %>
+</div>
--- a/app/controllers/providers/interrupt/blocks_controller.rb
+++ b/app/controllers/providers/interrupt/blocks_controller.rb
@@ -1,0 +1,7 @@
+module Providers
+  module Interrupt
+    class BlocksController < ProviderBaseController
+      def show; end
+    end
+  end
+end
--- a/app/helpers/providers_helper.rb
+++ b/app/helpers/providers_helper.rb
@@ -4,7 +4,11 @@ module ProvidersHelper
   end
 
   def url_for_application(legal_aid_application)
-    name = legal_aid_application.provider_step.presence || :proceedings_types
+    name = if legal_aid_application.expired?
+             :providers_blocked
+           else
+             legal_aid_application.provider_step.presence || :proceedings_types
+           end
 
     Flow::ProviderFlowService.new(
       legal_aid_application:,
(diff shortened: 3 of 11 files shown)


app/views/providers/interrupt/blocks/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+      <p class="govuk-body"><%= t(".options") %></p>
+      <ul class="govuk-list govuk-list--bullet">
+        <% t(".option_bullets").each do |bullet| %>
+          <li><%= bullet %></li>
+        <% end %>
+      </ul>
+    </div>
+
+    <div class='govuk-button-group'>
+      <%= govuk_button_link_to t(".applications"), providers_legal_aid_applications_path, class: "white-button", role: "button" %>
```
agoldstone93, line 16: Is the `white-button` class just styling the button to look like [this inverse button that's in the govuk components](https://govuk-components.netlify.app/helpers/button/#inverse-buttons)? Non-blocking, but we could consider removing the custom css at some point

app/views/providers/interrupt/blocks/show.html.erb
```diff
@@ -0,0 +1,19 @@
...
+    <%= page_heading(size: "l") %>
+
+    <div class="maximize-text-width">
+      <p class="govuk-body"><%= t(".body_text") %></p>
+      <p class="govuk-body"><%= t(".options") %></p>
+      <ul class="govuk-list govuk-list--bullet">
+        <% t(".option_bullets").each do |bullet| %>
+          <li><%= bullet %></li>
+        <% end %>
+      </ul>
```
agoldstone93, line 12: Similarly, in the latest version of the components, there's [this way of adding lists](https://govuk-components.netlify.app/helpers/list/) which looks like it might be useful for us. Presumably, it would be as simple as something like:
```
<%= govuk_list t(".option_bullets"), type: :bullet %>
```

Non-blocking, but something to consider in the future perhaps?

spec/helpers/providers_helper_spec.rb
```diff
@@ -112,5 +136,31 @@
...
     end
+
+    context "when the application predates the 2023 surname at birth issue" do
+      let(:legal_aid_application) do
+        travel_to Date.parse("2023-12-25") do
+          create(:legal_aid_application, :with_multiple_proceedings_inc_section8, provider_step:)
+        end
+      end
+
+      context "and the provider step should expire it" do
```
agoldstone93, line 147: Sorry if this is too picky, but I don't think it's clear to me what 'should expire it' means here. Same for the next test (line 156)

spec/helpers/providers_helper_spec.rb
```diff
@@ -112,5 +136,31 @@
...
+        it "routes to the block page" do
+          application_id = legal_aid_application.id
+          expect(url_for_application(legal_aid_application)).to eq("/providers/applications/#{application_id}/out-of-date-application?locale=en")
+        end
+      end
+
+      context "and the provider step should not expire it" do
+        let(:provider_step) { "submitted_applications" }
+
+        it "routes to the block page" do
```
agoldstone93, line 159: should this say something like "routes to the submitted application page" instead?

agoldstone93 (approved):
Some minor comments about wording in specs and possible changes we could make that don't have to be part of this PR. Overall LGTM!

app/views/providers/interrupt/blocks/show.html.erb
colinbruce: Yes, and no, we have some additional styling (bolder text, drop shadows etc) but the styling will affect multiple pages that use it.
We could add a ticket to remove the `white-button` class and replace all uses with inverse

app/views/providers/interrupt/blocks/show.html.erb
colinbruce: Another good shout for a future PR - main is only on 5.2.1 of govuk.components and the govuk_list was added in 5.2.2

app/views/providers/interrupt/blocks/show.html.erb
colinbruce: Ticket created

naseberry (approved):

PR #6543, opened 2024-04-09
Fix: Resolve flickering address test

colinbruce:
## What

The default factory for address populates address_line_one with a building number.  
Sometimes the randomly generated building number is included elsewhere in the HTML rendering this test failing.  

In this failing screenshot, the house number was set to 268, which is part of the colour polyfill for the govuk logo!
image

This PR replaces the one off flickery test with a building_name to prevent it recurring.  
Changing the factory could have affected many more tests and caused more fragility so was left as-is for now. 
If we see more of this type of flickering, it may be worth switching the factory to always use street_name.

## Checklist


kmahern (approved):
LGTM 👍

PR #6575, opened 2024-04-17
AP-4828: Add copy case question

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4828)

This adds the copy case question page required by the link/copy case feature
When yes is selected it sets a flash message to be displayed on the next page, when no is selected it routes to the next step in the standard flow, proceedings

## Checklist

### DIFF
--- a/app/views/providers/link_application/copies/show.html.erb
+++ b/app/views/providers/link_application/copies/show.html.erb
@@ -1,0 +1,38 @@
+<%= form_with(
+      model: @form,
+      url: providers_legal_aid_application_link_application_copy_path,
+      method: :patch,
+      local: true,
+    ) do |form| %>
+
+  <%= page_template(
+        page_title: t(".question", lead_application_reference: @lead_application_reference),
+        form:,
+        template: :basic,
+      ) do %>
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-1"><%= t(".heading") %></h1>
+    <%= govuk_summary_list(actions: false, classes: "govuk-summary-list--no-border") do |summary_list|
+          summary_list.with_row do |row|
+            row.with_key { t(".reference") }
+            row.with_value { @lead_application.application_ref }
+          end
+          summary_list.with_row do |row|
+            row.with_key { t(".client") }
+            row.with_value { @lead_application.applicant_full_name }
+          end
+          summary_list.with_row do |row|
+            row.with_key { t(".proceedings") }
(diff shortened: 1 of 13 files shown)


app/views/providers/link_application/copies/show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      local: true,
+    ) do |form| %>
+
+  <%= page_template(
+        page_title: t(".question", lead_application_reference: @lead_application_reference),
+        form:,
+        template: :basic,
+      ) do %>
+    <h1 class="govuk-heading-xl govuk-!-margin-bottom-1"><%= t(".heading") %></h1>
+    <%= govuk_summary_list(actions: false, classes: " govuk-summary-list--no-border govuk-!-margin-bottom-20") do |summary_list|
```
agoldstone93, line 14: Is there such thing as a `govuk-!-margin-bottom-20`? I thought it only went up to 9?

spec/services/flow/steps/linked_applications/copy_step_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+RSpec.describe Flow::Steps::LinkedApplications::CopyStep do
+  let(:legal_aid_application) { create(:legal_aid_application, copy_case:) }
+  let(:copy_case) { true }
+
+  describe "#path" do
+    subject { described_class.path.call(legal_aid_application) }
+
+    it { is_expected.to eql "/providers/applications/#{legal_aid_application.id}/link_application/copy?locale=en" }
```
agoldstone93, line 10: If we agree to proceed with url helpers (_path) rather than explicit urls, maybe this could be changed?

agoldstone93 (approved):
A couple of minor comments

spec/services/flow/steps/linked_applications/copy_step_spec.rb
colinbruce: Good spot - I wrote this before we had the conversation! 😄

app/views/providers/link_application/copies/show.html.erb
colinbruce: Weird - I know I was having spacing issues. I think this was _supposed_ to be a 2 bit I added 0.
A bit of playing now and it's not needed at all now! 🎉

agoldstone93 (approved):

PR #6583, opened 2024-04-22
AP-4679: Flow refactor applicant details steps

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4679)

## Checklist


app/services/flow/steps/providers_home_step.rb
```diff
@@ -0,0 +1,7 @@
+module Flow
+  module Steps
+    ProvidersHomeStep = Step.new(
+      path: ->(_application) { urls.providers_legal_aid_applications_path },
```
colinbruce, line 4: Can we use the same pattern for paths that don't need an application?

either `->(_application)` or ` ->(_)` that was used on applicants_step

spec/requests/providers/applicants_controller_spec.rb
```diff
@@ -59,11 +59,6 @@
       )
     end
 
-    it "redirects to next page" do
```
colinbruce, line 62: Should we still test that it redirects, rather than re-render?

Does it add any value?

spec/services/flow/steps/addresses/correspondence_address_lookups_step_spec.rb
```diff
@@ -0,0 +1,29 @@
+require "rails_helper"
+
+RSpec.describe Flow::Steps::Addresses::CorrespondenceAddressLookupsStep, type: :request do
```
colinbruce, line 3: Do these new step tests need `, type: :request`?

colinbruce (commented):
Looking good, a couple of questions it would be good to discuss, but no majot blockers

spec/services/flow/steps/addresses/correspondence_address_lookups_step_spec.rb
agoldstone93: I added this in because without it, the '_path' helpers didn't work

PR #6597, opened 2024-04-25
AP-4952: Bug - move any step with no path into new step pattern

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4952)

The bug listed was caused by a provider reaching a page with no flow step and then closing the webpage or returning to the application list.  The list then tried to reach a flow step with no path, errored and re-directed the user to the first page.

This corrects the error seen and all other steps with no path.

It might be easier to review each commit in turn as the changes get scattered in the whole PR review.

The correspondence manual controller tests had already been updated to the new flow and required no additional changes, in case you wonder why it's the only controller with no tests changed!

## Checklist

### DIFF
--- a/app/services/flow/flows/provider_capital.rb
+++ b/app/services/flow/flows/provider_capital.rb
@@ -55,15 +55,7 @@ module Flow
             end
           end,
         },
-        remove_vehicles: {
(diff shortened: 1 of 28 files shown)


kmahern (approved):
lgtm 👍

PR #6655, opened 2024-05-10
AP-4992: Update linking case back button handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4992)

This PR updates the views, forms and flow for each of the linking cases pages and updated to allow the use of back buttons, it also updates the CYA page to use the new values 

>[!NOTE]
A future PR will have to delete the linked_application if user has picked no

## Checklist

### DIFF
--- a/app/forms/providers/link_application/confirm_link_form.rb
+++ b/app/forms/providers/link_application/confirm_link_form.rb
@@ -1,20 +1,15 @@
 module Providers
   module LinkApplication
     class ConfirmLinkForm < BaseForm
-      form_for LegalAidApplication
+      form_for LinkedApplication
 
-      attr_accessor :link_case
+      attr_accessor :confirm_link
 
-      validates :link_case, presence: true, unless: :draft?
+      validates :confirm_link, presence: true, unless: :draft?
 
       def save
-        if link_case == "No"
-          model.update!(link_case: nil)
-          return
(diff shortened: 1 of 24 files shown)


app/forms/providers/link_application/confirm_link_form.rb
```diff
@@ -1,20 +1,15 @@
...
   module LinkApplication
     class ConfirmLinkForm < BaseForm
-      form_for LegalAidApplication
+      form_for LinkedApplication
 
-      attr_accessor :link_case
+      attr_accessor :confirm_link
 
-      validates :link_case, presence: true, unless: :draft?
+      validates :confirm_link, presence: true, unless: :draft?
```
kmahern, line 8: Are we still using the link_case column on the legal_aid_applications table for anything? If not should it be removed?

kmahern (approved):
One comment but could be done in a separate pr if necessary. Otherwise lgtm 👍

app/forms/providers/link_application/confirm_link_form.rb
colinbruce: I was thinking of re-using it for once the change is confirmed we set it true or false for reporting purposes, while deleting the linked_application object if it's false :+1:

PR #6661, opened 2024-05-13
Fix: Flickery linked Application proceeding test

colinbruce:
## What

This was caused by a race condition when setting up the proceedings, the test expected `A` -> `B`, 
but the setup sometimes created `B` a fraction of a second before `A`

To replicate the failure on main, duplicating the setup from the failing Circle job:
```shell
TEST_FILES=(spec/services/ccms/requestors/case_add_requestor_xml_blocks/non_passported_attributes_spec.rb
           spec/requests/providers/partners/bank_statements_controller_spec.rb
           spec/requests/pages_controller_spec.rb
           spec/models/legal_aid_application_spec.rb)

bundle exec rspec --seed 56013 --fail-fast -- ${TEST_FILES}
```

This changes the test to expect the text in the same order

## Checklist


kmahern (approved):
👍

PR #6716, opened 2024-05-23
AP-5048: Add notification banner to check merits answer page

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5048)

Add notification banner to check merits answer page to show user how the information was populated if they have copied an application. Added notification banner as an attribute to the page template so that it is in right place within the HTML, so that the banner is the correct width, and as we are increasingly using banners in the service.

## Screenshot

image

## Checklist


app/views/providers/check_merits_answers/show.html.erb
```diff
@@ -1,6 +1,10 @@
+<% notification_banner_title = t("generic.important") if @legal_aid_application.lead_linked_application %>
```
kmahern, line 1: Is this correct. Should it be if @legal_aid_application.copy_case?

app/views/providers/check_merits_answers/show.html.erb
```diff
@@ -1,6 +1,10 @@
...
+
 <%= page_template(
       page_title: t(".h1-heading"),
       back_link: { path: reset_providers_legal_aid_application_check_merits_answers_path(@legal_aid_application), method: :patch },
+      notification_banner_title:,
+      notification_banner_text: t(".banner_text", application_ref: @legal_aid_application.lead_linked_application&.lead_application&.application_ref),
```
kmahern, line 7: Should this be this be something like LegalAidApplication.find(@legal_aid_application.copy_case_id).application_ref ?

kmahern (approved):
👍

PR #6718, opened 2024-05-23
Ap 4683 flow refactor proceedings and limitations

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4683)

## Checklist

### DIFF
--- a/app/services/flow/steps/provider_start/has_other_proceedings_step.rb
+++ b/app/services/flow/steps/provider_start/has_other_proceedings_step.rb
@@ -1,0 +1,16 @@
+module Flow
+  module Steps
+    module ProviderStart
+      HasOtherProceedingsStep = Step.new(
+        path: ->(application) { Steps.urls.providers_legal_aid_application_has_other_proceedings_path(application) },
+        forward: lambda do |application, add_another_proceeding|
+          if add_another_proceeding
+            :proceedings_types
+          else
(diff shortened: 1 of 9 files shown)


app/services/flow/steps/provider_start/has_other_proceedings_step.rb
```diff
@@ -0,0 +1,16 @@
+module Flow
+  module Steps
+    module ProviderStart
+      HasOtherProceedingsStep = Step.new(
+        path: ->(application) { Steps.urls.providers_legal_aid_application_has_other_proceedings_path(application) },
+        forward: lambda do |application, add_another_proceeding|
+          if add_another_proceeding
+            :proceedings_types
+          else
+            Flow::ProceedingLoop.next_step(application)
```
agoldstone93, line 10: I know this has just been moved without being changed, but why is this path not named? If it is not always `client_involvement_type`, is the corresponding spec correct?

agoldstone93 (approved):

app/services/flow/steps/provider_start/has_other_proceedings_step.rb
skdejong: The path was not named in this flow because `Flow::ProceedingLoop` has around 18 different steps. It may redirect the user to different location depending on the prior options selected.

PR #6808, opened 2024-06-12
AP-5076: Remove copy info from CYA when legal link

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5076)

- Copy case info removed from CYA page when legal link is selected
- When family link is chosen and case is copied, but then link is changed to legal link, copying information is removed

## Checklist

### DIFF
--- a/spec/forms/providers/link_application/make_link_form_spec.rb
+++ b/spec/forms/providers/link_application/make_link_form_spec.rb
@@ -15,23 +15,29 @@ RSpec.describe Providers::LinkApplication::MakeLinkForm, type: :form do
   describe "#save" do
     subject(:call_save) { instance.save }
 
-    before { call_save }
-
     context "with family link type chosen" do
       let(:link_type_code) { "FC_LEAD" }
 
       it "sets the link_type_code" do
+        call_save
         expect(legal_aid_application.lead_linked_application).to have_attributes(link_type_code: "FC_LEAD",
                                                                                  confirm_link: nil)
       end
     end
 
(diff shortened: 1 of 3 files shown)


colinbruce (approved):
LGTM 👍

spec/forms/providers/link_application/make_link_form_spec.rb
```diff
@@ -65,6 +70,16 @@
         expect(linked_application.confirm_link).to be false
       end
     end
+
+    context "when the answer is changed from family link to legal link" do
+      let(:linked_application) { build(:linked_application, associated_application_id: legal_aid_application.id, link_type_code: "FC_LEAD") }
+      let(:link_type_code) { "LEGAL" }
+
+      it "resets the linked_application model" do
```
kmahern, line 78: Is this test checking that changing the link type to family changes copy_case and copy_case_id to nil if they are already set? Would these not already be  set to nil before the test is run in any case? (i.e. would we not need something like `let(:legal_aid_application) { create(:legal_aid_application, copy_case: true, copy_case_id:  id_of_another_legal_aid_applicaton) }` to test the values are actually reset?

kmahern (approved):
One question about a test, other looks good

spec/forms/providers/link_application/make_link_form_spec.rb
agoldstone93: I've rewritten this test which should be much better now

colinbruce (approved):

PR #6826, opened 2024-06-14
AP-4667: flow refactor of capital property and vehicles steps

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4667)

Refactor the required steps to use the new flow pattern

## Checklist

### DIFF
--- a/app/services/flow/steps/provider_capital/own_homes_step.rb
+++ b/app/services/flow/steps/provider_capital/own_homes_step.rb
@@ -1,0 +1,12 @@
+module Flow
+  module Steps
+    module ProviderCapital
+      OwnHomesStep = Step.new(
+        path: ->(application) { Steps.urls.providers_legal_aid_application_means_own_home_path(application) },
+        forward: ->(application) { application.own_home_no? ? :vehicles : :property_details },
+        carry_on_sub_flow: ->(application) { !application.own_home_no? },
+        check_answers: ->(application) { application.checking_non_passported_means? ? :check_capital_answers : :check_passported_answers },
+      )
+    end
+  end
+end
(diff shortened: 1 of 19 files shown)


app/services/flow/steps/provider_capital/own_homes_step.rb
```diff
@@ -0,0 +1,12 @@
+module Flow
+  module Steps
+    module ProviderCapital
+      OwnHomesStep = Step.new(
+        path: ->(application) { Steps.urls.providers_legal_aid_application_means_own_home_path(application) },
+        forward: ->(application) { application.own_home_no? ? :vehicles : :property_details },
```
agoldstone93, line 6: `own_home_no?` doesn't seem like a great name. I'm assuming it means they don't own their own home. I know it's not your code, but if it's straightforward to do it, could we rename it in this PR? If not, don't worry about it.

agoldstone93 (approved):
One non-blocking comment

app/services/flow/steps/provider_capital/own_homes_step.rb
colinbruce: It's an odd one this... it's part of the Rails Magic :tm: that adds helper methods.  The code in question is
```ruby 

  enum(
    own_home: {
      no: "no".freeze,
      mortgage: "mortgage".freeze,
      owned_outright: "owned_outright".freeze,
    },
    _prefix: true,
  )
```
So the method `own_home_no?` says does the `own_home` enum match `no`?
Also valid would be `own_home_owned_outright?` which would return false, etc

I agree it's not clear but changing it would involve added new methods that run thi sunder the hood!

PR #6837, opened 2024-06-18
AP-5096: SCA means test interrupt

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5096)

Add a means test interrupt page for SCA proceedings.

When the application is non-means tested, route to the interrupt page and then on to the merits task list page

Note: The merits task list page will be empty as we have not added those into LFA yet

## Checklist

### DIFF
--- a/spec/models/legal_aid_application_spec.rb
+++ b/spec/models/legal_aid_application_spec.rb
@@ -1443,7 +1443,7 @@ RSpec.describe LegalAidApplication do
     end
   end
 
-  describe "#special_children_act_proceedings??" do
+  describe "#special_children_act_proceedings?" do
     context "with special children act proceedings" do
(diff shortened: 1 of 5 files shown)


spec/models/legal_aid_application_spec.rb
```diff
@@ -1443,6 +1443,25 @@
     end
   end
 
+  describe "#special_children_act_proceedings??" do
```
agoldstone93, line 1446: Is this supposed to be a double question mark?

```suggestion
  describe "#special_children_act_proceedings?" do
```

agoldstone93 (approved):
Very minor comment, otherwise LGTM

spec/models/legal_aid_application_spec.rb
colinbruce: 🤦‍♂️ merged suggestion!

PR #6852, opened 2024-06-20
AP-4622: Remove NINO hint text when there is SCA proceeding

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4622)

Remove NINO hint text when there is SCA proceeding
Move hint text into govuk fieldset component

## Checklist

### DIFF
--- a/app/views/providers/has_national_insurance_numbers/show.html.erb
+++ b/app/views/providers/has_national_insurance_numbers/show.html.erb
@@ -8,11 +8,8 @@
   <%= page_template page_title: t(".page_title"), template: :basic, form: do %>
 
     <%= form.govuk_radio_buttons_fieldset(:has_national_insurance_number,
-                                          legend: { text: page_title, size: "xl", tag: "h1" }) do %>
-
-      <p class="govuk-body govuk-!-margin-top-3 govuk-!-margin-bottom-6">
-        <%= t(".hint") %>
-      </p>
+                                          legend: { text: page_title, size: "xl", tag: "h1" },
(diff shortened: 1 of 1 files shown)


app/views/providers/has_national_insurance_numbers/show.html.erb
```diff
@@ -8,11 +8,8 @@
   <%= page_template page_title: t(".page_title"), template: :basic, form: do %>
 
     <%= form.govuk_radio_buttons_fieldset(:has_national_insurance_number,
-                                          legend: { text: page_title, size: "xl", tag: "h1" }) do %>
-
-      <p class="govuk-body govuk-!-margin-top-3 govuk-!-margin-bottom-6">
-        <%= t(".hint") %>
-      </p>
+                                          legend: { text: page_title, size: "xl", tag: "h1" },
+                                          hint: { text: Setting.special_childrens_act? ? nil : t(".hint") }) do %>
```
kmahern, line 12: Is this correct? Is the logic of whether or not to display the hint determined by the setting or whether the legal aid application has special children act proceedings? (ie something like legal_aid_application.special_children_act_proceedings?)

agoldstone93: I checked with Quay and yes, it's dependent on the sca feature flag. It'll be removed for all cases when the flag is enabled

kmahern: Ah OK, 👍

kmahern (approved):

PR #6853, opened 2024-06-20
ap-4621: add check for special_children_act_proceedings? to substanti…

kmahern:
…ve_cost_overridable?

## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4621)

Added a check that there are not special children act proceedings to substantive_cost_overridable? method.
Added a check to not display emergency certificate if there are special children act proceedings (see ticket for discussion).

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -402,7 +402,7 @@ class LegalAidApplication < ApplicationRecord
   end
 
   def substantive_cost_overridable?
-    substantive_cost_limitation.present? && default_substantive_cost_limitation < MAX_SUBSTANTIVE_COST_LIMIT
+    substantive_cost_limitation.present? && default_substantive_cost_limitation < MAX_SUBSTANTIVE_COST_LIMIT && !special_children_act_proceedings?
   end
 
   def substantive_cost_limitation
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -401,6 +401,10 @@ class LegalAidApplication < ApplicationRecord
     default_substantive_cost_limitation
   end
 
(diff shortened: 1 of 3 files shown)


app/models/legal_aid_application.rb
```diff
@@ -401,8 +401,12 @@ def default_cost_limitation
     default_substantive_cost_limitation
   end
 
+  def display_emergency_certificate?
+    used_delegated_functions? && !special_children_act_proceedings?
+  end
+
   def substantive_cost_overridable?
-    substantive_cost_limitation.present? && default_substantive_cost_limitation < MAX_SUBSTANTIVE_COST_LIMIT
+    substantive_cost_limitation.present? && default_substantive_cost_limitation < MAX_SUBSTANTIVE_COST_LIMIT && !special_children_act_proceedings?
```
colinbruce, line 409: Does this need to be added? We only seem to be calling the new `display_emergency_certificate?` below?

kmahern: Not sure that I understand? (probably missing something!) We are still calling substantive_cost_overridable in multiple places emergency_cost_override_form, limitations_controller, checked_provider_ansers/shared.html.erb, limitations/cost_limit.html.erb, merits_reports/show.html.erb and review_application/questions_and_answers.html.erb. The display_emergency_certificate was intended to be the logic to determine whether or not to display the emergency certificate which Sara said we shouldn't do for SCA applications. The substantive_cost_overridable was intended to be the logic whether or not the provider can override the substantive cost limit.

colinbruce: 🤦‍♂️ In that case, feel free to ignore, this is fine!

colinbruce (approved):

PR #6854, opened 2024-06-20
AP-5083: Update CCMS home address payloads

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5083)

This PR updates the applicant model to update and add home_ and correspondence_address_for_ccms methods

These implement logic that should work for applications created before, during, or after the home_address flag is enabled. 
CCMS address flows

It updates CCMS tests to use, and expect, the new methods rather than just `applicant.address` or `applicant.home_address`
It also adds a temporary trait and uses it to make some tests pass

### Still to do:

- [x] Ensure the logic is updated for new home/correspondence mandatory changes
- [x] Ensure existing tests all pass
- [x] Add the care of line to addresses
- [x] Add the, optional, third address line to the payload

## Checklist

### DIFF
--- a/app/services/ccms/requestors/case_add_requestor.rb
+++ b/app/services/ccms/requestors/case_add_requestor.rb
@@ -171,6 +171,7 @@ module CCMS
         xml.__send__(:"common:CoffName", applicant.correspondence_address_for_ccms.care_of_recipient) unless applicant.correspondence_address_for_ccms.care_of_recipient.nil?
         xml.__send__(:"common:AddressLine1", applicant.correspondence_address_for_ccms.address_line_one)
         xml.__send__(:"common:AddressLine2", applicant.correspondence_address_for_ccms.address_line_two)
+        xml.__send__(:"common:AddressLine3", applicant.correspondence_address_for_ccms.address_line_three) unless applicant.correspondence_address_for_ccms.address_line_three.nil?
         xml.__send__(:"common:City", applicant.correspondence_address_for_ccms.city)
         xml.__send__(:"common:County", applicant.correspondence_address_for_ccms.county)
         xml.__send__(:"common:Country", "GBR")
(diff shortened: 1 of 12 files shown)


agoldstone93 (approved):
LGTM

kmahern (approved):
lgtm

app/services/ccms/requestors/case_add_requestor.rb
```diff
@@ -168,12 +168,14 @@ def generate_provider_details(xml)
...
-        xml.__send__(:"common:County", applicant.address.county)
+        xml.__send__(:"common:CoffName", applicant.correspondence_address_for_ccms.care_of_recipient) unless applicant.correspondence_address_for_ccms.care_of_recipient.nil?
+        xml.__send__(:"common:AddressLine1", applicant.correspondence_address_for_ccms.address_line_one)
+        xml.__send__(:"common:AddressLine2", applicant.correspondence_address_for_ccms.address_line_two)
+        xml.__send__(:"common:AddressLine3", applicant.correspondence_address_for_ccms.address_line_three) unless applicant.correspondence_address_for_ccms.address_line_three.nil?
+        xml.__send__(:"common:City", applicant.correspondence_address_for_ccms.city)
+        xml.__send__(:"common:County", applicant.correspondence_address_for_ccms.county)
         xml.__send__(:"common:Country", "GBR")
-        xml.__send__(:"common:PostalCode", applicant.address.pretty_postcode)
+        xml.__send__(:"common:PostalCode", applicant.correspondence_address_for_ccms.pretty_postcode) if applicant.correspondence_address_for_ccms.postcode.present?
```
naseberry, line 178: Maybe it's my migraine, however, I presume `post code` is always required, regardless of the correspondence preference, therefore, do we need the conditional statement?

colinbruce: It's been a while, but I think this may have been added because the postcode validation is added in the form, not the model.  So test harnesses could have been created with no postcodes and then failed here... maybe

naseberry: 👍  not a blocker

PR #6879, opened 2024-06-27
AP-5107: Update case link handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5107)

This PR extends the linked_application model to set a target_application as well as a lead_application.

This should allow us to display the link to data on the link search result page:
image

As well as having a flattened lead_application structure:
image

So, when creating Application G, we could link to, and optionally copy data from, Application C. While setting Application A as the lead application as CCMS requires this

## Checklist

### DIFF
--- a/app/forms/providers/link_application/find_link_application_form.rb
+++ b/app/forms/providers/link_application/find_link_application_form.rb
@@ -34,7 +34,8 @@ module Providers
       end
 
       def save
-        attributes[:lead_application_id] = @found_application&.id
+        attributes[:target_application_id] = @found_application&.id
+        attributes[:lead_application_id] = @found_application&.lead_application&.id.presence || @found_application&.id
         super
       end
       alias_method :save!, :save
--- a/db/migrate/20240627104142_add_target_application_to_linked_applications.rb
+++ b/db/migrate/20240627104142_add_target_application_to_linked_applications.rb
@@ -1,0 +1,7 @@
+class AddTargetApplicationToLinkedApplications < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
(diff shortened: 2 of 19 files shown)


app/forms/providers/link_application/find_link_application_form.rb
```diff
@@ -34,7 +34,8 @@ def exclude_from_model
       end
 
       def save
-        attributes[:lead_application_id] = @found_application&.id
+        attributes[:target_application_id] = @found_application&.id
+        attributes[:lead_application_id] = @found_application&.lead_application&.id.presence || @found_application&.id
```
naseberry, line 38: Discussion: I like the safe nav operator, however, the chaining of them here feels off. As we don't have a guide/standard, let's discuss.

db/migrate/20240627104142_add_target_application_to_linked_applications.rb
```diff
@@ -0,0 +1,7 @@
+class AddTargetApplicationToLinkedApplications < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
```
naseberry, line 2: 🤔 Can we discuss this?

naseberry (approved):

app/forms/providers/link_application/find_link_application_form.rb
colinbruce: Rubocop enforces it.  If you use a safe navigation operator on the first object you _have_ to use it on all subsequent calls.  If you remove the second one, you get this 
```shell
Lint/SafeNavigationChain: Do not chain ordinary method call after safe navigation operator.
        attributes[:lead_application_id] = @found_application&.lead_application.id.presence || @found_application&.id
                                                                               ^^^
```

PR #6955, opened 2024-07-22
AP-5136: Store dwp partner override on applicant too

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5136)

This stores the shared benefit override answer on the applicant as well as the partner

This caused issues with reporting when the partner has a contrary interest and the answer was only stored on the partner model, it was unavailable

Now we record on, and check, both models and improve the reporting accuracy

We _could_ remove the partner handling but that should probably be a separate ticket as it would be out of the scope of a bug fix ticket

## Checklist

### DIFF
--- a/db/migrate/20240722082922_add_shared_benefit_column_to_applicant.rb
+++ b/db/migrate/20240722082922_add_shared_benefit_column_to_applicant.rb
@@ -1,0 +1,12 @@
+class AddSharedBenefitColumnToApplicant < ActiveRecord::Migration[7.1]
+  def up
+    add_column :applicants, :shared_benefit_with_partner, :boolean
+    Partner.where.not(shared_benefit_with_applicant: nil).each do |partner|
+      partner.legal_aid_application.applicant.update!(shared_benefit_with_partner: partner.shared_benefit_with_applicant)
+    end
+  end
+
+  def down
+    remove_column :applicants, :shared_benefit_with_partner
+  end
+end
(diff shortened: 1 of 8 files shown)


db/migrate/20240722082922_add_shared_benefit_column_to_applicant.rb
```diff
@@ -0,0 +1,12 @@
...
+  def up
+    add_column :applicants, :shared_benefit_with_partner, :boolean
+    Partner.where.not(shared_benefit_with_applicant: nil).each do |partner|
+      partner.legal_aid_application.applicant.update!(shared_benefit_with_partner: partner.shared_benefit_with_applicant)
+    end
+  end
+
+  def down
+    remove_column :applicants, :shared_benefit_with_partner
+  end
```
naseberry, line 11: Up/Down vs Change method.

Is there a preferred method for migration is there a difference, I thought the change method was the new approach?

naseberry (approved):
looks good, just a question, not a blocker

db/migrate/20240722082922_add_shared_benefit_column_to_applicant.rb
colinbruce: Change is fine if you are making a simple change, add/remove a column.
As I needed to migrate the data in one direction, I used a separate up/down split :+1:

PR #6998, opened 2024-08-02
AP-5195: PDF generation bug fix

colinbruce:
## What

After chromium updated to 127+ grover stopped generating PDFs, it would hang and eventually fail.  After testing the interactions between Alpine, Chromium, Puppeteer and Grover for many days, we identified this as a working fix

Further work is needed to ensure that the issue cannot get deployed to production, but this fix will allow us to resume merging and deploying fixes, while we investigate that solution more

## Checklist

### DIFF
--- a/config/initializers/grover.rb
+++ b/config/initializers/grover.rb
@@ -19,6 +19,6 @@ Grover.configure do |config|
       left: "10mm",
       right: "20mm",
     },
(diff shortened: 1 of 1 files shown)


skdejong (approved):

naseberry (approved):
:boom:

PR #7032, opened 2024-08-13
Chore: Update Rails to 7.2.1

colinbruce:
## What

Dependabot attempted to update the individual components and _all_ the PRs failed.

This PR is to centralise the updates, address issues that are seen and allow collaboration on the fixes

https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-1-to-rails-7-2

### TODO list

- [x] update the Rails gems
- [x] Fix the SAML issue in the feature tests
Switching DatabaseCleaner from `:transaction` to `:truncation` has fixed the gross errors with features
- [x] Fix the failing features, mainly focussed around evidence/file uploads
Excluding the document_categories from the truncation rule fixed those
- [x] Fix the mock saml error
now the database is truncated, ensure the provider is created when we start the test
- [x] investigate `DEPRECATION WARNING: Passing nil to the :model argument is deprecated and will raise in Rails 8.0`
mainly in `app/views/providers/uploaded_evidence_collections/show.html.erb` but other places too

## Checklist


features/support/env.rb
```diff
@@ -92,7 +92,7 @@ def feature
 # Remove/comment out the lines below if your app doesn't have a database.
 # For some databases (like MongoDB and CouchDB) you may need to use :truncation instead.
 begin
-  DatabaseCleaner.strategy = :transaction
+  DatabaseCleaner.strategy = :truncation, { except: %w[document_categories] }
```
naseberry, line 95: Why are we excluding `document_categories`?

app/views/providers/uploaded_evidence_collections/show.html.erb
```diff
@@ -25,7 +25,7 @@
   <%= render partial: "upload_evidence" %>
 
   <%= form_with(
-        model: @submission_form,
+        model: @submission_form.presence || false,
```
naseberry, line 28: Does this only affect file upload forms?

features/support/env.rb
colinbruce: The switch to truncation, essentially, clears the database after each feature file.
Because most of the features set up the required values in the DB, i.e. create(:provider), etc, but document categories are only created in the db:seed call that is run once.
As document_categories is a lookup, there is no danger of new records being created, or duplicates being added. 
The other option is manually adding the document categories in every scenario that needs them

app/views/providers/uploaded_evidence_collections/show.html.erb
colinbruce: Yes. in certain circumstances the page is re-rendered, purposefully without a submission form.
There is a deprecation warning that Rails 8 will fail if that occurs.  It will accept false, but not nil.
This allows the tests to run without the warning and is acceptable to rails 8

colinbruce:
> What was the SAML issue in the feature tests?
> 
> Just a little concerned we need to truncate as this slows the test suite down [further]. Plus do we know if the `:deletion` strategy is faster or slower in our case - [What strategy is fastest?](https://github.com/DatabaseCleaner/database_cleaner?tab=readme-ov-file#what-strategy-is-fastest)

Everything failed, because the login_as helper would fail on every test! 
I did not try deletion, as truncation worked... I'm happy to try a comparison locally and record the results 
Also happy to pair on my workings out if anyone can think of an alternative 👍

colinbruce:
Testing different DatabaseCleaner strategies on my local machine
type | time taken | sceanrios passed
--- | --- | --- 
truncation | 39m | 193 scenarios passed
deletion | 39m | 193 scenarios passed
transaction | xx | xx

jsugarman:
> Testing different DatabaseCleaner strategies on my local machine
> 
> type	time taken	scenarios passed
> truncation	39m	193 scenarios passed
> deletion	39m	193 scenarios passed
> transaction	29m	3 passed 190 failed (this was with cache_classes set to true to remove warnings)

nicely done.

be good to see if we could fix before adding 10 minutes to the test suite run and happy to pair on that.

PR #7043, opened 2024-08-14
Ap 5211 user removing core proceeding

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5211)

Added confirm_delete_core_proceedings route, controller and view

TODO

Further work on back link functionality? (see comment for discussion)

## Checklist

### DIFF
--- a/app/controllers/providers/proceedings_sca/confirm_delete_core_proceedings_controller.rb
+++ b/app/controllers/providers/proceedings_sca/confirm_delete_core_proceedings_controller.rb
@@ -1,0 +1,17 @@
+module Providers
+  module ProceedingsSCA
+    class ConfirmDeleteCoreProceedingsController < ProviderBaseController
+      prefix_step_with :proceedings_sca
+
+      def show
+        proceeding
+      end
+
+    private
+
+      def proceeding
+        @proceeding = Proceeding.find(params["proceeding_id"])
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


app/controllers/providers/proceedings_sca/confirm_delete_core_proceedings_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+    class ConfirmDeleteCoreProceedingsController < ProviderBaseController
+      prefix_step_with :proceedings_sca
+
+      def show
+        proceeding
+      end
+
+      def destroy
+        remove_proceedings
+        replace_last_page_in_history(providers_legal_aid_applications_path)
```
kmahern, line 12: Discussion: This is the pattern for how we handle the back link when removing something in other places in the application e.g. in remove_dependants_controller, remove_state_benefits_controller and remove_vehicles_controller. It takes the user back to their list of applications (in this case, if they click the back link after removing a core proceeding) and then they can't click back again. I think this might be the best we can do for this use case as they could have come to the confirm_delete_core_proceedings page via various routes e.g. by creating related proceedings which we are also deleting so the page history would also be invalid for these pages. I think trying to untangle that logic would get messy?

colinbruce: Seems sensible to me :+1:

colinbruce (approved):
LGTM :+1:

PR #7072, opened 2024-08-21
Ap 5119 provider details api integration

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5119)

- Created new PDA::ProviderDetailsRetriever and PDA::ProviderDetailsCreator classes and associated specs to enable calls to be made to the new Provider Details API.
- These  follow the patterns of the existing ProviderDetailsRetrieveer and ProviderDetailsCreator classes and have been added to a new PDA namespace so that when we want to switch over to the new Provider Details API we can simply change all refererences to ProviderDetailsCreator to PDA::ProviderDetailsCreator and everything should 🤞 work.
- In the meantime, so that we can dual-run the old and new provider detail apis and compare results, a new class PDA::CompareProviderDetails class has been added.
- This is called asynchronously by a new CompareProviderDetailsWorker class after provider details have been updated with the results from the existing CCMS pda call.
- The new PDA::CompareProviderDetails class logs (to Kibana - with prefix PDA::CompareProviderDetails) whether or not the details returned by the new ProviderDetailsAPI match the provider details that have already been updated by the existing existing CCMS pda call and if not where there are differences.
- It also logs the start and end times (and calculated duration) of the api call to the new Provider Details API.
- This final commit on this pr is purely to allow it to be tested in uat. This pr should not be merged until this commit is removed.

## Checklist


colinbruce (approved):
:+1:

PR #7074, opened 2024-08-21
AP-5220: Remove 'client has home address?' question from CYA

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5220)

Remove question from check your answers page if it has not been answered by applicant
Delete address information if applicant has no fixed address
Delete 'care of' information if user answers no to 'care of' question
Fix routing issues when clicking change link for "Where to send client's mail"
Delete `correspondence_address` when changing `correspondence_address_choice` if it is no longer relevant.

## Screenshots

When sending client's mail to "My client's UK home address":
image

When sending client's mail to "Another UK residential address":
image

When sending client's mail to "A UK office address":
image

## Checklist

### DIFF
--- a/app/controllers/providers/correspondence_address/care_ofs_controller.rb
+++ b/app/controllers/providers/correspondence_address/care_ofs_controller.rb
@@ -10,19 +10,10 @@ module Providers
       def update
         @form = Addresses::CareOfForm.new(form_params)
 
-        delete_care_of_information if form_params[:care_of].eql?("no")
-
         render :show unless save_continue_or_draft(@form)
       end
 
     private
-
-      def delete_care_of_information
-        address.care_of_first_name = ""
-        address.care_of_last_name = ""
-        address.care_of_organisation_name = ""
-        address.save!
-      end
 
       def address
         @address ||= legal_aid_application.applicant.address
--- a/app/controllers/providers/correspondence_address/care_ofs_controller.rb
+++ b/app/controllers/providers/correspondence_address/care_ofs_controller.rb
@@ -10,10 +10,19 @@ module Providers
       def update
         @form = Addresses::CareOfForm.new(form_params)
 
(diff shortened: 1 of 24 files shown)


colinbruce (approved):
Looks okay, but the complexity may draw things out during UAT!

app/controllers/providers/correspondence_address/care_ofs_controller.rb
```diff
@@ -10,11 +10,20 @@ def show
       def update
         @form = Addresses::CareOfForm.new(form_params)
 
+        delete_care_of_information if form_params[:care_of].eql?("no")
```
kmahern, line 13: Does it follow our standard pattern to do this update in the form class?

app/controllers/providers/correspondence_address/choices_controller.rb
```diff
@@ -10,6 +10,12 @@ def show
       def update
         @form = Addresses::ChoiceForm.new(form_params)
 
+        delete_no_fixed_residence if form_params[:correspondence_address_choice].eql?("home")
```
kmahern, line 13: Does it follow our standard pattern to do this update in the form class?

app/controllers/providers/correspondence_address/care_ofs_controller.rb
agoldstone93: Updated in final commit

app/controllers/providers/correspondence_address/choices_controller.rb
agoldstone93: Updated in final commit

kmahern (approved):

kmahern (approved):
LGTM but agree there is a lot of complexity here and will need a thorough uat.

PR #7183, opened 2024-09-17
AP-4454: Fix for DWP Override means report review reason

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4454)

In production a caseworker found an application where it had been flagged for caseworker review as the provider had overridden the dwp response.

In reality, they must have said yes, gone back and changed the client record and moved onto the passported flow.

A DWPOverride record was created but the questions had not been answered.

I was unable to replicate the steps needed to get the application into that state so this PR adds addition checks to the manual review determiner to ensure that we only flag an application when the object is present _and_ the questions have been answered

## Checklist

### DIFF
--- a/app/services/ccms/manual_review_determiner.rb
+++ b/app/services/ccms/manual_review_determiner.rb
@@ -32,7 +32,7 @@ module CCMS
     end
 
     def manual_review_required?
-      dwp_override.present? ||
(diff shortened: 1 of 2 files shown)


kmahern (approved):
lgtm

PR #7208, opened 2024-09-20
Chore: puppeteer update

colinbruce:
## What

Update the puppeteer version in the ApplyBase docker image

It got locked at 22.15 when the updates started to break everything.

We are now running 23.x and this PR includes the dependabot update to 23.4 to see if it matches what is used in the app

>[!IMPORTANT]
>We know that 22.15 was working, the temp commit allows us to have confidence 
that if we do a new daily apply build with the latest version, it will also work
>The temp commit will be removed before merging

## Checklist


jsugarman (approved):
tested means and merits report - which i assume are the ones using this - and all good.

Just need to remove the temp commit before merge but 👍

PR #7215, opened 2024-09-23
AP-5315: Fix SCA proceeding search bug

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5315)

When LFA returns a code that is not present on the page, i.e. an SCA related proceeding on an initial search, previously it showed the failure message and returned from the loop.

This updates the proceedings javascript so that it counts successfully displayed proceedings, using a new, `shown`, variable. After the loop is complete it then sets the ariaText and shows/hides the no-proceeding-items element accordingly.

## Checklist

### DIFF
--- a/app/javascript/src/modules/proceedings.js
+++ b/app/javascript/src/modules/proceedings.js
@@ -160,13 +160,16 @@ function showResults (results, inputText) {
       ariaText = `${codes.length} ${pluralizedMatches} found for ${inputText}, use tab to move to options`
       hide(document.querySelector('.no-proceeding-items'))
     } else {
-      show(document.querySelector('.no-proceeding-items'))
-      ariaText = `No results found matching ${inputText}`
+      showNoResults(inputText)
     }
   } else {
-    show(document.querySelector('.no-proceeding-items'))
-    ariaText = `No results found matching ${inputText}`
+    showNoResults(inputText)
   }
+}
+
+function showNoResults (inputText) {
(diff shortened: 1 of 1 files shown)


app/javascript/src/modules/proceedings.js
```diff
@@ -153,11 +151,18 @@ function showResults (results, inputText) {
...
+    // once we have looped over all returned codes, check if any were shown to the user
+    // and set the appropriate ariaText and show/hide the no-proceedings found element
+    if (shown > 0) {
+      // the below alerts screen reader users that results appeared on the page
+      const pluralizedMatches = pluralize(codes.length, 'match', 'matches')
+      ariaText = `${codes.length} ${pluralizedMatches} found for ${inputText}, use tab to move to options`
+      hide(document.querySelector('.no-proceeding-items'))
+    } else {
+      show(document.querySelector('.no-proceeding-items'))
+      ariaText = `No results found matching ${inputText}`
```
jsugarman, line 164: No biggee. I see this was repeated before in any event but would be nice to have one place where this is set/handled, even if just a method that is just called in the two places (here and line 167.

jsugarman (approved):

app/javascript/src/modules/proceedings.js
colinbruce: Hmm, I can't see this actually being used 😕 
I might summon @naseberry for a consultation! 😄

naseberry (approved):

PR #7235, opened 2024-09-30
AP-5180: Client address tracking data

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5180)

Add tracking of `No fixed address` responses to the Application Digest and Admin reports

## Checklist

### DIFF
--- a/app/services/reports/mis/application_detail_csv_line.rb
+++ b/app/services/reports/mis/application_detail_csv_line.rb
@@ -171,6 +171,7 @@ module Reports
           "Legal Linked?",
           "Legal link lead?",
           "Number of legal links",
+          "No fixed address",
         ]
       end
 
@@ -205,6 +206,7 @@ module Reports
         banking_data
         partner
         linked_applications
+        home_address
         sanitise
       end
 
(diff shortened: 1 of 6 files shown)


agoldstone93 (approved):
Looks good

app/services/reports/mis/application_detail_csv_line.rb
```diff
@@ -442,6 +444,10 @@ def linked_applications
         end
       end
 
+      def home_address
+        @line << yesno(laa.applicant.no_fixed_residence?)
```
jsugarman, line 448: This looks like it will be "Yes" for " Home address" column when `no_fixed_residence?` is true. Is that what is needed?

jsugarman (approved):
One question on whether the boolean logic is the right way round but otherwise 👍

app/services/reports/mis/application_detail_csv_line.rb
colinbruce: I think it's right, let's see what Mike thinks when testing! 😄

PR #7245, opened 2024-10-01
AP-4707: Prevent none plus other checkbox selection when JS not enabled

jsugarman:
## What
Prevent none plus other checkbox selection when JS not enabled

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4707)

We rely on JS to deselect other regular income and outgoings checkboxes if the "none" checkbox option
is selected. Without JS, therefore, we need a validation that ensures "none" plus any other income/outgoing
cannot be saved.

This PR also ensures that selected checkboxes are re-rendered with user input when an error is encountered
and links the "none plus other" error to the checkbox list itself.

This change impacts the following forms:
- client regular income
- client regular outgoings 
- partner regular income
- partner regular outgoings (shown below)

Similar functionality applied to truelayer path  form equivalents - identify_types_of_income|outgoings

### After changes

non_js_regular_transaction_with_none_plus_other_selected

## Checklist

### DIFF
--- a/app/controllers/providers/means/identify_types_of_incomes_controller.rb
+++ b/app/controllers/providers/means/identify_types_of_incomes_controller.rb
@@ -7,24 +7,25 @@ module Providers
 
       def update
         synchronize_credit_transaction_types
+        validate
 
-        if none_selected?
-          legal_aid_application.update!(no_credit_transaction_types_selected: true)
+        if legal_aid_application.errors.present?
+          return continue_or_draft if draft_selected?
 
-          return continue_or_draft
-        elsif transaction_types_selected?
-          legal_aid_application.update!(no_credit_transaction_types_selected: false)
+          render :show
+        else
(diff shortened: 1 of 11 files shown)


RoseSAK (approved):
Looks good to me, two comments though:
1) Did you want to also add a test for `partner/regular_income_form` ? It's the only one that doesn't have one
2) Might also be worth just running those error messages via content/design if you haven't already (they look pretty standard but you never know)

jsugarman:
> Looks good to me, two comments though:
> 
> 1. Did you want to also add a test for `partner/regular_income_form` ? It's the only one that doesn't have one
> 2. Might also be worth just running those error messages via content/design if you haven't already (they look pretty standard but you never know)

Thanks.
I have added a simple validation spec for the validation of the partner income form. The error messages exist in that "format" in a lot of forms so if they need changing i would say that should be done across the whole app as a separate ticket.

PR #7257, opened 2024-10-03
AP-4616 Property values bug

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4616)

BUG: Property values were not being reset when the `own_home` answer was changed. 

Scenario: the provider selects that their client owns their home, and enters property values for their home, but later returns to the `own_home` page (either via back buttons or CYA link) and changes the `own_home` answer to 'No'. The previously entered property values are not removed and were being sent to CFE and used in the calculation. 

There was a service `CleanupCapitalAttributes` which was called by another service `ApplicantCompleteMeans` at the end of the applicant journey via a state transition hook. Both these files have not been touched since 2019 (!). Previously the capital journey was part of the applicant journey, so it looks like when the capital questions were moved to the provider journey over two years ago, the calls to these services were never updated/moved. So this bug has been around for at least two years. 

The only thing that `CleanupCapitalAttributes` does is reset property attributes. Therefore I decided to just remove this service as it no longer needs to be called at the end of the applicant journey, and added the code to reset property attributes into the `own_home` form itself, where it is more obvious, less obscured and does not depend on state transitions. 

## Checklist


colinbruce (approved):
LGTM :+1:

PR #7269, opened 2024-10-07
Stop updating student_loan `archived_at` date

jsugarman:
## What
Stop updating student_loan `archived_at` date

[Came out of looking at story](https://dsdmoj.atlassian.net/browse/AP-5346)

This was causing the archived_at date to be updated with the current
date time everytime it is deployed.

Arguably we could/should remove this line entirely?!

## Checklist

### DIFF
--- a/app/services/populators/transaction_type_populator.rb
+++ b/app/services/populators/transaction_type_populator.rb
@@ -32,7 +32,7 @@ module Populators
 
     def mark_old_as_archived
       TransactionType.active.where.not(name: TransactionType::NAMES.values.flatten).update!(archived_at: Time.current)
-      TransactionType.find_by(name: "student_loan").update!(archived_at: Time.current)
+      TransactionType.active.find_by(name: "student_loan")&.update!(archived_at: Time.current)
     end
 
     def update_other_income_types


app/services/populators/transaction_type_populator.rb
```diff
@@ -32,7 +32,7 @@ def find_or_create(name, operation)
 
     def mark_old_as_archived
       TransactionType.active.where.not(name: TransactionType::NAMES.values.flatten).update!(archived_at: Time.current)
-      TransactionType.find_by(name: "student_loan").update!(archived_at: Time.current)
+      TransactionType.active.find_by(name: "student_loan")&.update!(archived_at: Time.current)
```
colinbruce, line 35: This works but, as you said in the comment, not sure it should be here.
If you remove student loan from the arrays above, would that not have the same effect?

jsugarman: I mean removing this line, 35, entirely from the seeder. That should have the same effect as student_loan not active. Removing it from the `TransactionType::NAMES` would require more understanding of what that array so I would be concerned about doing that.

colinbruce (approved):

app/services/populators/transaction_type_populator.rb
colinbruce: Agreed it's more work... is it worth dropping a tech-debt ticket in the backlog so this isn't re-discovered in a year!  😄

jsugarman: https://dsdmoj.atlassian.net/browse/AP-5373

PR #7277, opened 2024-10-08
AP-5269: Update delete application handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5269)

Remove the modal JS dialog and update the non-js path to match the requirement.

## Checklist


features/providers/deleting_applications.feature
```diff
@@ -1,34 +1,15 @@
 Feature: Deleting applications
 
   @javascript
```
jsugarman, line 3: does feature work pass without the javascript tag?

jsugarman (approved):
LGTM. Are we using an existing delete application page then

colinbruce:
> LGTM. Are we using an existing delete application page then

Yes, there was always a standard delete page.  The JS version git added over the top

features/providers/deleting_applications.feature
colinbruce: No, the features hang at the first step without the tag! 🤷‍♂️

PR #7282, opened 2024-10-09
ap-5158: migrate postgres, redis, secretkeybase secrets

kmahern:
## What

## Checklist


jsugarman (commented):
Looks fine to me. Some notes however,

- We may not need to put UAT redis or postgres pod passwords in secrets. They are ephemeral objects that we should be able to destroy as and when we like. NSCC [did not bother as you can see here they are in the values files unencrypted](https://github.com/ministryofjustice/laa-submit-crime-forms/blob/main/helm_deploy/values-dev.yaml)
- open question: Could swapping the source of the secret_key_base cause an issue that results in sessions getting lost. I assume the value is the same but am thinking of quote issues in aws secrets compared to what we currently have in the k8s or some suttlety that we missed resulting in this.

- open question: the same issue as above but for the ENCRYPTION_PRIMARY_KEY

kmahern:
Closing this pr to migrate secrets in separate prs.

PR #7290, opened 2024-10-10
Ap 5348 mtr capital discretionary disregards

kmahern:
## What

## Checklist

### DIFF
--- a/db/migrate/20241009151640_create_discretionary_disregards.rb
+++ b/db/migrate/20241009151640_create_discretionary_disregards.rb
@@ -1,0 +1,18 @@
+class CreateDiscretionaryDisregards < ActiveRecord::Migration[7.2]
+  def change
+    create_table :discretionary_disregards, id: :uuid do |t|
+      t.boolean :backdated_benefits
+      t.boolean :compensation_for_personal_harm
+      t.boolean :criminal_injuries_compensation
+      t.boolean :grenfell_tower_fire_victims
+      t.boolean :london_emergencies_trust
+      t.boolean :national_emergencies_trust
+      t.boolean :loss_or_harm_relating_to_this_application
+      t.boolean :victims_of_overseas_terrorism_compensation
+      t.boolean :we_love_manchester_emergency_fund
+      t.boolean "none_selected"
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -427,6 +427,23 @@
     t.index ["legal_aid_application_id"], name: "index_dependants_on_legal_aid_application_id"
   end
 
+  create_table "discretionary_disregards", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
kmahern, line 430: question: are we happy with this as the data model for discretionary_disregards given the work on ap-5349 to come?

jsugarman: given that it follows the naming convention for policy_disregards table and model I think this is ok, but might be better named as `discretionary_policy_disregards`, to separate this from "disregarded benefits" naming. 

Though I do not have strong feelings on these I mention them here for completeness

- we could rename the policy_disregards table as mandatory_policy_disregards if that naming is logical (I am not sure)
- its a shame we do not have a table structure that allows us to extend the existing table more easily by having a `legal_aid_application_policy_disregards` join table  and  a `policy_disregards` table that is seed with records with `name` and `mandatory:boolean` or similar to achieve this functionality more elegantly - like the `transaction_types` table and equivalent `legal_aid_application_transaction_types` join table. Perhaps we should have a tech debt ticket 🤔 ?!

db/migrate/20241009151640_create_discretionary_disregards.rb
```diff
@@ -0,0 +1,18 @@
+class CreateDiscretionaryDisregards < ActiveRecord::Migration[7.2]
+  def change
+    create_table :discretionary_disregards, id: :uuid do |t|
```
jsugarman, line 3: as mentioned below disrectionary_policy_disregards would separate this naming from the notion of "disregarded benefits"

jsugarman (commented):
Should we not be using the route, controller and step naming of discretionary [policy] disregards? The PaymentsToReview naming does not seem to emphasise its purpose, unless i am missing something.

PR #7292, opened 2024-10-10
AP-5335: Remove ability to add disregarded benefits

jsugarman:
## What
Remove ability to add disregarded benefits on open banking (truelayer) journey

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5335)

The BA has identified that the caseworkers do not find the addition of transactions pertaining to disregarded benefits, via the clients truelayer journey, helpful when assessing means. The option to add transactions that are “disregarded benefits” should therefore be removed.

- [X] check mid-flow and historical data consequences ok
_agreed with BA that we can ignore the impact on the few applications that will be impacted_
- [x] add housing benefits question to truelayer path
_this came out of discussion with BA. Existing system has a bug in this regard as HB needs to be discounted from housing payments to reach an actual housing payment amount. Further, since we are removing disregards we need to ask the provider for HB amounts and send the respetive value to CFE just as for the bank statement upload flow._

## Note on approach and impact
The disregarded benefits type, `excluded_benefits` was explicitly archived by setting its 
`archived_at` value. This means `excluded_benefits` transaction type continues
to exist but is not used. 

1. excluded_benefits transaction_type will exist but be inactive
2. tests will run with production like data in this respect
2. fresh UAT branches will have production like data in this respect

NOTE: Removing the transaction type from the NAMES constant was _not_ done
because it will cause tests and fresh UAT branches to NOT have the
excluded_benefits transaction type at all - i.e. non-production like data

Having excluded_benefits be inactive will make

- the option to selected disregarded benefits disappear from the
  transaction select page (/income_summary) - mid flow
- the disregarded benefits output to disappear from CYA means page (/check_income_answers) - mid flow
- the disregarded benefits output to disappear from review and print page (/review_and_print_application) - mid-flow
- the disregarded benefits output to disappear from submitted_applications review and print page (/submitted_application) - post submission
- means report has a deductions which may also need removing

## Checklist

### DIFF
--- a/features/step_definitions/check_your_answers_steps.rb
+++ b/features/step_definitions/check_your_answers_steps.rb
@@ -100,7 +100,7 @@ def expect_matching_questions_and_answers(actual_selector:, expected_table:)
   expected = expected_table.hashes.map(&:symbolize_keys)
   actual = actual_questions_and_answers_in(selector: actual_selector)
 
-  expect(actual).to match_array(expected)
+  expect(actual).to match_array(expected), SuperDiff.diff(expected, actual)
 end
 
 def actual_questions_and_answers_in(selector:)
--- a/app/controllers/providers/means/housing_benefits_controller.rb
+++ b/app/controllers/providers/means/housing_benefits_controller.rb
@@ -28,7 +28,7 @@ module Providers
 
       def individual
         show_for_partner = legal_aid_application.housing_payments_for?("Partner")
-        show_for_client = legal_aid_application.housing_payments_for?("Applicant") && legal_aid_application.uploading_bank_statements?
+        show_for_client = legal_aid_application.housing_payments_for?("Applicant")
         if show_for_partner && show_for_client
           I18n.t("generic.client_or_partner")
         else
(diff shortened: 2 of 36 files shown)


agoldstone93 (approved):
Looks good to me

features/step_definitions/check_your_answers_steps.rb
```diff
@@ -100,7 +100,7 @@ def expect_matching_questions_and_answers(actual_selector:, expected_table:)
   expected = expected_table.hashes.map(&:symbolize_keys)
   actual = actual_questions_and_answers_in(selector: actual_selector)
 
-  expect(actual).to match_array(expected)
+  expect(actual).to match_array(expected), SuperDiff.diff(expected, actual)
```
colinbruce, line 103: SuperDiff.diff? Is this a new pattern?

colinbruce (approved):
This is a massively complex change.  I cannot see anything obviously wrong so passing but I think a thorough UAT shakedown will be needed!

features/step_definitions/check_your_answers_steps.rb
jsugarman: We already use super diff which integrates into rspec and gives a diff when an error is raised. However in this instance in cucumber output it was not giving a diff and i was struggling to see the failure. Adding this output the actual diff (which immediately showed me I had neglected to add a `)` to a translation which i was then able to fix 🥳 . 

in short, it is helps when using this cucumber wrapper step expectation so i left it in

PR #7337, opened 2024-10-21
ap-5080: Do not alert apply team if citizen_start_email is undeliverable

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5080)

Turn off alerts and emails to apply team for undeliverable citizen_start emails only.

## Checklist

### DIFF
--- a/app/services/govuk_emails/monitor.rb
+++ b/app/services/govuk_emails/monitor.rb
@@ -28,9 +28,9 @@ module GovukEmails
     def send_undeliverable_alerts
       return unless HostEnv.production?
 
-      AlertManager.capture_message("Unable to deliver mail to #{scheduled_mail.addressee} - ScheduledMailing record #{scheduled_mail.id}")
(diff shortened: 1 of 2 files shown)


app/services/govuk_emails/monitor.rb
```diff
@@ -28,9 +28,9 @@ def response
     def send_undeliverable_alerts
       return unless HostEnv.production?
 
-      AlertManager.capture_message("Unable to deliver mail to #{scheduled_mail.addressee} - ScheduledMailing record #{scheduled_mail.id}")
+      AlertManager.capture_message("Unable to deliver mail to #{scheduled_mail.addressee} - ScheduledMailing record #{scheduled_mail.id}") unless @scheduled_mail.mailer_method == "citizen_start_email"
```
jsugarman, line 31: given the repetition might be worth extracting to a method 

```ruby
def citizen_start_email?
  @scheduled_mail.mailer_method == "citizen_start_email"
end
```

and use

```ruby
blah unless citizen_start_email?
```

jsugarman (approved):
one refactor suggestion, otherwise 👍

app/services/govuk_emails/monitor.rb
kmahern: 👍 have made this refactor

PR #7347, opened 2024-10-22
AP-5426 Create capital disregards DB table and add namespace

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5426)

Add a new table `CapitalDisregards` to record both mandatory and discretionary disregards.
Add the `capital_disregards` namespace 

## Checklist

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.2].define(version: 2024_10_18_081039) do
+ActiveRecord::Schema[7.2].define(version: 2024_10_22_133251) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -427,6 +427,19 @@
     t.index ["legal_aid_application_id"], name: "index_dependants_on_legal_aid_application_id"
   end
 
+  create_table "disregards", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
jsugarman, line 430: should we not be calling this policy_disregards to avoid confusion with benefit disregards?

jsugarman: or perhaps `capital_disregards` (if it makes business sense) if we are avoiding the `policy_regards` name due to the other table

RoseSAK: Hmm, yeah probably, we already have an existing PolicyDisregards table though which we were going to leave for a while until we migrate the data. Perhaps we'd need to rename this to `LegacyPolicyDisregards`? Or alternatively migrate all the old policy disregards data in this PR. 

@kmahern and @skdejong what do you guys think?

skdejong: I think `capital_disregards` would make the most sense out of all those options.

kmahern: I think I agree with @skdejong that capital_disregards might be the best option. For consistency should we also make the namespace capital_disregards?

skdejong: Yes, I think we should keep things consistent.

colinbruce (approved):

PR #7350, opened 2024-10-23
Ap 5348 discretionary policy disregards

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5348)

## Checklist


app/forms/providers/capital_disregards/discretionary_form.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      attr_accessor :discretionary_capital_disregards, :none_selected
+
+      validate :validate_any_checkbox_checked, unless: :draft?
+      validate :validate_no_account_and_another_checkbox_not_both_checked, unless: :draft?
+
+      def save
+        return false unless valid?
+
+        model.capital_disregards.where(mandatory: false).destroy_all
```
RoseSAK, line 26: Just wondering whether all existing disregards should definitely be destroyed here - what if someone has already entered details of disregards on the next page, has come back to this page (e.g. via CYA) and just wants to add one more, they would have to re-add details for all of them.

Thinking it might be better to compare the new selected disregards with the existing disregards. I will try to implement this on my branch and copy over to here - if you are in agreement @kmahern ?

jsugarman: yes, I'd have to agree. especially since we may want to have a hook which deletes any related details.

RoseSAK (approved):

app/forms/providers/capital_disregards/discretionary_form.rb
RoseSAK: @agoldstone93 , @skdejong  and I have agreed that in order not to block this branch from being merged, this change will be implemented on my branch and I will ensure it's also copied over to this `discretionary_form` as well

jsugarman: @RoseSAK I have implemented a fix for the discretionary on my branch off of this as part of the bug fix. This is all getting rather messy though as this branch has just been renamespaced and broken a lot of stuff.

PR #7351, opened 2024-10-23
AP-5347 Mandatory disregards page

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5347)

Add a new page for mandatory capital disregards, behind the MTR2A feature flag. The old PolicyDisregards page still exists. 

## Checklist


colinbruce (commented):
This looks fine code, logic and spelling wise to me... 🏆 
You may want a second, more MTR focussed, pair of eyes on it though!

app/forms/providers/means/capital_disregards/mandatory_form.rb
```diff
@@ -0,0 +1,115 @@
...
+
+        def error_message_for_none_and_another_option_selected
+          I18n.t("activemodel.errors.models.mandatory_capital_disregards.attributes.base.#{error_key('none_and_another_option_selected')}")
+        end
+
+        def error_key(key_name)
+          return "#{key_name}_with_partner" if legal_aid_application&.applicant&.has_partner_with_no_contrary_interest?
+
+          key_name
+        end
```
naseberry, line 111: Does this form inherit `app/forms/base_form.rb`

If yes, could/should we use the `error_key` method from the base form?

app/services/flow/steps/provider_capital_disregards/mandatory_step.rb
```diff
@@ -0,0 +1,11 @@
+module Flow
+  module Steps
+    module ProviderCapitalDisregards
+      MandatoryStep = Step.new(
+        path: ->(application) { Steps.urls.providers_legal_aid_application_means_capital_disregards_mandatory_path(application) },
+        forward: ->(application) { application.passported? ? :check_passported_answers : :check_capital_answers },
```
naseberry, line 6: Forward: I presume this is because the next page is not yet developed?

RoseSAK: Yes, that's right

app/forms/providers/means/capital_disregards/mandatory_form.rb
colinbruce: No it doesn't inherit from base_form, there was a conversation yesterday about the base form having too much 'rails magic' causing issues!

RoseSAK: It does not inherit BaseForm, it follows the same pattern as the (now merged) `discretionary_disregards` page and the existing transactions pages , which are a bit different to the model-based forms

app/views/providers/means/policy_disregards/show.html.erb
```diff
@@ -13,17 +13,16 @@
...
-        <%= form.govuk_check_box checkbox, true, "", multiple: false, link_errors: true, label: { text: t(".#{checkbox}") } %>
-      <% end %>
-    </div>
-
-    <%= form.govuk_radio_divider %>
-    <%= form.govuk_check_box :none_selected, true, "", multiple: false, label: { text: t(".none_of_these") } %>
+      <div class="deselect-group" data-deselect-ctrl="#policy-disregards-none-selected-true-field">
+        <% Providers::PolicyDisregardsForm::SINGLE_VALUE_ATTRIBUTES.each do |checkbox| %>
+          <%= form.govuk_check_box checkbox, true, "", multiple: false, link_errors: true, label: { text: t(".#{checkbox}") } %>
+        <% end %>
```
naseberry, line 19: Form builder guidance: "the first checkbox link_errors must be set to true"

Line 18, would would set true on all checkboxes.

Consider call `.each_with_index` on `PolicyDisregardsForm::SINGLE_VALUE_ATTRIBUTES`
then using `link_errors: index.zero?` to set the first checkbox link_errors to true.

RoseSAK: This is the old page, I think I just changed the indentation so it was more readable. I guess it was never implemented correctly at the time 🤷‍♀️ I could update it but this page will be removed/unused pretty soon anyway!

naseberry (approved):

app/controllers/providers/means/capital_disregards/mandatory_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+        def update
+          @form = Providers::Means::CapitalDisregards::MandatoryForm.new(form_params)
+          render :show unless save_continue_or_draft(@form)
+        end
+
+      private
+
+        def form_params
+          merge_with_model(legal_aid_application) do
```
jsugarman, line 19: i do not believe you need this `merge_with_model` block - i removed it from discretionary - as we explicitly extract legal_aid_application in the intializer of the form...but am wondering if the update depends on it now 🤔

jsugarman (approved):
be good to have a feature test too to lockdown the behaviour (prevent regressions) I added one you can copy and amend at `features/providers/mtr_accelerated/discretionary_capital_disregards.feature`

jsugarman:
Have sanity tested and all looks good 🏅

app/forms/providers/means/capital_disregards/mandatory_form.rb
```diff
@@ -0,0 +1,115 @@
...
+
+      private
+
+        # create new records if they don't already exist
+        def create_new_disregards!
+          return unless mandatory_capital_disregards
+
+          mandatory_capital_disregards&.reject { |disregard|
+            existing_mandatory_disregards.include?(disregard)
+          }&.each { |disregard| legal_aid_application.capital_disregards.create!(name: disregard, mandatory: true) }
```
jsugarman, line 67: for consistency it might be good to use the same pattern as used in discretionary and identify_types_of_income/outgoings_controllers but either way

PR #7352, opened 2024-10-23
AP-5436: Stop redirecting to page that no longer exists

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5436)

If e.g 11 applications exist, and there are 10 on each page, when the 11th is deleted, the application tried to redirect to page 2. Address this so it always redirects to page 1

## Checklist

### DIFF
--- a/config/initializers/pagy.rb
+++ b/config/initializers/pagy.rb
@@ -1,1 +1,5 @@
+require "pagy/extras/overflow"
+
 Pagy::I18n.load(locale: "en", filepath: Rails.root.join("config/locales/en/pagy.yml"))
+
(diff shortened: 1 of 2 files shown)


jsugarman (commented):
Be good to spec the behaviour for documentary reasons and to prevent regressions.

I wonder if we could instead render the "last" page as this would have been the page that had 1 item, but would "now" be the page one before that.

example here
https://github.com/ddnexus/pagy/issues/68#issuecomment-403586502

colinbruce (approved):

colinbruce (approved):

PR #7368, opened 2024-10-25
AP-5392: SCA update check merits answers

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5392)

Add a loop skip when the proceeding being displayed has no tasks on the merits_task_list

## Checklist

### DIFF
--- a/app/views/shared/check_answers/_merits_proceeding_section.html.erb
+++ b/app/views/shared/check_answers/_merits_proceeding_section.html.erb
@@ -1,5 +1,5 @@
 <% @legal_aid_application.proceedings.each do |proceeding| %>
-  <% next if @legal_aid_application.legal_framework_merits_task_list.task_list.tasks.dig(:proceedings, proceeding.ccms_code.to_sym)[:tasks].empty? %>
+  <% next if merits_task_list_empty_for?(@legal_aid_application, proceeding) %>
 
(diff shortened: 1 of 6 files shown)


jsugarman (approved):

app/views/shared/check_answers/_merits_proceeding_section.html.erb
```diff
@@ -1,4 +1,6 @@
 <% @legal_aid_application.proceedings.each do |proceeding| %>
+  <% next if @legal_aid_application.legal_framework_merits_task_list.task_list.tasks.dig(:proceedings, proceeding.ccms_code.to_sym)[:tasks].empty? %>
```
jsugarman, line 2: perhaps a helper

```
#
def mertis_task_list_empty_for?(proceeding_ccms_code)
  legal_framework_merits_task_list.task_list.tasks.dig(:proceedings, proceeding_ccms_code.to_sym)[:tasks].empty?
end
```

colinbruce: As these are due to be replaced as the CYA refresh - we could do it then?

colinbruce: Had second thoughts... moved it into the task_list_helper, can I get a re-review?

PR #7380, opened 2024-10-28
Ap 5349 build add details page

skdejong:
## What

[MTR2 A: Build 'Add details' page for mandatory Capital disregards](https://dsdmoj.atlassian.net/browse/AP-5349)
[MTR2 A: Add CYA, R&P and means report](https://dsdmoj.atlassian.net/browse/AP-5424)

## Checklist

### DIFF
--- a/app/controllers/providers/means/capital_disregards/add_details_controller.rb
+++ b/app/controllers/providers/means/capital_disregards/add_details_controller.rb
@@ -1,0 +1,43 @@
+module Providers
+  module Means
+    module CapitalDisregards
+      class AddDetailsController < ProviderBaseController
+        prefix_step_with :capital_disregards
+        def show
+          @form = AddDetailsForm.new(model: capital_disregard)
+        end
+
+        def update
+          @form = AddDetailsForm.new(form_params)
+          if @form.valid?
+            @form.save!
+            go_forward(next_disregard)
+          else
+            render :show
+          end
+        end
+
+      private
+
+        def next_disregard
+          @capital_disregard.mandatory? ? @legal_aid_application.next_incomplete_mandatory_disregard : @legal_aid_application.next_incomplete_discretionary_disregard
+        end
+
+        def capital_disregard
+          @capital_disregard = CapitalDisregard.find(capital_disregard_id_param)
+        end
+
+        def capital_disregard_id_param
+          params.require(:id)
+        end
+
+        def form_params
+          merged_params = merge_with_model(capital_disregard) do
+            params.require(:capital_disregard).permit(:payment_reason, :amount, :account_name, :date_received)
+          end
+          convert_date_params(merged_params)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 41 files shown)


app/controllers/providers/means/capital_disregards/mandatory_controller.rb
```diff
@@ -10,7 +10,11 @@ def show
 
         def update
           @form = Providers::Means::CapitalDisregards::MandatoryForm.new(form_params)
-          render :show unless save_continue_or_draft(@form)
+
```
jsugarman, line 13: You may be able to use this pattern

```ruby
ender :show unless save_continue_or_draft(@form, @legal_aid_application.next_incomplete_mandatory_disregard)
```

I believe any args are passed to go_forward anyway if form is valid

colinbruce: You can namespace it with a key
```ruby
render :show unless save_continue_or_draft(@form, next: @legal_aid_application.next_incomplete_mandatory_disregard)
```
and then update the flow to expect arguments

app/controllers/providers/means/capital_disregards/add_details_controller.rb
```diff
@@ -0,0 +1,43 @@
+module Providers
+  module Means
+    module CapitalDisregards
+      class AddDetailsController < ProviderBaseController
+        prefix_step_with :capital_disregards
+        def show
+          @form = AddDetailsForm.new(model: capital_disregard)
```
naseberry, line 7: Should this meet the current pattern, that is to chain the module definitions `Providers::Means::CapitalDisregards::AddDetailsForm`?

app/controllers/providers/means/capital_disregards/add_details_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+
+        def update
+          @form = AddDetailsForm.new(form_params)
```
naseberry, line 11: Should this meet the current pattern, that is to chain the module definitions `Providers::Means::CapitalDisregards::AddDetailsForm`?

laa-apply-for-legal-aid
naseberry, line 1: What has happened here, why are we using submodules?

app/controllers/providers/means/capital_disregards/add_details_controller.rb
colinbruce: The forms are namespace as such, but rails doesn't _need_ to chain the modules when the controller is in the same namespace.

The other two controllers are 50/50, one does, one doesn't. I think this one sets the pattern! 😁

laa-apply-for-legal-aid
colinbruce: This looks like an empty folder was created! :/ 
I think it can just be deleted

jsugarman (approved):
It looks good 👍 

Only point for possible improvement I can see would be if the controllers could be made to use the `render :show unless save_continue_or_draft(@form)` pattern but pass in the next disregard as necessary. But no worries.

I have tested both the client only bank statement and truelayer journeys plus the client bank statement with partner (with contrary interest) journey and not seen any issues, the text reflecting what i would expect 🥇

app/controllers/providers/means/capital_disregards/add_details_controller.rb
skdejong: After a discussion today with all of the Civil Apply devs, it was decided to keep this as it is for now.

PR #7387, opened 2024-10-29
AP-5395: Update SCA means_tested display on Means Report

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5395)

Create ClientHelper for means_tested display text and call it from the _client_details partial

## Checklist

### DIFF
--- a/app/helpers/client_helper.rb
+++ b/app/helpers/client_helper.rb
@@ -1,0 +1,8 @@
+module ClientHelper
+  def means_tested_description(legal_aid_application)
+    return t("generic.means_tested.no_sca") if legal_aid_application.special_children_act_proceedings?
+    return t("generic.means_tested.no_under_18") if legal_aid_application.applicant.under_18?
+
+    t("generic.yes")
+  end
+end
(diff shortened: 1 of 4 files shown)


app/helpers/client_helper.rb
```diff
@@ -0,0 +1,8 @@
+module ClientHelper
+  def means_tested?(legal_aid_application)
```
jsugarman, line 2: I would expect `method?` to return a boolean, or yes/no possibly. perhaps `means_tested_answer`, would be less confusing for our future selves

jsugarman (approved):

app/helpers/client_helper.rb
colinbruce: Cool, thanks Joel, good point.  Originally it _was_ a boolean response, then I refactored it! 🤦‍♂️ 
I'll re-work and re-submit before moving into UAT 👍

colinbruce: Updated to means_tested_description

PR #7340, opened 2024-10-21
AP-5416: Fix truelayer client with partner means CYA and flow

jsugarman:
## What

> [!WARNING]
> Need to verify correct flow for scenario and ensure housing benefit question works as expected. Designs for flow may be missing so needs verifying too. 
> ✅ flow checked with BA and designs 
> - truelayer income > identify income types > cash income transactions > bank transaction categorisation > CYA; 
> - truelayer outgoings > identify outgoing types > cash outgoing transactions > bank transaction categorisation > [Housing benefit question if client OR partner has housing payments] > CYA; 

Fix truelayer client with partner CYA income page and flow

[Link to main bug story](https://dsdmoj.atlassian.net/browse/AP-5416)
[Link to older related story](https://dsdmoj.atlassian.net/browse/AP-5405)

Currently the CYA for this scenario shows the `regular_transactions` and their values, which are always going to be empty/blank/invalid for the truelayer flow. Plus, while the Change link goes to the correct starting point, `identify_types_of_income` it, after cash_income page, does not take them to bank transaction categorisation page, thereby preventing them ever being able to edit previously selected bank transaction categorisations. 

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing: `bundle exec rake`
- Github should not be reporting conflicts; you should have recently run `git rebase main`.
- The standards in the [Git Workflow document on Confluence](https://dsdmoj.atlassian.net/wiki/spaces/ATPPB/pages/4602855954/Git+Workflow) should be followed
- There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- The PR description should say what you changed and why, with a link to the JIRA story.
- You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
- You should have checked that the commit messages say why the change was made.

PR #7396, opened 2024-10-31
AP-5453: CYA truelayer transaction bug

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5453)

This PR addresses a bug found while working on #7340 where a user returning to the Payments your client receives page from CYA would see payments that had been selected for a partner too

While resolving that, another bug was found where a payment that has been assigned to a category which was subsequently removed, cannot be re-assigned to a category (new or the previous one if re-enabled)

- Update checkbox selection on incoming truelayer, CYA path
- Update meta-data handling after incoming and outgoing checkboxes are cleared
- Update checkbox selection on outgoing truelayer, CYA path
- Update CYA routing when a provider selects no payments
- Update routing after the identify_* pages, there were routes that _could_ be reached that were weird!

## Checklist

### DIFF
--- a/features/providers/regressions/client_truelayer_partner_bank_statements_check_means_answer.feature
+++ b/features/providers/regressions/client_truelayer_partner_bank_statements_check_means_answer.feature
@@ -34,6 +34,13 @@ Feature: Bank statement upload check your answers
     And the checkbox for Maintenance payments from a former partner should be unchecked
 
   @javascript
+  Scenario: Viewing client outgoings
+    When I click Check Your Answers Change link for "payments your client makes"
+    Then I should be on a page with title "Which of these payments does your client pay?"
+    And the checkbox for Childcare payments should be checked
+    And the checkbox for Maintenance payments to a former partner should be unchecked
+
+  @javascript
   Scenario: Viewing partner payments
     When I click Check Your Answers Change link for "payments the partner receives"
     Then I should be on the "regular_incomes" page showing "Which of these payments does the partner get?"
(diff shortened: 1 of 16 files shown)


features/providers/regressions/client_truelayer_partner_bank_statements_check_means_answer.feature
```diff
@@ -0,0 +1,84 @@
...
+    And I click "Save and continue"
+    Then I should be on the "income_summary" page showing "Sort your client's income into categories"
+
+    When I click link "View statements and add transactions"
+    Then I should be on the "incoming_transactions/friends_or_family" page showing "Select payments from friends or family"
+
+    When I select "babysitting"
+    And I click "Save and continue"
+    Then I should be on the "income_summary" page showing "Sort your client's income into categories"
+    And I should see "babysitting"
```
jsugarman, line 84: might be worth adding a CYA result check step to see the updated amount, if there is a fixed monetary value associated with babysitting, or if the selector can take regexes?!

```
    And the "Payments your client receives" section's questions and answers should match:
      | question | answer |
      | Benefits, charitable or government payments | None |
      | Financial help from friends or family | "/£\d+" |
      | Maintenance payments from a former partner | None |
      | Income from a property or lodger | None |
      | Pension | None |
 ```

jsugarman (approved):

jsugarman (approved):
🎖️Looks good

I would be tempted to use the fuller name of `applicant_transaction_type_ids`, to avoid (further) confusion with various uses of the term transaction (bank_, cash_) etc, but either way 🤷 as it is a lonnnng name

colinbruce:
Good shout on the rename - it was a temp fix to see if it would work, but the updated name is more accurate :+1:

features/providers/regressions/client_truelayer_partner_bank_statements_check_means_answer.feature
colinbruce: That's not the same page, this would be on the income summary page:
image

We could do something like expect `Babysitting\t£44.00` but it would be a bit fragile

PR #7397, opened 2024-11-04
AP-5418: Update save as draft handling in the means pages

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5418)

Previously there was no handler for the Save and come back later button

Now we save the form and redirect to the application list as expected

I have removed the save_as_draft button from the new_benefit_form as the database has been configured to not allow empty fields and this means we cannot save part completed data on this page

I will create a follow up ticket to investigate the impact of changing the database table.  It will have a note to reinstate the save as draft function on this ticket if the DB can be updated

## Checklist

### DIFF
--- a/app/controllers/providers/means/receives_state_benefits_controller.rb
+++ b/app/controllers/providers/means/receives_state_benefits_controller.rb
@@ -7,12 +7,8 @@ module Providers
 
       def update
         @form = ReceivesStateBenefitsForm.new(form_params)
-        if @form.save
-          remove_state_benefits unless applicant.receives_state_benefits?
-          go_forward(applicant.receives_state_benefits?)
-        else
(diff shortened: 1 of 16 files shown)


jsugarman (approved):

jsugarman (approved):
Code looks fine to me 🥇 

It might be worth considering this bugfix that went in a while ago and whether the "new" page for benefits and its flow needs to handle it. essentially new pages can sometimes/somehow store an id with value "new" against a particular model. 

https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/7264

colinbruce:
Good spot, but I'll spin out another ticket as this has bloated enough! :D

PR #7398, opened 2024-11-04
Ap 5389 update heard as alternatives page

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5389)

- Changed the colour for visited link on interrupt page to white.
- Delete the current proceeding and redirect to the proceeding search page if the user selects to Check for another matter type. Also add applications index page to page history as user will no longer be able to navigate back (as the proceeding has been deleted).

## Checklist

### DIFF
--- a/app/controllers/providers/proceedings_sca/interrupts_controller.rb
+++ b/app/controllers/providers/proceedings_sca/interrupts_controller.rb
@@ -5,6 +5,12 @@ module Providers
         type
       end
 
+      def destroy
+        legal_aid_application.proceedings.order(:created_at).last.destroy!
+        replace_last_page_in_history(submitted_providers_legal_aid_applications_path)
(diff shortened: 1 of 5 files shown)


app/controllers/providers/proceedings_sca/interrupts_controller.rb
```diff
@@ -5,6 +5,12 @@ def show
         type
       end
 
+      def destroy
+        legal_aid_application.proceedings.order(:created_at).last.destroy!
```
colinbruce, line 9: order_by(:created_at).last feels a bit fragile, is there no way we can pass the proceeding to the page and be sure?

colinbruce (commented):
Looks good, one comment about fragility

colinbruce (approved):
Discussed and agreed that this is the best way to deal with the situation :+1:

PR #7399, opened 2024-11-04
AP-5398: Update date_field_builder to handle spaces

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5398)

When the builder is reading the form values, it now strips leading and trailing spaces from the individual parts of the date

## Checklist


jsugarman (approved):
👍 

wondered if we need to account for it not being a string given its a form param, or if so might be shorter (and more confusing, coz we love that) as 

```ruby
form.send__(field).to_s.strip!
```

but either way

PR #7405, opened 2024-11-06
ap-5338: add sca fields to application details report

kmahern:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5338)

Added biological_parent?, parental_responsibility_order?, child_subject? and autogranted? columns to the application digest and additionally child_subject_client_involvement_type?  to the application details report.

Whether or not the application is an SCA application can already be determined from the matter_types column and similarly the applicant age from the applicant_age column.

TODO

After discussions, it was decided that it was best to wait to populate the auto_grant_special_children_act? on the LegalAidApplication model until after the autogrant work is complete.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -650,6 +650,18 @@ class LegalAidApplication < ApplicationRecord
     proceedings.where(sca_type: "related")
   end
 
+  def biological_parent?
+    proceedings.any? { |proceeding| proceeding.relationship_to_child.eql?("biological") }
+  end
+
+  def parental_responsibility_order?
+    proceedings.any? { |proceeding| proceeding.relationship_to_child.eql?("court_order") }
+  end
+
+  def child_subject?
+    proceedings.any? { |proceeding| proceeding.relationship_to_child.eql?("child_subject") }
+  end
+
 private
 
   def expired_by_2023_surname_at_birth_issue?
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -650,15 +650,15 @@ class LegalAidApplication < ApplicationRecord
     proceedings.where(sca_type: "related")
   end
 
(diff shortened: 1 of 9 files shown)


colinbruce (dismissed):
LGTM 👍

app/models/legal_aid_application.rb
```diff
@@ -650,6 +650,23 @@ def related_proceedings
...
   end
 
+  def biological_parent_relationship?
+    proceedings.any? { |proceeding| proceeding.relationship_to_child.eql?("biological") }
+  end
+
+  def parental_responsibility_order_relationship?
+    proceedings.any? { |proceeding| proceeding.relationship_to_child.eql?("court_order") }
+  end
+
```
colinbruce, line 660: I think you need to track parental responsibility agreements as well
```ruby
  def parental_responsibility_agreement_relationship?
    proceedings.any? { |proceeding| proceeding.relationship_to_child.eql?("parental_responsibility_agreement") }
  end
```

This will need adding at each stage.

In the mapping there is a requirement for
>Have they stated the supporting evidence for this (follow up questions) is either via court order OR via a PR agreement.

My CCMS code already has something similar.  Happy to pair if needed

colinbruce (changes requested):

app/models/legal_aid_application.rb
kmahern: 👍 I have added a new method to the legal_aid_application model parental_responsibility_evidence? and am reporting this on both the digest and the application_details report - do you think this meets the requirement?

kmahern: 👍 added new parental_responsibility_court_order_relationship? and parental_responsibility_agreement_relationship? methods to the legal_aid_application model and am reporting both of these on the digest and application details report.

colinbruce (approved):
LGTM :+1:

PR #7406, opened 2024-11-06
AP-5434: CYA update - convert existing summary lists to cards

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5435)

- Update summary lists on each CYA page and means/merits reports to include a flag which makes them a summary card
- Convert non summary list/card (often custom) markup to use summary card component
- Remove redundant logic, unnecessary spacing and unused partials

NB. Feature test in features/providers/partner_means_assessment/means_report.feature has been updated and is offering the same test coverage as before. There is a bug (AP-5479) which has been raised and when it is addressed, the values in this feature test will change
## Checklist


app/assets/stylesheets/application.scss
```diff
@@ -31,6 +31,7 @@ $govuk-assets-path: "/";
 @import "table-sort";
 @import "tables";
 @import "sort-transactions";
+@import "summary-card";
```
naseberry, line 34: Import order: are we certain this import order is correct?

app/assets/stylesheets/summary-card.scss
```diff
@@ -0,0 +1,3 @@
+.bold-border-top {
+  border-top: 2px solid #000000
```
naseberry, line 2: Design system: should we use the GDS Sass variables or colour palette instead of the hex value?

app/views/shared/check_answers/_partner_details.html.erb
```diff
@@ -1,7 +1,7 @@
 <% read_only = false unless local_assigns.key?(:read_only)
    heading_class = local_assigns.key?(:heading) ? local_assigns[:heading] : "govuk-heading-m" %>
 <h2 class="<%= heading_class %>"><%= t ".section_heading" %></h2>
-<%= govuk_summary_list(classes: "govuk-!-margin-bottom-9",
+<%= govuk_summary_list(card: { title: "Partner's details" },
```
naseberry, line 4: Internationalization: minor

app/views/shared/check_answers/_proceedings_details.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= govuk_summary_list(actions: false, classes: "govuk-!-margin-bottom-9") do |summary_list| %>
+<%= govuk_summary_list(card: { title: "Proceedings" }, actions: false) do |summary_list| %>
```
naseberry, line 1: Internationalization: minor

app/views/providers/means_reports/_with_cfe_result_details.html.erb
```diff
@@ -90,13 +86,11 @@
...
-<div class="govuk-!-padding-bottom-8"></div>
+  <section class="print-no-break">
+    <h3 class="govuk-heading-m"><%= t(".bank_statement_heading") %></h3>
 
-<section class="print-no-break">
-  <h3 class="govuk-heading-m"><%= t(".bank_statement_heading") %></h3>
-  <% if @legal_aid_application.uploading_bank_statements? %>
     <% has_partner = @legal_aid_application.applicant.has_partner_with_no_contrary_interest? %>
-    <%= govuk_summary_list(actions: false, classes: "govuk-!-margin-bottom-9") do |summary_list| %>
+    <%= govuk_summary_list(card: { title: t(".bank_statement_heading") }, actions: false) do |summary_list| %>
```
colinbruce, line 93: Non blocking question... does adding the title value change the bottom margin or do the new designs call for sections to be closer together?

colinbruce: Hmm, just seen the next page still has the ```classes: "govuk-!-margin-bottom-9"``` so perhaps this was an oversight?

app/views/providers/proceeding_merits_task/check_who_client_is/show.html.erb
```diff
@@ -7,7 +7,9 @@
     <p class="govuk-body"><%= t(".paragraph") %></p>
   <% end %>
 
-  <%= render("shared/check_answers/relationship_to_child", proceeding: @proceeding, relationship_to_child: @relationship_to_child, read_only: false) %>
+  <%= govuk_summary_list(card: { title: page_title }, actions: true) do |summary_list| %>
+    <%= render("shared/check_answers/relationship_to_child", summary_list:, proceeding: @proceeding, relationship_to_child: @relationship_to_child) %>
```
colinbruce, line 11: this no longer has the ```, read_only: false``` attribute

agoldstone93: I think it's because it's not being used in the partial anywhere. And, even though i've changed it, it doesn't look like it was before

app/assets/stylesheets/application.scss
agoldstone93: Where would you think it should go? At the end? I must admit, i put it in kind of at random

app/views/shared/check_answers/_income_result.html.erb
```diff
@@ -1,47 +1,30 @@
...
-
-    <%= check_answer_link(
-          name: :disposable_income_contribution,
-          question: t(".disposable_income_contribution"),
-          answer: gds_number_to_currency(@legal_aid_application.cfe_result.income_contribution),
-          read_only:,
-        ) %>
-  </dl>
+  <%= govuk_summary_list(card: { title: heading },
+                         actions: false,
```
colinbruce, line 3: If this is check_answers, should they not have change links?
I'd expect the reports to not have change links.
The deleted code had ```<% read_only = false unless local_assigns.key?(:read_only) %>``` to check

app/views/shared/means_report/_outgoings_details.html.erb
```diff
@@ -0,0 +1,26 @@
+<section>
+  <% partner = @legal_aid_application.applicant.has_partner_with_no_contrary_interest? %>
```
colinbruce, line 2: minor nit pick: this could be clearer as `partner?` to show it's a boolean check 
Further down it reads as partner object
```if partner && !partner_exclude_items.include?(item.value_method)```

colinbruce (approved):
Slightly 🤯 but nothing serious in my review... 
I think UAT will be the proof in the pudding on this one though! 😄

app/views/shared/check_answers/_income_result.html.erb
agoldstone93: @colinbruce Looks like the only place this is ever called is in partials in the means report, which is why change links are gone. With that in mind, do you think this partial would be better off in the `/means_reports` folder instead?

app/assets/stylesheets/application.scss
naseberry: For now, the order works, based on the class in this new file.
We need to rethink the css architecture to ensure this doesnt become a problem.

colinbruce (approved):
Reviewed additional changes - 👍

PR #7416, opened 2024-11-11
AP-5393: Update SCA client declaration page

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5393)

Update the handling on the view to be reliant on the age of the applicant rather than proceeding types (SCA)

## Checklist

### DIFF
--- a/app/views/providers/confirm_client_declarations/show.html.erb
+++ b/app/views/providers/confirm_client_declarations/show.html.erb
@@ -8,7 +8,7 @@
   <%= page_template page_title: t(".h1-heading"), form: do %>
 
     <% provider_firm_name = @legal_aid_application.provider.firm.name %>
-    <% if @legal_aid_application.non_means_tested? %>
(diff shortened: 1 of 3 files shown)


app/views/providers/confirm_client_declarations/show.html.erb
```diff
@@ -8,7 +8,7 @@
   <%= page_template page_title: t(".h1-heading"), form: do %>
 
     <% provider_firm_name = @legal_aid_application.provider.firm.name %>
-    <% if @legal_aid_application.non_means_tested? %>
+    <% if @legal_aid_application.applicant&.no_means_test_required? %>
```
agoldstone93, line 11: I'm happy with it regardless but would it make more sense to use `under_18?` rather than `no_means_test_required?` because there's no means test if they're over 18 too with SCA, right?

agoldstone93 (approved):
One non-blocking comment

PR #7431, opened 2024-11-13
Fix: nullify student_finance_amount when no

jsugarman:
## What
Fix: nullify student_finance_amount when provider selects "Yes" with amount then amends to "No" 

[Came out of looking at](https://dsdmoj.atlassian.net/browse/AP-5496)

Currently saying yes, adding an amount then coming back and answering No 
will not nullify the student_finance_amount. This is because while it updates the
record it subsequently saves the form values from the form params, which still
contain the `student_finance_amount` value.

### In addition
This refactors the form and partial to transparently use the `student_finance` boolean attribute, via the BaseForm pattern, value rather than use a specific instance variable, `@receives_student_finance`, to set the radio option - taken from `has_national_insurance_number_form` controller and forms pattern.

## Checklist

### DIFF
--- a/spec/requests/providers/partners/student_finances_controller_spec.rb
+++ b/spec/requests/providers/partners/student_finances_controller_spec.rb
@@ -86,12 +86,31 @@ RSpec.describe Providers::Partners::StudentFinancesController do
       end
 
       context "when the provider selects `No`" do
+        let(:student_finance) { "false" }
+
         it "updates the partner" do
           login_as provider
           request
 
           expect(partner.reload.student_finance).to be false
           expect(partner.reload.student_finance_amount).to be_nil
+        end
+      end
+
+      context "when the provider selects `No` after previously selecting `Yes` and adding an amount" do
+        before do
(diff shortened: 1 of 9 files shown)


colinbruce (approved):
LGTM 👍

spec/requests/providers/partners/student_finances_controller_spec.rb
```diff
@@ -95,6 +97,22 @@
...
       end
 
+      context "when the provider selects `No` after previously selecting `Yes` and adding an amount" do
+        before do
+          partner.update!(student_finance: true, student_finance_amount: 1_000.22)
+        end
+
+        let(:student_finance) { "false" }
+
+        it "updates the applicant" do
```
jsugarman, line 107:
```suggestion
        it "updates the partner" do
```

colinbruce (commented):
Changes approved 🏆

colinbruce (approved):
Reviewed form tests :+1:

PR #7435, opened 2024-11-14
AP-5143: initiate state machine after create

jsugarman:
## What
Initiate state machine after application creation

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5143)

With the removal of this hook an inadvertent side effect has been
removed
```
  after_create do
    ActiveSupport::Notifications.instrument "dashboard.application_created", id:, state:
  end
````

The calling of the `state` method in this code means it actually calls
`state_machine_proxy` method, which creates a `state_machine` relation
on the application if it does not have one. Its removal therefore opens
up issues both in tests and possibly functioning of the app.

By replacing it with an explicit call to `state_machine_proxy` we avoid
the issue.

## Checklist

### DIFF
--- a/app/models/legal_aid_application.rb
+++ b/app/models/legal_aid_application.rb
@@ -70,6 +70,8 @@ class LegalAidApplication < ApplicationRecord
   before_save :set_open_banking_consent_choice_at
   before_create :create_app_ref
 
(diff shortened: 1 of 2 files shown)


skdejong (approved):
LGTM

PR #7442, opened 2024-11-18
AP-5268: Add primary nav

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5268)

Add your applications and search to primary nav, displayed on each page. Add individual primary nav component from moj design system. Additional mixins are added too as they are required.

## Screenshot

Updated page looks like this:

image

## Checklist


colinbruce (approved):
LGTM 👍 
Shame there isn't a govuk helper for the navigation tree!

naseberry:
> LGTM 👍 Shame there isn't a govuk helper for the navigation tree!

One for UX, but could we use [GOVUK's Service navigation](https://design-system.service.gov.uk/components/service-navigation/) instead?

This would mean less assets to import, and we have the benefit of using the helper from GOVUK Components for Ruby

agoldstone93:
> > LGTM 👍 Shame there isn't a govuk helper for the navigation tree!
> 
> One for UX, but could we use [GOVUK's Service navigation](https://design-system.service.gov.uk/components/service-navigation/) instead?
> 
> This would mean less assets to import, and we have the benefit of using the helper from [GOVUK Components](https://govuk-components.netlify.app/components/service-navigation/) for Ruby

I did raise this but it was decided that we'd go with the MoJ element. I think it's because this work came about by aiming for consistency across teams and that's what everyone else is using. Definitely something to consider though?

agoldstone93:
> @agoldstone93 Be considerate of the sass import order. At present, any shared selector across across our imports means MoJ styles would override GOVUK, then our custom style would then override both the above.

@naseberry That seems like the order we'd want, isn't it?

PR #7447, opened 2024-11-19
AP-5524: status 404 for page not found

jsugarman:
## What
Change response for page not found to 404

These were previously redirecting/302 and then rendering
a page successfully/200.

In order to capture 404s using prometheus custom rules
we need respond with that status. This could help to
identify attacks or other issues.

### Discussion

This mechanism returns 404 for the page_not_found page response. The actual cause of the redirect and therefore the 404 is lost however. A scripted attack, using curl for example, to probe known paths will not follow the redirect response necessarily and therefore not 404 at all.

This, for example, does not generate any 404s in OpenSearch logs, as each just gets a 302

```shell
for i in {2..10}; do curl -I https://ap-5524-status-404-for-page-no-applyforlegalaid-uat.cloud-platform.service.justice.gov.uk/nada; done
```

but using the browser to go to https://ap-5524-status-404-for-page-no-applyforlegalaid-uat.cloud-platform.service.justice.gov.uk/nada will follow the redirect and result in a 404.

**So, it might be worth spiking changing the entire way we we handle 404s, possibly to use a more rails based approach?**

## Checklist

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -2,6 +2,13 @@ class ErrorsController < ApplicationController
   before_action :update_locale
   def show
     @error = error_param
+
+    case @error
+    when "page_not_found"
+      render status: :not_found
+    else
+      render status: :ok
+    end
   end
 
   def page_not_found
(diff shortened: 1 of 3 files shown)


jsugarman:
superceded

PR #7451, opened 2024-11-19
AP-5371: Update check provider answers page

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5371)

- Code refactor
  - Reduce number of partials to make future development easier.
  - Move text that appears before page heading into the template so
ordering is correct.
- Make changes to `check_provider_answers` to match figma designs

## Checklist


app/views/shared/check_answers/_linking_and_copying.html.erb
```diff
@@ -1,73 +1,55 @@
...
-        <h2 class="govuk-heading-m"><%= t ".other_links_section.#{@legal_aid_application.lead_linked_application.link_type_code}" %></h2>
-      </div>
-    </div>
-    <%= govuk_summary_list(card: { title: t(".other_links_section.#{@legal_aid_application.lead_linked_application.link_type_code}") }, actions: false) do |summary_list|
+      end
+    end %>
+
+<% if @legal_aid_application&.lead_linked_application&.confirm_link? && all_linked_applications_details(@legal_aid_application).present? %>
+  <%= govuk_summary_card(title: t(".other_links_section.#{@legal_aid_application.lead_linked_application.link_type_code}")) do |card|
+        card.with_summary_list(actions: false) do |summary_list|
```
colinbruce, line 29: This is `actions: false`, but line 3 has `actions: !read_only` should they be the same or do we never want a change link for this?

app/views/shared/check_answers/_linking_and_copying.html.erb
```diff
@@ -1,73 +1,55 @@
...
-              )
-            end %>
-      </div>
-    </div>
-    <%= govuk_summary_list(card: { title: t(".section_copying.heading") },
-                           actions: !read_only,
-                           html_attributes: { id: "app-check-your-answers__copying_items" }) do |summary_list|
+<% if @legal_aid_application&.lead_linked_application&.confirm_link? && @legal_aid_application&.lead_linked_application&.link_type_code.eql?("FC_LEAD") %>
+  <%= govuk_summary_card(title: t(".section_copying.heading")) do |card|
+        card.with_action do
```
colinbruce, line 42: This is a third way of handling actions on the same view, can they be normalised or are they all unique?

config/locales/en/shared.yml
```diff
@@ -440,9 +440,11 @@ en:
         first_name: First name
         last_name: Last name
         changed_last_name: Has your client ever changed their last name?
-        last_name_at_birth: What was your client's last name at birth?
+        last_name_at_birth: Last name at birth
+        last_name_at_birth_question: What was your client's last name at birth?
+        same_as_last_name: Same as last name
         nino: National Insurance number
-        has_partner: Does your client have a partner?
+        has_partner: Client has a partner
```
colinbruce, line 447: Should this lose the question mark?  442 still has one

colinbruce (commented):
Looking good, a few questions but nothing to blocking, I hope?!

config/locales/en/shared.yml
agoldstone93: Good question. It doesn't on the design but you're right that it seems a bit inconsistent so I'll double check

app/views/shared/check_answers/_linking_and_copying.html.erb
agoldstone93: This section shows the following
```
other_links_section:
  FC_LEAD: All applications with a family link to this one
  LEGAL: All applications with a legal link to this one
````

As it's showing other applications linked to this, I would expect it to never have a change link

File: config/locales/en/shared.yml (hunk 3 above)
Comment (agoldstone93) [reply config/locales/en/shared.yml]: I've updated to add a question mark

File: app/views/shared/check_answers/_linking_and_copying.html.erb (hunk 2 above)
Comment (colinbruce) [reply app/views/shared/check_answers/_linking_and_copying.html.erb:42]: Three I meant was 
```ruby
actions: !read_only
```
```ruby 
actions: false ```
and
```ruby 
unless read_only
  # add link
```
As long as they are all serving their different purposes, it fine.  I just wanted to make sure we weren't using different patterns randomly! :)

colinbruce (approved):

PR #7453, opened 2024-11-19
AP-5177: Switch to new PDA after login

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5177)

This implements calls to the new, name-spaced PDA, versions of the provider classes, PDA::ProviderDetailsCreator

I had to add some `# :nocov:` tags to the old ProviderDetailsRetriever as it was only tested via other invocations. 
As the intention is to remove this class once we are happy with the new classes, this seemed like a small infraction.  
Feel free to argue in the comments! 😆  

I have tested locally by re-creating my database, re-seeding and deleting a test account, it's firm and the offices. Once pointed at a portal instance, it creates all three objects successfully.

This can be re-created on the branch for testing purposes

## Checklist

### DIFF
--- a/helm_deploy/apply-for-legal-aid/values-uat.yaml
+++ b/helm_deploy/apply-for-legal-aid/values-uat.yaml
@@ -103,7 +103,7 @@ laa_portal:
   idpSsoTargetUrl: "https://portal.uat.legalservices.gov.uk/oamfed/idp/samlv20"
   idpSloTargetUrl: "https://portal.uat.legalservices.gov.uk/oam/server/logout"
   idpCertFingerprintAlgorithm: "<idp-cert-fingerprint-alg-goes-here>"
-  mockSaml: "true"
+  mockSaml: "false"
 
 provider_details:
   url: "https://ccms-pda.stg.legalservices.gov.uk/api/providerDetails"
--- a/spec/requests/saml_sessions_controller_spec.rb
+++ b/spec/requests/saml_sessions_controller_spec.rb
@@ -70,10 +70,9 @@ RSpec.describe SamlSessionsController do
 
           it "updates the record with firm and offices" do
             post_request
-            firm = Firm.find_by(ccms_id: raw_details_response.dig(:firm, :ccmsFirmId))
-            expect(provider.firm_id).to eq firm.id
-            expect(firm.offices.size).to eq 1
-            expect(firm.offices.first.ccms_id).to eq raw_details_response[:officeCodes].first[:ccmsFirmOfficeId].to_s
+            expect(provider.firm.ccms_id).to eq "10835831"
+            expect(provider.firm.offices.size).to eq 1
+            expect(provider.firm.offices.first.ccms_id).to eq "34406595"
           end
 
           context "when mock_saml is activated" do
@@ -181,10 +180,9 @@ RSpec.describe SamlSessionsController do
 
           it "updates the record with firm and offices" do
             post_request
-            firm = Firm.find_by(ccms_id: raw_details_response.dig(:firm, :ccmsFirmId))
-            expect(provider.firm_id).to eq firm.id
-            expect(firm.offices.size).to eq 1
-            expect(firm.offices.first.ccms_id).to eq raw_details_response[:officeCodes].first[:ccmsFirmOfficeId].to_s
+            expect(provider.firm.ccms_id).to eq "10835831"
+            expect(provider.firm.offices.size).to eq 1
+            expect(provider.firm.offices.first.ccms_id).to eq "34406595"
           end
 
           it "displays the start page" do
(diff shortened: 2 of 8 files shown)


helm_deploy/apply-for-legal-aid/values-uat.yaml
```diff
@@ -103,7 +103,7 @@ laa_portal:
   idpSsoTargetUrl: "https://portal.uat.legalservices.gov.uk/oamfed/idp/samlv20"
   idpSloTargetUrl: "https://portal.uat.legalservices.gov.uk/oam/server/logout"
   idpCertFingerprintAlgorithm: "<idp-cert-fingerprint-alg-goes-here>"
-  mockSaml: "true"
+  mockSaml: "false"
```
jsugarman, line 106: I guess this is just for testing. remove prior to merge.

spec/requests/saml_sessions_controller_spec.rb
```diff
@@ -221,34 +224,74 @@
...
+      },
+      user: {
+        userId: 454_031,
+        ccmsContactId: 0,
+        userLogin: "DT_SCRIPT_USER1",
+        name: "Dts User1",
+        emailAddress: "deepak.tanna@digital.justice.gov.uk",
+        portalStatus: "Active",
+      },
+      officeCodes: [
```
jsugarman, line 241: personally I prefer if the stubs only contain the bear minimum information to get the test to pass. This is more informative to our future selves and sets an expectation of only what we need from the API.

spec/requests/saml_sessions_controller_spec.rb
```diff
@@ -221,34 +224,74 @@
...
+          creationDate: "2023-11-10",
+          lscRegion: "Midlands",
+          lscBidZone: "Leicester",
+          lscAreaOffice: "Nottingham",
+          cjsForceName: "Leicestershire",
+          localAuthority: "City of Leicester",
+          policeStationAreaName: "Leicester",
+          dutySolicitorAreaName: "Leicester",
+        },
+      ],
```
jsugarman, line 294: As above, personally I prefer if the stubs only contain the bear minimum information to get the test to pass. This is more informative to our future selves and sets an expectation of only what we need from the API.

spec/requests/saml_sessions_controller_spec.rb
```diff
@@ -1,13 +1,15 @@
...
   let(:firm) { create(:firm, offices: [office]) }
   let(:office) { create(:office) }
   let(:provider) { create(:provider, firm:, selected_office: office, offices: [office], username:) }
   let(:username) { "bob the builder" }
-  let(:provider_details_api_url) { "#{Rails.configuration.x.provider_details.url}#{username.gsub(' ', '%20')}" }
-  let(:provider_details_api_reponse) { api_response.to_json }
+  let(:firm_id) { "3959183" }
+  let(:provider_details_api_url) { "#{Rails.configuration.x.pda.url}/provider-users/#{username.gsub(' ', '%20')}/provider-offices" }
+  let(:provider_firms_api_url) { "#{Rails.configuration.x.pda.url}/provider-firms/#{firm_id}/provider-offices" }
+  let(:provider_details_api_response) { api_response.to_json }
```
jsugarman, line 12: is the api response being used here [this one](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/9e4220d3e8551ed25c880297a320befd249f395f/spec/services/provider_details_creator_spec.rb#L270)? Is that the correct payload to use for the new PDA?

spec/requests/saml_sessions_controller_spec.rb
```diff
@@ -61,16 +64,16 @@
...
           it "calls the Provider details api" do
-            expect(ProviderDetailsCreator).to receive(:call).with(provider).and_call_original
+            expect(PDA::ProviderDetailsCreator).to receive(:call).with(provider).and_call_original
             post_request
           end
 
           it "updates the record with firm and offices" do
             post_request
-            firm = Firm.find_by(ccms_id: raw_details_response[:providerFirmId])
+            firm = Firm.find_by(ccms_id: raw_details_response.dig(:firm, :ccmsFirmId))
```
jsugarman, line 73: My preference for the line below (deleting this line). feel free to ignore. I generally prefer to use literals for comparison. These can be search for/greped easily enough if necessary, and in the case of text values in particular (not these), are more readable and understandable.

```suggestion
  expect(provider.firm_id).to eq 10_835_831
```

jsugarman (commented):
some questions and comments on the spec mainly

helm_deploy/apply-for-legal-aid/values-uat.yaml
colinbruce: It's in the `TEMP: Use UAT portal login, DO NOT MERGE - testing only` commit! :D

spec/requests/saml_sessions_controller_spec.rb
colinbruce: No, it should be set in individual test blocks, e.g. 
```ruby 
  let(:api_response) { raw_details_response }
```
Like [here](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/7453/files#diff-589b797e671e70b24d3566c55ac9dda4eacba6297ab202df568cbfd786e6ac33L59)

colinbruce: I've checked and removing the linked code does not make these tests fail

jsugarman: cool, yes had not noticed the let as it was not part of the changes itself 👍

jsugarman (approved):
LGTM 👍

PR #7454, opened 2024-11-19
AP-5524: Replace error handling with exceptions_app

jsugarman:
## What 
Implement custom error handling using a rails approach

This uses application config `exceptions_app` to wire up our own exception handling via the same controller. it enables a page not found to raise 404 and not redirect, so that logs will informs us what caused the 404. It also handles internal servers errors, which currently aren't, by displaying the custom error page but still resulting in 500s as there are no redirects. Access denied and "Assessment already completed" are handled as they were previously.

### Notes
Idea from https://dev.to/ayushn21/custom-error-pages-in-rails-4i43

See rails documentation for [exceptions_app](https://guides.rubyonrails.org/configuring.html#config-exceptions-app)

### To test
- Access denied can be tested on this branch by attempting to login to /admin using googleauth
- Page not found can be tested by a) amending url to /nada or anything other nonsense. b) amending the id of the application in the url (this exercises the RecordNotFound angle)
- Internal Server error can be tested by creating a new application on this branch and entering the name "joely" for the first name - this will trigger an Argument error. This relies on the TEMP commit.
- Assessment already completed can be excercised by following the open banking journey and completed the mock bank sharing of financial information and then attempting to go back and "reshare" by the browser or the email ink.

### TODO

- test logging for 404s works as expected ✅ 
- handle welsh translation issues for controller ✅ 
- handle internal server errors ✅ 
- consider need to handle `ActionDispatch::Http::MimeNegotiation::InvalidType` as [mentioned here](https://guides.rubyonrails.org/configuring.html#config-exceptions-app). _I have tried this using curl and it results in an unhandled 500 as is (browser handled only). It may be worth extra work but a quick spike proved it to be problematic._ 

### Discussion

This would fundamentally change the way we handle errors but does mean a call to a non-existent endpoint returns a 404

### Tests

**Assessment already completed test**: I have tested the citizen path for open banking and it continues to handle the need to redirect citizens to the "already shared your financial information" page if they back page or click the link in the email again after having completed the flow.

Screenshot 2024-11-20 at 15 27 29

**Access denied test**. This was performed using admin google auth admin login to simulate the failure

Screenshot 2024-11-20 at 16 07 44

**Internal server error**
This is performed by starting a new application and using the name "joely" for the first name and then saving
Screenshot 2024-11-21 at 15 48 15

## Checklist

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,18 +4,7 @@ class ApplicationController < ActionController::Base
   include OmniauthPathHelper
   helper_method :omniauth_login_start_path
 
-  # See also catch all route at end of config/routes.rb
-  rescue_from ActiveRecord::RecordNotFound, with: :page_not_found
-
 private
-
-  def page_not_found(exception)
-    # TODO: alerting and logging added to handle bug AP-5365 but could keep in?!
-    AlertManager.capture_exception(exception)
-    Rails.logger.error(exception.message)
-    update_locale
-    redirect_to error_path(:page_not_found, default_url_options)
-  end
 
   def default_url_options
     { locale: I18n.locale }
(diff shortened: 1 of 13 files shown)


colinbruce (approved):
LGTM :+1:

PR #7464, opened 2024-11-21
AP-5490 add inset text to CYA pages

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5490)

Add inset text to inform users that they cannot go back once they have gone past a CYA page

## Checklist

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -1344,6 +1344,7 @@ en:
     check_capital_answers:
       show:
         h1-heading: Check your answers
+        inset_text: You cannot change the answers on this page once you save and continue.
         capital-section-heading: Your client's capital
         capital-section-heading_with_partner: Your client's and their partner's capital
         what_happens_next:
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -1344,7 +1344,6 @@ en:
     check_capital_answers:
       show:
         h1-heading: Check your answers
-        inset_text: You cannot change the answers on this page once you save and continue.
         capital-section-heading: Your client's capital
         capital-section-heading_with_partner: Your client's and their partner's capital
         what_happens_next:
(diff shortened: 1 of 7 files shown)


jsugarman (approved):
👍 

Might be worth:

1 . using an anchored version for the translations as they are all the same
2. Having a basic feature test step in appropriate CYA features (i think there is a `check_capital_answers` and `check_provider_answers` feature)

config/locales/en/providers.yml
```diff
@@ -268,6 +268,7 @@ en:
       show:
         banner_text: The merits information was copied from %{application_ref}.
         h1-heading: Check your answers
+        inset_text: You cannot change the answers on this page once you save and continue.
```
colinbruce, line 271: Is it worth moving this to the `generic` block so it's only stored onece, then it can be used on pages with 
```ruby
t(".generic.inset_text")
```

RoseSAK: Thanks, I have just updated the text to be stored in `generic.yml` and have added a line to the CYA feature test (as suggested by @jsugarman )

config/locales/en/providers.yml
```diff
@@ -1344,6 +1344,7 @@ en:
     check_capital_answers:
       show:
         h1-heading: Check your answers
+        inset_text: You cannot change the answers on this page once you save and continue.
```
jsugarman, line 1347: not a big deal but do we need this, since we are now calling `t("generic.cya_inset_text")` directly in the views. 

Apologies it it was not clear but what I was intending was something like 

```yml
   # anchor somewhere like generic.yml
   cya_inset_text: &cya_inset_text You cannot change the answers on this page once you save and continue.

   # splat/use in each view as relative translation
   inset_text: *cya_inset_text
```

colinbruce: No, it looks like it got missed in the amendment tidy up. I'll remove and merge this :+1:

PR #7469, opened 2024-11-22
AP-5524: Change access denied status to 403

jsugarman:
## What
Change status code for the access_denied page redirect to 403

So the response status matches the intended behaviour of being
forbidden/403 for the current user.

A subsequent commit removes sentry capturing of details of failures
because the originating PR suggests doing so and there does not appear
to have been any action taken on these, if they are still occuring (which I cannot
see).

## Checklist


agoldstone93 (approved):
Looks good to me

PR #7483, opened 2024-11-27
Chore: Fix Welsh locale flicker

colinbruce:
## What

A recent change to the error controller removed the around block for the welsh language checks.  This was needed as, otherwise, it leaves the locale set to CY for subsequent tests.  This caused flickers in CI.

This PR creates a new rspec flag, :use_welsh_locale, to reduce the noise in individual rspec files, while still allowing the locale to be set for an individual group of tests

## Checklist


kmahern (approved):
LGTM

PR #7484, opened 2024-11-27
Discussion: Allow CSP connect to localhost

jsugarman:
## What
Allow CSP connect to localhost

[Came out of wanting to tes story](https://dsdmoj.atlassian.net/browse/AP-5377)

Allows communication with LFA, at least, via JS so
we can test proceeding types and merits work locally.

## Checklist


colinbruce (commented):
I have a stashed commit that I use to do the same, when needed
I was never sure if this was "acceptable" 🤷‍♂️ 
I think the problem is not everyone runs the server on 3000, I'm sure I've paired with people who use `5000`

PR #7495, opened 2024-12-02
AP-5566: Prevent infinite proceeding loop

jsugarman:
## What
Prevent infinite proceeding loop

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5566)

When the provider says they do not wish to accept the
substantive default level of service then the Defaults
are cleared and the are taken to the page to select
non default level of service. However if there is only
one level of service then they are unable to select that
one and so the level of service remains cleared.

On ending the proceeding loop for this proceeding it is checking
for completeness, which it fails because it has no level of service
`stage` details being set, only the "name" and "id".

This fix simply adds the `stage` at that point.

## TODO
- [X] assign/use the default substantive levels of service when only 1 available
- [X] feature test proceeding loop substantive levels of service when only 1 available
- [X] unit test proceeding loop substantive levels of service when only 1 available?
- [x] check assign of level of service name impact when only 1 available - see `SubstantiveLevelOfServiceForm` line 22
- [X] assign/use the default emergency levels of service when only 1 available
- [X] feature test proceeding loop emergency levels of service when only 1 available
- [X] unit test proceeding loop emergency levels of service when only 1 available?
- [x] check assign of level of service name impact when only 1 available - see `EmergencyLevelOfServiceForm` line 22

## Checklist


colinbruce (approved):
LGTM 👍

PR #7500, opened 2024-12-03
AP-5529: Add  Join our research panel link to the footer

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5529)

This was raised by the central UR team.  They have requested that the url be hidden from github crawlers, so the form link is stored in AWS Secret Manager for each environment.

It is then loaded in _ENVS.tpl and passed to the footer meta items.

This routes to root_path if the value is missing, this should only affect devs without a RESEARCH_PANEL_FORM_LINK in their .env files, as each environment has the value set

## Checklist


kmahern (approved):
lgtm

jsugarman (approved):

PR #7502, opened 2024-12-03
AP-5162 Remove git-crypt

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5162)

Remove all references and usages of git-crypt from config, setup and README

## Checklist


.gitignore
```diff
@@ -31,8 +31,7 @@
 # Ignore node_modules
 node_modules/
 
-# Ignore git-crypt key and Rails master key
-git-crypt.key
```
jsugarman, line 35: i'd be tempted to leave this in for past, future and general safety. I do not have this file locally nor the  `config/master.key` below but it can be useful to have things in ,gitignore that others may have in their root still or to mitigate against accidents.

jsugarman (approved):
looks good but, comment about perhaps leaving the `.gitignore` line in

.gitignore
RoseSAK: Okay good point, I will add back in. I also don't have either of these so thought it might be a legacy thing. But the point still stands

colinbruce (approved):
LGTM :+1:

PR #7507, opened 2024-12-04
AP-5589: Stop preselecting emergency level of service

jsugarman:
## What
Stop preselecting emergency level of service

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5589)

Came out of UAT of another bug.

The emergency levels of service were being cleared but not saved
due to the location of the `super` call in the local form `save`
method, I believe.

This duplicates what the substantive defaults form does more
closely which does ensure that the clearing of defaults is followed
by a save.

## Checklist


colinbruce (approved):
LGTM :+1:

PR #7512, opened 2024-12-05
AP-4409: Add the related_orders page to the PLF proceeding loop

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4409)

Add new interrupt pages for PLF proceedings

This includes a new variety of proceeding interrupt pages.  I deliberately did not put them into the proceedings_sca namespace

Should we re-factor the other proceeding interrupts into a more generic namespace to keep them contained, but not tied explicitly to a certain proceeding type?

## Checklist

### DIFF
--- a/app/controllers/providers/proceeding_interrupts/related_orders_controller.rb
+++ b/app/controllers/providers/proceeding_interrupts/related_orders_controller.rb
@@ -1,0 +1,39 @@
+module Providers
+  module ProceedingInterrupts
+    class RelatedOrdersController < ProviderBaseController
+      before_action :proceeding
+      def show
+        @form = RelatedOrdersForm.new(model: proceeding)
+      end
+
+      def update
+        @form = RelatedOrdersForm.new(form_params)
+        return redirect_to providers_legal_aid_application_sca_interrupt_path(legal_aid_application, "plf_none_selected") if redirect_on_none_selected?
+
+        render :show, status: :unprocessable_content unless save_continue_or_draft(@form, proceeding:)
+      end
+
+    private
+
+      def redirect_on_none_selected?
+        form_params[:none_selected].present? && !params.key?(:draft_button)
+      end
+
+      def proceeding
+        @proceeding ||= Proceeding.find(proceeding_id_param)
+      end
+
+      def proceeding_id_param
+        params.require(:id)
+      end
+
+      def form_params
+        merge_with_model(proceeding) do
+          params
+            .require(:proceeding)
+            .permit(:none_selected, related_orders: [])
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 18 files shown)


app/controllers/providers/proceeding_interrupts/related_orders_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  module ProceedingInterrupts
+    class RelatedOrdersController < ProviderBaseController
+      before_action :proceeding
+      def show
+        @form = RelatedOrdersForm.new(model: proceeding)
+      end
+
+      def update
+        @form = RelatedOrdersForm.new(form_params)
+        return redirect_to providers_legal_aid_application_sca_interrupt_path(legal_aid_application, "plf_none_selected") if redirect_on_none_selected?
```
kmahern, line 11: Not a blocker: I find it slightly confusing that we are redirecting to a path in the SCA namespace for a PLF proceeding. Would it be worth moving the interrupts pages out of a specifically SCA namespace (could be a separate ticket)?

kmahern (approved):
One non-blocking comment for consideration otherwise LGTM 👍

app/controllers/providers/proceeding_interrupts/related_orders_controller.rb
colinbruce: AP-5607 has been created

app/forms/providers/proceeding_interrupts/related_orders_form.rb
```diff
@@ -0,0 +1,63 @@
...
+      ORDER_TYPES = %i[
+        care
+        supervision
+        secure_accommodation
+        placement
+        adoption
+        emergency_protection
+        child_assessment
+        parenting
+        child_safety
```
jsugarman, line 15: BTW, what was the outcome of the conversation with design about asking whether this "Placement order" proceeding relates to a "placement order" issue?

jsugarman (approved):
On the refactor. Makes sense to me. How many other interrupts controllers are there for SCA specifically, or other matter types? be good to move all them in to the new structure as part of this or as soon as possible thereafter would be my only observation.

app/forms/providers/proceeding_interrupts/related_orders_form.rb
colinbruce: It doesn't! There was a step missing from the Figma... only certain PLF proceedings hit this check page, placement orders (as an example) wouldn't! :smilr

colinbruce:
> On the refactor. Makes sense to me. How many other interrupts controllers are there for SCA specifically, or other matter types? be good to move all them in to the new structure as part of this or as soon as possible thereafter would be my only observation.

AP-5607 has all the affected pages, there are currently 9 different proceeding interrupts

PR #7520, opened 2024-12-10
Fix: Flickering CCMS case requestor test

colinbruce:
## What

This is caused by the set-up creating many opponents and a race condition meaning they sometimes get created in a different sequence.

By converting the two hashes into json we can run a, sequence idempotent, match_json_expression comparison

This ensures the payloads match, regardless of sequence

### How?
To track this down I re-created a failing CircleCI run

Adding the tests from the split job and re-using the same seed value.  Then, with a binding.pry, ensured the test woudl fail and then tried different match methods until I could consistently get it to pass

## Checklist


jsugarman (approved):
LGTM 👍

PR #7530, opened 2024-12-11
LnD: Update Alpine and ruby of base image

colinbruce:
## What

Update Alpine to 3.21 and ruby to 3.3.6

This should reduce the noise from the ruby parser about using a 3.3.6 parser with 3.3.5
It _may_ also reduce security warnings if the go update in alpine addresses some of the issues

## Checklist


jsugarman (approved):
LGTM - we should check pdf reports creation manually at least I would say, in case of grover/puppeteer system lib dependencies 🤷

colinbruce:
image

PR #7532, opened 2024-12-11
AP-5530: Add original judge level question

jsugarman:
## What
Add original judge level question

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5530)

Inline with requirements and designs for PLF

- [x] question is added following the design
- [x] ~~merits task list label is added~~
- [x] check your answers is updated
- [x] merits report to be updated
- [x] review and print is updated
- [x] certificate is updated

### Question
Screenshot 2024-12-13 at 11 08 26

### Display (in CYA, R&P, Certificate page and merits report equivalent)
Screenshot 2024-12-13 at 11 11 55

### Merits report 
Screenshot 2024-12-13 at 11 11 10

## Checklist

### DIFF
--- a/features/step_definitions/merits_task_list_steps.rb
+++ b/features/step_definitions/merits_task_list_steps.rb
@@ -1,0 +1,16 @@
+Given("I complete the journey as far as merits task list for a PLF proceeding with second appeal question") do
+  @legal_aid_application = create(
+    :legal_aid_application,
+    :with_non_means_tested_state_machine,
+    :with_positive_benefit_check_result,
+    :with_proceedings,
+    :with_applicant,
+    :provider_entering_merits,
+    explicit_proceedings: %i[pbm01a],
+    set_lead_proceeding: :pbm01a,
+  )
+  create(:legal_framework_merits_task_list, :pbm01a_as_applicant, legal_aid_application: @legal_aid_application)
+
+  login_as @legal_aid_application.provider
+  visit(providers_legal_aid_application_merits_task_list_path(@legal_aid_application))
+end
--- a/app/controllers/providers/application_merits_task/original_judge_levels_controller.rb
+++ b/app/controllers/providers/application_merits_task/original_judge_levels_controller.rb
@@ -1,0 +1,29 @@
+module Providers
+  module ApplicationMeritsTask
+    class OriginalJudgeLevelsController < ProviderBaseController
+      def show
+        @form = OriginalJudgeLevelForm.new(model: appeal)
+      end
+
+      def update
+        @form = OriginalJudgeLevelForm.new(form_params)
+
+        render :show unless update_task_save_continue_or_draft(:application, :second_appeal)
+      end
+
+    private
+
+      def appeal
+        legal_aid_application.appeal
+      end
+
+      def form_params
+        merge_with_model(appeal) do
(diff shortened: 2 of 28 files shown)


features/step_definitions/merits_task_list_steps.rb
```diff
@@ -0,0 +1,25 @@
...
+  create(:legal_framework_merits_task_list, :pbm01a_as_applicant, legal_aid_application: @legal_aid_application)
+
+  # Mark other merits questions as complete to enable CYA flow for second appeal
+  # %i[opponent_name statement_of_case children_application second_appeal].each do |merit_task|
+  #   @legal_aid_application.legal_framework_merits_task_list.mark_as_complete!(:application, merit_task)
+  # end
+
+  # %i[children_proceeding chances_of_success].each do |merit_task|
+  #   @legal_aid_application.legal_framework_merits_task_list.mark_as_complete!(:PBM01A, merit_task)
+  # end
```
jsugarman, line 21: remove

app/models/application_merits_task/appeal.rb
```diff
@@ -1,5 +1,15 @@
...
   class Appeal < ApplicationRecord
     belongs_to :legal_aid_application
+
+    before_save :sync_original_judge_level
+
+  private
+
+    def sync_original_judge_level
+      if second_appeal?
+        self.original_judge_level = nil
```
colinbruce, line 11: I don't understand this, why set the value to nil if changing second_appeal?
I'm sure there's a good reason, but maybe a comment to clarify it would help future devs?

colinbruce (approved):
Looks okay, one method that could maybe be clearer, but non-blocking

app/models/application_merits_task/appeal.rb
jsugarman: sure, will add. it's because the original judge question is not required for second appeals, only for "first" appeals. So if they change their answer from "No", to "Yes" then we want to clear the now irrelevant original_judge_level

jsugarman: yeah, i see other examples in forms. I guess my database centric brain just immediately went for a model callback, of which there are a fair few in the codebase (before_save, after_commit). I guess i should follow the "form" based pattern as it is more prevalent, even if it feels less safe to me - as someone who loves updating data in the console 😄.

jsugarman: @colinbruce I have added a commit that does the same thing via a form based approach, with some comments on impact. I am not fussed but while it does adhere to the pattern of similar forms and is less "magical" than callbacks, it has disadvantages too. 

What do you prefer. Happy to go with this regardless, but may be worth a wider discussion going forward (have put a placeholder in dev time trello for now regardless)

jsugarman: @colinbruce have added a second alternative which:

1. uses custom callbacks so we don't not have to override default `def save` 
2. uses attribute assignment rather than model update* 

_*which i think would have worked via the `def save` override approach too, but does not allow for "where" to do the assignment except before or after whatever `super` would call_

I rather like the callback approach to help keep form objects small/thin

jsugarman: Agreed by developers to introduce this new pattern

PR #7548, opened 2024-12-18
AP-4790: Update accessibility statement and change approach to translations for it

jsugarman:
## What
Update accessibility statement and change approach to translations

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4790)

Following handling of user accessibility report we need to update the
accessibiility statement.

This also changes the approach for rendering the accessibility statement 
from using a single view that uses separate translation files in favour of
separate english and welsh views. This is because large [almost entirely]
static html pages are more easily amended than calls to translation files,
especially when we do not actually have any actual welsh translations.

## TODO

- [X] change approach to translations
- [x] update statement 
- [x] Add number of days for response times
- [x] Look at amending content for "Content that’s not within the scope of the accessibility regulations"

## Checklist


RoseSAK (approved):

app/views/accessibility_statement/index.en.html.erb
```diff
@@ -0,0 +1,81 @@
...
+
+  <p class="govuk-body">
+    <%= govuk_link_to("AbilityNet has advice on making your device easier to use", "https://mcmw.abilitynet.org.uk/") %> if you have a disability.
+  </p>
+
+  <h2 class="govuk-heading-m">Feedback and contact information</h2>
+
+  <p class="govuk-body">
+    We’re always looking to improve the accessibility of this service. If you find any problems that are not listed on this page or think we’re not
+    meeting accessibility requirements, email apply-for-civil-legal-aid@digital.justice.gov.uk giving details of the issue and any assistive technology you are using.
```
naseberry, line 32: Nit: would it be ideal for to use the feedback form instead of the team email?

naseberry (approved):

app/views/accessibility_statement/index.en.html.erb
jsugarman: 🤷 - More a question for PMs/Design i'd say but the in-app feedback form has unrelated questions that might be irrelevant/unrelated to accessibility.

PR #7571, opened 2025-01-10
Chore: Address transaction type todo

colinbruce:
## What

While removing the citizen values and update tests to stop referring to them, I discovered that the only place referring to the citizen values were the user_transaction_helper specs.

When I investigated the helper, I could find no references to it's methods in the codebase.

Therefore I have removed the user_transaction_helper entirely and removed the citizen section of the locale file. The citizen path was phased out many years ago and they were left in place because of the now unused helper.

## Checklist


jsugarman (approved):
🔥 LGTM

PR #7573, opened 2025-01-13
AP-4439: Add a "removed bank transaction" flash message

jsugarman:
## What
Add a "removed bank transaction" flash message

[Link to story](https://dsdmoj.atlassian.net/browse/AP-4439)

For accessibility we need to announce the removal of the
record to the page/screen reader. For progressive web app
reasons we need to ensure removal is also available and
announced to screen readers when javascript is disabled.

This provides mechanism that should be both accessible and
progressive. 

The remove action is **Accessible** because it adds a 
govuk notification banner (which is accessible) whenever a bank 
transaction is removed. In addition [all] flash messages have had an `aria-live: assertive`
and `aria-atomic: true` attribute added to their outer `div` inline with suggestions from
[a11yonrails](https://www.a11yonrails.com/pages/flash.html)

The remove action is also **Progressive** because without javascript it also 
adds the flash and reloads [redirects back to] the page using an idiomatic rails 
approach.

The approach uses a `js.erb` in order to share the flash view partials logic
rather than have a javascript version and standard request version.

## Testing

The **accessibility** aspect requires testing with screen reader technologies that I do not have
so will need to be checked by our tester using one or more of the suggested tooling - JAWs, NvDA,
Voiceover, Dragon speaking naturally. I have only tested with Mac Voiceover which is not very reliable 
apparently.

## Screenshot After

Screenshot 2025-01-14 at 12 08 47

## Checklist

### DIFF
--- a/app/controllers/providers/bank_transactions_controller.rb
+++ b/app/controllers/providers/bank_transactions_controller.rb
@@ -2,13 +2,33 @@ module Providers
   class BankTransactionsController < ProviderBaseController
     def remove_transaction_type
       bank_transaction.update!(transaction_type: nil, meta_data: nil)
-      head :ok
+
+      flash[:hash] = success_hash
+
+      respond_to do |format|
+        format.js
+        format.html { redirect_back fallback_location: providers_root_path }
+      end
     end
 
   private
 
     def bank_transaction
       @bank_transaction ||= @legal_aid_application.bank_transactions.find(params[:id])
+    end
+
+    def success_hash
+      {
+        title_text: t("generic.success"),
+        success: true,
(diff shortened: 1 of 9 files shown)


colinbruce (approved):
I can't see any obvious coding issues but UAT will be very important! :)

jsugarman:
> I can't see any obvious coding issues but UAT will be very important! :)

Thanks @colinbruce., yeah we discussed with design and UR and it was underlined that testing would need to check the accessibility aspect using a "proper" screen reader (JAWs, NvDA. Dragon...) to check the accessibility.

PR #7576, opened 2025-01-13
Chore: Remove next_action_link helper method

agoldstone93:
## What

Remove next_action_link as the functionality can just as easily be provided with the govuk_button_link_to method from the govuk_components gem

## Checklist

### DIFF
--- a/app/views/shared/_worker_complete.html.erb
+++ b/app/views/shared/_worker_complete.html.erb
@@ -4,14 +4,10 @@
 <% continue_text = locals[:continue_text] || nil %>
 
 <%= page_template page_title: heading do %>
-  <p>
-    <%= intro_text %>
-  </p>
-  <p>
-    <%= balance_text %>
-  </p>
-  <div class="govuk-!-padding-top-3">
-    <%= next_action_link(continue_text:) %>
-  </div>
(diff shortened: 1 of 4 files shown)


app/views/shared/_worker_complete.html.erb
```diff
@@ -4,14 +4,10 @@
...
-  <p>
-    <%= balance_text %>
-  </p>
-  <div class="govuk-!-padding-top-3">
-    <%= next_action_link(continue_text:) %>
-  </div>
+  <p class="govuk-body"><%= intro_text %></p>
+  <p class="govuk-body"><%= balance_text %></p>
+
+  <%= govuk_button_link_to(continue_text, forward_path) %>
```
jsugarman, line 10: does this need the `govuk-!-padding-top-3` still? as did this one `app/views/citizens/accounts/index.html.erb`

jsugarman (approved):

app/views/shared/_worker_complete.html.erb
agoldstone93: @jsugarman I removed it intentionally because it was just a standard button below a standard paragraph so I thought it should have the default padding. Happy to put it back in if you think I should though

PR #7578, opened 2025-01-14
AP-5647: Update proceeding captions

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5467)

Switch from proceedings.last to proceedings.order(:created_at).last

Because we use GUIDs as the id, there is no value in a plain .last call as it orders by GUID text.  Adding the order allows us to get the most recently created proceeding, which is what is expected, without having to make the URLs and routing more complex by adding specific proceeding IDs into the routes.

This also matches the existing use in the ProceedingIssueStatusesController

## Checklist

### DIFF
--- a/app/controllers/providers/change_of_names_controller.rb
+++ b/app/controllers/providers/change_of_names_controller.rb
@@ -1,14 +1,14 @@
 module Providers
   class ChangeOfNamesController < ProviderBaseController
     def show
-      @proceeding = legal_aid_application.proceedings.last
+      @proceeding = legal_aid_application.proceedings.order(:created_at).last
       form
     end
 
     def update
       return continue_or_draft if draft_selected?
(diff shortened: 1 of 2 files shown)


agoldstone93 (approved):
This looks like an improvement. Happy to approve. I'm wondering if there's any weirdness when you add and delete proceedings but fingers crossed if there is, it'll come out in UAT

colinbruce:
> This looks like an improvement. Happy to approve. I'm wondering if there's any weirdness when you add and delete proceedings but fingers crossed if there is, it'll come out in UAT

I _think_ it should be okay, because it should always be for the most recently added proceeding, i.e. I have just added X - now ask the question but, as you say, the proof will be in the UAT! 🙂

PR #7597, opened 2025-01-27
AP-5379: Update chances of success handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5379)

This PR updates the chances/prospects of success handling.  It merges the two pages into a simple yes/no question with a reason for a no answer.

The original second page with multiple choices for less then likely, marginal, poor, etc is deleted

The model still holds the same values as previously, just saved differently, meaning that no changes are needed for the CCMS submissions

I have left the options for less than likely in the model and CCMS submission builders as we have seen logic and flow reversals in the past... in case you are looking at this in the future and wondering why there is a lot of unused code!

## Checklist


jsugarman (commented):
It looks good to me code/test wise.

Are you setting the success_prospect to not_known so we can distinguish new ones from old ones? Was thinking `nil` might suffice since we are not collecting that data?!

It seems from what you have said, the code and the stats that existing data mid-flow is unlikely to be affected, especially as 99.9% are "likely" 
```
{"marginal" => 4, nil => 2, "likely" => 27249, "borderline" => 2}
```
-

jsugarman:
Not a blocke by any means but I meant to also mention that we could take the opportunity to change the chances of success page to a show action/page instead of its current index action/page, bringing it inline with the other merits questions.

jsugarman (approved):

colinbruce:
> It looks good to me code/test wise.
> 
> Are you setting the success_prospect to not_known so we can distinguish new ones from old ones? Was thinking `nil` might suffice since we are not collecting that data?!
> 
> It seems from what you have said, the code and the stats that existing data mid-flow is unlikely to be affected, especially as 99.9% are "likely"
> 
> ```
> {"marginal" => 4, nil => 2, "likely" => 27249, "borderline" => 2}
> ```

This is by design.  The `not_known` value is used by CCMS to populate the payload and was previously used on the second page to record a text entry explanation

PR #7604, opened 2025-01-31
AP-5701: PLF proceeding loop limit

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5701)

Limit the ability to proceed if the provider has chosen _all_ applicable proceedings

## Checklist


app/services/legal_framework/proceeding_types/all.rb
```diff
@@ -37,7 +39,10 @@ def initialize(legal_aid_application)
       end
 
       def call
-        result = JSON.parse(request.body).map { |pt_hash| ProceedingTypeStruct.new(pt_hash) }
+        parsed_body = JSON.parse(request.body)
+        raise NoMatchingProceedingsFoundError, "nope!" if parsed_body.is_a?(Hash) && parsed_body["success"] == false
```
jsugarman, line 43: 😄, perhaps a "proceeding not found!"

jsugarman: Do we need to test it is a hash? it was not being before but either way

colinbruce: 😬 that was a debug step that should have been removed! 🤦‍♂️

colinbruce: The problem is that a successful return is an array. An unsuccessful result is a Hash, so `if parsed_body["success"]` will error on a successful call 😖

jsugarman (approved):
👍 

Agreed to write up ticket to convert LFA's `ProceedingTypes::FilterController`, as well other other LFA endpoints to have a consistent response structure where possible and sensible. These would return a top level attribute and value of `"success": true|false`.

This could be used to allow `LegalFramework::ProceedingTypes::All.call(model).count` to return 0 directly rather than having a throw/catch mechanism to return 0.

app/services/legal_framework/proceeding_types/all.rb
colinbruce: spun out into new ticket

PR #7607, opened 2025-02-03
AP-5725: Update state change for linked cases

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5725)

The method of changing the state from initiated to entering_applicant_details was only being called on certain address and the limitation controllers.  When a provider chose a business address and linked a case, it bypassed all of the state change mechanisms.

This PR moves the address state change to the first address page which should not be bypassed in any of our current flows

## Checklist

### DIFF
--- a/app/controllers/providers/address_selections_base_controller.rb
+++ b/app/controllers/providers/address_selections_base_controller.rb
@@ -5,8 +5,6 @@ module Providers
     def show
       return redirect_to back_path unless address.postcode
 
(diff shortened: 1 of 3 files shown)


jsugarman (approved):
Makes good logical sense to me. 

Somewhat perturbed by the fact that no tests broke when this was moved 😬  - is it worth a request spec [extension] to cover this [not to mention similar state changes in controllers]

PR #7628, opened 2025-02-17
AP-5669: Rails update to 8.0.1

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5669)

- Update config files rails v7.2
- Perform all updates needed to update to rails v8.0.1
- Fix error in routes: combined two routes that only need to be one and updated singleton resource to use `show` action rather than `index` action
- Remove inline styling as javascript wasn't loading it properly, because of our content security policy

## Checklist


colinbruce (approved):
Nice work, LGTM  👍

config/environments/production.rb
```diff
@@ -11,46 +11,63 @@
...
 
   # Specifies the header that your server uses for sending files.
-  # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
-  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
+  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
+  # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
 
-  # Mount Action Cable outside main process or domain
+  # Store uploaded files on the local file system (see config/storage.yml for options).
+  config.active_storage.service = :local
```
jsugarman, line 25: This looks wrong. should this not be 

```ruby
config.active_storage.service = :amazon
```

config/environments/production.rb
```diff
@@ -1,66 +1,78 @@
...
 
   # Prepend all log lines with the following tags.
   config.log_tags = [:request_id]
 
-  # Use a different cache store in production.
+  # Change to "debug" to log everything (including potentially personally-identifiable information!)
+  config.log_level = ENV.fetch("RAILS_LOG_LEVEL", "info")
+
+  # Prevent health checks from clogging up the logs.
+  config.silence_healthcheck_path = "/up"
```
jsugarman, line 47: This should be using our healthcheck endpoint `/status` - `/up` is the modern rails default healthcheck endpoint provided out of the box nowadays

```suggestion
  config.silence_healthcheck_path = "/status"
```

config/environments/production.rb
```diff
@@ -85,6 +97,18 @@
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
 
+  # Only use :id for inspections in production.
+  config.active_record.attributes_for_inspect = [:id]
```
jsugarman, line 101: To emulate current behaviour we may need this. which i had applied to assure. Without it I _think_ it may not allow you to not display record details expect for id   - see https://github.com/rails/rails/issues/52728

```suggestion
  config.active_record.attributes_for_inspect = :all
```

config/environments/test.rb
```diff
@@ -1,40 +1,44 @@
...
-  # The test environment is used exclusively to run your application's
-  # test suite. You never need to work with it otherwise. Remember that
-  # your test database is "scratch space" for the test suite and is wiped
-  # and recreated between test runs. Don't rely on the data there!
   config.cache_classes = false
 
-  # Eager loading loads your whole application. When running a single test locally,
-  # this probably isn't necessary. It's a good idea to do in a continuous integration
-  # system, or in some way before deploying your code.
+  config.enable_reloading = true
```
jsugarman, line 10: I believe for test runs, including cucumber, you do not need to reload during test runs, and may slow down test runs 🤷 
```suggestion
# While tests run files are not watched, reloading is not necessary.
config.enable_reloading = false
```


```

File: config/environments/test.rb [hunk 5]
```diff
@@ -1,40 +1,44 @@
...
+# and recreated between test runs. Don't rely on the data there!
+
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
-
-  # The test environment is used exclusively to run your application's
-  # test suite. You never need to work with it otherwise. Remember that
-  # your test database is "scratch space" for the test suite and is wiped
-  # and recreated between test runs. Don't rely on the data there!
   config.cache_classes = false
```
jsugarman, line 8: this option can be removed as has been replaced by `config.enable_reloading`

agoldstone93: @jsugarman Everything fails when I do that (see last commit)

agoldstone93: I've now reverted the change as everything broke (again) when I set this to true

config/environments/test.rb
agoldstone93: I tried this and everything failed so I reverted back to this

config/environments/test.rb
agoldstone93: Oops, replied to the wrong comment above. [Put it here instead](https://github.com/ministryofjustice/laa-apply-for-legal-aid/pull/7628/commits/e938c579f41180fb0e0bac9cc506a999091b672a#r1967188836)

PR #7636, opened 2025-02-20
AP-5429: CYA update - review and print and submitted applications pages

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5429)

Update review and print and submitted applications pages to follow updated design:
- Increase page to full width
- Correctly show the information as in the figma the designs

Reuse partials that being used in earlier CYA pages and remove partials that are no longer being used

Extend / update feature tests to test the headings that should be show on the cya pages

Add missing caption to means CYA

## Checklist


app/controllers/providers/provider_base_controller.rb
```diff
@@ -21,5 +21,15 @@ def link_banner_display
 
       I18n.t("providers.submitted_applications.show.link_banner_one")
     end
+
+    def set_transaction_types
```
colinbruce, line 25: could this be a concern rather than on _all_ provider controllers?  Seems a bit off an overhead!

colinbruce (commented):
Looks good, only a slight concern about the overhead of including the `set_transaction_types` method on all controllers

app/controllers/providers/provider_base_controller.rb
agoldstone93: Have updated the last commit to use a concern instead

colinbruce (approved):
LGTM :+1:

PR #7644, opened 2025-02-26
AP-5580 part 2: remove proceeding parental responsibility questions

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5580)

This PR:
* removes the original, proceeding level, question controllers, views, etc 
* updates the url handler to remove the original requirements
* removes the original column from the database

The column was tested with ignore_columns to highlight errors and then removed 

## Checklist


config/locales/en/shared.yml
```diff
@@ -505,7 +505,7 @@ en:
           child_care_assessment_heading: Assessment of your client
           linked_children_heading: Children covered
           linked_children: Children covered under this proceeding
-          relationship_to_child_heading: Who your client is in the proceeding
+#          relationship_to_child_heading: Who your client is in the proceeding
```
jsugarman, line 508: can this be removed?

app/services/ccms/attribute_value_generator.rb
```diff
@@ -121,7 +121,7 @@ def child_subject_of_proceeding?(_options)
       # as Client Involvement Type or using the merits task, just that there is a child subject on the application
       legal_aid_application.proceedings.any? do |proceeding|
         proceeding.special_childrens_act? &&
-          (proceeding.relationship_to_child.eql?("child_subject") || proceeding.client_involvement_type_ccms_code.eql?("W"))
+          (legal_aid_application.applicant.relationship_to_children.eql?("child_subject") || proceeding.client_involvement_type_ccms_code.eql?("W"))
```
jsugarman, line 124: just wondering, from a business logic point of view whether we need to check that proceeeding is `special_childrens_act?`. 🤔 If proceeding has a child subject `client_involvement_type_ccms_code`, or [now] the application has a child subject `relationship_to_children` does that qualify it regardless of matter type :shrug:


If so could refactor, but either way. The name of the method also seems less aligned now. `child_subject_of_application?` or just `child_subject?`

```ruby
 def child_subject?(_options)
   return true if legal_aid_application.applicant.relationship_to_children.eql?("child_subject")
   
   legal_aid_application.proceedings.any? do |proceeding|
     proceeding.relationship_to_child.eql?("child_subject")
   end
 end
```

spec/helpers/task_list_helper_spec.rb
```diff
@@ -69,27 +69,4 @@
       end
     end
   end
-
-  describe "#proceeding_task_url" do
```
jsugarman, line 73: **nit/not-a-blocker** I see this is excecising functionality that is no longer part of the method but do we not want a unit test for it at all? happy to not have if this is covered elsewhere but i guess we have a lot of mixtures of unit tests of things that are covered elsewhere generally  and no "standard" that we adhere to (integration tests -v- unit tests)

jsugarman (approved):
👍 Some non-blocking comments and suggestions

spec/helpers/task_list_helper_spec.rb
colinbruce: Nope, this was a bit of greedy deletion... only the client relationship _was_ tested, but restored and changed to now test the generic helper :+1:

app/services/ccms/attribute_value_generator.rb
colinbruce: I can see the point of the logic update and will investigate

With regards to renaming, this method is only used to populate a CCMS field named `CLIENT_CHILD_SUBJECT_OF_PROC` so I think it makes sense to leave it as is for now :+1:

PR #7650, opened 2025-02-28
FIX: remove webdack gem

colinbruce:
## What

This is a follow up/replacement for #5727 

The [webdack_uuid-migration gem](https://github.com/kreatio-sw/webdack-uuid_migration) was used 5 years ago to change tables and has only been referenced in migration files, not actively used, since.

This version (as opposed to the original!) removes the gem and modifies the migration files affected.
It adds `id: :uuid` to the affected tables and removes the contents of the webdack migrations

## Benefits

* A new user will be able to spin up a local database using `rails db:migrate`
* removes a little used gem and reduces our lock file and dependabot updates
* reduces our dependency attack surface

## Cons

* should we be amending old migrations?  It won't affect production as the tables already exist

## Discussion
How do people feel about older migrations being modified?
The opinion online is that devs should use rails db:schema:load or rails db:prepare instead of rails db:migrate and running everything since day one.

I found one medium article saying they deleted migrations older than 3 months!? 🤔
https://medium.com/clutter-engineering/cleaning-up-old-rails-migrations-1b55b638abb5

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,7 +26,6 @@ gem "sentry-ruby"
 gem "sentry-sidekiq"
 gem "simple_command"
 gem "tzinfo-data"
-gem "webdack-uuid_migration", "~> 1.5.0"
 
 # Reduces boot times through caching; required in config/boot.rb
 gem "bootsnap", ">= 1.1.0", require: false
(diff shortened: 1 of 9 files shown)


jsugarman (approved):
👍 LGTM, lets burn with fire!

PR #7653, opened 2025-03-03
AP-5791(ext): New interrupt card pattern

jsugarman:
## What
Change pattern for interrupt card

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5791)

Investigations into font-size issues identified during PLF playback raised other MOJ design pattern
infringements for the [interrupt card](https://design-patterns.service.justice.gov.uk/components/interruption-card/) component. These included:

- There should be only one `moj-interruption-card__body` element/div. Elements within that such as `<p>` then inherit relevant properties.
- font-size for body text should be 1.5 rem generally
- actions should be in separate div to `moj-interruption-card__body`, not nested within it.

## CFE result
**BEFORE**
Screenshot 2025-03-04 at 13 17 20

**AFTER**
Screenshot 2025-03-04 at 13 17 04

## use CCMS
**BEFORE**
Screenshot 2025-03-04 at 13 03 40

**AFTER**
Screenshot 2025-03-04 at 09 32 09

## No national insurance number
**BEFORE**
Screenshot 2025-03-04 at 13 05 49

**AFTER**
Screenshot 2025-03-04 at 09 39 31

## Checklist

### DIFF
--- a/app/views/providers/check_benefits/_actions.html.erb
+++ b/app/views/providers/check_benefits/_actions.html.erb
@@ -1,0 +1,7 @@
+   <%= next_action_buttons_with_form(
+         url: providers_legal_aid_application_check_benefit_path,
+         method: :patch,
+         show_draft: true,
+         inverse_continue: true,
+         continue_button_text: t("generic.continue"),
+       ) %>
--- a/app/views/shared/_use_ccms.html.erb
+++ b/app/views/shared/_use_ccms.html.erb
@@ -1,13 +1,7 @@
 <%= page_template(page_title: t(".title_html"), head_title: t(".head_title"), template: :basic, column_width: "full") do %>
-  <%= interruption_card(heading: t(".title_html")) do %>
+  <%= interruption_card(heading: t(".title_html"), actions: -> { render "shared/use_ccms_actions" }) do %>
+    <p><%= body_text %></p>
 
-    <p class="govuk-body moj-interruption-card__body"><%= body_text %></p>
-
-    <%= govuk_list list, type: :bullet, classes: "moj-interruption-card__body" if local_assigns[:list] %>
-
-    <div class="govuk-button-group moj-interruption-card__actions">
-      <%= govuk_button_link_to t(".continue_in_ccms_html"), "https://portal.legalservices.gov.uk", inverse: true, role: "button" %>
-      <%= govuk_link_to t(".back_to_applications"), submitted_providers_legal_aid_applications_path, inverse: true %>
-    </div>
+    <%= govuk_list(list, type: :bullet) if local_assigns[:list] %>
   <% end %>
 <% end %>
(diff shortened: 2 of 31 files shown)


app/views/providers/check_benefits/_actions.html.erb
```diff
@@ -0,0 +1,8 @@
+   <%= next_action_buttons_with_form(
+         url: providers_legal_aid_application_check_benefit_path,
+         method: :patch,
+         show_draft: true,
+         inverse_continue: true,
+         continue_button_text: t("generic.continue"),
+         container_class: "moj-interruption-card__actions",
```
naseberry, line 7: Does the container also require the class `govuk-button-group`?

app/views/shared/_use_ccms.html.erb
```diff
@@ -1,13 +1,17 @@
...
+        heading: t(".title_html"),
+        actions: lambda {
+          safe_join(
+            [
+              govuk_button_link_to(t("shared.use_ccms.continue_in_ccms_html"), "https://portal.legalservices.gov.uk", inverse: true, role: "button"),
+              govuk_link_to(t("shared.use_ccms.back_to_applications"), submitted_providers_legal_aid_applications_path, inverse: true),
+            ],
+          )
+        },
+      ) do %>
```
naseberry, line 12: I presume this method action call is different because of the need to change the link.
Can we not use an action partial like the other views?

app/views/shared/moj_components_templates/_interruption_card_template.html.erb
```diff
@@ -3,7 +3,16 @@
...
+        <% if body.present? %>
+        <div class="govuk-body moj-interruption-card__body">
+          <%= body %>
+        </div>
+        <% end %>
+        <% if actions.present? %>
+        <div class="govuk-button-group moj-interruption-card__actions">
+          <%= actions.call %>
+        </div>
+        <% end %>
```
naseberry, line 15: Nit: indentation seems incorrect.

app/views/shared/_use_ccms.html.erb
jsugarman: This was considered. We could but it would mean an unnested partial called `_use_ccms_actions.html.erb`, or similar, and moving the translations. It felt more icky to me but I don't mind either way so your call?

app/views/shared/moj_components_templates/_interruption_card_template.html.erb
jsugarman: This was considered. Indenting here actually indents the divs in the markup an extra unneeded level when they are rendered which did not seem appropriate. I went with removing unneeded indent in markup  but don't mind either way - unneeded indent in rendered markup, or no indent in unrendered erb? Your call.

app/views/shared/_use_ccms.html.erb
naseberry: I would expect a single pattern, therefore could recommend having the partial.

jsugarman: done

naseberry:
Thanks, my comments have been reviewed and resolved.

naseberry (approved):

colinbruce (approved):
The `-> { render "actions" }` is a little counter-intuitive but there are lots of examples to follow if adding new interrupt pages in the future 👍

PR #7666, opened 2025-03-06
Chore: update govuk rubocop

colinbruce:
## What

This update includes a lot of cop changes... so I span it up separately to manage it pro-actively

I updated the gem and then updated each of the infractions in turn

## Checklist


app/controllers/providers/means/capital_disregards/discretionary_controller.rb
```diff
@@ -24,7 +24,7 @@ def update
 
         def form_params
           params
-            .expect(providers_means_capital_disregards_discretionary_form: [:none_selected, discretionary_capital_disregards: []])
+            .expect(providers_means_capital_disregards_discretionary_form: [:none_selected, {discretionary_capital_disregards: []}])
```
jsugarman: i would have expected/hoped it would have a space inside the `{}` but 🤷

app/controllers/providers/means/capital_disregards/discretionary_controller.rb
```diff
@@ -24,7 +24,7 @@ def update
 
         def form_params
           params
-            .expect(providers_means_capital_disregards_discretionary_form: [:none_selected, {discretionary_capital_disregards: []}])
+            .expect(providers_means_capital_disregards_discretionary_form: [:none_selected, { discretionary_capital_disregards: [] }])
```
jsugarman, line 27: 😂 scratch  my previous comment

jsugarman (approved):
👍 Cannot see anything wrong after a commit by commit go through.

The `expect` instead of `require`/`permit` change for strong params feels like we need to give it decent once over, aside from the test suite but i'd hope the test suite would have identified any issues from that.

PR #7676, opened 2025-03-11
AP-5789: Prevent phantom vehicle submissions

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5789)

We have seen a few occurrences of providers getting stuck on CFE submission because partially complete vehicles were being submitted.

On the last incident, we tracked the logs and the user started a new vehicle and hit save as draft.  the next log shows them restarting from the capital introductions page.  It's not clear if this was a glitch or if they used the browser back button.

This PR takes a couple of steps to mitigate this...

It updates the add another vehicle page to show that a record is incomplete
image

It also re-routes the page so that if the provider has said yes to vehicle, they will not see the yes/no question on a second pass through, they will see the above page with an incomplete record warning

I discussed validating this page so that they cannot submit if there are incomplete records, but agreed with design that we would investigate that in the future if people ignore the incomplete table entries

## Checklist

### DIFF
--- a/app/models/vehicle.rb
+++ b/app/models/vehicle.rb
@@ -1,6 +1,14 @@
 class Vehicle < ApplicationRecord
   belongs_to :legal_aid_application
 
+  def complete?
+    [
+      estimated_value.present?,
+      more_than_three_years_old.in?([true, false]),
+      used_regularly.in?([true, false]),
+    ].all?(true)
+  end
+
   def purchased_on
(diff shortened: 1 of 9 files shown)


colinbruce:
> Looks good to me!
> 
> Looks like you've missed a fixup commit in a rebase, but otherwise all looks good

Thank you!

PR #7696, opened 2025-03-13
AP-5239: Remove SCA feature flag

agoldstone93:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5239)

- Remove all references to SCA feature flag as it is now on by default
- Rename everything from special_childrens_act to special_children_act, to keep codebase consistent and to match the name of the act
- Update view on start page to use list component from govuk components library

## Checklist

### DIFF
--- a/config/locales/en/providers.yml
+++ b/config/locales/en/providers.yml
@@ -838,10 +838,6 @@ en:
     has_national_insurance_numbers:
       show:
         page_title: Does your client have a National Insurance number?
-        hint: |
-          We'll use your client's National Insurance number to check if they receive
-          a passported benefit. If you do not have it you may have to do a means
-          assessment for them.
         nino_label: Enter National Insurance number
     has_other_involved_children:
       show:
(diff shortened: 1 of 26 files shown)


config/locales/en/providers.yml
```diff
@@ -838,10 +838,6 @@ en:
     has_national_insurance_numbers:
       show:
         page_title: Does your client have a National Insurance number?
-        hint: |
```
colinbruce, line 841: I wonder if this was a typo, should the switch have been on proceeding.special_children_act/non_means_tested rather than setting.special_children_act?  Is it worth checking with design?
It just feels like this hint should be shown for DA, PLF etc

colinbruce (approved):
Generally looks great (and thanks for addressing my naming nightmare!) but one question about the NI hint

config/locales/en/providers.yml
agoldstone93: Double checked with them and it's good to be removed for everything

PR #7705, opened 2025-03-18
AP-5776 Fix bypassing evidence validation bug

RoseSAK:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5776)

Having come from the Checking Merits Answers page, redirect to supporting evidence page when the `copy of court order` answer is changed from `No` to `Yes` so the provider can upload the court order. 

Same for the `child_care_assessment` page and question.

## Checklist

### DIFF
--- a/app/forms/providers/application_merits_task/plf_court_order_form.rb
+++ b/app/forms/providers/application_merits_task/plf_court_order_form.rb
@@ -6,6 +6,10 @@ module Providers
       attr_accessor :plf_court_order
 
       validates :plf_court_order, inclusion: %w[true false], unless: :draft?
+
+      def copy_of_court_order?
+        plf_court_order == "true"
+      end
     end
   end
 end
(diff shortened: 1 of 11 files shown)


app/forms/providers/application_merits_task/plf_court_order_form.rb
```diff
@@ -6,6 +6,10 @@ class PLFCourtOrderForm < BaseForm
       attr_accessor :plf_court_order
 
       validates :plf_court_order, inclusion: %w[true false], unless: :draft?
+
+      def copy_of_court_order?
+        plf_court_order == "true"
```
agoldstone93, line 11: I'm interpreting this ticket as:
if their current answer is `no` and then they change their answer to `yes` they'll see evidence upload, but if their current answer is `yes` and they still choose `yes` they'll just go to CYA

They way this seems to work is:
if their current answer is `yes` and they still choose `yes`, they go to evidence upload

Not a major issue to me either way, but just flagging

agoldstone93 (approved):
One non-blocking comment about the interpretation of the ticket but not a major issue. Otherwise, everything else looks good to me

Also, not all your commits start with the AP number but i don't know how strict we are on that

PR #7709, opened 2025-03-19
Chore: make merits report summary card column widths one-third/two-thirds

jsugarman:
## What
Make merits report summary card column widths one-third/two-thirds

Inline with means report, CYA and R&P pages. Agreed with devs

## Checklist

### DIFF
--- a/app/views/providers/merits_reports/_applied_previously.html.erb
+++ b/app/views/providers/merits_reports/_applied_previously.html.erb
@@ -6,11 +6,11 @@
                          html_attributes: { id: "app-check-your-answers__applied_previously__card" }) do |card|
         card.with_summary_list(actions: false, html_attributes: { id: "app-check-your-answers__applied_previously__summary" }) do |summary_list|
           summary_list.with_row(html_attributes: { id: "app-check-your-answers__applied_previously" }) do |row|
-            row.with_key(text: t(".has_applied_before"), classes: "govuk-!-width-one-half")
+            row.with_key(text: t(".has_applied_before"), classes: "govuk-!-width-one-third")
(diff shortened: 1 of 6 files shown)


colinbruce (commented):
Why are some deleted and some changed.  Do subsequent rows inherit from the first? Seems a bit odd for CSS 😕 

What's the default display with *no* width set?  Could _all_ the *third classes be removed too?

jsugarman:
> Why are some deleted and some changed. Do subsequent rows inherit from the first? Seems a bit odd for CSS 😕

Yes if one row with the value exists for the card the others will behave the same. It is odd. This was pattern Adam had introduced and is already used in means and other pages. I guess to remove some unnecessary CSS. Happy to go either way but this would make the approach inconsistent. 

If we put the one-third back in we should do it for all the other places too really.

ultimately this makes the merits report consistently sized within all its summary cards and consistent with means and CYA/R&P pages.


examples.

[merits_report-one-third.pdf](https://github.com/user-attachments/files/19361571/merits_report-one-third.pdf)
[merits_report-half-half-current-example.pdf](https://github.com/user-attachments/files/19361572/merits_report-half-half-current-example.pdf)

jsugarman:
> What's the default display with _no_ width set? Could _all_ the *third classes be removed too?

Yes, removing them results in one-third split anyway! damn! I can apply in this PR but can do another to remove all of these from all summary cards as a separate PR?!

colinbruce:
I'd say maybe stick with row one as one-third, to maintain the current pattern and prevent future changes :+1:
Will approve

colinbruce (approved):

PR #7724, opened 2025-03-27
AP-5877: Add ApplicationDigest csv to admin report

naseberry:
## What

[Add ApplicationDigest CSV to admin reports](https://dsdmoj.atlassian.net/browse/AP-5877)

## Checklist

### DIFF
--- a/app/controllers/admin/reports_controller.rb
+++ b/app/controllers/admin/reports_controller.rb
@@ -46,6 +46,24 @@ module Admin
       end
     end
 
+    def download_application_digest_report
+      expires_now
+      respond_to do |format|
+        format.csv do
+          tempfile = Tempfile.new("application_digest_report")
+          headers = ApplicationDigest.first.attributes.keys - %w[id created_at updated_at]
+
+          CSV.open(tempfile, "w", write_headers: true, headers: headers) do |csv|
+            ApplicationDigest.order(created_at: :desc).each do |record|
+              csv << record.attributes.slice(*headers).values
+            end
+          end
+
+          send_data tempfile.read, filename: "application_digest_#{timestamp}.csv", content_type: "text/csv"
+        end
+      end
+    end
+
     def timestamp
       Time.current.strftime("%FT%T")
     end
@@ -69,19 +87,25 @@ module Admin
           report_title: "Application Details report ",
           report_link: { href: "https://dsdmoj.atlassian.net/wiki/x/JwBOKQE", text: "About this report (opens in new tab)", class: "govuk-link govuk-link--no-visited-state", rel: "noreferrer noopener", target: "_blank" },
           path: :admin_application_details_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">application details</span>CSV".html_safe,
+        },
+        application_digest_download: {
+          report_title: "Application Digest report",
+          report_link: nil,
+          path: :admin_application_digest_csv_path,
+          path_text: "Download <span class=\"govuk-visually-hidden\">application digest</span>CSV".html_safe,
         },
         provider_download: {
           report_title: "Provider email report",
           report_link: nil,
           path: :admin_provider_emails_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">provider emails</span>CSV".html_safe,
         },
         user_feedback_download: {
           report_title: "User feedback report",
           report_link: nil,
           path: :admin_user_feedbacks_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">user feedback</span>CSV".html_safe,
         },
       }
     end
--- a/spec/requests/admin/reports_controller_spec.rb
+++ b/spec/requests/admin/reports_controller_spec.rb
@@ -75,4 +75,28 @@ RSpec.describe Admin::ReportsController do
       expect(response.body).to match(/My suggested improvement/)
     end
   end
+
+  describe "GET application digest" do
+    subject(:get_request) { get admin_application_digest_csv_path(format: :csv) }
+
+    before do
+      create(
+        :application_digest,
+        firm_name: "An Awesome Firm",
+        provider_username: "Joe Bloggs",
+      )
+    end
+
+    it "renders successfully" do
+      get_request
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "sends the data" do
+      get_request
+      expect(response.body)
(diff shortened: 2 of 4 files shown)


app/controllers/admin/reports_controller.rb
```diff
@@ -46,6 +46,24 @@ def download_user_feedbacks_report
...
 
+    def download_application_digest_report
+      expires_now
+      respond_to do |format|
+        format.csv do
+          tempfile = Tempfile.new("application_digest_report")
+          headers = ApplicationDigest.first.attributes.keys - %w[id created_at updated_at]
+
+          CSV.open(tempfile, "w", write_headers: true, headers: headers) do |csv|
+            ApplicationDigest.order(create_at: :desc).find_each do |record|
```
jsugarman, line 57: I think this should be `created_at`. 
```suggestion
            ApplicationDigest.order(created_at: :desc).find_each do |record|
```

I am surprised tests did not breaks as if this was exercised it should have caused an error:
```sh
ApplicationDigest.order(create_at: :desc)
=> An error occurred when inspecting the object: #<ActiveRecord::StatementInvalid....
```

This implies we need more testing around this

app/controllers/admin/reports_controller.rb
```diff
@@ -69,19 +87,25 @@ def reports
...
           report_link: { href: "https://dsdmoj.atlassian.net/wiki/x/JwBOKQE", text: "About this report (opens in new tab)", class: "govuk-link govuk-link--no-visited-state", rel: "noreferrer noopener", target: "_blank" },
           path: :admin_application_details_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">application details</span> CSV".html_safe,
+        },
+        application_digest_download: {
+          report_title: "Application Digest report",
+          report_link: nil,
+          path: :admin_application_digest_csv_path,
+          path_text: "Download <span class=\"govuk-visually-hidden\">application digest</span> CSV".html_safe,
```
jsugarman, line 96: One too many spaces?!
```suggestion
          path_text: "Download <span class=\"govuk-visually-hidden\">application digest</span>CSV".html_safe,
```

app/controllers/admin/reports_controller.rb
```diff
@@ -69,19 +87,25 @@ def reports
...
         },
         provider_download: {
           report_title: "Provider email report",
           report_link: nil,
           path: :admin_provider_emails_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">provider emails</span> CSV".html_safe,
```
jsugarman, line 102: One too many spaces?!
```suggestion
          path_text: "Download <span class=\"govuk-visually-hidden\">provider emails</span>CSV".html_safe,
```

app/controllers/admin/reports_controller.rb
```diff
@@ -69,19 +87,25 @@ def reports
...
         },
         user_feedback_download: {
           report_title: "User feedback report",
           report_link: nil,
           path: :admin_user_feedbacks_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">user feedback</span> CSV".html_safe,
```
jsugarman, line 108: One too many spaces?!
```suggestion
          path_text: "Download <span class=\"govuk-visually-hidden\">user feedback</span>CSV".html_safe,
```

app/controllers/admin/reports_controller.rb
```diff
@@ -69,19 +87,25 @@ def reports
           report_title: "Application Details report ",
           report_link: { href: "https://dsdmoj.atlassian.net/wiki/x/JwBOKQE", text: "About this report (opens in new tab)", class: "govuk-link govuk-link--no-visited-state", rel: "noreferrer noopener", target: "_blank" },
           path: :admin_application_details_csv_path,
-          path_text: "Download CSV",
+          path_text: "Download <span class=\"govuk-visually-hidden\">application details</span> CSV".html_safe,
```
jsugarman, line 90: One too many spaces?!
```suggestion
          path_text: "Download <span class=\"govuk-visually-hidden\">application details</span>CSV".html_safe,
```

spec/requests/admin/reports_controller_spec.rb
```diff
@@ -75,4 +75,20 @@
...
+    before { create(:application_digest) }
+
+    it "renders successfully" do
+      get_request
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "sends the data" do
+      get_request
+      expect(response.body).to match(/^legal_aid_application_id,firm_name,provider_username,date_started,date_submitted,days_to_submission,use_ccms,matter_types,proceedings,passported,df_used,earliest_df_date,df_reported_date,working_days_to_report_df,working_days_to_submit_df,employed,hmrc_data_used,referred_to_caseworker,true_layer_path,bank_statements_path,true_layer_data,has_partner,contrary_interest,partner_dwp_challenge,applicant_age,non_means_tested,family_linked,family_linked_lead_or_associated,number_of_family_linked_applications,legal_linked,legal_linked_lead_or_associated,number_of_legal_linked_applications,no_fixed_address,biological_parent,parental_responsibility_agreement,parental_responsibility_court_order,child_subject,parental_responsibility_evidence,autogranted,ecct_routed/)
```
jsugarman, line 91: could we extend to test some of the content as well as the headers. You can use compound matchers

```ruby
expect(response.body)
  .to match(/this/)
  .and match(/that/)
```

jsugarman (commented):
A few comments, one essential, several minor content issues

app/controllers/admin/reports_controller.rb
naseberry: So it seems this is the default behaviour in Rails, it ignores any custom ordering specified in the query. Therefore, even if there's a typo in the order clause, find_each disregards it and proceeds without raising an error.

So, I can remove the order clause or change to the each method.

I prefer the more performant option, so removing the order clause.

"NOTE: It's not possible to set the order. That is automatically set to ascending on the primary key (“id ASC”) to make the batch ordering work." - https://api.rubyonrails.org/v5.2/classes/ActiveRecord/Batches.html#method-i-find_each

app/controllers/admin/reports_controller.rb
naseberry: 🤔  the space seems correct

app/controllers/admin/reports_controller.rb
colinbruce: That causes us problems as the primary key is a GUID so there is no intrinsic 'order'

I would use Joel's suggestion as that would return the records in an order that makes sense :+1:

app/controllers/admin/reports_controller.rb
colinbruce: Because the span is hidden the output will be 
Download<space><hidden tag><space>CSV
So there will be a double white space on screen for ```Download  CSV```

colinbruce (approved):
Updates look good to me

jsugarman (approved):
updates LGTM 👍

PR #7754, opened 2025-04-23
AP-5646: Update two year digit handling

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5646)

This was written a long time ago, when only date of birth was entered

We now have multiple date inputs that require future dates to be entered, e.g. hearing dates in the future, and this handler would reset a future date to be 19xx.

This update only applies a rollback on dates when the requested field is a date of birth entry. For all others it will roll forward

## Checklist

### DIFF
--- a/app/forms/concerns/date_field_builder.rb
+++ b/app/forms/concerns/date_field_builder.rb
@@ -105,7 +105,8 @@ private
     when 4
       year
     when 2
-      year > Time.current.strftime("%y") ? "19#{year}" : "20#{year}"
+      century = @method.eql?(:date_of_birth) && year > Time.current.strftime("%y") ? "19" : "20"
+      "#{century}#{year}"
     else
       raise YearError, "Year is incorrect length"
     end
--- a/app/forms/concerns/date_field_builder.rb
+++ b/app/forms/concerns/date_field_builder.rb
@@ -105,7 +105,7 @@ private
     when 4
       year
     when 2
-      century = @method.eql?(:date_of_birth) && year > Time.current.strftime("%y") ? "19" : "20"
+      century = @method.eql?(:date_of_birth) && year.to_i > Time.current.strftime("%y").to_i ? "19" : "20"
       "#{century}#{year}"
     else
       raise YearError, "Year is incorrect length"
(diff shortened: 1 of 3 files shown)


app/forms/concerns/date_field_builder.rb
```diff
@@ -105,7 +105,8 @@ def four_digit_year(year)
     when 4
       year
     when 2
-      year > Time.current.strftime("%y") ? "19#{year}" : "20#{year}"
+      century = @method.eql?(:date_of_birth) && year > Time.current.strftime("%y") ? "19" : "20"
```
jsugarman, line 108: should we compare as integers for safety

```suggestion
      century = @method.eql?(:date_of_birth) && year.to_i > Time.current.strftime("%y").to_i ? "19" : "20"
```

```
> "6" > Time.current.strftime("%y")
 => true
> "6".to_i > Time.current.strftime("%y").to_i
 => false
 ```
 not that single digits are possible here anyway but....

File: spec/forms/proceedings/final_hearing_form_spec.rb [hunk 2]
```diff
@@ -103,6 +103,15 @@ module Proceedings
               date: Time.zone.today,
             )
           end
+
+          context "and the year is greater than the current year" do
+            let(:current_year) { Time.current.strftime("%y").to_i }
+            let(:date_1i) { current_year + 1 }
+
+            it "sets the date of birth to be 19xx" do
```
jsugarman, line 111:
```suggestion
            it "sets the [final hearing] date to be 20xx" do
```

jsugarman: And, as above, be tempted to extend test coverage with its opposite, even though for this scenario the functional code will always return 20xx

```
          context "and the year is less than the current year" do
            let(:current_year) { Time.current.strftime("%y").to_i }
            let(:date_1i) { current_year - 1 }

            it "sets the [final hearing] date to be 20xx" do
              expect(final_hearing).to have_attributes(date: Date.new("20#{current_year + 1}".to_i, Time.zone.today.month, Time.zone.today.day))
            end
          end
```

jsugarman (approved):
✅ 

Non-blocking spec typo and couple of test extension suggestions for coverage, just because date maths is always confusing.

colinbruce:
Cheers @jsugarman - tests updated and extended

naseberry (approved):

PR #7792, opened 2025-05-16
AP-5910: Split statement of case into two pages

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5910)

This splits the statement of case page into two, an initial page with the choice of upload or typed. Typed expands with a textarea for the manually typed input. Upload takes you to the next page where multiple files can be uploaded

It removes the original form that handled both options on a single page

Update: This version includes changes to the migration script to better handle actual data, not the anonymised version that gets pulled down!

## Checklist


lib/tasks/migrate_statement_of_case.rake
```diff
@@ -0,0 +1,45 @@
...
+    upload_target = Attachment.statement_of_case.pluck(:legal_aid_application_id).uniq.count
+    Rails.logger.info "Affected applications: #{records.count}"
+    Rails.logger.info "Number with text statement: #{text_target}"
+    Rails.logger.info "Number with files uploaded: #{upload_target}"
+
+    ActiveRecord::Base.transaction do
+      records.find_each do |statement|
+        typed = statement.statement.present?
+        upload = statement.legal_aid_application.statement_of_case_uploaded?
+        statement.update!(typed:, upload:)
```
jsugarman, line 35: I just noted/remember somthing reviewing Adams rake task - it's probabkly best to not "touch" the statement of case to avoid delaying purging of the application. I don't think it will matter because it is not updating the application _but_ does activerecord cascade changes from associated records to the parent (the legal_aid application) 🤷 .



```
  statement.update!(typed:, upload:, touch: false) # this prevents the updated_at date being changed and delaying purging of stale records
```

I don't _think_ it will unless the belongs_to is defined as `touch: true`, which it is not.

colinbruce: `touch: false` doesn't work on update, only `save`

Changed to `update_columns` which doesn't trigger callbacks or timestamps

jsugarman: damn, wasn't sure if it worked on update or not - now i know.

Not normally keen on update_columns but can't see any related callbacks and validations are relevant here and assume this all worked against prod dump, in which case 👍

jsugarman (approved):
👍

PR #7814, opened 2025-05-27
Fix: fix flaky tests

jsugarman:
## What
Couple of flaky tests need fixing

The feature test flakiness is caused by an ordering issues on 
proceedings, resulting from use a less than specific search term.

The helper specs use of `eq` makes the tests order dependant, possibly 
unnecesarily - to check with Adam.

## Checklist


colinbruce (commented):
LGTM, but may be worth getting confirmation from @agoldstone93

PR #7831, opened 2025-06-02
AP-5519: Ensure consistent currency validation

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-5519)

Use the CurrencyCleaner class in the forms and handlers that had a different solution used.  This ensures the same substitutions are used in all forms and a consistent return is provided by the system.

This is, possibly, a short term solution as we are using many different ways of validating, storing and displaying currency values between different forms and handlers

This PR also fixes the display of currency amounts on the Check your Answers pages, to display without trailing zeros i.e. `£45` instead of `£45.00`. Amounts with pence still display as `£45.01`.

## Checklist

### DIFF
--- a/app/forms/providers/regular_transaction_form.rb
+++ b/app/forms/providers/regular_transaction_form.rb
@@ -168,7 +168,7 @@ module Providers
     def all_regular_transactions_valid
       regular_transactions.each do |transaction|
         transaction_type = transaction.transaction_type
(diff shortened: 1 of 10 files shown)


arthurashman:
Only additional thought is do we want a test or 2 that covers when there are pence?

PR #7858, opened 2025-06-06
AP-6071: Separate ping ingress

jsugarman:
## What
Separate ping ingress

[Link to story](https://dsdmoj.atlassian.net/browse/AP-6071)

To allow or ping to continue unwhitelisted behind the velvet rope

## Checklist


helm_deploy/apply-for-legal-aid/values-production.yaml
```diff
@@ -54,6 +54,14 @@ ingress:
     - "apply-for-legal-aid.service.justice.gov.uk"
   secretName: "apply-for-legal-aid-tls-secret"
 
+publicIngress:
+  enabled: true
+  className: modsec
```
colinbruce, line 59: The PR description says
> Note that this also disables modsec on the ingress for /ping and /status.

But this seems to say it _is_ implemented :/

colinbruce (approved):
LGTM 👍 
One non-blocking observation on the description

helm_deploy/apply-for-legal-aid/values-production.yaml
jsugarman: @colinbruce thanks for the review. I have left `modsec` as the `className` so that it is "enablable" but staging and production both have an annotation, as below, that disables it.
```
nginx.ingress.kubernetes.io/enable-modsecurity: "false"
```

I was not sure whether we wanted to use modsec or not for this ingress 🤷 so left it enableable in case. What do you think? Should we completely remove it? `#ping` is a very basic action so do not see how modesec adds anything.

PR #7899, opened 2025-07-03
AP-6089: Fix task list navigation impact

jsugarman:
## What
Fix navigation to from task list impact on flow

[Link to  story](https://dsdmoj.atlassian.net/browse/AP-6089)

Was causing oddity in that if you went to task list then it was set as your provider_step, but because the 
application list generation of links uses the flow service and the task list is is not a recognised step
of the flow service it errors and sets the link to a fallback (the applicant details page). 

This excludes task list from provider step updates because provider_step is updated 
on every page that uses the application_dependable  module but the task list is not 
currently part of any flow or step and so should be excluded. 

Also, this amend the task list "Back to applications" button. It was pointing to submitted tab of application list but should be the "In progress" tab, inline with other "Save and continue later" buttons.
## Checklist


colinbruce (approved):
LGTM :+1:

PR #7909, opened 2025-07-10
AP-6140: Restrict access via port 80

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-6140)

This PR will block access on port 80, instead of redirecting to the https endpoint

## Checklist


colinbruce:
> Do we not have to do the same on production and staging, or will that be future prs? (I am probably missing something!)

We do... and I added them to a commit which I thought I pushed! 🤦‍♂️ 
I also didn't update the comment - yesterday did not go well!

jsugarman (approved):
👍  Looks like you have an empty commit as the first commit and the ticket number on the second commit is 6041 instead of 6140 but otherwise all good.

PR #7913, opened 2025-07-15
rediret 403 to holding page on UAT

jsugarman:
## What
Redirect 403 to holding page on UAT

When not on whitelist you receive and nginx ingress 403 that looks
ugly. This redirects to the holding page for testing purposes but 
could point to our own internal 403 or other page.

## Checklist


jsugarman:
> This suggested change does not work unfortunately. Also tried removing the `$request_uri;` part and no joy. have let Folarin know

**For discussion:** This now works, and combining the two server-snippets into one (there can be only one per ingress) works for both rules so all good. We may want to redirect to another page, like our own application-specific 403 page.

jsugarman:
does not work, no velvet rope anymore and not a suitable implementation in any event. in service 403 page similar to assurance tool or NSCC would be more suitable

PR #7975, opened 2025-08-19
AP-6144: Benefit Checker Fallback

colinbruce:
## What

[Link to story](https://dsdmoj.atlassian.net/browse/AP-6144)

This adds a setting to turn off Benefit Checker if we know it will be off for some time, it matches the `collect_hmrc_data?` name and pattern.

This will redirect to a new set of pages when the switch is off, or a call to benefit checker call fails.

It could be expanded to handle the whole DWP flow, but (while work was started) it was taking too long so this is the initial, simplified version

## Checklist

### DIFF
--- a/app/controllers/providers/dwp/overrides_controller.rb
+++ b/app/controllers/providers/dwp/overrides_controller.rb
@@ -1,0 +1,82 @@
+module Providers
+  module DWP
+    class OverridesController < ProviderBaseController
+      include ApplicantDetailsCheckable
+
+      prefix_step_with :dwp
+
+      helper_method :display_hmrc_text?
+
+      def show
+        @form = Providers::DWP::OverridesForm.new(model: partner)
+      end
+
+      def update
+        return continue_or_draft if draft_selected?
+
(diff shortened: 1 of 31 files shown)


app/controllers/providers/dwp/overrides_controller.rb
```diff
@@ -0,0 +1,82 @@
+module Providers
```
kmahern, line 1: Is it still our pattern to create request specs for new controllers? (or maybe we are now creating system specs instead?)

app/services/flow/steps/provider_dwp/dwp_overrides_step.rb
```diff
@@ -0,0 +1,18 @@
+module Flow
```
kmahern, line 1: Would it be worth creating a dwp_overrides_step_spec?

app/forms/providers/dwp/overrides_form.rb
```diff
@@ -0,0 +1,37 @@
+module Providers
```
kmahern, line 1: Would it be worth creating a overrides_form_spec for this form?

